PR #45, opened 2014-10-07
Add bitpos command.

vmihailenco:
One problem with this implementation is that it is impossible to issue command like:

```
bitpos key 1 0 0
```


vmihailenco:
Closing in favor of https://github.com/go-redis/redis/pull/46

PR #59, opened 2015-01-15
Migrated test suite to ginkgo/gomega

dim:
Also, looks like the sentinel test is now failing consistently


dim:
Damn, wrong base, will submit a new one

PR #61, opened 2015-01-15
Normalise `pubsub numsub` results

dim:
Due to https://github.com/antirez/redis/issues/1561 the result of PubSubNumSub may differ at the moment and breaks tests.


vmihailenco:
This workaround will not work with pipelines when command is not immediately proccessed. Probably you have to create new `PubSubNumSubCmd` that contains this workaround.

dim:
You're right, sorry didn't think of that.

PR #65, opened 2015-01-25
Better fix for pubsub numsub

dim:
Closes #61


vmihailenco:
Looks good. This is an API change, so we probably we should release v3.

dim:
I think the change is reasonably minor, 2.4 seems more appropriate

dim:
Ultimately, your call :)

Issue #66, opened 2015-01-26
set command with XX options


bronze1man:
Is there any command in the library that only set the key if it already exist?
I found set with xx options can fix this problem. But it is not in the library.

please look at http://redis.io/commands/set
I can use CustomCommand to workaround this right now.

vmihailenco:
Closed by https://github.com/go-redis/redis/pull/69

PR #67, opened 2015-01-30
Added SetSpecial command 

dim:
Proposal for issue #66

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -429,6 +429,29 @@ func (c *Client) Set(key, value string) *StatusCmd {
 	return cmd
 }
 
+type SetOpt struct {
+	EX, PX time.Duration
+	NX, XX bool
+}
+
+func (c *Client) SetSpecial(key, value string, opt SetOpt) *StatusCmd {
+	args := append(make([]string, 0, 6), "SET", key, value)
+	if opt.PX > 0 {
+		args = append(args, "PX", strconv.FormatInt(int64(opt.PX/time.Millisecond), 10))
+	} else if opt.EX > 0 {
+		args = append(args, "EX", strconv.FormatInt(int64(opt.EX/time.Second), 10))
+	}
+	if opt.NX {
+		args = append(args, "NX")
+	} else if opt.XX {
+		args = append(args, "XX")
+	}
+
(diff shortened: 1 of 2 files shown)


vmihailenco:
We could add some logic to automatically use EX or PX as needed. And use more descriptive name :)

As for the NX|XX we can have SetNX(key, value, expiration) and SetXX(key, value, expiration). Since `SetNX` is already used we should probably release it as v3.

vmihailenco:
@bronze1man @dim any thoughts?

dim:
As the redis documentation suggests:

> Note: Since the SET command options can replace SETNX, SETEX, PSETEX, it is possible that in future versions of Redis these three commands will be deprecated and finally removed.

That's the primary reason why I went for a `SetSpecial`. I don't really have a name preference though.

bronze1man:
I think it is a good idea to add SetSpecial,and rename to set in v4.

vmihailenco:
I've sent https://github.com/go-redis/redis/pull/69 to demonstrate what I mean. So far I like it much more than `SetSpecial` approach.

>That's the primary reason why I went for a SetSpecial

When Redis will remove support for `SETNX` go-redis will use `SET key value NX` as a replacement. No API changes will be needed.

bronze1man:
I think #69 is better.#69 is simpler then #67.
But I think #69 need API changes.

vmihailenco:
Yeah, it is supposed to be released in new version. I don't mind to merge this PR into v2 branch if there is interest.

commands.go
```diff
@@ -429,6 +429,29 @@ func (c *Client) Set(key, value string) *StatusCmd {
 	return cmd
 }
 
+type SetOpt struct {
+	EX, PX time.Duration
+	NX, XX bool
+}
+
+func (c *Client) SetSpecial(key, value string, opt SetOpt) *StatusCmd {
```
vmihailenco: I would name it `SetExt` or `SetExtended` if we are going to add it into v2 branch.

vmihailenco:
Closing in favor of https://github.com/go-redis/redis/pull/69

PR #69, opened 2015-01-31
Refactor Set, SetNX and SetXX with expiration.

vmihailenco:


vmihailenco:
I think this PR is good enough to be merged. But I don't mind if we leave `Set(key, value)` unchanged and add `SetExpiration(key, value, expiration)` for better backwards compatibility if there is interest in it. Merging.

PR #70, opened 2015-01-31
Faster, lock-free pool

dim:
Before:
```
$ go test ./... -run=NONE -bench=Pool -benchmem -cpu=1,2,4,8 -benchtime=2s
BenchmarkPool   10000000         293 ns/op         0 B/op        0 allocs/op
BenchmarkPool-2  3000000        1127 ns/op         0 B/op        0 allocs/op
BenchmarkPool-4  2000000        1187 ns/op         0 B/op        0 allocs/op
BenchmarkPool-8  2000000        1459 ns/op         0 B/op        0 allocs/op
```

After:
```
$ go test ./... -run=NONE -bench=Pool -benchmem -cpu=1,2,4,8 -benchtime=2s
BenchmarkPool   10000000         330 ns/op         0 B/op        0 allocs/op
BenchmarkPool-2 10000000         288 ns/op         0 B/op        0 allocs/op
BenchmarkPool-4 10000000         401 ns/op         0 B/op        0 allocs/op
BenchmarkPool-8 10000000         403 ns/op         0 B/op        0 allocs/op
```

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -251,13 +251,13 @@ func (p *connPool) Close() (err error) {
 
 	for {
 		if p.Size() < 1 {
-			return
+			break
 		}
 		if e := p.remove(<-p.conns); e != nil {
 			err = e
 		}
 	}
-	panic("not reached")
+	return
 }
 
 //------------------------------------------------------------------------------
--- a/pool.go
+++ b/pool.go
@@ -1,19 +1,20 @@
 package redis
 
 import (
-	"container/list"
 	"errors"
 	"fmt"
 	"log"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"gopkg.in/bufio.v1"
 )
 
 var (
-	errClosed = errors.New("redis: client is closed")
+	errClosed      = errors.New("redis: client is closed")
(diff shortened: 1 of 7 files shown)


pool.go
```diff
@@ -195,92 +197,67 @@ func (p *connPool) Put(cn *conn) error {
...
-	p.closed = true
 	p.rl.Close()
-	var retErr error
+
 	for {
-		e := p.conns.Front()
-		if e == nil {
-			break
+		if p.Size() < 1 {
+			return
```
vmihailenco: You couild replace `return` with `break` and get rid of `panic`

redis.go
```diff
@@ -167,6 +168,7 @@ type Options struct {
 	WriteTimeout time.Duration
 
 	PoolSize    int
+	PoolTimeout time.Duration
```
vmihailenco: I know that I am bad at documenting my code, but could you please add brief comment explaining this option? Something like `maximum duration before timing out waiting for connection from the pool`.

dim:
Updated

vmihailenco:
Thanks! Finally package got some docs :)

PR #71, opened 2015-03-08
Cluster intergration

dim:
This is a my attempt at integrating cluster features. I tried to keep the API consistent and make as few changes as possible. Looking forward to your review.


cluster.go
```diff
@@ -0,0 +1,464 @@
...
+// Closes the pool
+func (c *ClusterClient) Close() error {
+	c.cachemx.Lock()
+	defer c.cachemx.Unlock()
+
+	return c.reset()
+}
+
+// GetMasterAddrBySlot finds the current master address for a given hash slot
+func (c *ClusterClient) GetMasterAddrBySlot(hashSlot int) string {
```
vmihailenco: Can we unexport this?

API seems to be low-level, e.g. it is unclear how user should obtain hashSlot.

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+	return ""
+}
+
+// GetSlaveAddrsBySlot finds the current slave addresses for a given hash slot
+func (c *ClusterClient) GetSlaveAddrsBySlot(hashSlot int) []string {
+	if addrs := c.slots[hashSlot]; len(addrs) > 0 {
+		return addrs[1:]
+	}
+	return []string{}
```
vmihailenco: return nil?

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+// GetMasterAddrBySlot finds the current master address for a given hash slot
+func (c *ClusterClient) GetMasterAddrBySlot(hashSlot int) string {
+	if addrs := c.slots[hashSlot]; len(addrs) > 0 {
+		return addrs[0]
+	}
+	return ""
+}
+
+// GetSlaveAddrsBySlot finds the current slave addresses for a given hash slot
+func (c *ClusterClient) GetSlaveAddrsBySlot(hashSlot int) []string {
```
vmihailenco: unexport?

cluster.go
dim: Yep, makes sense

cluster.go
dim: The use case for this is creating pipelines. I was thinking about adding an `OmniPipeline` which would use the code above to group commands by address, then execute in individual Pipelines against the actual master, but until then, this is the only way to access this functionality

cluster.go
dim: Same as above, it's for creating pipelines. `HashSlot` function is public for the very same reason.

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+		WriteTimeout: opt.WriteTimeout,
+
+		PoolSize:    opt.getPoolSize(),
+		IdleTimeout: opt.IdleTimeout,
+	}
+}
+
+//------------------------------------------------------------------------------
+
+const HashSlots = 16384
```
vmihailenco: unexport const and HashSlot function?

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+			key = key[s+1 : s+e+1]
+		}
+	}
+	if key == "" {
+		return rand.Intn(HashSlots)
+	}
+	return int(crc16sum(key)) % HashSlots
+}
+
+// CRC16 implementation according to CCITT standards.
```
vmihailenco: Please move to separate file, e.g. `crc16.go`

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+func crc16sum(key string) (crc uint16) {
+	for i := 0; i < len(key); i++ {
+		crc = (crc << 8) ^ crc16tab[(byte(crc>>8)^key[i])&0x00ff]
+	}
+	return
+}
+
+// ------------------------------------------------------------------------
+
+// Shamelessly copied and adapted from:
```
vmihailenco: Separate file, e.g. lru.go

dim: Tried to keep all cluster-related code in one file, but sure

cluster.go
dim: Tried to keep all cluster-related code in one file, but sure

cluster.go
dim: OK with const, but keeping HashSlot makes sense

command.go
```diff
@@ -632,3 +692,47 @@ func (cmd *ScanCmd) parseReply(rd *bufio.Reader) error {
 
 	return nil
 }
+
+//------------------------------------------------------------------------------
+
+type ClusterSlotInfo struct {
```
vmihailenco: Probably `ClusterNodeInfo`?

command.go
```diff
@@ -632,3 +692,47 @@ func (cmd *ScanCmd) parseReply(rd *bufio.Reader) error {
...
+	Min, Max int
```
vmihailenco: `MinSlot`, `MaxSlot` I guess?

command.go
dim: The redis command is called CLUSTER SLOTS - http://redis.io/commands/cluster-slots

commands.go
```diff
@@ -17,76 +17,85 @@ func readTimeout(sec int64) time.Duration {
 	return time.Duration(sec+1) * time.Second
 }
 
+type commandable struct {
```
vmihailenco: `processor` is more obvious name.

command.go
dim: Sure, easy to change

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+
+type ClusterClient struct {
+	commandable
+
+	addrs []string
+	slots [][]string
+	conns *lruPool
+	opt   *ClusterOptions
+
+	cachemx sync.RWMutex
```
vmihailenco: Please add a comment explaining what this mutex protects.

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+// GetSlaveAddrsBySlot finds the current slave addresses for a given hash slot
+func (c *ClusterClient) GetSlaveAddrsBySlot(hashSlot int) []string {
+	if addrs := c.slots[hashSlot]; len(addrs) > 0 {
+		return addrs[1:]
+	}
+	return []string{}
+}
+
+// GetNodeClientByAddr returns a node's client for a given address
+func (c *ClusterClient) GetNodeClientByAddr(addr string) *Client {
```
vmihailenco: unexport?

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+	return c.conns.Fetch(addr)
+}
+
+// ------------------------------------------------------------------------
+
+func (c *ClusterClient) process(cmd Cmder) {
+	var hashSlot int
+
+	c.reloadIfDue()
+	ask := false
```
vmihailenco: `var ask bool` and move it to the top.

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+func (opt *ClusterOptions) getAddrs() ([]string, error) {
+	if len(opt.Addrs) < 1 {
+		return nil, errNoAddrs
+	}
+	return opt.Addrs, nil
+}
+
+func (opt *ClusterOptions) ClientOptions(addr string) *Options {
```
vmihailenco: I am not sure why this is exported...

command.go
```diff
@@ -24,18 +24,22 @@ var (
...
 	setErr(error)
+	reset()
 
 	writeTimeout() *time.Duration
 	readTimeout() *time.Duration
+	firstKey() string
 
 	Err() error
 	String() string
+	SetKeyArgPos(int)
```
vmihailenco: This is ugly. I suggest to
- remove `SetKeyArgPos`
- rename `firstKey` to `clusterKey` to make it clear that it is used to cluster commands
- later we can consider exporting `clusterKey` if user needs it

Instead of

```
 cmd := NewStatusCmd("CONFIG", "SET", parameter, value)
 cmd.SetKeyArgPos(0)
```

use something like

```
 cmd := NewStatusCmd("CONFIG", "SET", parameter, value)
 cmd._base._clusterKeyPos = 0
```

command.go
vmihailenco: Looking at http://redis.io/commands/cluster-slots I think it should be called `Start` and `End`

vmihailenco:
My biggest concern so far is that we export a lot of names. Can you try to unexport as many funcs as possible?

cluster.go
vmihailenco: Can we add `OmniPipeline` to the package when it is ready and unexport this?

cluster.go
dim: Pipelines again.

cluster.go
dim: k

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+// ------------------------------------------------------------------------
+
+func (c *ClusterClient) process(cmd Cmder) {
+	var hashSlot int
+
+	c.reloadIfDue()
+	ask := false
+
+	key := cmd.firstKey()
+	if key != "" {
```
vmihailenco: Looks like empty key is handled by `HashSlot`.

cluster.go
dim: Because we required it in one of our projects :)

command.go
dim: OK, but I really think users should have option to set cluster keys for custom commands

dim:
I know you like to keep your exported API as small as possible, but there are reasons for the exports, and they are all based on real-world production usage.

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+		return
+	}
+
+	var infos []ClusterSlotInfo
+
+	c.cachemx.Lock()
+	defer c.cachemx.Unlock()
+
+	for _, addr := range c.addrs {
+		c.reset()
```
vmihailenco: Should not it be placed before the loop?

cluster.go
dim: I find this feature really quite useful. What if you wanted to do a MULTI operation on a specific key or just get the info from a particular node.

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+	return err
+}
+
+// Forces a cache reload on next request
+func (c *ClusterClient) forceReload() {
+	atomic.StoreUint32(&c._reload, 1)
+}
+
+// Find the next unseen address
+func (c *ClusterClient) next(seen map[string]struct{}) string {
```
vmihailenco: More descriptive name would be welcome, e.g. `nextAddr`

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+	for _, addr := range c.addrs {
+		if _, ok := seen[addr]; !ok {
+			return addr
+		}
+	}
+	return ""
+}
+
+// Fetch slot information
+func (c *ClusterClient) fetch(addr string) ([]ClusterSlotInfo, error) {
```
vmihailenco: `fetchClusterSlots`?

cluster.go
dim: Sorry, which one?

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+
+		for _, addr := range info.Addrs {
+			if _, ok := known[addr]; !ok {
+				c.addrs = append(c.addrs, addr)
+				known[addr] = struct{}{}
+			}
+		}
+	}
+
+	// Shuffle addresses
```
vmihailenco: Please add a comment explaining why this is useful (probably link to the doc?).

cluster.go
```diff
@@ -0,0 +1,464 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+type ClusterClient struct {
+	commandable
+
+	addrs []string
+	slots [][]string
```
vmihailenco: This seems to have fixed size, so why not make it obvious:

```
slots [HashSlots][]string
```

command.go
vmihailenco: I don't mind, but I don't want to block this PR picking good name for this.

cluster.go
vmihailenco: Seems like you iterate over all addresses to fetch slot info, but you reset it (slot info) on each iteration (`c.reset`). Probably I misunderstand the code.

Also `c.update` shuffles `c.addrs` slice and I am not sure how it affects `range c.addrs` here.

cluster.go
dim: Fixed in #4bbb0cf

cluster.go
dim: Fixed in https://github.com/go-redis/redis/commit/4bbb0cf19944b2d56f6bb894eef0b1f91c7f5b53

cluster.go
dim: Fixed in https://github.com/go-redis/redis/commit/4bbb0cf19944b2d56f6bb894eef0b1f91c7f5b53

cluster.go
dim: Fixed in https://github.com/go-redis/redis/commit/4bbb0cf19944b2d56f6bb894eef0b1f91c7f5b53

cluster.go
dim: This is based on the reference implementation: https://github.com/antirez/redis-rb-cluster/blob/master/cluster.rb

dim: It resets the cache on each attempt https://github.com/antirez/redis-rb-cluster/blob/master/cluster.rb#L60 and then breaks on the first success https://github.com/antirez/redis-rb-cluster/blob/master/cluster.rb#L83

cluster.go
dim: Because it's easier to reset slices than arrays.

dim:
What do you think?
https://github.com/go-redis/redis/compare/641a75e...ef1d746

cluster.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	addrs := make(map[string]struct{}, size)
+	for _, addr := range opt.Addrs {
+		addrs[addr] = struct{}{}
+	}
+	return addrs, nil
+}
+
+func (opt *ClusterOptions) ClientOptions(addr string) *Options {
```
vmihailenco: I would prefer we unexport this. But if you think that it must be exported can you please remove `addr` argument from this func? Instead you can always do:

```
clopt = opt.ClientOptions()
clopt.Addr = theAddr
```

This way `ClientOptions() *Options` will have the same signature on `ClusterOptions` and `FailoverOptions`.

cluster.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+}
+
+//------------------------------------------------------------------------------
+
+const hashSlots = 16384
+
+// HashSlot returns a consistent slot number between 0 and 16383
+// for any given string key
+func HashSlot(key string) int {
```
vmihailenco: Can we unexport this now? :)

command.go
vmihailenco: Probably `ClusterSlot`? I don't insist.

cluster.go
```diff
@@ -0,0 +1,316 @@
...
+	Password string
+
+	// The maximum number of MOVED/ASK redirects to follow, before
+	// giving up. Default: 16
+	MaxRedirects int
+
+	// The maximum number of open connections to cluster nodes. Default: 10.
+	// WARNING: Each connection maintains its own connection pool. The maximum
+	// possible number of socket connections is therefore MaxConns x PoolSize
+	MaxConns int
```
vmihailenco: This name is a bit confusing. Probably `MaxClients`? `MaxNodes`?

vmihailenco: I understand that you copied this setting from the Ruby client, but do you really think that it is useful? Just wondering.

vmihailenco:
Sorry for delay and thanks for unexporting most of the funcs! It will give us freedom to refactor the code later. I will try to fix Travis tomorrow and then we can think about merging it :) Great work!

cluster.go
dim: It's really quite useful in debugging apps, I don't insist but this implementation will never change

cluster.go
dim: Will remove it for the initial release. There are some problems with the LRU, let's not get stuck with details for now.

vmihailenco:
What do you think about "MOVED 10374 127.0.0.1:8222" error in the ClusterPipeline? It seems inconvenient that we don't handle it, but what is the alternative? Repeat the whole pipeline? Theoretically it should be safe...

vmihailenco:
I am going to branch this PR, remove `ClusterPipeline` and merge it. `ClusterPipeline` requires more work and can be done in a separate PR.

dim:
Repeat the whole pipeline? That wouldn't be safe. You could only collect the MOVED responses of the failed commands, group them by address and re-issue them to the relevant servers. I could add that bit in, as I rather prefer having a pipelining option in the next release

dim:
Closing this one

PR #75, opened 2015-03-20
Improve rate-limiter

dim:
Minor optimisation: this one is a little less precise (which is hardly relevant), but doesn't require a separate goroutine.


rate_limit.go
```diff
@@ -5,49 +5,38 @@ import (
...
-		v: int64(bucketSize),
+func newRateLimiter(limit time.Duration, rate int) *rateLimiter {
+	if limit < time.Second {
+		limit = time.Second
 	}
-	go rl.loop(limit, int64(bucketSize))
-	return rl
-}
+	return &rateLimiter{
+		rate:  int64(rate) * 1e9,
```
vmihailenco: 1e9 should be a constant for readability.

rate_limit.go
```diff
@@ -5,49 +5,38 @@ import (
...
-	atomic.StoreInt64(&rl._closed, 1)
-	return nil
-}
-
-func (rl *rateLimiter) closed() bool {
-	return atomic.LoadInt64(&rl._closed) == 1
+	current := atomic.AddInt64(&rl.allowance, passed*rl.rate/rl.limit)
+	if current > rl.rate {
+		atomic.StoreInt64(&rl.allowance, rl.rate)
+		current = rl.rate
```
vmihailenco: Comment please. I don't understand what you are doing here.

rate_limit.go
```diff
@@ -5,49 +5,38 @@ import (
...
+	}
+	if current < 1e9 {
```
vmihailenco: And here.

rate_limit.go
```diff
@@ -5,49 +5,38 @@ import (
...
+		return true
+	}
+	atomic.AddInt64(&rl.allowance, -1e9)
```
vmihailenco: And here...

rate_limit.go
```diff
@@ -5,49 +5,38 @@ import (
...
-		if v := atomic.LoadInt64(&rl.v); v > 0 {
-			if atomic.CompareAndSwapInt64(&rl.v, v, v-1) {
-				return true
-			}
-		} else {
-			return false
-		}
-	}
-}
+func (rl *rateLimiter) Limit() bool {
```
vmihailenco: Please document return value since it is not obvious what true/false means in this context.

rate_limit_test.go
```diff
@@ -9,17 +8,14 @@ import (
 )
 
 var _ = Describe("RateLimiter", func() {
-	var n = 100000
-	if testing.Short() {
-		n = 1000
-	}
+	n := 1000
```
vmihailenco: Is this change relevant?

dim:
I've published a more generic, accurate and simplified version of the algorithm here https://github.com/bsm/ratelimit. It's also well-documented. Do you want me to vendor it to e.g. `_vendor/github.com/bsm/ratelimit` or just copy and paste?

vmihailenco:
I don't mind if we add it as a dependency. 

I think you should check `rl.allowance` for overflow. It can happen when client does not open new connections for several days and therefore `RateLimiter.Limit` is not called too. So `passed` can have very large value.

Also comments you added are not very helpful, because they say "what" not "why". E.g. instead of

```
// Ensure our allowance is not over maximum
if current > rl.max {
```

you can say

```
// Ensure our allowance is not over maximum.
// It can happen because allowance initialized to maximum.
```

dim:
I thought about an overflow guard to `rl.allowance`, but then left it out in favour of code simplicity. The chance of overflows is relatively small after idleness. Overflows are much more likely when large rates and intervals are used in `New()`.

vmihailenco:
I think overflow checking is as simple as:

```
if current < 0 {
    allowance = max
    current = max
}
```

That does not add much complexity.

dim:
Can you please submit a PR with a test? :) I couldn't find a sensible use case where an overflow would be caused 'purely' by high `passed`. As said, it's usually a combination of high `rate` and `per` input values to `New`.

vmihailenco:
I mean something like this http://play.golang.org/p/JfUuc_nL84 . Number does not look too big to me. I can write a test, but that requires fake time.Now implementation...

dim:
But it's not `fmt.Println(idle * unit)` but `fmt.Println(idle + unit)`

dim:
Actually, it's `current := atomic.AddUint64(&rl.allowance, passed)`, but it's still an addition, not a multiplication.

dim:
Hey, pushed another update.

vmihailenco:
LGTM. Thanks.

PR #77, opened 2015-03-23
Feature/cluster pipeline rebas

vmihailenco:


mpipeline.go
```diff
@@ -0,0 +1,129 @@
+package redis
+
+// Not thread-safe.
+type MultiPipeline struct {
```
vmihailenco: Can we rename it to `ClusterPipeline`?

dim: Sure, go ahead

dim:
I think this one's ready now, I've been using it now for a while and don't have anything to add.

vmihailenco:
@dim I tried to rebase this branch on master, but forgot to pull you latest changes first... As a result your last commits are lost. Can you please push them again?

dim:
Force-pushed again

cluster_pipeline.go
```diff
@@ -0,0 +1,130 @@
...
+	cmds   []Cmder
+	client *ClusterClient
+	closed bool
+}
+
+// ClusterPipeline creates a new pipeline which is able to execute commands
+// against multiple shards. Be aware that multi-pipelines are
+// not able to follow redirects and you may receive errors if a node
+// fails or if keys are moved before/during pipeline execution.
+func (c *ClusterClient) ClusterPipeline() *ClusterPipeline {
```
vmihailenco: Can we rename this to just `ClusterClient.Pipeline`? There is no need to repeat cluster twice...

cluster_pipeline.go
```diff
@@ -0,0 +1,130 @@
...
+
+	pipes := &clusterPipelines{make(map[string]*Pipeline), c.client}
+
+	// Assign commands to pipelines
+	for _, cmd := range cmds {
+		pipes.process(cmd)
+	}
+	pipes.exec()
+
+	// Detect MOVED errors, try again
```
vmihailenco: Please add a comment explaining why we don't handle `ASK` response here.

cluster_pipeline.go
```diff
@@ -0,0 +1,130 @@
...
+		if err := cmd.Err(); err != nil {
+			return cmds, err
+		}
+	}
+	return cmds, nil
+}
+
+// ------------------------------------------------------------------------
+
+type clusterPipelines struct {
```
vmihailenco: Can we merge this helper class into `ClusterPipeline`? Its usefulness is doubtful...

vmihailenco:
I think we should handle ask/moved in the `Pipeline`. @dim what do you think?

vmihailenco:
Hm, we can't move ask/moved to the Pipeline, because we don't have access to the slots metadata there.

cluster_pipeline.go
dim: I don't mind, we had a originally a `ClusterClient.Pipeline(key)` which was creating a simple pipeline against a single node, happy to rename.

cluster_pipeline.go
dim: I didn't want to assign this to the pipeline directly as this only lives on Exec and is much easier to test on its own.

dim:
@vmihailenco please check, I'd prefer if you could merge this soon(ish)

vmihailenco:
I have simpler approach to ClusterPipeline, but I am waiting for https://github.com/go-redis/redis/pull/82 to get merged. Sorry that I did not let you know.

vmihailenco:
@dim PTAL.

cluster_pipeline.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (c *ClusterPipeline) Exec() (cmds []Cmder, retErr error) {
+	if c.closed {
+		return nil, errClosed
+	}
+
+	cmds = c.cmds
+	c.cmds = make([]Cmder, 0, 10)
+	if len(cmds) == 0 {
```
dim: superminor: you could avoid the (re-)allocation of c.cmds by changing the code to:

```go
if c.closed {
  return nil, errClosed
} else if len(c.cmds) == 0 {
  return []Cmder{}, nil
}
```

cluster_pipeline.go
```diff
@@ -0,0 +1,134 @@
...
+
+		var addr string
+		if len(addrs) > 0 {
+			addr = addrs[0] // First address is master.
+		}
+
+		cmdsMap[addr] = append(cmdsMap[addr], cmd)
+	}
+
+	for attempt := 0; attempt < c.cluster.opt.getMaxRedirects(); attempt++ {
```
dim: In this case, it should be `attempt <= c.cluster.opt.getMaxRedirects()`.

cluster_pipeline.go
```diff
@@ -0,0 +1,134 @@
...
+		failedCmds := make(map[string][]Cmder)
+
+		for addr, cmds := range cmdsMap {
+			var client *Client
+			if addr != "" {
+				client = c.cluster.getClient(addr)
```
dim: Should we make `getClient("")` return `randomClient()`?

dim: Or at least move that block into a helper method. We are using a very similar logic in `ClusterClient.process`

dim:
That's it, not much to add. Looks good otherwise :)

vmihailenco:
@dim addressed all comments. Merging.

PR #80, opened 2015-03-30
Better race protection

dim:


cluster.go
```diff
@@ -16,11 +16,13 @@ type ClusterClient struct {
 
 	addrs map[string]struct{}
 	slots [][]string
+	sLock sync.RWMutex // protects slots & addrs cache
```
vmihailenco: Can you please use full name, e.g. `slotsLock`? Difference between `sLock` and `cLock` is too subtle.

Also the usual name is `slotsMutex` or `slotsMtx` to avoid double "lock" word in `theLock.Lock`.

dim:
Updated

vmihailenco:
Thanks.

PR #82, opened 2015-04-04
Attempt to cleanup cluster logic.

vmihailenco:
@dim I tried to refactor code a bit to learn more about Redis cluster. Changes:
- `NewClusterClient` does not return error any more, because `NewClient` does not too. I personally think that app can't do anything useful except exiting when `NewClusterClient` returns an error. So panic should be a good alternative.
- Now `ClusterClient.process` tries next available replica before falling back to the `randomClient`. I am not sure that this change is correct, but I hope so :)
- `randomClient` is completely rewritten so it does not require allocating `seen map[string]struct{}{}` on every request. It also checks that node is online before returning.


dim:
So I am not sure about this one, at least not in its entirety.

> NewClusterClient does not return error any more, because NewClient does not too.

Fully agree, I already wanted to address this, but was waiting for #77 to be merged. Should we address this in a separate PR?

> Now ClusterClient.process tries next available replica before falling back to the randomClient

While this seems to make sense, I don't think it's right. In general, I tried to follow the instructions in http://redis.io/topics/cluster-spec as strictly as possible.

> randomClient is completely rewritten so it does not require allocating ...

Cool, will review and comment inline. Could you please move this into a separate PR?

cluster.go
```diff
@@ -28,94 +40,119 @@ type ClusterClient struct {
...
 	}
-	return ""
-}
+	c.clientsMx.RUnlock()
 
-// Returns a node's client for a given address
-func (c *ClusterClient) getNodeClientByAddr(addr string) *Client {
-	c.connsMx.Lock()
-	client, ok := c.conns[addr]
+	c.clientsMx.Lock()
```
dim: There's (technically) a little race condition here, but I doubt it will actually matter

cluster.go
```diff
@@ -28,94 +40,119 @@ type ClusterClient struct {
...
 		}
 
-		// On connection errors, pick a random, previosuly untried connection
-		// and request again.
-		if _, ok := err.(*net.OpError); ok || err == io.EOF {
-			if addr = c.findNextAddr(tried); addr == "" {
-				return
+		// On network errors try another node.
+		if isNetworkError(err) {
+			if replicaIndex < len(addrs) {
```
dim: Again, I wouldn't worry about this.

cluster.go
```diff
@@ -28,94 +40,119 @@ type ClusterClient struct {
...
 }
 
-// Close closes the cluster connection
+// Close closes the cluster client.
 func (c *ClusterClient) Close() error {
-	c.slotsMx.Lock()
-	defer c.slotsMx.Unlock()
-
-	return c.reset()
+	// TODO: close should make client unusable
```
dim: Don't you want to call `resetClients` here? Otherwise, you might leave open connections behind, or am I missing something here?

cluster.go
```diff
@@ -142,88 +179,55 @@ func (c *ClusterClient) process(cmd Cmder) {
...
-	// https://github.com/antirez/redis-rb-cluster/blob/fd931ed/cluster.rb#L157
-	for addr := range c.addrs {
-		c.reset()
-
-		infos, err = c.fetchClusterSlots(addr)
-		if err == nil {
-			c.update(infos)
-			break
-		}
+	client := c.randomClient()
```
dim: I followed antirez' reference implementation here, which suggests to iterate over all known cluster addresses in random order and reset before each step. What if you have no more (working) clients left? It will panic.

cluster.go
vmihailenco: I don't see race here... Can you please point me to it?

cluster.go
vmihailenco: Are replicas read-only? Otherwise I don't see reasons to not use them... AFAIK `randomClient` can pick replica too...

cluster.go
vmihailenco: I will add check for `nil`, but `randomClient` will try 10 random nodes before returning `nil`. So I think there is no degradation here.

cluster.go
vmihailenco: Done.

cluster.go
vmihailenco: Me research shows that replicas are read-only (just a plain Redis slave) and Ruby Redis Cluster client does not use them at all: https://github.com/antirez/redis-rb-cluster/blob/fd931ed/cluster.rb#L64 . @dim do you agree? Meanwhile I will update this PR to ignore replicas.

vmihailenco:
@dim PTAL.

cluster.go
```diff
@@ -28,94 +25,108 @@ type ClusterClient struct {
...
-	addrs, err := opt.getAddrSet()
-	if err != nil {
-		return nil, err
-	}
-
+func NewClusterClient(opt *ClusterOptions) *ClusterClient {
 	client := &ClusterClient{
-		addrs:   addrs,
-		conns:   make(map[string]*Client),
+		addrs:   opt.getAddrs(),
```
dim: Do we still need `getAddrs`?

cluster.go
dim: Sorry, ignore, all good, missed a line when I was looking before

cluster.go
```diff
@@ -28,94 +25,108 @@ type ClusterClient struct {
...
+		c.clients[addr] = client
 	}
-	c.connsMx.Unlock()
+	c.clientsMx.Unlock()
+
 	return client
 }
 
+// randomClient returns a Client for the first live node.
+func (c *ClusterClient) randomClient() (client *Client, err error) {
```
dim: What's the advantage of this, compared to storing addrs as a `map[string]struct{}`. Instead of having N attempts, trying to pick a random address (why 10?), it seems cleaner to simply iterate over the addrs map, don't you think?

cluster.go
```diff
@@ -143,82 +154,65 @@ func (c *ClusterClient) process(cmd Cmder) {
...
+func (c *ClusterClient) setSlots(slots []ClusterSlotInfo) {
 	c.slotsMx.Lock()
-	defer c.slotsMx.Unlock()
 
-	// Try known addresses in random order (map interation order is random in Go)
-	// http://redis.io/topics/cluster-spec#clients-first-connection-and-handling-of-redirections
-	// https://github.com/antirez/redis-rb-cluster/blob/fd931ed/cluster.rb#L157
-	for addr := range c.addrs {
-		c.reset()
+	c.addrs = c.addrs[:0]
```
dim: I remember this being discouraged by the cluster spec. You should always remember all known addresses. The edge case is `slots` being empty because of a broken or detached node. Then, you would have nothing to fall back on.

cluster_client_test.go
```diff
@@ -1,95 +0,0 @@
-package redis
-
-import (
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-)
-
-var _ = Describe("ClusterClient", func() {
```
dim: This test was actually quite useful to get the cluster basics right. I have no 'personal attachment' but keeping it in some way would probably improve my sleep :)

cluster.go
vmihailenco: I really think that you abuse implementation details of the language. I would even say that randomness of the iteration order of the maps is much weaker compared to `math/rand` and is not specified anywhere.

cluster.go
vmihailenco: No, removing.

cluster.go
vmihailenco: Ok, reverting. I just feel uncomfortable that `randomClient` will use nodes that are not in the cluster any more.

vmihailenco:
@dim PTAL

dim:
So yes, much better, the only remaining 'issue' for me is the choice of `10` for the randomClient attempts. Not that we retain all addrs forever again, there could be quite a few legacy ones (especially in cloud environments). Trying max 10x may just be too simple, that's why I preferred the iteration approach, but I understand your concern about randomness. The alternative is to shuffle the list of addrs, like I did in my first attempt: https://github.com/bsm/redis-cluster/blob/master/cluster.go#L175

vmihailenco:
How shuffle solves your concern?

We can increase 10 to 100, but that is too much to my taste

dim:
First shuffle, then iterate until you find the first working one. Alternatively, instead of hard-coding 10 or 100, you could use a probabilistic approach and set the number e.g. to `len(c.addrs)` or `len(c.addrs)*2`.

vmihailenco:
I think that number of random nodes we try should be fixed. I doubt that we want to try 100 or 1000 random nodes, because that can be slow and creates a lot of bad connections with `CLOSING` state so soon whole app will be unresponsible.

I honestly don't understand why we can't remove dead nodes from the list. They will be added back on the next `scheduleReload` if they come online. If needed we can schedule reload every 5-10 minutes in a separate goroutine.

vmihailenco:
@dim I am waiting for your LGTM or objections :)

dim:
> I honestly don't understand why we can't remove dead nodes from the list.

Because you could end up with no nodes at all. Even if you tracked the downtime and evicted the ones that are failing persistently, there is a chance that they might come back online. I know of various cloud setups where this could be a real problem.

> They will be added back on the next scheduleReload if they come online

Not if you wipe them from c.addrs. Assuming we only keep the addrs from `CLUSTER SLOTS` and drop all previous seeds, what if you reshard/failover all your nodes to e.g. a different zone? None of your known addrs would work anymore.

> I think that number of random nodes we try should be fixed

How about this: 
```go
// randomClient returns a Client for the first live node.
func (c *ClusterClient) randomClient() (client *Client, err error) {
  for i, addr := range c.addrs {
    client = c.getClient(c.addrs[n])
    err = client.Ping().Err()
    if err == nil {
      c.addrs[0], c.addrs[i] = c.addrs[i], c.addrs[0]
      return client, nil
    }    
  }
  return nil, err
}
```

It would ensure that we push the known-to-be-good ones to the top of the adds queue.

vmihailenco:
Your code is not random any more and therefore it does not scale, because there can be situation when all clients hit the same very first node.

>drop all previous seeds

I suggest only drop unreachable nodes, i.e. those that don't response to PING command.

>there is a chance that they might come back online

Client will eventually reload slot info. We can also reload it every 1 or 5 minutes.

dim:
I can see some edge cases, but I don't have a strong opinion here, as they may never trigger in practice. Dropping unreachable nodes should be ok, but I woudn't do it in `randomClient` but in `setSlots`, i.e. replace `c.addrs` there but only if we have a new list of addrs, ie:

```go
func (c *ClusterClient) setSlots(slots []ClusterSlotInfo) {
  c.slotsMx.Lock()

  c.slots = make([][]string, hashSlots)
  c.resetClients()

  addrs := make([]string, len(c.addrs))
  for _, info := range slots {
    for slot := info.Start; slot <= info.End; slot++ {
      c.slots[slot] = info.Addrs
    }

    for _, addr := range info.Addrs {
      if _, ok := seen[addr]; !ok {
        addrs = append(addrs, addr)
        seen[addr] = struct{}{}
      }
    }
  }

  if len(addrs) != 0 {
    c.addrs = addrs
  }
  c.slotsMx.Unlock()
}
```

vmihailenco:
The code you pasted drops unknown nodes (i.e. nodes that are not in the cluster). By unreachable I mean nodes that did not respond to PING command.

AFAIK the only concern you have is about 10 retries in `randomClient`. I think that 10 retries is enough and more importantly safe. I agree with you that it does not work when you have 100 nodes down, but I doubt that adding more attempts will fix it. In such case we should monitor state of the nodes and we can do it in a separate PR. Sounds good?

dim:
Sorry, let's rewind, there are two separate concerns here:

1. `10` retries are probably enough if we only remembered working addrs, but it may not be enough if we keep a full node history. It's really not a big deal at this pint, but redis cluster was built with thousands of nodes in mind and if we decided to keep a full history of `addrs`, we should make the number dynamic and base it on a fraction/multiple of `len(addrs)`.

2. The bigger topic here is: should we removing broken addrs or not? I am generally in favour, but I don't believe we should remove nodes from the list of `c.addrs` in `randomClient`, when a `Ping()` fails. Let's assume the client is disconnected from the network, a call to `randomClient` would instantly mark 10 addrs as down and remove them. We will end up with an empty `c.addrs` in virtually no time, even tiny downtimes would make our cluster client unusable.

vmihailenco:
1. It probably dependends on your usage pattern. I personally think that it is better to fail fast rather then check 100 nodes, because I mainly use Redis as cache and shared counters. I will not mind if you make this option configurable, but I don't want to make it depend on number of nodes in cluster.

As I said previously if cluster is down and every your Redis request checks all 100 nodes in the Cluster you will end up with unresponsible app, because it will eat all resources (at least it will open too many connections).

2. We can preserve original list of addresses and fallback to it when all nodes are down. Or we can maintain 2 lists of up and down nodes. Anyway this PR does not remove nodes from the list, so this concern can be addressed separately.

dim:
Well, merge it then, it's functional and we can submit a separate PRs to address remaining issues

PR #83, opened 2015-04-05
Add reaper that closes idle connections to the cluster.

vmihailenco:
@dim what do you think?


dim:
Cool, will review tomorrow.

dim:
Sorry, reviewed already, just to clarify: what's the main intention here? To reduce the number of current connections? In that case, looks good, feel free to merge.

vmihailenco:
Yes, the case is to reduce number of idle connections. This is very useful for PostgreSQL connection pool, but should be moderately useful for Redis cluster too.

Issue #86, opened 2015-04-13
Cluster pipeline panic


dim:
After plugging the new master into a real-world app, I get these:

```
2015/04/13 12:34:20 http: panic serving 127.0.0.1:47736: runtime error: index out of range
goroutine 7 [running]:
net/http.func·011()
        $GOROOT/src/net/http/server.go:1130 +0xbb
gopkg.in/redis%2ev2.(*ClusterClient).slotAddrs(0xc208096000, 0x13e8, 0x0, 0x0, 0x0)
        $GOPATH/src/gopkg.in/redis.v2/cluster.go:78 +0xee
gopkg.in/redis%2ev2.(*ClusterPipeline).Exec(0xc2086153e0, 0xc208746000, 0x2710, 0x3000, 0x0, 0x0)
        $GOPATH/src/gopkg.in/redis.v2/cluster_pipeline.go:56 +0x2b8
```

Will debug too.

dim:
After a restart, I am now getting tons of: `redis: you open connections too fast (last error: nil)`.

vmihailenco:
As a quick fix we should put `reloadIfDue` back to the `NewClusterClient` so `slots` are initialized.

dim:
I'm not really sure what is going on, trying to trace it down. None of the issues were there before :(

vmihailenco:
1. first request comes and sees that `_reload = 1`
2. reloadIfDue is called and sets `_reload = 0`
3. second request comes and sees that `_reload = 0` so it tries to access `slots`. But `reloadIfDue` has not finished its work yet, so `slots` is empty

dim:
Sounds like it's best solved with a mutex, but that doesn't solve the `you open connections too fast` error. Is it right to call `freeConn` in the pipeline's Exec?

dim:
Or, you could just call `setSlots(nil)` in the constructor.

vmihailenco:
Your original version had `reloadIfDue` call in constructor. So `ClusterClient` never had uninitialized `slots`. I think that is a good quick fix.

`setSlots(nil)` will not help, because it resets slots.

dim:
`setSlots(nil)` initializes c.slots so we will never end up with an `index out of range` but `reloadIfDue` sounds good to me too. Either way, what about `you open connections too fast` issue?

dim:
[...]

PR #87, opened 2015-04-13
Ensure slots are initialised. Return non-failing connections to pool

dim:
Fixes for #86

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -28,11 +28,14 @@ type ClusterClient struct {
 func NewClusterClient(opt *ClusterOptions) *ClusterClient {
 	client := &ClusterClient{
 		addrs:   opt.Addrs,
+		slots:   make([][]string, hashSlots),
 		clients: make(map[string]*Client),
 		opt:     opt,
 		_reload: 1,
 	}
 	client.commandable.process = client.process
+	client.reloadIfDue()
+
 	go client.reaper(time.NewTicker(5 * time.Minute))
 	return client
 }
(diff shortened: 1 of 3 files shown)


cluster.go
```diff
@@ -176,14 +179,15 @@ func (c *ClusterClient) resetClients() (err error) {
...
 
-	c.slots = make([][]string, hashSlots)
 	c.resetClients()
-
 	seen := make(map[string]struct{})
 	for _, addr := range c.addrs {
 		seen[addr] = struct{}{}
 	}
 
+	for i := 0; i < hashSlots; i++ {
```
vmihailenco: This should be much more slower than `make([][]string, hashSlots)`. But I don't insist.

vmihailenco:
LGTM

cluster.go
dim: You would be surprised:

```go
package x

import "testing"

func BenchmarkMake(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = make([][]string, 20000)
	}
}

func BenchmarkUnset(b *testing.B) {
	x := make([][]string, 20000)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for j := 0; j < 20000; j++ {
			x[j] = x[j][:0]
		}
	}
}
```

```bash
PASS
BenchmarkMake	    5000	    234386 ns/op
BenchmarkUnset	   50000	     35669 ns/op
```

vmihailenco: Ha, sure I am wrong. `x[j][:0]` just sets slice length, but `make([][]string, 20000)` allocates new slice. No surprise here :)

dim: Cool, trying to avoid allocations where possible, old habit :)

PR #89, opened 2015-04-17
Implement non-blocking connection pool

dim:
We are seeing quite substantial latency issues with the current pool implementation. The main reason is that connections are not always returned back to the pool and `pool.wait` might wait for a connection that will never appear, eventually timing out after `PoolTimeout`. Example:

1. `pool.Get()` is called
2. `pool.First()` has no free connection available
3. `pool.size` is max, so resume by `pool.wait()`
4. Previously checked out connections are removed via `pool.Remove()`, `pool.size` is decremented
5. `pool.wait()` still waits for a connection until it times out

This patch implements a non-blocking pool and removes PoolTimeout. If no idle connection is available, we don't `wait` but go straight to `new`. Pools are not allowed to over-commit themselves, which seems an OK compromise given that we are rate-limiting new connections anyway.

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -17,8 +17,7 @@ var _ = Describe("Commands", func() {
 
 	BeforeEach(func() {
 		client = redis.NewTCPClient(&redis.Options{
-			Addr:        redisAddr,
-			PoolTimeout: 30 * time.Second,
+			Addr: redisAddr,
 		})
 	})
 
--- a/pipeline_test.go
+++ b/pipeline_test.go
@@ -1,6 +1,7 @@
 package redis_test
 
 import (
+	"fmt"
 	"strconv"
 	"sync"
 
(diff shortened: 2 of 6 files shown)


vmihailenco:
Looks like this implementation makes it possible to have 1000 open connections with `pool_size = 10`. It does not looks good and rate-limiting makes situation even worse.

I think we should either:
1. revert https://github.com/go-redis/redis/pull/70
2. rewrite Remove to open new connection and send it to the `freeConns` channel. `pool.Close` should be updated accordingly.

What do you think?

vmihailenco:
FYI I am preparing PR with fix described in 2nd item.

dim:
How will ratelimiter make it worse, it's set to 2*PoolSize per second? You cannot easily open a 1000 connections without hitting it. 

Pushing a new connection into `freeConns` on `Remove` is doable but I am a bit afraid of the added complexity and possible edge cases. My PR seems a reasonable compromise. I also searched github for pool implementation and the one that seems most 'battle-tested' is https://github.com/fatih/pool. It also uses the same algorithm.

vmihailenco:
>My PR seems a reasonable compromise

Your new pool returns `errRateLimited` error when old pool would process requests as expected. That does not look as a compromise to me.

Issue #88, opened 2015-04-16
Increased frequency of pool.Put failed: redis: client is closed


dim:
Since last revision of cluster-client was deployed, we see many, many `pool.Put failed: redis: client is closed` in the logs, any idea why this could be?

vmihailenco:
That was a bug in the pool:
1. client is closed
2. operation finishes and connection is returned to the pool
3. pool says that client is closed is logs the message you mentioned

Should be fixed by https://github.com/go-redis/redis/pull/90/files

dim:
Yep, seen it in the patch

PR #90, opened 2015-04-17
Fix pool waiting for connection that was removed.

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/88

@dim PTAL.

I also changed default `PoolTimeout` because 5 seconds it too much for Redis and usually indicates that you have to increase pool size.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -256,29 +256,13 @@ type ClusterOptions struct {
 	// giving up. Default: 16
 	MaxRedirects int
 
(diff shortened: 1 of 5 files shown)


dim:
OK, looks reasonable, will test it in the real world. Minor but imporatnt: could you please add PoolTimeout to ClusterOptions then, forgot about that one.

vmihailenco:
@dim done. PTAL.

dim:
LGTM

dim:
Looks good

PR #92, opened 2015-04-22
Ensure all new connections are initialised

dim:
We have noticed that `Password` and `DB` settings are not correctly applied to connections being created as part of `pool.Put`, here's a proposed solution. PTAL

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -178,8 +178,8 @@ func (p *connPool) new() (*conn, error) {
 		return nil, err
 	}
 
-	if err := p.init(cn); err != nil {
-		p.Remove(cn)
+	if err := p.initConn(cn); err != nil {
+		cn.Close()
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


pool.go
```diff
@@ -171,23 +171,59 @@ func (p *connPool) new() (*conn, error) {
...
+	if err := p.init(cn); err != nil {
+		p.Remove(cn)
+		return nil, err
+	}
+
+	return cn, nil
+}
+
+// Initialize connection
+func (p *connPool) init(cn *conn) error {
```
vmihailenco: Please rename to `initConn`. Just `init` should initialize pool, not connection.

pool.go
```diff
@@ -171,23 +171,59 @@ func (p *connPool) new() (*conn, error) {
...
+
 	cn, err := p.dial()
 	if err != nil {
 		p.lastDialErr = err
+		return nil, err
 	}
-	return cn, err
+
+	if err := p.init(cn); err != nil {
+		p.Remove(cn)
```
vmihailenco: This looks wrong and dangerous. Please remove.

dim: `p.Remove` just closes the connection, do you prefer to `cn.Close()` instead?

vmihailenco: Yes. `Remove` also opens new connection, so we can have an infinitive loop here.

dim:
Updated, good to merge?

vmihailenco:
LGTM. Please merge.

PR #93, opened 2015-04-28
cluster: don't reset clients on MOVED.

vmihailenco:
I think there no need to reset clients, because we have connection reaper that will close connections and associated resources if client is idle. @dim ?

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -35,7 +35,6 @@ func NewClusterClient(opt *ClusterOptions) *ClusterClient {
 	}
 	client.commandable.process = client.process
 	client.reloadIfDue()
-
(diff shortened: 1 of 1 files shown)


dim:
Seems OK, but what's the rationale? Do you want to reduce latency spikes due to `MOVED`? Also, I think our reaper is leaking goroutines, we should have a closer channel which triggers on `Close`.

vmihailenco:
>Do you want to reduce latency spikes due to MOVED?

Yes. Also it just seems unneeded to close every client on MOVED.

>we should have a closer channel which triggers on Close

Will do in a separate PR.

vmihailenco:
Good to merge?

dim:
Sure, go ahead

PR #94, opened 2015-04-28
Add very basic cluster benchmark. Cleanup other benchmarks.

vmihailenco:
```
BenchmarkRedisClusterPing	   30000	     39906 ns/op
BenchmarkRedisPing	   50000	     33728 ns/op
```


dim:
:ok_hand:

PR #95, opened 2015-05-01
Implement Close and fix reaper goroutine leak.

vmihailenco:
@dim please review

Also do you any plans how to release v3 with latest API changes and Cluster client?

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -241,7 +241,7 @@ func (c *ClusterClient) scheduleReload() {
 
 // reaper closes idle connections to the cluster.
 func (c *ClusterClient) reaper() {
-	ticker := time.NewTicker(5 * time.Minute)
+	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
 	for _ = range ticker.C {
 		c.clientsMx.RLock()
--- a/cluster.go
+++ b/cluster.go
@@ -13,10 +13,11 @@ type ClusterClient struct {
 
 	addrs   []string
 	slots   [][]string
-	slotsMx sync.RWMutex // protects slots & addrs cache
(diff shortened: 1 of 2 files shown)


dim:
You are not unlocking clientMx when c.closed on two occasions. Sometimes it's worth accepting the additional allocation and simply deferring the Unlock to improve code readability, I think

dim:
I also think you should cache the ticker on the client instance and `close(c.ticker.C)` when closing the client to make sure the reaper goroutine doesn't linger around unnecessarily.

vmihailenco:
@dim thanks, fixed. PTAL.

vmihailenco:
PTAL

cluster.go
```diff
@@ -35,20 +36,27 @@ func NewClusterClient(opt *ClusterOptions) *ClusterClient {
...
 
-// Close closes the cluster client.
+// Close closes the cluster client, releasing any open resources.
+//
+// It is rare to Close a Client, as the Client is meant to be
+// long-lived and shared between many goroutines.
 func (c *ClusterClient) Close() error {
-	// TODO: close should make client unusable
-	c.setSlots(nil)
+	defer c.clientsMx.Unlock()
```
dim: Not a problem, just out of interest: what's the benefit of deferring the unlock before acquiring the lock? I've seen you doing it on a few occasions. Looks like no one else is doing it this way 'round.

dim:
Cool, ship it! :)

cluster.go
vmihailenco: It should be slightly faster (I believe), because `defer c.clientsMx.Unlock()` is evaluated before the lock is acquired. It does not matter too much and I have to stop doing this.

PR #96, opened 2015-05-01
Reload slots in background goroutine.

vmihailenco:
Since this is Go client we can use goroutine to reload slots info in background.

@dim ptal


vmihailenco:
@dim ping in case you missed this

dim:
Looks OK, will test it in the wild once merged

PR #97, opened 2015-05-02
Fix pool to close all connections when client is closed.

vmihailenco:
This is very useful when you want application to exit gracefully. Existing `BenchmarkPool` numbers are not changed.

Pool was refactored a bit since I want to reuse it in https://github.com/go-pg/pg/pull/52

/cc @dim

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -4,13 +4,11 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"net"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"gopkg.in/bsm/ratelimit.v1"
-	"gopkg.in/bufio.v1"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pool.go
```diff
@@ -28,103 +26,133 @@ type pool interface {
...
-	buf := cn.buf[:0]
-	for _, cmd := range cmds {
-		buf = appendArgs(buf, cmd.args())
+func (l *connList) Len() int {
+	return int(atomic.LoadInt32(&l.len))
+}
+
+// Full reports whether list is full. If list is not full the caller must add or
+// remove connection from the pool.
+func (l *connList) Full() bool {
```
dim: Minor: how about calling this method `Reserve`? Seems like it would better reflect the purpose.

pool.go
```diff
@@ -259,49 +254,53 @@ func (p *connPool) Put(cn *conn) error {
...
+func (p *connPool) FreeLen() int {
+	return len(p.freeConns)
 }
 
-func (p *connPool) Close() (retErr error) {
-	if !atomic.CompareAndSwapInt32(&p.closed, 0, 1) {
-		return nil
+func (p *connPool) Close() error {
+	if !atomic.CompareAndSwapInt32(&p._closed, 0, 1) {
+		return errClosed
```
dim: I think I returned nil for a reason here, but I can't remember why. Should be OK though

dim:
Looked over briefly, can't see anything obvious: :+1:

pool.go
vmihailenco: Thanks, sounds much better.

PR #101, opened 2015-05-10
Add auto-retry and MaxRetries option. Fixes #84.

### LINKED ISSUE #84: Redis EOF after a period of time - probably due to timeout
Hello,

While running in production an app which uses go-redis library, we observed that the first request to Redis after a longer period of inactivity fails. It seems that the connection "dies" if unused with err = EOF.

What is the suggested mechanism to re-connect transparently and to prevent failure for the failing initial request? Can the library auto-reconnect in case of failed connection?

The code is super-simple. It does a MGET command for several keys. Redis connection is initialized using:

```go
func InitRedis(host string, db int64) {
    client = redis.NewClient(&redis.Options{
        Network:"tcp4",
        Addr: host,
        DB: db,
    })
}

//request to get
  v, err := client.MGet("cp:"+date, "ob:" + date).Result()
    if(err!=nil || err == redis.Nil){
        log.Println("Error while getting Redis keys: ", err)
        panic(err)
    }

```

Thanks,
Robert

vmihailenco:
Things to consider when reviewing this PR:
- we have MaxRetries and MaxRedirects. Merge them?
- MultiClient and PubSubClient don't auto-retry
- Pipelines are auto-retried. Should we?

### DIFF
--- a/error.go
+++ b/error.go
@@ -26,7 +26,7 @@ func (err redisError) Error() string {
 }
 
 func isNetworkError(err error) bool {
-	if _, ok := err.(*net.OpError); ok || err == io.EOF {
+	if _, ok := err.(net.Error); ok || err == io.EOF {
 		return true
 	}
 	return false
@@ -52,4 +52,12 @@ func isMovedError(err error) (moved bool, ask bool, addr string) {
 	}
 
 	return
+}
+
+// shouldRetry reports whether failed command should be retried.
+func shouldRetry(err error) bool {
+	if err == nil {
+		return false
+	}
+	return isNetworkError(err)
 }
--- a/pipeline.go
+++ b/pipeline.go
@@ -50,26 +50,38 @@ func (c *Pipeline) Discard() error {
 
 // Exec always returns list of commands and error of the first failed
 // command if any.
-func (c *Pipeline) Exec() ([]Cmder, error) {
+func (c *Pipeline) Exec() (cmds []Cmder, retErr error) {
 	if c.closed {
 		return nil, errClosed
 	}
 	if len(c.cmds) == 0 {
-		return []Cmder{}, nil
+		return c.cmds, nil
 	}
 
-	cmds := c.cmds
+	cmds = c.cmds
 	c.cmds = make([]Cmder, 0, 0)
 
-	cn, err := c.client.conn()
-	if err != nil {
-		setCmdsErr(cmds, err)
-		return cmds, err
+	for i := 0; i <= c.client.opt.MaxRetries; i++ {
+		if i > 0 {
+			resetCmds(cmds)
+		}
+
+		cn, err := c.client.conn()
+		if err != nil {
+			setCmdsErr(cmds, err)
+			return cmds, err
+		}
+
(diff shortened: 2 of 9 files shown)


error.go
```diff
@@ -26,7 +26,7 @@ func (err redisError) Error() string {
 }
 
 func isNetworkError(err error) bool {
-	if _, ok := err.(*net.OpError); ok || err == io.EOF {
+	if _, ok := err.(net.Error); ok || err == io.EOF {
```
vmihailenco: @dim was there a specific reason to use `*net.OpError`?

dim: Not really, `net.Error` seems a better fit

redis.go
```diff
@@ -105,6 +119,10 @@ type Options struct {
 	// than specified in this option.
 	// Default: 0 = no eviction
 	IdleTimeout time.Duration
+
+	// MaxRetries specifies maximum number of times client will retry
+	// failed command. Default is to not retry failed command.
+	MaxRetries int
```
dim: MaxRetries and MaxRedirects are two different things really. We would e.g. never allow to set `MaxRetries` on ClusterOptions, right?

pipeline.go
```diff
@@ -50,26 +50,38 @@ func (c *Pipeline) Discard() error {
 
 // Exec always returns list of commands and error of the first failed
 // command if any.
-func (c *Pipeline) Exec() ([]Cmder, error) {
+func (c *Pipeline) Exec() (cmds []Cmder, retErr error) {
```
dim: I prefer your implmentation where pipeline is using MaxRetries too. Can you see any edge case where this would not be desirable?

vmihailenco: No, just wanted to hear another opinion.

redis.go
vmihailenco: Right, though `MaxRedirects` retries command on network errors too. Will update doc in a separate PR.

pipeline.go
dim: My worry is that `isNetworkError` only focusses on the kind of error but not if it occurred during request or response. There is a chance of a network disconnect while reading a response, which could still mean that the request succeeded. If you have e.g. 1M `INCR` commands in your pipeline, retrying suddenly becomes a dangerous business.

error.go
dim: Actually, not sure if this is even a concern, but would that include read/write timeouts? A tiny ReadTimeout could trigger every command to be retried although it actually succeeds.

dim:
After some careful thought, I wonder if retrying at this level is a good idea in the first place. What problem are we trying to solve? I thought it's all about avoiding EOFs on stale connections.

vmihailenco:
This PR is about retrying failed commands when possible. io.EOF is just a specific case.

I don't want to differentiate request and response errors until you have a good example of the lib that already does this. AFAIK there is no way to tell if connection is live/dead except using PING/PONG.

vmihailenco:
@dim I am merging this since retrying is disabled by default and we should start from something. Let me know if you still have any concerns about this PR.

PR #102, opened 2015-05-10
cluster: don't reset command when there no attempts left.

vmihailenco:
Newly added test fails without the fix.

/cc @dim


dim:
Well spotted

vmihailenco:
I guess you closed this by mistake. Reopening and merging.

PR #103, opened 2015-05-11
commands: Reduce memory allocations in ZAdd

abustany:
When adding lots of data in Redis, the allocations in ZAdd can put quite some pressure on the GC... Using the benchmark in https://gist.github.com/abustany/8f2cf2ee3b968020b75e , I get the following results:

```
benchmark              old ns/op     new ns/op     delta
BenchmarkRedisZAdd     22371         21779         -2.65%

benchmark              old allocs     new allocs     delta
BenchmarkRedisZAdd     10             9              -10.00%

benchmark              old bytes     new bytes     delta
BenchmarkRedisZAdd     373           343           -8.04%
```

There are probably a couple of places in the library where we could use a sync.Pool, for example for the commands, to reduce GC pressure, but that'd be the subject of another pull request...

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -870,9 +870,12 @@ type ZStore struct {
 }
 
 func (c *commandable) ZAdd(key string, members ...Z) *IntCmd {
-	args := []string{"ZADD", key}
-	for _, m := range members {
-		args = append(args, formatFloat(m.Score), m.Member)
+	args := make([]string, 2+2*len(members))
+	args[0] = "ZADD"
+	args[1] = key
+	for i, m := range members {
+		args[2+2*i] = formatFloat(m.Score)
+		args[2+2*i+1] = m.Member
 	}
 	cmd := NewIntCmd(args...)
 	c.Process(cmd)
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -870,9 +870,13 @@ type ZStore struct {
 }
 
 func (c *commandable) ZAdd(key string, members ...Z) *IntCmd {
-	args := []string{"ZADD", key}
-	for _, m := range members {
-		args = append(args, formatFloat(m.Score), m.Member)
+	args := make([]string, 2+2*len(members))
+	args[0] = "ZADD"
+	args[1] = key
+
```
vmihailenco: Remove space.

vmihailenco:
Looks good, but please add benchmark from https://gist.github.com/abustany/8f2cf2ee3b968020b75e to the redis_test.go file.

vmihailenco:
Thanks. I rebased your changes and created new PR https://github.com/go-redis/redis/pull/108. Closing this one.

PR #104, opened 2015-05-13
cluster: user ClusterInfo instead of Ping to find live node.

vmihailenco:
This check is much stricter since it verifies that node at least knows how to handle `CLUSTER INFO` command.


dim:
LGTM

PR #105, opened 2015-05-14
Release redis.v3 beta.

vmihailenco:
@dim is there anything you want to do before releasing v3 with ClusterClient?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,5 @@ install:
   - go get github.com/onsi/ginkgo
   - go get github.com/onsi/gomega
   - mkdir -p $HOME/gopath/src/gopkg.in
-  - mv $HOME/gopath/src/github.com/go-redis/redis $HOME/gopath/src/gopkg.in/redis.v2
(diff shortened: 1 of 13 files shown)


dim:
Only a `Readonly bool` on the cluster options, which is then applied on `initConn`

vmihailenco:
Looks like that can be added at any point and should not block v3. FYI I plan to tag v3 this weekend after I switch our internal projects to use new API.

epelc:
@vmihailenco Does v2 of this package work with redis 3.0?

vmihailenco:
Yes, it works. go-redis client version numeration has nothing to do with Redis Server version numeration. It is just an accident that they (will) match.

epelc:
@vmihailenco Thanks I got a few more questions. Are there some major changes in v3? Would you recommend using v2 or v3 if I was going to switch from redigo today?

vmihailenco:
I would say v3. It has nicer API and has Redis Cluster support.

PR #107, opened 2015-05-14
Remove PSetEx and SetEx. Set should be used instead.

vmihailenco:


TwinProduction:
Challenged by https://github.com/go-redis/redis/pull/1546

Issue #111, opened 2015-05-15
Are MULTI/EXEC transactions thread-safe?


albrow:
I was looking at the code that handles MULTI/EXEC. [This comment](https://github.com/go-redis/redis/blob/v2.3.2/multi.go#L10) seems to say that issuing transactions with MULTI/EXEC is not thread-safe. I.e., if you have multiple connections issuing transactions in different threads, it could result in commands being received in the wrong order (e.g. redis server receiving MULTI twice before receiving EXEC). I just wanted to make sure. Is the comment still correct?

justonia:
Pretty sure that just means that you can't share the multi object itself between concurrent goroutines. As far as I'm aware, under the hood that multi reserves a single connection for it's use and thus cannot collide with any other goroutine using the redis connection pool.

vmihailenco:
What @justonia said.

I don't mind to make MULTI/EXEC thread-safe (goroutine-safe?) if there is a real use case. But I can't imagine any.

albrow:
Okay. That makes complete sense, thank you. I can't imagine a use-case where you would need to share a Multi object between multiple goroutines.

PR #112, opened 2015-05-18
multi: remove confusing comment.

vmihailenco:
Remove comment since there are no use cases when `Multi` can be used from different goroutines. Should help with https://github.com/go-redis/redis/issues/111


dim:
OK

PR #117, opened 2015-05-25
Add Redis Ring.

vmihailenco:
I called it Ring, because client uses consistent hash ring to distribute keys.

@dim would be nice to hear you opinion.

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -1,0 +1,237 @@
+package redis
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	"github.com/golang/groupcache/consistenthash"
+)
+
+var (
+	errRingShardsDown = errors.New("redis: all ring shards are down")
+)
+
+// RingOptions are used to configure a ring client and should be
+// passed to NewRing.
+type RingOptions struct {
+	// A map of name => host:port addresses of ring shards.
+	Addrs map[string]string
+
+	// Following options are copied from Options struct.
+
+	DB       int64
+	Password string
+
+	DialTimeout  time.Duration
+	ReadTimeout  time.Duration
+	WriteTimeout time.Duration
+
+	PoolSize    int
+	PoolTimeout time.Duration
+	IdleTimeout time.Duration
+}
+
+func (opt *RingOptions) clientOptions() *Options {
+	return &Options{
+		DB:       opt.DB,
+		Password: opt.Password,
+
+		DialTimeout:  opt.DialTimeout,
+		ReadTimeout:  opt.ReadTimeout,
+		WriteTimeout: opt.WriteTimeout,
+
+		PoolSize:    opt.PoolSize,
+		PoolTimeout: opt.PoolTimeout,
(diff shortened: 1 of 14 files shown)


dim:
@vmihailenco will try to look through today/tomorrow. I get the idea, but wouldn't it make more sense to wrap this into a separate package/repo? It's not a core redis concept, especially not since cluster support was added. It also introduces a new dependency, which is not the end of the world, but a new package seems just cleaner overall.

vmihailenco:
I am fine with separate package, but that requires exporting `process` method and doing something about `cl.baseClient.process` ambiguity.

vmihailenco:
@dim ping.

Even though `process` is already exported I would like to keep Ring in this repo. Probably let's move it to subpackage?

vmihailenco:
Moving to subpackage requires exporting `Commandable` struct, which I don't want to do (we did not have that struct in v2 and there are chances we will not have it in v4). Going to merge this as is tomorrow if I don't hear anything from you :/

ring.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		cmd.setErr(err)
+		return
+	}
+	cl.baseClient.process(cmd)
+}
+
+// rebalance removes dead shards from the ring.
+func (ring *Ring) rebalance() {
+	defer ring.mx.Unlock()
```
dim: Again :)

ring.go
```diff
@@ -0,0 +1,237 @@
...
+
+		PoolSize:    opt.PoolSize,
+		PoolTimeout: opt.PoolTimeout,
+		IdleTimeout: opt.IdleTimeout,
+	}
+}
+
+type ringShard struct {
+	Client *Client
+	down   int
```
dim: Checked briefly, looks OK. ringShards don't need to be thread-safe since they are guarded by `mx` in Ring

ring.go
```diff
@@ -0,0 +1,237 @@
+package redis
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	"github.com/golang/groupcache/consistenthash"
```
dim: Any change you could copy and paste this across? Looks like it's short enough to be embedded. I prefer not to have external dependencies, especially when it comes to hashing.

dim:
Really sorry about the delay, feel free to merge. The only (minor) concerns I have are:

1. importing hashing algorithms seems risky, I would prefer to embed it to be *really* consistent
2. how relevant are ring implementations, given that redis cluster is now available?

ring.go
vmihailenco: Will do.

vmihailenco:
>how relevant are ring implementations, given that redis cluster is now available?

Redis Cluster requires using replication and slaves so it can do failover when master dies. That is a bit expensive for us atm. So it is relevant at least for me :)

dim:
Fair point

vmihailenco:
Done. Package `consistenthash` was copied to internal folder.

PR #118, opened 2015-05-28
Allow setting and scaning interface{} values.

vmihailenco:
Should make client a little bit more convenient to use. @dim?

Fixes https://github.com/go-redis/redis/issues/121. Before

```
BenchmarkRedisBytes	   10000	    120101 ns/op	   52809 B/op	      16 allocs/op
BenchmarkRedisSet	   30000	     45606 ns/op	   10666 B/op	       7 allocs/op
```

After

```
BenchmarkRedisBytes	   10000	    108434 ns/op	   32022 B/op	      21 allocs/op
BenchmarkRedisSet	   30000	     46389 ns/op	   10762 B/op	      10 allocs/op
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -1,6 +1,7 @@
 package redis
 
 import (
+	"bytes"
 	"fmt"
 	"strconv"
 	"strings"
@@ -28,7 +29,7 @@ var (
 )
 
 type Cmder interface {
-	args() []string
+	args() []interface{}
 	parseReply(*bufio.Reader) error
 	setErr(error)
 	reset()
@@ -38,7 +39,7 @@ type Cmder interface {
 	clusterKey() string
 
 	Err() error
-	String() string
+	fmt.Stringer
 }
 
 func setCmdsErr(cmds []Cmder, e error) {
@@ -54,12 +55,21 @@ func resetCmds(cmds []Cmder) {
 }
 
 func cmdString(cmd Cmder, val interface{}) string {
-	s := strings.Join(cmd.args(), " ")
+	var ss []string
+	for _, arg := range cmd.args() {
+		ss = append(ss, fmt.Sprint(arg))
+	}
+	s := strings.Join(ss, " ")
 	if err := cmd.Err(); err != nil {
 		return s + ": " + err.Error()
 	}
 	if val != nil {
-		return s + ": " + fmt.Sprint(val)
+		switch vv := val.(type) {
+		case []byte:
+			return s + ": " + string(vv)
+		default:
+			return s + ": " + fmt.Sprint(val)
+		}
 	}
 	return s
 
@@ -68,7 +78,7 @@ func cmdString(cmd Cmder, val interface{}) string {
 //------------------------------------------------------------------------------
 
 type baseCmd struct {
-	_args []string
+	_args []interface{}
 
 	err error
 
@@ -84,7 +94,7 @@ func (cmd *baseCmd) Err() error {
 	return nil
 }
 
-func (cmd *baseCmd) args() []string {
+func (cmd *baseCmd) args() []interface{} {
 	return cmd._args
 }
 
@@ -102,7 +112,7 @@ func (cmd *baseCmd) writeTimeout() *time.Duration {
 
 func (cmd *baseCmd) clusterKey() string {
 	if cmd._clusterKeyPos > 0 && cmd._clusterKeyPos < len(cmd._args) {
-		return cmd._args[cmd._clusterKeyPos]
+		return fmt.Sprint(cmd._args[cmd._clusterKeyPos])
 	}
 	return ""
 }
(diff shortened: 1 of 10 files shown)


vmihailenco:
@dim ping :) I am not worried about the code, but more about approach in general.

command.go
```diff
@@ -411,6 +411,102 @@ func (cmd *StringCmd) Float64() (float64, error) {
 	return strconv.ParseFloat(cmd.val, 64)
 }
 
+func (cmd *StringCmd) Scan(vv interface{}) error {
```
dim: Don't you want to declare `(err error)`, would simplify the code below, no need for declarations, returns, etc

commands.go
```diff
@@ -54,6 +50,61 @@ func formatSec(dur time.Duration) string {
...
+	case float64:
+		return formatFloat(v), nil
+	case bool:
+		if v {
+			return "1", nil
+		} else {
+			return "0", nil
+		}
+	default:
+		b, err := Marshal(vv)
```
dim: I would probably define a `type Encodable interface { EncodeRedis() (string, error) }` and return an error on `default`

command.go
```diff
@@ -411,6 +411,102 @@ func (cmd *StringCmd) Float64() (float64, error) {
...
+		return err
+	case *float64:
+		var err error
+		*v, err = strconv.ParseFloat(cmd.val, 64)
+		return err
+	case *bool:
+		*v = cmd.val == "1"
+		return nil
+	default:
+		return Unmarshal([]byte(cmd.val), v)
```
dim: I would probably define a `type Decodable interface { DecodeRedis(string, interface{}) (error) }` and return an error on `default`

redis.go
```diff
@@ -1,12 +1,22 @@
...
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"net"
 	"time"
 )
 
+var (
```
dim: I don't think this is flexible enough, I like the approach of e.g. http://godoc.org/github.com/eaigner/jet#ComplexValue

command.go
vmihailenco: What use-case you have in mind?

I think most users will not invent special encoding to be used with Redis client. Most likely they will stick with well-known formats like json, gob, protobuf or msgpack.

Besides that requiring users to implement yet another `Decodable` interface on their types seems to be too tough... And most likely implementation will look like:

```
func DecodeRedis(s string, v interface{}) error {
    return msgpack.Unmrshal([]byte(s), v)
}
```

command.go
vmihailenco: I think it is 50/50. E.g. it will simplify first 4 cases, but in the 5-10th cases I will have to define `var n int64`. So I will trade return for var declaration, which does not make much difference for me...

command.go
dim: I think this is actually quite a nice and flexible abstraction. I don't think the overhead is particularly worrying.

commands.go
```diff
@@ -425,9 +476,14 @@ func (c *commandable) GetRange(key string, start, end int64) *StringCmd {
 	return cmd
 }
 
-func (c *commandable) GetSet(key, value string) *StringCmd {
-	cmd := NewStringCmd("GETSET", key, value)
-	c.Process(cmd)
+func (c *commandable) GetSet(key string, value interface{}) *StringCmd {
```
dim: I wonder if we should have a `GetSet` and a separate `GetSetValue`?

command.go
dim: Ignore, not important :)

commands.go
dim: Please see #121 too

command.go
vmihailenco: It is nice and flexible abstraction, but I don't see a use case for it...

commands.go
vmihailenco: Do you mean `GetSet(string, string)` and `GetSetValue(string, interface{})`? Just to save overhead of `interface{}`?

dim: http://www.darkcoding.net/software/go-the-price-of-interface/ :) Assuming you are doing 300k GETs per second: 120ns * 300,000 = 3.6ms. Not the end of the world, but again, my goal for this client is to be 'as fast as possible', but it may not apply to everyone else.

vmihailenco:
@dim PTAL

Before

```
BenchmarkRedisSet	   30000	     44599 ns/op	   10666 B/op	       7 allocs/op
BenchmarkRedisSetBytes	   30000	     51482 ns/op	   21163 B/op	       8 allocs/op
BenchmarkAppendArgs	10000000	       155 ns/op	       0 B/op	       0 allocs/op
```

After

```
BenchmarkRedisSet	   30000	     45297 ns/op	   10746 B/op	      10 allocs/op
BenchmarkRedisSetBytes	   30000	     45215 ns/op	   10762 B/op	      10 allocs/op
BenchmarkAppendArgs	10000000	       223 ns/op	       0 B/op	       0 allocs/op
```

vmihailenco:
@dim ping

dim:
Hey, sorry, I'm away until next week, but results look good

vmihailenco:
Np, this can wait.

vmihailenco:
@dim last time ping :)

dim:
Sorry, traveling (again), will review this week, promise, but should be OK
to merge in the meantime anyway

PR #138, opened 2015-07-18
Use unsafe for bytes->string conversion.

vmihailenco:
Benchmark

```
benchmark                       old ns/op     new ns/op     delta
BenchmarkRedisSetGet64Bytes     103732        100645        -2.98%
BenchmarkRedisSetGet1KB         104926        101539        -3.23%
BenchmarkRedisSetGet10KB        107164        102861        -4.02%
BenchmarkRedisSetGet1MB         5206102       3871532       -25.63%

benchmark                       old allocs     new allocs     delta
BenchmarkRedisSetGet64Bytes     16             15             -6.25%
BenchmarkRedisSetGet1KB         16             15             -6.25%
BenchmarkRedisSetGet10KB        16             15             -6.25%
BenchmarkRedisSetGet1MB         17             16             -5.88%

benchmark                       old bytes     new bytes     delta
BenchmarkRedisSetGet64Bytes     720           656           -8.89%
BenchmarkRedisSetGet1KB         3664          2640          -27.95%
BenchmarkRedisSetGet10KB        3664          2640          -27.95%
BenchmarkRedisSetGet1MB         4211153       3162576       -24.90%
```

@dim?

### DIFF
--- a/command.go
+++ b/command.go
@@ -400,7 +400,7 @@ func (cmd *StringCmd) reset() {
 }
 
 func (cmd *StringCmd) Val() string {
-	return string(cmd.val)
+	return bytesToString(cmd.val)
 }
 
 func (cmd *StringCmd) Result() (string, error) {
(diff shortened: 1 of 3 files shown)


dim:
Two things I am not sure about. One is GC: if you create an unsafe reference to a byte slice and that byte slice is garbage collected, what then? Just a thought, may not be relevant. 

Also, would like to see a GET/SET benchmark with a relatively large value, eg a few MBs.

vmihailenco:
I believe GC correctly handles unsafe.Pointer.

Will add more benchmarks.

vmihailenco:
Updated benchmarks. See first message.

dim:
LGTM

PR #142, opened 2015-08-04
Add pub/sub support to ClusterClient. Fixes #141.

### LINKED ISSUE #141: Publish / Subscribe for ClusterClient
Is there currently a way to subscribe to a cluster? The Publish / Subscribe methods are not exposed to the ClusterClient.

The spec (http://redis.io/topics/cluster-spec) says...

"In a Redis Cluster clients can subscribe to every node, and can also publish to every other node. The cluster will make sure that published messages are forwarded as needed."

The workaround that I see at the moment is to call `client.ClusterNodes()` and parse for a random master, then create a standard client and subscribe.  This is fine, but it would finer to be able to have this done in the background.

Thanks for this tremendous project.

vmihailenco:


vmihailenco:
@dim PTAL

cluster.go
```diff
@@ -108,8 +109,15 @@ func (c *ClusterClient) slotMasterAddr(slot int) string {
 // randomClient returns a Client for the first live node.
 func (c *ClusterClient) randomClient() (client *Client, err error) {
 	for i := 0; i < 10; i++ {
-		n := rand.Intn(len(c.addrs))
-		client, err = c.getClient(c.addrs[n])
+		var addr string
+		if len(c.masters) > 0 {
```
vmihailenco: I had to use masters here, becuase it looks like slaves don't propagate pub/sub messages to masters.

vmihailenco:
The issue with current design is that
1. we have to use master to send messages
2. we update masters list when we receive moved/ask response, which is not very reliable
3. Redis throws away message sent to slaves, which can happen when we have outdated list of masters

An improvement would be to refresh masters list every second. A solution is to fix Redis to return an error when message is sent to slave. Thoughts?

dim:
Sorry for the delay, will review tomorrow morning.

dim:
So that patch LGTM, but I am not entirely sure messages sent to slaves are discarded, I haven't tested it myself, but the spec doesn't say anything about that: http://redis.io/topics/cluster-spec#publishsubscribe

vmihailenco:
Tests reliably fail when messages are sent to slaves. So looks like cluster client should renew subscription when master is downgraded to slave...

dim:
Sounds like a bug worth mentioning on https://github.com/antirez/redis/issues. I don't see anyhing mentioned there, surprised no one else noticed it. Do you get an error from the slave when trying to publish?

vmihailenco:
No, but the Redis returns that there are 0 clients that received the message (publish returns 0).

vmihailenco:
Blocked by https://github.com/antirez/redis/issues/2740

PR #153, opened 2015-08-23
Implemented ZRangeByLex and ZRevRangeByLex with tests.

jeffpierce:
- Added isLex boolean to zRangeByScore.
- Added if statement to zRangeByScore which checks isLex.
  * If isLex is true, runs ZRANGEBYLEX rather than ZRANGEBYSCORE against the zset.
- Implemented ZRangeByLex.
- Added tests for ZRangeByLex to commands_test.go.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -1094,8 +1094,14 @@ type ZRangeByScore struct {
 	Offset, Count int64
 }
 
-func (c *commandable) zRangeByScore(key string, opt ZRangeByScore, withScores bool) *StringSliceCmd {
-	args := []interface{}{"ZRANGEBYSCORE", key, opt.Min, opt.Max}
+func (c *commandable) zRangeByScore(key string, opt ZRangeByScore, withScores, isLex bool) *StringSliceCmd {
+	var zcmd string
+	if isLex {
+		zcmd = "ZRANGEBYLEX"
+	} else {
+		zcmd = "ZRANGEBYSCORE"
+	}
+	args := []interface{}{zcmd, key, opt.Min, opt.Max}
 	if withScores {
 		args = append(args, "WITHSCORES")
 	}
@@ -1113,7 +1119,11 @@ func (c *commandable) zRangeByScore(key string, opt ZRangeByScore, withScores bo
 }
 
 func (c *commandable) ZRangeByScore(key string, opt ZRangeByScore) *StringSliceCmd {
-	return c.zRangeByScore(key, opt, false)
+	return c.zRangeByScore(key, opt, false, false)
+}
+
+func (c *commandable) ZRangeByLex(key string, opt ZRangeByScore) *StringSliceCmd {
+	return c.zRangeByScore(key, opt, false, true)
 }
 
 func (c *commandable) ZRangeByScoreWithScores(key string, opt ZRangeByScore) *ZSliceCmd {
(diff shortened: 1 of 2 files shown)


jeffpierce:
@vmihailenco If you prefer, I can reimplement this without piggy-backing off of zRangeByScore, but it seemed a little redundant to me to make a ZRangeByLex struct that's identical to the ZRangeByScore struct, then basically copy/paste zRangeByScore() while changing one word, which is why I did it this way.

Your package, your call though.  :)

commands.go
```diff
@@ -1094,8 +1094,14 @@ type ZRangeByScore struct {
 	Offset, Count int64
 }
 
-func (c *commandable) zRangeByScore(key string, opt ZRangeByScore, withScores bool) *StringSliceCmd {
-	args := []interface{}{"ZRANGEBYSCORE", key, opt.Min, opt.Max}
+func (c *commandable) zRangeByScore(key string, opt ZRangeByScore, withScores, isLex bool) *StringSliceCmd {
```
dim: Instead of passing `isLex`, why not just rename the function to `zRangeByGeneric` and pass the actual `zcmd` string?

commands.go
```diff
@@ -1113,7 +1119,11 @@ func (c *commandable) zRangeByScore(key string, opt ZRangeByScore, withScores bo
 }
 
 func (c *commandable) ZRangeByScore(key string, opt ZRangeByScore) *StringSliceCmd {
-	return c.zRangeByScore(key, opt, false)
+	return c.zRangeByScore(key, opt, false, false)
+}
+
+func (c *commandable) ZRangeByLex(key string, opt ZRangeByScore) *StringSliceCmd {
```
dim: I would also prefer to rename `ZRangeByScore` to `ZRangeByOpts`, but it would break the API. @vmihailenco: thoughts?

commands.go
jeffpierce: That'd work too, if you prefer it that way.

vmihailenco: +1, please refactor it so following is possible
- zRange("ZRANGEBYLEX", key, opts, withScores)
- zRange("ZRANGEBYSCORE", key, opts, withScores)

commands.go
vmihailenco: I think breaking API in this case is not worth it. Besides that I prefer to have `ZRangeByLex` and `ZRangeByScore` rather than `ZRangeByOpts`. It reads better and compiler helps finding typos in the name :)

vmihailenco:
Looks good, but please address comment regarding `isLex` and `zRangeByGeneric`

commands.go
jeffpierce: Can do.  I'll update it this afternoon.

jeffpierce:
@vmihailenco @dim Here you go.

commands.go
```diff
@@ -1089,13 +1089,13 @@ func (c *commandable) ZRangeWithScores(key string, start, stop int64) *ZSliceCmd
 	return cmd
 }
 
-type ZRangeByScore struct {
+type ZRangeBy struct {
```
vmihailenco: Please revert this and add a TODO to rename later (in v4). I don't want to break users code compilation.

vmihailenco:
Also can you please add `ZREVRANGEBYLEX` too?

commands.go
jeffpierce: Can do.

jeffpierce:
Sure thing.

jeffpierce:
@vmihailenco @dim Changed back the ZRangeByScore struct, added the TODO, and implemented ZRevRangeByLex as requested.

vmihailenco:
Excellent. Thanks.

Issue #148, opened 2015-08-11
No MaxRetries field in FailoverOptions struct


arnaudbriche:
HI,

I noticed that there is a MaxRetries field in the Options struct, but not in the FailoverOptions.
Is this deliberate ? And if so, could you explain why you choose to not implement auto-retry with
a failover client ?

vmihailenco:
Hi,

Sorry, I missed this issue.

`MaxRetries` was added after `FailoverClient`, so this is simply an overlook. Also you probably have to use new `ClusterClient` with Redis Cluster, which supports `MaxRedirects` option.

PR #154, opened 2015-08-24
Add MaxRetries to FailoverOptions

arnaudbriche:
To follow on https://github.com/go-redis/redis/issues/148, my use case does'nt allow me to make use of redis cluster, so I make this small PR to fix my issue.


vmihailenco:
Thanks.

PR #168, opened 2015-09-20
Added binding for GEOADD and GEORADIUS

icha024:


command.go
```diff
@@ -783,3 +783,80 @@ func (cmd *ClusterSlotCmd) parseReply(cn *conn) error {
 	cmd.val = v.([]ClusterSlotInfo)
 	return nil
 }
+
+//------------------------------------------------------------------------------
+
+type Location struct {
+	Name string
+	Distance string
```
vmihailenco: Probably use float64? Users will have to parse it anyway...

command.go
```diff
@@ -783,3 +783,80 @@ func (cmd *ClusterSlotCmd) parseReply(cn *conn) error {
...
+	GeoHash int64
+	Coordinates [2]string
```
vmihailenco: What about replacing this with 2 fields: `longitude` and `latitude`?

command.go
```diff
@@ -783,3 +783,80 @@ func (cmd *ClusterSlotCmd) parseReply(cn *conn) error {
 	cmd.val = v.([]ClusterSlotInfo)
 	return nil
 }
+
+//------------------------------------------------------------------------------
+
+type Location struct {
```
vmihailenco: Please rename to `GeoLocation` and probably add brief comment.

commands.go
```diff
@@ -1671,3 +1671,33 @@ func (c *commandable) ClusterAddSlotsRange(min, max int) *StatusCmd {
 	}
 	return c.ClusterAddSlots(slots...)
 }
+
+//------------------------------------------------------------------------------
+
+func (c *commandable) GeoAdd(key string, longLatMember...string) *IntCmd {
```
vmihailenco: Probably

```
GeoAdd(key, &GeoLocation{
  Name: "member",
  Longitude: 123,
  Latitude: 123,
})
```

?

commands.go
```diff
@@ -1671,3 +1671,33 @@ func (c *commandable) ClusterAddSlotsRange(min, max int) *StatusCmd {
...
+	args[1] = key
+	for i, triplet := range longLatMember {
+		args[2+i] = triplet
+	}
+	cmd := NewIntCmd(args...)
+	c.Process(cmd)
+	return cmd
+}
+
+func (c *commandable) GeoRadius(key, longitude, latitude, radius, unit string, options ...string) *GeoCmd {
```
vmihailenco: Oh, this new Redis API with lots of options...What about

```
GeoRadius(&GeoRadious{
  Key: "abc",
  Longitude: 123.123,
  Latitude: 123.123,
  WithCord: true,
  Sort: "ASC",
  Count: 3,
})
```

icha024: Done

commands.go
icha024: Done

command.go
icha024: Done.

command.go
icha024: Done

command.go
icha024: 0 would mean at the exact point. If it's a float64 and user didn't ask for distance calculation - it will be 0 but doesn't mean it's a the exact point - it should be undefined instead.

(Geohash does not have this problem, we know 0 is invalid)

command.go
```diff
@@ -783,3 +783,92 @@ func (cmd *ClusterSlotCmd) parseReply(cn *conn) error {
 	cmd.val = v.([]ClusterSlotInfo)
 	return nil
 }
+
+//------------------------------------------------------------------------------
+
+// Location type for GEO operations in Redis
+type GeoLocation struct {
+	Longitude, Latitude, Name, Distance string
```
vmihailenco: I still don't fully understand why we can't make `Longitude, Latitude, Distance float64`. I think your point is that `Distance = 0` when `WithDist = false` is confusing, which is probably true, but I think it is a minor point and having correct types is more important.

Also what about `Longitude, Latitude`?

command.go
```diff
@@ -783,3 +783,92 @@ func (cmd *ClusterSlotCmd) parseReply(cn *conn) error {
...
+	locations []GeoLocation
+}
+
+// Query type for geo radius
+type GeoRadiusQuery struct {
+	Key string
+	Longitude, Latitude, Radius string
+	// Unit default to km when nil
+	Unit string
+	WithCoord, WithDist, WithHash bool
```
vmihailenco: We have `GeoLocation.Distance`, but `GeoRadiusQuery.WithDist`. Can you please change names to match?

command.go
```diff
@@ -783,3 +783,92 @@ func (cmd *ClusterSlotCmd) parseReply(cn *conn) error {
...
+type GeoCmd struct {
+	baseCmd
+
+	locations []GeoLocation
+}
+
+// Query type for geo radius
+type GeoRadiusQuery struct {
+	Key string
+	Longitude, Latitude, Radius string
```
vmihailenco: string -> float64?

commands.go
```diff
@@ -1671,3 +1671,75 @@ func (c *commandable) ClusterAddSlotsRange(min, max int) *StatusCmd {
...
+		options++
+	}
+
+	args := make([]interface{}, 6 + options)
+	args[0] = "GEORADIUS"
+	args[1] = query.Key
+	args[2] = query.Longitude
+	args[3] = query.Latitude
+	args[4] = query.Radius
+	if len(query.Unit) > 0 {
```
vmihailenco: `query.Unit != ""`

commands.go
```diff
@@ -1671,3 +1671,75 @@ func (c *commandable) ClusterAddSlotsRange(min, max int) *StatusCmd {
...
+		args[5] = query.Unit
+	} else {
+		args[5] = "km"
```
vmihailenco: Please document that "km" is default value.

commands.go
```diff
@@ -1671,3 +1671,75 @@ func (c *commandable) ClusterAddSlotsRange(min, max int) *StatusCmd {
...
+		args[6+optionsCtr] = "WITHHASH"
+		optionsCtr++
+	}
+	if query.Count > 0 {
+		args[6+optionsCtr] = "COUNT"
+		optionsCtr++
+		args[6+optionsCtr] = query.Count
+		optionsCtr++
+	}
+	if len(query.Sort) > 0 {
```
vmihailenco: `query.Sort != ""`

commands.go
```diff
@@ -1671,3 +1671,75 @@ func (c *commandable) ClusterAddSlotsRange(min, max int) *StatusCmd {
...
+	if query.WithDist {
+		options++
+	}
+	if query.WithHash {
+		options++
+	}
+	if query.Count > 0 {
+		options += 2
+	}
+	if len(query.Sort) > 0 {
```
vmihailenco: `query.Sort != ""`

command.go
icha024: WithCoordinates, WithDistance, WithGeoHash (will let the 'with' prefix stay for bool)

command.go
icha024: Fair enough. Done.

command.go
icha024: Done.

commands.go
icha024: Done.

commands.go
icha024: Done.

commands.go
icha024: Done.

icha024:
CI is failing because it uses the stable version of Redis (does not yet have this feature). Should we hold off the pull until release or should we leave it/pull into an unstable branch?

vmihailenco:
I fixed the build in https://github.com/go-redis/redis/pull/168 . I still plan to do few changes before merging this, but thanks for your work!

PR #178, opened 2015-10-20
Implement SetName and GetName

wjessop:
Allows setting and getting the client connection name.

http://redis.io/commands/client-setname
http://redis.io/commands/client-getname

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -1367,6 +1367,18 @@ func (c *commandable) ClientPause(dur time.Duration) *BoolCmd {
 	return cmd
 }
 
+func (c *commandable) SetName(name string) *StatusCmd {
+	cmd := NewStatusCmd("CLIENT", "SETNAME", name)
+	c.Process(cmd)
+	return cmd
+}
+
+func (c *Client) GetName() *StringCmd {
+	cmd := NewStringCmd("CLIENT", "GETNAME")
+	c.Process(cmd)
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -1367,6 +1367,18 @@ func (c *commandable) ClientPause(dur time.Duration) *BoolCmd {
 	return cmd
 }
 
+func (c *commandable) SetName(name string) *StatusCmd {
+	cmd := NewStatusCmd("CLIENT", "SETNAME", name)
```
vmihailenco: You can also use `NewBoolCmd` here and the result will be true if Redis returns "OK". But I am fine with `StatusCmd` too.

vmihailenco:
Docs say that `SET NAME` assigns a name to the current connection. But go-redis client works with a pool of connections, so `redis.SetName` will assign name to the random connection in the pool. Can you please describe what you try to achieve?

wjessop:
I need to open a single connection to a redis server then 1) kill all other connections then 2) alter some config vars.

The thinking behind setting the name is that I can loop through all connections and CLIENT KILL them if they aren't the name I set.

vmihailenco:
Okay. Merging, but I will rename it to ClientSetName/ClientGetName and add a comment.

wjessop:
Thanks!

PR #201, opened 2015-11-27
Fix sporadic pool timeouts with IdleTimeout != 0. Fixes #195.

### LINKED ISSUE #195: connection pool timeout
I am using Redis as a caching layer for long running web services. I initialize the connection like so:
```
var (
	Queues  *redis.Client
	Tracker *redis.Client
)

func Connect(url string) {
	// cut away redis://
	url = url[8:]

	// connect to db #0
	Queues = redis.NewClient(&redis.Options{
		Addr:     url,
		Password: "",
		DB:       0,
	})

	_, err := Queues.Ping().Result()
	if err != nil {
		panic(err)
	}

	// connect to db #1
	Tracker = redis.NewClient(&redis.Options{
		Addr:     url,
		Password: "",
		DB:       1,
	})

	_, err = Tracker.Ping().Result()
	if err != nil {
		panic(err)
	}
}
```

Abeit in an upcoming patch (sysadmin is deploying a Redis cluster) it will be like so:

```
var (
	Cluster *redis.ClusterClient
)

func ConnectCluster(cluster, password string) {
	addresses := strings.Split(cluster, ",")
	Cluster = redis.NewClusterClient(&redis.ClusterOptions{
		Addrs: addresses,
		// Password: password,
	})

	_, err := Cluster.Ping().Result()
	if err != nil {
		panic(err)
	}
}
```

The above code gets run once when service boots up `in main.go` and the `*redis.ClusterClient` is being used for the lifetime of the process.

I realize there is an iherent problem with this approach, which is manifesting itself in connections timing out after a few days, and crashing the application:
`redis: connection pool timeout`
[See logs here](http://i.imgur.com/zlOCV2y.png)

Please advise, what would be a proper approach to use go-redis in this situation?

vmihailenco:
I will add test later :)

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -163,8 +163,12 @@ func (p *connPool) First() *conn {
 		select {
 		case cn := <-p.freeConns:
 			if p.isIdle(cn) {
-				p.conns.Remove(cn)
-				continue
+				var err error
+				cn, err = p.replace(cn)
+				if err != nil {
+					log.Printf("redis: replace failed: %s", err)
+					continue
+				}
 			}
 			return cn
 		default:
@@ -181,8 +185,12 @@ func (p *connPool) wait() *conn {
 		select {
 		case cn := <-p.freeConns:
 			if p.isIdle(cn) {
-				p.Remove(cn)
-				continue
+				var err error
+				cn, err = p.replace(cn)
+				if err != nil {
+					log.Printf("redis: replace failed: %s", err)
+					continue
+				}
 			}
 			return cn
 		case <-deadline:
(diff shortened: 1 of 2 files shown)


pool.go
```diff
@@ -163,8 +163,11 @@ func (p *connPool) First() *conn {
 		select {
 		case cn := <-p.freeConns:
 			if p.isIdle(cn) {
-				p.conns.Remove(cn)
```
vmihailenco: This had to be `p.Remove` so connection is replaced with new one.

Issue #203, opened 2015-11-30
Watch support for ClusterClient


imkira:
I want to issue a Watch to a specific key but when using ClusterClient it is not possible to do so. It's not even possible to work around that using getClient (and others), since they are all unexported.

I would like to ask you first, if there is any reason why Watch is not yet supported by ClusterClient.
If there isn't any in particular, I could start working on a PR to address this issue.

If you have any directions on how this should be done, please let me know.
I am simply thinking on adding a Watch function to cluster.go that just picks the Client for the slot of the first key, and returns the Multi for the corresponding client.Watch(keys...).

Something like:

```go
func (c *ClusterClient) Watch(keys ...string) (*Multi, error) {
        slot := hashSlot(keys[0])
	addr := c.slotMasterAddr(slot)
	client, err := c.getClient(addr)
	if err != nil {
		return nil, err
	}
        return client.Multi(), nil
}
```

Limitations with this:

- All keys should go to the same node.
- No support for ASKING (I guess not having this would be critical)

What do you think?

vmihailenco:
Support for `Multi` can be done similar to Pipeline https://github.com/go-redis/redis/blob/master/cluster_pipeline.go. But as you said it can require some special logic to handle MOVED/ASKING redirection...

imkira:
What do you think of a general-purpose 
```ClientForKey(key string) (*Client, error)```
where we avoid the compromise of handling the MOVED/ASKING logic to the user, so they either do it all themselves or, are at least aware of the limitations of just working with one client.
As far as I noticed, go-redis Client doesn't require you to put it back into the connection pool, so no problem with connection leaking either.

siriuszhuang:
I agree with @imkira . A `ClientForKey(key string) (*Client, error)` will be great for handle transaction on cluster.

siriuszhuang:
I agree with @imkira . A `ClientForKey(key string) (*Client, error)` will be great for handle transaction on cluster.

vmihailenco:
Sent https://github.com/go-redis/redis/pull/219

PR #217, opened 2015-12-14
Add `func ClientForKey(string) (*Client, error)` for ClusterClient.

siriuszhuang:
patch for #203


vmihailenco:
Thanks, but I prefer much smaller API change in https://github.com/go-redis/redis/pull/219

PR #219, opened 2015-12-16
cluster: add Watch support.

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/203
Fixes https://github.com/go-redis/redis/pull/217

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -44,6 +44,17 @@ func NewClusterClient(opt *ClusterOptions) *ClusterClient {
 	return client
 }
 
+// Watch creates new transaction and marks the keys to be watched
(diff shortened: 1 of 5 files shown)


siriuszhuang:
@vmihailenco that's great. But I found another problem that is I cannot run `ClusterClient.FlushALL()` to wipe all data on different nodes. It only can wipe one node. Is there a `ClusterServer.MasterClients() []*ClusterClient` like API? Or there is another way to flush cluster?

vmihailenco:
@siriuszhuang please file a separate issue so I don't forget about. But I think we need a special case in `ClusterClient.process` for `FlushDb`/`FlushAll`.

Issue #222, opened 2015-12-21
Pool rate limit should only trigger on dial errors


dim:
Hi, before I submit a PR, I wanted to discuss this first. One of our apps uses redis cluster and periodically throws a series of `redis: you open connections too fast (last error: nil)` errors for no obvious reason. The affected shards seem to be stable and last error is always `nil`. I tried to debug this several times but it is only happening at a very high frequency and I can't pin it down.

Happy to hear more debug ideas, but the most obvious one is: should we really limit new connections, or should we rather limit connection failures. Assuming dial is not failing, do we need to limit? Is there another use case?

vmihailenco:
Technically rate limiting only on dial errors is fine, but it does not fix the root cause (opening/closing connections).

I suggest we improve the error message to say something like this: `redis: you open connections too fast (last_dial_error: %q last_error: %q)`, where `last_error` says why last connection was closed. I will send a PR.

vmihailenco:
I hope this is not needed any more, so closing.

PR #228, opened 2015-12-22
Better rate limited message.

vmihailenco:
Updates https://github.com/go-redis/redis/issues/222

@dim PTAL

### DIFF
--- a/multi.go
+++ b/multi.go
@@ -46,16 +46,15 @@ func (c *Client) Multi() *Multi {
 	return multi
 }
 
-func (c *Multi) putConn(cn *conn, ei error) {
-	var err error
-	if isBadConn(cn, ei) {
+func (c *Multi) putConn(cn *conn, err error) {
+	if isBadConn(cn, err) {
 		// Close current connection.
-		c.base.connPool.(*stickyConnPool).Reset()
+		c.base.connPool.(*stickyConnPool).Reset(err)
 	} else {
-		err = c.base.connPool.Put(cn)
-	}
-	if err != nil {
-		log.Printf("redis: putConn failed: %s", err)
+		err := c.base.connPool.Put(cn)
+		if err != nil {
+			log.Printf("redis: putConn failed: %s", err)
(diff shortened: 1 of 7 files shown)


multi.go
```diff
@@ -46,16 +46,15 @@ func (c *Client) Multi() *Multi {
...
 		// Close current connection.
-		c.base.connPool.(*stickyConnPool).Reset()
+		c.base.connPool.(*stickyConnPool).Reset(err)
 	} else {
-		err = c.base.connPool.Put(cn)
-	}
-	if err != nil {
-		log.Printf("redis: putConn failed: %s", err)
+		err := c.base.connPool.Put(cn)
+		if err != nil {
```
dim: gofmt?

pool.go
```diff
@@ -204,15 +204,15 @@ func (p *connPool) wait() *conn {
...
+			"redis: you open connections too fast (last_error=%q)",
+			p.loadLastErr(),
 		)
 		return nil, err
 	}
 
 	cn, err := p.dialer()
 	if err != nil {
-		p.lastDialErr = err
+		p.storeLastErr(err.Error())
```
dim: Wouldn't it make sense to store the full `error` and not just a `string`?

pool.go
```diff
@@ -330,6 +333,17 @@ func (p *connPool) reaper() {
 	}
 }
 
+func (p *connPool) storeLastErr(err string) {
+	p.lastErr.Store(err)
+}
+
+func (p *connPool) loadLastErr() string {
```
dim: And return `error` here, instead of `string` too?

dim:
LGTM overall, see few notes above. thanks, dim

pool.go
vmihailenco: `atomic.Value` only allows to store values of one type and `error` interface can have different underlying types.

multi.go
vmihailenco: This is gofmted...

PR #229, opened 2015-12-23
Support multiple keys for the PFCOUNT command

hongrich:
From http://redis.io/commands/pfcount:
> When called with multiple keys, returns the approximated cardinality of the union of the HyperLogLogs passed, by internally merging the HyperLogLogs stored at the provided keys into a temporary HyperLogLog.


vmihailenco:
Thanks!

PR #233, opened 2015-12-30
Bugfix: pubsub/psubscribe should register a pattern rather than channel

Freeaqingme:
Noticed that when the connection got interrupted, my PSUBSCRIBE pubsub connection had turned into a SUBSCRIBE pubsub connection. I believe this fixes it.


vmihailenco:
Thanks, released v3.2.26

PR #236, opened 2016-01-06
Allow to instrument client pool sizes

dim:


cluster.go
```diff
@@ -61,6 +61,20 @@ func (c *ClusterClient) Close() error {
 	return nil
 }
 
+// PoolLen returns the number of active connections in the pool, for instrumentation
```
vmihailenco: What do you mean by "instrumentation"?

Also description is not very accurate since you return max number of active connections...

dim: I would like measure and periodically report the actual number of used pool connections

vmihailenco: Are you sure that you don't want to return the sum? If not - please update the comment. Your implementation is *very* unusual and specific...

dim: So it all goes back to the fact that `ClusterOptions` accept a `PoolSize` which is actually per-client and not a per-cluster-client. Returning the `PoolLen()` is supposed to signal the user that the `PoolSize` setting is probably too low, therefore using the maximum seems the right approach to me, I will update the comment.

vmihailenco:
LGTM, thanks.

Let me know if you find that default limit of 10 connections is not enough...

vmihailenco:
...and feel free to merge this and release new version :)

dim:
Closing in favour of https://github.com/go-redis/redis/pull/241

Issue #235, opened 2016-01-05
func (c *Client) Info() should accept a section parameter


Shamanoid:
func (c *Client) Info() *StringCmd should become :
func (c *Client) Info(section string) *StringCmd, in order to respect the INFO command reference such as defined in http://redis.io/commands/info

It would also permit to avoid requesting all default sections each time and would ease the creation of monitoring tools.

vmihailenco:
Done https://godoc.org/gopkg.in/redis.v3#Client.Info

Shamanoid:
Thanks a lot !

PR #239, opened 2016-01-08
commands.go: add section parameter to Info function.

anmic:


vmihailenco:
Fixes https://github.com/go-redis/redis/issues/235

PR #241, opened 2016-01-13
Added pool instrumentation

dim:
This is actually much better than https://github.com/go-redis/redis/pull/236. PTAL

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -56,6 +56,22 @@ func (c *ClusterClient) Watch(keys ...string) (*Multi, error) {
 	return client.Watch(keys...)
 }
 
+// PoolMetrics returns accumulated connection pool metrics
+func (c *ClusterClient) PoolMetrics() *PoolMetrics {
+	acc := PoolMetrics{}
+	c.clientsMx.RLock()
+	for _, client := range c.clients {
+		m := client.PoolMetrics()
+		acc.Conns.Open += m.Conns.Open
+		acc.Conns.Free += m.Conns.Free
+		acc.Counts.Requests += m.Counts.Requests
+		acc.Counts.Waits += m.Counts.Waits
+		acc.Counts.Timeouts += m.Counts.Timeouts
+	}
+	c.clientsMx.RUnlock()
+	return &acc
+}
+
 // Close closes the cluster client, releasing any open resources.
 //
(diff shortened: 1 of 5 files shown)


pool.go
```diff
@@ -16,6 +16,19 @@ var (
 	errPoolTimeout = errors.New("redis: connection pool timeout")
 )
 
+// PoolMetrics contains pool state information and accumulated stats
+type PoolMetrics struct {
```
vmihailenco: Let's rename to `PoolStats` to match https://golang.org/pkg/runtime/#MemStats?

pool.go
```diff
@@ -16,6 +16,19 @@ var (
...
+	Conns struct {
+		Free int // the number of free connections in the pool
```
vmihailenco: uint64 here and below? see https://golang.org/pkg/runtime/#MemStats

pool.go
```diff
@@ -16,6 +16,19 @@ var (
...
+		Open int // the number of open connections in the pool
+	}
+	Counts struct {
```
vmihailenco: `Counts` name is not very descriptive... Probably change struct to have flat layout:

```
Gets uint64
Waits uint64
Timeouts uint64

TotalConns uint64
FreeConns uint64
```

pool.go
```diff
@@ -298,6 +319,14 @@ func (p *connPool) FreeLen() int {
 	return len(p.freeConns)
 }
 
+func (p *connPool) Stats() poolStats {
```
vmihailenco: Can't we return `PoolMetrics` (`PoolStats`) from this method?

cluster.go
```diff
@@ -56,6 +56,22 @@ func (c *ClusterClient) Watch(keys ...string) (*Multi, error) {
 	return client.Watch(keys...)
 }
 
+// PoolMetrics returns accumulated connection pool metrics
+func (c *ClusterClient) PoolMetrics() *PoolMetrics {
```
vmihailenco: Probably return map[string]*PoolStats?

Or it is good time to add iterator like you suggested? E.g.

```
clients := cluster.Masters()
for _, client := range clients {
    fmt.Println(client.PoolStats())
}
```

vmihailenco:
Closing in favor of https://github.com/go-redis/redis/pull/244

PR #249, opened 2016-01-25
Add ClusterClient.Clients.

vmihailenco:
@dim what do you think about this?

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -17,11 +17,11 @@ type ClusterClient struct {
 
 	addrs   []string
 	slots   [][]string
-	slotsMx sync.RWMutex // Protects slots and addrs.
+	slotsMx sync.RWMutex // protects slots and addrs
 
 	clients   map[string]*Client
 	closed    bool
-	clientsMx sync.RWMutex // Protects clients and closed.
+	clientsMx sync.RWMutex // protects clients and closed
 
 	opt *ClusterOptions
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
@dim ping

dim:
Sorry, travelling again, will take a look ASAP

dim:
Sorry that it took so long, but I am not sure I really like the idea. It seems like something we may want to keep private, as it may cause a lot of confusion and unpredictable behaviour. It requires the users to be responsible, to not e.g. cause any (accidentally) damage. 

I can imagine the simplest of all cases, where users would setup a background goroutine, to e.g. iterate over `client.Clients()` and call `Info()` on each. At some point, the user will call `Close()` on the main ClusterClient and the background goroutine will panic. It will require really complicated sync between the two, which is hardly obvious from a user's point of view. 

I am actually not sure about the best way to solve this, that's why I thought retuning an iterator object may be safer than just a collection of clients, as we could at least check for `p.closed` on each `Next()` call an return early.

vmihailenco:
Theoretically iterator is exploitable too, e.g.

```
var clients []*redis.Client
cluster.Iterate(func(cl *redis.Client) {
    clients = append(clients, cl)
})
for _, cl := range clients {
    cl.Close()
}
```

>At some point, the user will call Close() on the main ClusterClient and the background goroutine will panic

I did not check, but I expect cluster to return `errClosed`, not to panic.

>It will require really complicated sync between the two, which is hardly obvious from a user's point of view.

If we speak about `errClosed` then we can simply check for that error and remove closed client (which will cause reconnection on next request / retry). It should be fairly simple.

vmihailenco:
@dim I made this API private and going to merge it. We can rework API later when we have more info.

dim:
I still wonder about its usefulness. I'd rather implement a special object
e.g. like a special Pipeline that can execute commands on every node within
three cluster. Seems cleaner and less prone to errors

PR #253, opened 2016-02-03
Declare and use a package-level Logger

### LINKED ISSUE #250: library logs messing around with application logs
go-redis is a library, and as a library, it should not be logging anything. But it's possible to see some calls to log.Printf here: https://github.com/go-redis/redis/search?q=log.Printf&type=Code&utf8=%E2%9C%93.

If removing logs is simply not an option, would you be open for accepting a Logger in the Client and using this Logger? Replacing ``log.Printf`` calls with ``c.logger.Printf``. And users willing to disable logging could simply use ioutil.Discard for that.

fsouza:
This allow users of the API to override the Logger. Fix #250.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1,7 +1,6 @@
 package redis
 
 import (
-	"log"
 	"math/rand"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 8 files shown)


fsouza:
Hmm, the build failure seems to be unrelated, but I will investigate.

vmihailenco:
Thanks, I will fix build in separate PR.

PR #254, opened 2016-02-06
Fix cluster slots parsing.

vmihailenco:
Changed in https://github.com/antirez/redis/commit/c285862621b2db70ec9152f5e081a091d49a0a45

### DIFF
--- a/command.go
+++ b/command.go
@@ -725,12 +725,14 @@ func (cmd *ScanCmd) readReply(cn *conn) error {
 
 //------------------------------------------------------------------------------
 
+// TODO: rename to ClusterSlot
 type ClusterSlotInfo struct {
 	Start int
 	End   int
 	Addrs []string
 }
 
+// TODO: rename to ClusterSlotsCmd
 type ClusterSlotCmd struct {
 	baseCmd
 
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -725,12 +725,14 @@ func (cmd *ScanCmd) readReply(cn *conn) error {
 
 //------------------------------------------------------------------------------
 
+// TODO: rename to ClusterSlot
 type ClusterSlotInfo struct {
 	Start int
 	End   int
 	Addrs []string
```
vmihailenco: @dim new cluster-slots returns following reply

```
127.0.0.1:30001> cluster slots
1) 1) (integer) 0
   2) (integer) 5460
   3) 1) "127.0.0.1"
      2) (integer) 30001
      3) "09dbe9720cda62f7865eabc5fd8857c5d2678366"
   4) 1) "127.0.0.1"
      2) (integer) 30004
      3) "821d8ca00d7ccf931ed3ffc7e3db0599d2271abf"
```

I.e. it contains node id which requires us to change struct definition to something like

```
type ClusterNode struct {
    Addr string
    Id string
}

type ClusterSlot struct {
 	Start int
 	End   int
 	Nodes []ClusterNode
}
```

What do you think? Should we do it in redis.v4 or it is fine do it in redis.v3?

vmihailenco:
@dim merging this to fix build, but please see my comment https://github.com/go-redis/redis/pull/254#discussion_r52100335

dim:
:+1:

PR #269, opened 2016-03-09
Correct method accessors

dim:
Minor but nevertheless:

* the `SetXX` method was only on `Client`, not on `commandable`
* some methods we exposed on the element and not the pointer: 
  * https://godoc.org/gopkg.in/redis.v3#Client.Close
  * https://godoc.org/gopkg.in/redis.v3#Client.String 
  * https://godoc.org/gopkg.in/redis.v3#Client.ZRevRangeByLex


vmihailenco:
Good catch!

PR #272, opened 2016-03-11
Minimise connections

dim:
So this one serves two purposes: 

1. The connection pool currently slows down substantially, when many connections are used. If we raise the PoolSize above a certain threshold, we end up with more and more pool timeouts. Our timeout settings are relatively low, but increasing the pool size seems to make matters only worse. I suspect it's due to the iterations on `connList.Remove` and `connList.Replace`. As these happen within a mutex, a failing connection effectively takes down everything else, causing the pool to block. I have turned the slice into a map, which should reduce the time spent within the lock.
2. We are using channels to hold our `freeConn` and they are FIFO by design. It also means that we round-robin across all our connections in the pool and large pools never actually shrink. I have replaced the channel with a LIFO stack which forces us to reuse connections that were most recently returned and reap unused connections.

I would also like to default `IdleTimeout` to something like `time.Hour`, if not provided.

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -147,14 +147,15 @@ func (l *connList) Remove(cn *conn) error {
 	}
 
 	l.mx.Lock()
-	defer l.mx.Unlock()
 
 	if _, ok := l.cns[cn]; ok {
 		delete(l.cns, cn)
+		l.mx.Unlock()
 		atomic.AddInt32(&l.size, -1)
 		return cn.Close()
 	}
 
+	l.mx.Unlock()
 	if l.closed() {
 		return nil
 	}
@@ -164,13 +165,14 @@ func (l *connList) Remove(cn *conn) error {
 // Replace a connection with an old connection with a new one
 func (l *connList) Replace(cn, newcn *conn) error {
 	l.mx.Lock()
-	defer l.mx.Unlock()
 
 	if _, ok := l.cns[cn]; ok {
 		delete(l.cns, cn)
 		l.cns[newcn] = struct{}{}
+		l.mx.Unlock()
 		return cn.Close()
 	}
+	l.mx.Unlock()
 
 	if l.closed() {
 		return newcn.Close()
--- a/pool.go
+++ b/pool.go
@@ -37,6 +37,55 @@ type pool interface {
 	Stats() *PoolStats
 }
 
+// connStack is used as a LIFO to maintain free connection
+type connStack struct {
+	cns  []*conn
+	free chan struct{}
+	mx   sync.Mutex
+}
+
+func newConnStack(max int) *connStack {
+	return &connStack{
+		cns:  make([]*conn, 0, max),
+		free: make(chan struct{}, max),
+	}
+}
+
+func (s *connStack) Len() int { return len(s.free) }
+
+func (s *connStack) Push(cn *conn) {
+	s.mx.Lock()
+	s.cns = append(s.cns, cn)
+	s.mx.Unlock()
+	s.free <- struct{}{}
+}
+
+func (s *connStack) Pop() *conn {
+	select {
+	case <-s.free:
+		return s.pop()
+	default:
+		return nil
+	}
+}
+
+func (s *connStack) PopWithTimeout(d time.Duration) *conn {
+	select {
+	case <-s.free:
+		return s.pop()
+	case <-time.After(d):
+		return nil
+	}
+}
+
+func (s *connStack) pop() (cn *conn) {
+	s.mx.Lock()
+	ci := len(s.cns) - 1
+	cn, s.cns = s.cns[ci], s.cns[:ci]
+	s.mx.Unlock()
+	return
+}
+
 // connList stores all known connections, usable or not
 type connList struct {
 	cns  map[*conn]struct{}
@@ -139,9 +188,8 @@ type connPool struct {
 	rl        *ratelimit.RateLimiter
 	opt       *Options
 	conns     *connList
-	freeConns chan *conn
-	// freeConns *connStack
-	stats PoolStats
+	freeConns *connStack
+	stats     PoolStats
 
 	_closed int32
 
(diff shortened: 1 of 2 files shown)


dim:
```
$ benchcmp old.txt new.txt 
benchmark                         old ns/op     new ns/op     delta
BenchmarkParseReplyStatus-4       123           113           -8.13%
BenchmarkParseReplyInt-4          102           102           +0.00%
BenchmarkParseReplyError-4        227           215           -5.29%
BenchmarkParseReplyString-4       170           178           +4.71%
BenchmarkParseReplySlice-4        901           889           -1.33%
BenchmarkAppendArgs-4             227           218           -3.96%
BenchmarkRedisClusterPing-4       9100          9183          +0.91%
BenchmarkPool-4                   241           308           +27.80%
BenchmarkRedisPing-4              7638          7528          -1.44%
BenchmarkRedisSet-4               18755         18934         +0.95%
BenchmarkRedisGetNil-4            7676          7668          -0.10%
BenchmarkRedisSetGet64Bytes-4     16940         16321         -3.65%
BenchmarkRedisSetGet1KB-4         17856         17755         -0.57%
BenchmarkRedisSetGet10KB-4        40290         37613         -6.64%
BenchmarkRedisSetGet1MB-4         2195778       2157348       -1.75%
BenchmarkRedisSetGetBytes-4       43140         37612         -12.81%
BenchmarkRedisMGet-4              8281          7888          -4.75%
BenchmarkSetExpire-4              19366         16504         -14.78%
BenchmarkPipeline-4               9863          9942          +0.80%
BenchmarkZAdd-4                   8951          8828          -1.37%

benchmark                         old allocs     new allocs     delta
BenchmarkParseReplyStatus-4       1              1              +0.00%
BenchmarkParseReplyInt-4          1              1              +0.00%
BenchmarkParseReplyError-4        2              2              +0.00%
BenchmarkParseReplyString-4       1              1              +0.00%
BenchmarkParseReplySlice-4        8              8              +0.00%
BenchmarkAppendArgs-4             0              0              +0.00%
BenchmarkRedisClusterPing-4       4              4              +0.00%
BenchmarkPool-4                   0              0              +0.00%
BenchmarkRedisPing-4              4              4              +0.00%
BenchmarkRedisSet-4               7              7              +0.00%
BenchmarkRedisGetNil-4            5              5              +0.00%
BenchmarkRedisSetGet64Bytes-4     11             11             +0.00%
BenchmarkRedisSetGet1KB-4         11             11             +0.00%
BenchmarkRedisSetGet10KB-4        12             12             +0.00%
BenchmarkRedisSetGet1MB-4         13             13             +0.00%
BenchmarkRedisSetGetBytes-4       12             12             +0.00%
BenchmarkRedisMGet-4              13             13             +0.00%
BenchmarkSetExpire-4              13             13             +0.00%
BenchmarkPipeline-4               17             17             +0.00%
BenchmarkZAdd-4                   8              8              +0.00%

benchmark                         old bytes     new bytes     delta
BenchmarkParseReplyStatus-4       32            32            +0.00%
BenchmarkParseReplyInt-4          8             8             +0.00%
BenchmarkParseReplyError-4        32            32            +0.00%
BenchmarkParseReplyString-4       32            32            +0.00%
BenchmarkParseReplySlice-4        170           170           +0.00%
BenchmarkAppendArgs-4             0             0             +0.00%
BenchmarkRedisClusterPing-4       116           116           +0.00%
BenchmarkPool-4                   0             0             +0.00%
BenchmarkRedisPing-4              116           116           +0.00%
BenchmarkRedisSet-4               10707         10707         +0.00%
BenchmarkRedisGetNil-4            160           160           +0.00%
BenchmarkRedisSetGet64Bytes-4     418           418           +0.00%
BenchmarkRedisSetGet1KB-4         1378          1378          +0.00%
BenchmarkRedisSetGet10KB-4        21349         21349         +0.00%
BenchmarkRedisSetGet1MB-4         2113967       2113942       -0.00%
BenchmarkRedisSetGetBytes-4       21365         21365         +0.00%
BenchmarkRedisMGet-4              352           352           +0.00%
BenchmarkSetExpire-4              380           380           +0.00%
BenchmarkPipeline-4               800           800           +0.00%
BenchmarkZAdd-4                   224           224           +0.00%
```

pool.go
```diff
@@ -37,31 +37,97 @@ type pool interface {
...
+}
+
+func (s *connStack) ShiftStale(timeout time.Duration) *conn {
+	select {
+	case <-s.free:
+		s.mx.Lock()
+		defer s.mx.Unlock()
+
+		if cn := s.cns[0]; cn.IsStale(timeout) {
+			s.cns = s.cns[1:]
```
vmihailenco: If I understand correctly there is a memory leak here, because `s.cns` slice is slowly growing.

vmihailenco:
>we end up with more and more pool timeouts

Can you share DiatTimeout and ReadTimeout? I suspect that you are affected by combination of:
- read timeout -> connection is closed
- dial timeout -> new connection can't be opened

>I suspect it's due to the iterations on connList.Remove and connList.Replace

See https://github.com/go-redis/redis/pull/273. Remove is not very fast, but it is fast enough and does not become slower when pool size grows.

>As these happen within a mutex, a failing connection effectively takes down everything else, causing the pool to block.

We hold lock only for adding/removing connection from the list. Lock is not held when connection is recycled.

>It also means that we round-robin across all our connections in the pool and large pools never actually shrink.

True, but this does not look very important for Redis which can handle thousands of connections... I would like to fix this, but honestly I don't see how it can fix your issue.

dim:
Sure, I can give you some background. The settings are pretty standard (and generous):
```
opt := redis.ClusterOptions{
  PoolSize: 20,
  DialTimeout: 5*time.Second,
  ReadTimeout: time.Second,
  WriteTimeout: time.Second,
  PoolTimeout: time.Second,
}
```

Every now and then, we see "spikes" of `redis: connection pool timeout`, although there is no indication of a break-down on operational level. The cluster nodes are up, RDB/AOF are disabled, and there is nothing in the redis logs or in SLOWLOG that would indicate a "lock up". There is nothing that would suggest a network problem either.

dim:
The odd thing we have identified is that the frequency of these spikes is increasing when we increase the `PoolSize`. We bumped the value from `20` to `200` and the pool time outs instantly went through the roof.

dim:
> True, but this does not look very important for Redis which can handle thousands of connections...

I don't think it's important to Redis, but it's important to the Go client. Assuming a PoolSize of 10 connections with a cluster of 200 nodes, you (worst case) end up with 2,000 open TCP connections and none of these will ever be retired.

vmihailenco:
Added your changes in https://github.com/go-redis/redis/pull/277

vmihailenco:
Tagged v3.5.0 with fixes from this PR.

PR #285, opened 2016-03-17
Faster and simpler pool.

vmihailenco:
With recent pool changes we reintroduced `pool.Remove` which has flaw described in https://github.com/go-redis/redis/pull/89. So this PR fixes that flaw and also makes sure that `Dial` and `Conn.Close` is called without holding the lock.

Before

```
BenchmarkPoolGetPut10Conns-4      	 3000000	       439 ns/op	       0 B/op	       0 allocs/op
BenchmarkPoolGetPut100Conns-4     	 3000000	       409 ns/op	       0 B/op	       0 allocs/op
BenchmarkPoolGetPut1000Conns-4    	 5000000	       399 ns/op	       0 B/op	       0 allocs/op
BenchmarkPoolGetReplace10Conns-4  	  500000	      2925 ns/op	    8432 B/op	       7 allocs/op
BenchmarkPoolGetReplace100Conns-4 	  500000	      2918 ns/op	    8432 B/op	       7 allocs/op
BenchmarkPoolGetReplace1000Conns-4	  500000	      2938 ns/op	    8432 B/op	       7 allocs/op
```

After

```
BenchmarkPoolGetPut10Conns-4      	 5000000	       376 ns/op	       0 B/op	       0 allocs/op
BenchmarkPoolGetPut100Conns-4     	 5000000	       367 ns/op	       0 B/op	       0 allocs/op
BenchmarkPoolGetPut1000Conns-4    	 5000000	       377 ns/op	       0 B/op	       0 allocs/op
BenchmarkPoolGetReplace10Conns-4  	  300000	      4650 ns/op	    8419 B/op	       7 allocs/op
BenchmarkPoolGetReplace100Conns-4 	  300000	      4945 ns/op	    8419 B/op	       7 allocs/op
BenchmarkPoolGetReplace1000Conns-4	  300000	      4887 ns/op	    8419 B/op	       7 allocs/op
```

So it is slightly slower on remove, but I hope to make it faster in next PR.

PS Also I need to rewrite cluster reaper...

/cc @dim


dim:
Will review it later today

vmihailenco:
Cluster reaper fixed.

cluster.go
```diff
@@ -34,17 +35,29 @@ type ClusterClient struct {
...
 	client.commandable.process = client.process
 	client.reloadSlots()
-	go client.reaper()
 	return client
 }
 
+// getClients returns a snapshot of clients for cluster nodes
+// this ClusterClient has been working with recently.
+// Note that snapshot can contain closed clients.
+func (c *ClusterClient) getClients() map[string]*Client {
```
dim: I think the code gets a little messy (personal preference), how about a little helper:

```go
type clientMap struct {
	clients map[string]*Client
	mu      sync.RWMutex
}

func newClientsMap() *clientMap {
	return &clientMap{
		clients: map[string]*Client{},
	}
}

func (m *clientMap) Clients() []*Client {
	m.mu.RLock()
	clients := make([]*Client, 0, len(m.clients))
	for _, client := range m.clients {
		clients = append(clients, client)
	}
	m.mu.RUnlock()
	return clients
}

func (m *clientMap) Fetch(addr string, factory func() *Client) (*Client, error) {
	m.mu.RLock()
	client, ok := m.clients[addr]
	m.mu.RUnlock()
	if ok {
		return client, nil
	}

	m.mu.Lock()
	client, ok = m.clients[addr]
	if !ok {
		client = factory()
		m.clients[addr] = client
	}
	m.mu.Unlock()
	return client, nil
}

func (m *clientMap) Reset() (err error) {
	m.mu.Lock()
	defer m.mu.Unlock()

	for addr, client := range m.clients {
		if e := client.Close(); e != nil {
			err = e
		}
		delete(m.clients, addr)
	}
	return
}
```

cluster.go
```diff
@@ -16,15 +16,16 @@ import (
...
 	commandable
 
+	opt *ClusterOptions
+
+	slotsMx sync.RWMutex // protects slots and addrs
 	addrs   []string
 	slots   [][]string
-	slotsMx sync.RWMutex // Protects slots and addrs.
 
+	clientsMx sync.RWMutex // protects clients and closed
```
dim: Closed is now atomic, comment should be updated.

cluster.go
```diff
@@ -68,47 +80,47 @@ func (c *ClusterClient) PoolStats() *PoolStats {
...
-	if c.closed {
+	if !atomic.CompareAndSwapInt32(&c._closed, 0, 1) {
 		return pool.ErrClosed
 	}
-	c.closed = true
+
+	c.clientsMx.Lock()
 	c.resetClients()
 	c.setSlots(nil)
+	c.clientsMx.Unlock()
```
dim: So slots is protected by its own mutex. If you don't want to use my `clientMap` suggestion above, I think you want to unlock the `clientMx` a line earlier (not that it matters much).

internal/pool/pool.go
```diff
@@ -148,6 +124,42 @@ func (p *ConnPool) NewConn() (*Conn, error) {
 	return NewConn(netConn), nil
 }
 
+func (p *ConnPool) PopFree() *Conn {
+	timer := timers.Get().(*time.Timer)
+	if !timer.Reset(p.poolTimeout) {
+		<-timer.C
```
dim: Sorry, I don't understand this. How does it matter if this is a recycled timer or not?

internal/pool/pool.go
```diff
@@ -53,18 +62,23 @@ type ConnPool struct {
...
+	stats PoolStats
 
+	_closed int32 // atomic
 	lastErr atomic.Value
 }
 
 var _ Pooler = (*ConnPool)(nil)
 
-func NewConnPool(dial dialer, poolSize int, poolTimeout, idleTimeout time.Duration) *ConnPool {
+func NewConnPool(dial dialer, poolSize int, poolTimeout, idleTimeout, idleCheckFrequency time.Duration) *ConnPool {
```
dim: So pool looks good to me, although I would prefer to abstract `conns` and `freeConns` into their own structs as it makes testing easier and separates trivial and non-trivial logic (again, just my preference).

cluster.go
vmihailenco: I agree, but let's not refactor everything in one PR :)

vmihailenco: Filled https://github.com/go-redis/redis/issues/287 so code is not lost

cluster.go
vmihailenco: done

cluster.go
vmihailenco: done

internal/pool/pool.go
vmihailenco: This is needed, because if timer is already expired there will be 2 events in timer channel - http://play.golang.org/p/eq_bvcd0AB

internal/pool/pool.go
vmihailenco: It is a bit complicated since sometimes I hold lock to do 2 actions to max performance, e.g.:

```
	p.connsMu.Lock()
	if cn != nil {
		p.remove(cn, errConnStale)
	}
	p.conns = append(p.conns, newcn)
	p.connsMu.Unlock()
```

PR #290, opened 2016-03-19
Cleanup error handling code.

vmihailenco:
This change makes cluster to retry command only in 2 cases:
- network error
- MOVED/ASK error response

Previously it retried on all error replies such as:

```
-ERR unknown command 'foobar'
-WRONGTYPE Operation against a key holding the wrong kind of value
```

etc

I took a chance to speedup `isMovedError` since it is called more often now.


dim:
LGTM

PR #294, opened 2016-04-06
Disable logging by default.

vmihailenco:


vmihailenco:
Fixes https://github.com/go-redis/redis/issues/250#issuecomment-205871962

PR #298, opened 2016-04-09
Rename Multi to Tx to better reflect the purpose. Fixes #194.

### LINKED ISSUE #194: Regarding client.Multi deprecation
I noticed that client.Multi marked in docs as deprecated, correct me if I am wrong, but isn't it useful when you need to run a bunch of commands in "all or nothing" way and won't maintain consistency?

vmihailenco:


vmihailenco:
This is for v4... closing...

PR #302, opened 2016-04-09
Expose cluster node id in ClusterSlots.

vmihailenco:
Before

```
type ClusterSlot struct {
 	Start int
 	End   int
	Addrs []string
 }
```

After

```
type ClusterNode struct {
	Id   string
	Addr string
}

type ClusterSlot struct {
 	Start int
 	End   int
	Nodes []ClusterNode
 }
```

/cc @dim


dim:
LGTM

PR #303, opened 2016-04-09
Scan: swap return values and change cursor type.

vmihailenco:


dim:
I would like to add high level iterators for all scan commands, to simplify their use. I'll submit a PR

vmihailenco:
Okay, waiting for your PR, but keep in mind that this change is for v4 that I am going to keep in alpha/beta state for at least a month :)

PR #310, opened 2016-04-13
Scan iterators (v4)

dim:
Here you go, as promised in #303

### DIFF
--- a/command.go
+++ b/command.go
@@ -689,41 +689,42 @@ type ScanCmd struct {
 	baseCmd
 
 	cursor int64
-	keys   []string
+	page   []string
 }
 
 func NewScanCmd(args ...interface{}) *ScanCmd {
-	return &ScanCmd{baseCmd: baseCmd{_args: args, _clusterKeyPos: 1}}
+	return &ScanCmd{
+		baseCmd: baseCmd{_args: args, _clusterKeyPos: 1},
+	}
 }
 
 func (cmd *ScanCmd) reset() {
 	cmd.cursor = 0
-	cmd.keys = nil
+	cmd.page = nil
 	cmd.err = nil
 }
 
-// TODO: cursor should be string to match redis type
 // TODO: swap return values
 
 func (cmd *ScanCmd) Val() (int64, []string) {
-	return cmd.cursor, cmd.keys
+	return cmd.cursor, cmd.page
 }
 
 func (cmd *ScanCmd) Result() (int64, []string, error) {
-	return cmd.cursor, cmd.keys, cmd.err
+	return cmd.cursor, cmd.page, cmd.err
 }
 
 func (cmd *ScanCmd) String() string {
-	return cmdString(cmd, cmd.keys)
+	return cmdString(cmd, cmd.page)
 }
 
 func (cmd *ScanCmd) readReply(cn *pool.Conn) error {
-	keys, cursor, err := readScanReply(cn)
+	page, cursor, err := readScanReply(cn)
 	if err != nil {
 		cmd.err = err
 		return cmd.err
 	}
-	cmd.keys = keys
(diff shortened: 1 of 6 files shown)


example_test.go
```diff
@@ -314,3 +314,58 @@ func Example_customCommand() {
 	fmt.Printf("%q %s", v, err)
 	// Output: "" redis: nil
 }
+
+func ExampleScanIterator() {
+	iter := client.ScanIterator(nil)
+
+	for iter.Next() {
+		fmt.Println(iter.Value())
```
vmihailenco: I would rename `Value` to `Key`.

vmihailenco:
Did you consider following API:

```
scan := client.Scan(0, "", 100).Result()
for scan.Next() {
    fmt.Println(scan.Key())
}
fmt.Println(scan.Err())
```

or


```
scan := client.Scan(0, "", 100).Iterator()
for scan.Next() {
    fmt.Println(scan.Key())
}
fmt.Println(scan.Err())
```

example_test.go
dim: The trouble is that it's `Key` for `SCAN` and `Field` for all the others. I rename it to `Val` to keep it consistent with `Cmd`

dim:
So yes, I did, but the trouble is that you have to parse a cursor value, which isn't exactly intuitive. Happy to adjust, just not sure it's the best option. Let me know ...

dim:
@vmihailenco PTAL ^^

command.go
```diff
@@ -685,11 +685,22 @@ func (cmd *ZSliceCmd) readReply(cn *pool.Conn) error {
 
 //------------------------------------------------------------------------------
 
+// Scanner is a ScanCmd with Iterator capabilities
+type Scanner struct {
+	*ScanCmd
+	c *commandable
+}
```
vmihailenco: Why not to remove `Scanner` and add `Iterator() *ScanIterator` method to `ScanCmd`?

iterator.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Return if there is more data to fetch
+	if i.scn.cursor == 0 {
+		return false
+	}
+
+	// Fetch next page
+	i.scn._args[1] = i.scn.cursor
+	i.scn.reset()
+	i.scn.c.Process(i.scn.ScanCmd)
```
vmihailenco: Nice trick!

vmihailenco: There is also opportunity to improve var names, e.g.:
- it.cmd._args
- it.client.Process

command.go
dim: So that would require `ScanCmd` to be able to have `commandable` too as it would need to pass it to `ScanIterator`

vmihailenco: Works for me. It is essentially the same, but now you also have additional type Scanner for this...

dim: Yes, but there is a public `NewScanCmd` API which we would need to deprecate.

vmihailenco: We can do it in v4, but I don't see why you need it. Just add new private field to existing struct...

vmihailenco: We can also change API to `client.Scan(0, "", 20).Iterator(client)`

vmihailenco:
@dim I added some changes. PTAL

command.go
dim: That seems clumsy to me, that' why I added a new type wrapping ScanCmd. When client is the origin of your command, then you have an `Iterator` method, but you can also construct your own `ScanCmd` with `NewScanCmd` if you really want to. For me, the additional type is worth the flexibility. I am happy to rename `Scanner` to `ScanCmdIterable` or something more verbose.

vmihailenco:
@dim restored `Scanner`, PTAL

dim:
LGTM

PR #313, opened 2016-05-01
Set hash keys as Go map with HMSetMap

johanstokking:
`HMSetMap` allows you to set hash keys and values using by passing a `map[string]string`. This works nicely with `HGetAllMap`


vmihailenco:
Thanks, I squashed your commits and sent https://github.com/go-redis/redis/pull/314. Closing this one.

PR #318, opened 2016-05-06
Latency Based Queries

minjatJ:
Hi guys,

I ran into few latency issues when doing queries on a redis-cluster split across multiple datacenters. 
I find out the package was always calling the first master of the slot (Which was causing latencies as master wasn't always the closest node handling the slot)

I suggest the following:
- Calculate the latency to each node just after retrieving the slots configuration
- Sending every `read only` commands to the closest node (Slave or Master)

Please let me know what you think about that or if you think of another/nicer way to do it.
Thanks,

### DIFF
--- a/command.go
+++ b/command.go
@@ -38,6 +38,7 @@ type Cmder interface {
 	clusterKey() string
 
 	Err() error
+	IsWriter() bool
 	fmt.Stringer
 }
 
@@ -98,6 +99,8 @@ type baseCmd struct {
 	_clusterKeyPos int
 
 	_readTimeout *time.Duration
+
+	_isWriter bool
 }
 
 func (cmd *baseCmd) Err() error {
@@ -130,6 +133,25 @@ func (cmd *baseCmd) setErr(e error) {
 	cmd.err = e
 }
 
+func (cmd *baseCmd) IsWriter() bool {
+	return cmd._isWriter
+}
+
+func NewBaseCmd(args []interface{}) baseCmd {
+	cmd := baseCmd{}
+	if len(args) > 0 {
+		if isWriter, ok := args[len(args)-1].(bool); ok {
+			cmd._isWriter = isWriter
+			if len(args) > 1 {
+				cmd._args = args[0 : len(args)-1]
+			}
+		} else {
+			cmd._args = args
+		}
+	}
+	return cmd
+}
+
 //------------------------------------------------------------------------------
 
 type Cmd struct {
@@ -139,7 +161,7 @@ type Cmd struct {
 }
 
 func NewCmd(args ...interface{}) *Cmd {
-	return &Cmd{baseCmd: baseCmd{_args: args}}
+	return &Cmd{baseCmd: NewBaseCmd(args)}
 }
 
 func (cmd *Cmd) reset() {
@@ -183,7 +205,9 @@ type SliceCmd struct {
 }
 
 func NewSliceCmd(args ...interface{}) *SliceCmd {
-	return &SliceCmd{baseCmd: baseCmd{_args: args, _clusterKeyPos: 1}}
+	cmd := NewBaseCmd(args)
+	cmd._clusterKeyPos = 1
+	return &SliceCmd{baseCmd: cmd}
 }
 
 func (cmd *SliceCmd) reset() {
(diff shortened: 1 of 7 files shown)


redis.go
```diff
@@ -35,6 +35,18 @@ func (c *baseClient) conn() (*pool.Conn, error) {
 			_ = c.connPool.Remove(cn, err)
 			return nil, err
 		}
+		// If readonly then send readonly command for the connection
+		if c.opt.ReadOnly {
+			cmd := newKeylessStatusCmd("READONLY", isReadOnly)
```
minjatJ: `READONLY` is only for the connection it was sent on so every new connection won't allow readonly.
Not sure it's the best place for that though but I can't find a better one.

dim:
I had a few thoughts about this one before and there are several considerations. `READONLY` is in a sense similar to `SELECT`, it must be consistently maintained on the connection pool level. Every new connection must call `READONLY` on initialisation and the most sensible sensible solution would be to add a `ReadOnly` option to `ClusterOptions`. The limitation is that users would have to maintain separate clients for RW and RO, but that's certaintly favourable to maintaining multiple internal connection pools.

In regards to load-balancing, latency is only one measure that could be used and there are several other use cases. It would probably make most sense to allow customisation and do something along the lines of:

```go
type ReadOnlyMode interface {
	SelectNode(addrs []string) string
}

type ReadOnlyModeRandom struct{}

func (ReadOnlyModeRandom) SelectNode(addrs []string) string {
	size := len(addrs)
	if size == 0 {
	  return ""
	}

	return addrs[rand.Intn(size)]
}

// passed to NewClusterClient.
type ClusterOptions struct {
	// ...

	ReadOnly ReadOnlyMode
}
```

@vmihailenco thoughts?

vmihailenco:
@dim `READONLY` definitely should be done in connection init block, but AFAIK we don't have such block for ClusterClient. So we will need to add it.

>The limitation is that users would have to maintain separate clients for RW and RO, but that's certaintly favourable to maintaining multiple internal connection pools.

As far as I understand slave node will have separate pool of readonly connections. We just need the logic to pick slave nodes for read-only commands.

vmihailenco:
@minjatJ first thanks for the PR. But it is too big for review and contains 2 separate features:
- picking node by latency
- using slave nodes for readonly commands

Can you please create 2 PRs (one PR for each feature)? Otherwise reviewing this PR will be a mess.

command.go
```diff
@@ -130,6 +133,25 @@ func (cmd *baseCmd) setErr(e error) {
...
 }
 
+func (cmd *baseCmd) IsWriter() bool {
+	return cmd._isWriter
+}
+
+func NewBaseCmd(args []interface{}) baseCmd {
+	cmd := baseCmd{}
+	if len(args) > 0 {
+		if isWriter, ok := args[len(args)-1].(bool); ok {
```
vmihailenco: Can you please explain how this works?

command.go
```diff
@@ -130,6 +133,25 @@ func (cmd *baseCmd) setErr(e error) {
 	cmd.err = e
 }
 
+func (cmd *baseCmd) IsWriter() bool {
+	return cmd._isWriter
+}
+
+func NewBaseCmd(args []interface{}) baseCmd {
```
vmihailenco: should be unexported, because it returns unexported type

command.go
```diff
@@ -130,6 +133,25 @@ func (cmd *baseCmd) setErr(e error) {
...
+	cmd := baseCmd{}
```
vmihailenco: var cmd baseCmd

vmihailenco:
Hm, but probably this PR is good as is and we don't need to split it :) I am going to create new branch in redis repo so everyone can push commits with some edits. I think such workflow is much faster than question/answer ping-pong in comments.

vmihailenco:
@minjatJ I created new branch and gave you access to add commits https://github.com/go-redis/redis/pull/319. Hopefully all this makes some sense.

vmihailenco:
>@vmihailenco thoughts?

@dim regarding `ReadOnlyMode` - it looks good, but let's get something working first and then decide if we can/should add different strategies for selecting nodes. This PR is already big enough for me.

vmihailenco:
Let's move discussion to https://github.com/go-redis/redis/pull/319

PR #319, opened 2016-05-07
Feature/cluster node latency2

vmihailenco:
Copy of https://github.com/go-redis/redis/pull/318

```
Hi guys,

I ran into few latency issues when doing queries on a redis-cluster split across multiple datacenters. 
I find out the package was always calling the first master of the slot (Which was causing latencies as master wasn't always the closest node handling the slot)

I suggest the following:
- Calculate the latency to each node just after retrieving the slots configuration
- Sending every `read only` commands to the closest node (Slave or Master)

Please let me know what you think about that or if you think of another/nicer way to do it.
Thanks,
```

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -11,6 +11,12 @@ import (
 	"gopkg.in/redis.v4/internal/pool"
 )
 
+type clusterNode struct {
+	Addr    string
+	Latency int
+	Client  *Client
+}
+
 // ClusterClient is a Redis Cluster client representing a pool of zero
 // or more underlying connections. It's safe for concurrent use by
 // multiple goroutines.
@@ -19,14 +25,14 @@ type ClusterClient struct {
 
 	opt *ClusterOptions
 
-	slotsMx sync.RWMutex // protects slots and addrs
-	addrs   []string
-	slots   [][]string
+	mu     sync.RWMutex
+	addrs  []string
+	nodes  map[string]*clusterNode
+	slots  [][]*clusterNode
+	closed bool
 
-	clientsMx sync.RWMutex // protects clients and closed
-	clients   map[string]*Client
-
-	_closed int32 // atomic
+	cmdsInfo     map[string]*CommandInfo
+	cmdsInfoOnce *sync.Once
 
 	// Reports where slots reloading is in progress.
 	reloading uint32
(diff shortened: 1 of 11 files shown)


vmihailenco:
@minjatJ PTAL. I've refactored ClusterClient internals to work with *clusterNode struct you added. That allowed to remove `clients` map and simplify code. What do you think?

Also it looks like `IsWriter` is not implemented yet. Do you have any plans? My suggest is to add function like:

```
func isReadonlyCmd(cmd Cmd) bool {
    switch cmd.args()[0] {
        case "get", "ping", ...:
            return true
        default:
            return false
    }
}
```

minjatJ:
Hi @vmihailenco,

Thanks a lot for the refacto, I definitely agree with it.
I quite like @dim's idea of using a `ReadOnlyMode` interface (In #318), I don't have much time this week but I might look into it later.
Also IsWriter() is implemented by all the modifications in command.go and commands.go.
The basic idea behind that is to specify wether a command is ReadOnly or not using the last argument when building a new command.
```
NewStringCmd("GET", key, isReadOnly)
```
So the last argument of the list will be used by `NewBaseCmd` to fill the "_isWriter" attribut (https://github.com/go-redis/redis/blob/feature/cluster-node-latency2/command.go#L140-L153)

It's maybe not the best solution as it's require from the user to specify the command attribut every-time. I'm fine with the switch case but I fear it will be too heavy (That's only < 200 strings to compare but still).
I don't see much more other ways to do it though.

vmihailenco:
>The basic idea behind that is to specify wether a command is ReadOnly or not using the last argument when building a new command.

Interesting. My first thought is that we should change `NewStringCmd(flag int, args ...interface{})` and pass it as first arg `NewStringCmd(Readonly, ....)`. That is how I would do it by default.

An interesting option is to use http://redis.io/commands/command which provides all information we need:
- position of the cluster key
- readonly flag etc

Did you consider this? @dim what do you think about using it for picking cluster key?

minjatJ:
Passing the ReadOnly parameter as first arg makes more sense, it will at least avoid mistakes as the param will be mandatory.
I used the command list to know if a command was ReadOnly or not but didn't consider to integrate it in the package.  We can maybe build an internal map of commands from that then use it when building a command.

minjatJ:
@vmihailenco Any other thoughts on this ? Do you want me to move the ReadOnly param as a first argument ?

vmihailenco:
@minjatJ sorry for delay. I really want to try using `http://redis.io/commands/command` information to get fresh info about "position of first key in argument list" and readonly flag. E.g.

```
type command struct {
    FirstKey int
    Readonly bool
}
var commands map[string]*command
```

and then using that info in `ClusterClient`. I will try to build something this week.

vmihailenco:
@minjatJ @dim please take a final look. After that I will add some tests for `slotClosestNode` and `slotSlaveNode`.

minjatJ:
Nice one, LGTM

vmihailenco:
@minjatJ that for the PR, merging. And I feel sorry that I had to revert so many changes for `isWriter/isReader` logic, but hopefully this version is cleaner :) And I still plan to add some tests later...

PR #324, opened 2016-05-25
Update README.md to use redis.v4 documentation

tuxxy:
This changes the links from redis.v3 documentation to redis.v4 documentation.


vmihailenco:
v4 is still in beta and API can change before the release. So thanks for the PR but it will not be merged until v4 is stable.

PR #329, opened 2016-06-09
Add support for SPOP with a count argument

mattias-lundell:
This PR adds support for the `count` part of `SPOP key [count]` that was introduced in REDIS 3.2. The implementation follows the implementation of `SRandMemberN`.


vmihailenco:
Thanks!

PR #333, opened 2016-06-21
Negative max tries is infinite.

tommyp1ckles:
Is there any reason not to have infinite retries if the `MaxRetries` is set to a negative value such as this (not to mention negative values are completely unhandled as it stands).

### DIFF
--- a/pipeline.go
+++ b/pipeline.go
@@ -1,6 +1,8 @@
 package redis
 
 import (
+	"errors"
+	"fmt"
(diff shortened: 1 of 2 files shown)


vmihailenco:
I don't see a compelling use case for infinite retries. But it can easily waste your resources. So I don't like this changes.

You can use `MaxRetries = math.MaxInt32` to get something similar.

PR #337, opened 2016-07-01
Ensure to use pointer methods where appropriate. Tidy up godoc.

dim:
Please check https://godoc.org/gopkg.in/redis.v4

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -61,7 +61,7 @@ type statefulCmdable struct {
 
 //------------------------------------------------------------------------------
 
-func (c statefulCmdable) Auth(password string) *StatusCmd {
+func (c *statefulCmdable) Auth(password string) *StatusCmd {
 	cmd := NewStatusCmd("auth", password)
 	c.process(cmd)
 	return cmd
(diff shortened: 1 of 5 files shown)


dim:
Sorry, unintentionally messed up branch name.

commands.go
```diff
@@ -61,7 +61,7 @@ type statefulCmdable struct {
 
 //------------------------------------------------------------------------------
 
-func (c statefulCmdable) Auth(password string) *StatusCmd {
+func (c *statefulCmdable) Auth(password string) *StatusCmd {
```
vmihailenco: Is this needed? I think `c statefulCmdable` should be faster, because `statefulCmdable` contains only 1 field which is a function pointer. So:
- `c statefulCmdable` - value containing function pointer
- `c *statefulCmdable` - pointer to the value containing function pointer

ring.go
```diff
@@ -102,7 +102,7 @@ func (shard *ringShard) Vote(up bool) bool {
 // concurrent use by multiple goroutines.
 //
 // Ring monitors the state of each shard and removes dead shards from
-// the ring. When shard comes online it is added back to the ring. This
+// the c. When shard comes online it is added back to the c. This
```
vmihailenco: Please revert

commands.go
dim: While it may indeed be quicker, the difference will be in the pico-seconds range. I think we should follow https://golang.org/doc/faq#methods_on_values_or_pointers:

> Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used. See the section on method sets for details.

dim: There is also the worry that a call to `func (c statefulCmdable) Auth(password string) *StatusCmd` would copy the whole `Pipeline/Client` struct.  I cannot confirm it, but that's what it looks like in https://godoc.org/gopkg.in/redis.v4#Pipeline.Auth

PR #338, opened 2016-07-02
Make proto/parser an internal package

dim:
I'm working on a Disque client/sub-package and this is a pre-requisite


dim:
Oh, and I have also removed unsafe bytes to string conversions as these were micro-optimisations for Go <1.5 and have less significance these days

vmihailenco:
>these were micro-optimisations for Go <1.5 and have less significance these days

That is interesting because I've added similar micro-optimisations to go-pg several days ago https://github.com/go-pg/pg/pull/272 :) But okay, I guess we can restore them later :)

mvrhov:
@dim. Is this disque package available anywhere?

PR #339, opened 2016-07-02
Internalise rate-limiter

dim:
Given that this lib is widely used, it's probably worth avoiding external dependencies, especially since it's only a few LOC.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ matrix:
     - go: tip
 
 install:
-  - go get gopkg.in/bsm/ratelimit.v1
   - go get github.com/onsi/ginkgo
   - go get github.com/onsi/gomega
   - go get github.com/garyburd/redigo/redis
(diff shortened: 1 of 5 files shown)


vmihailenco:
Probably it should be in vendor/ folder? I don't have much experience with using vendor in libraries...

dim:
So no, that's not how vendor/ works. I think using internal is most appropriate here

vmihailenco:
Can you please fix/disable this test? If fails every 2nd time I retry build.

```
• Failure [0.110 seconds]

RateLimiter

/home/travis/gopath/src/gopkg.in/redis.v4/internal/ratelimit/ratelimit_test.go:116

  should correctly spread allowance [It]

  /home/travis/gopath/src/gopkg.in/redis.v4/internal/ratelimit/ratelimit_test.go:61

  Expected

      <int>: 50

  to be ~

      <int>: 54
```

vmihailenco:
I removed rate limiter from dialer, because it was rarely useful but caused a lot of confusion. Closing this PR.

Issue #341, opened 2016-07-19
Making cmdable an interface


artushin:
Would you be amenable to making cmdable an exported interface? I ask because I'm running a single instance redis docker container in dev and a full cluster in prod, but my code relies on using the global client established in my redis package that wraps go-redis. With the current setup, I can't do that because the ClusterClient and Client are separate types without a shared interface. Since you already use cmdable on both structs, it would be easy to make that an interface that would allow access only to the high level redis interface.

Let me know if that sounds good and I can do a PR.

vmihailenco:
Released in v4.1.0

artushin:
Thanks!

PR #342, opened 2016-07-21
Feature: Export cmdable as an interface

yzprofile:
https://github.com/go-redis/redis/issues/341

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -44,6 +44,209 @@ type cmdable struct {
 	process func(cmd Cmder) error
 }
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
I personally define much smaller interfaces like this - https://github.com/go-redis/rate/blob/v4/rate.go#L14-L16. Overall such huge interfaces are strongly discouraged in Go, but I don't see a better way...

So thanks for contribution - merging. But I will rename to `Cmdable` so it matches our internal naming.

Issue #259, opened 2016-02-24
Command Logging?


pyros2097:
Is there some way I can log the commands that are being sent to redis for debugging purposes? It would be great if we could attach a logger to the client.

vmihailenco:
No, currently such logger is not supported.

vmihailenco:
See https://github.com/go-redis/redis/pull/352 for possible solution. Closing.

PR #352, opened 2016-08-16
add WrapProcess to allow users to alter command processing

jrmarkle:
This could be used to update metrics or debug/logging (#259).

There are many ways this could be done. I chose a way that involves minimal changes.

Example usage:
```
	client = redis.NewClient(...)

	client.WrapProcess(func(oldProcess func(cmd redis.Cmder) error) func(cmd redis.Cmder) error {
		return func(cmd redis.Cmder) error {
			before := time.Now()

			err := oldProcess(cmd)

			if err != nil {
				log.Println(`error running redis command "`, cmd, `": `, err)
			} else {
				log.Println(`running redis command "`, cmd, `" took `, time.Since(before))
			}

			return err
		}
	})
```

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -245,6 +245,14 @@ type cmdable struct {
 	process func(cmd Cmder) error
 }
 
+// WrapProcess replaces the process func. It takes a function createWrapper
+// which is supplied by the user. createWrapper takes the old process func as
+// an input and returns the new wrapper process func. createWrapper should
+// use call the old process func within the new process func.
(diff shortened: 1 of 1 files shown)


vmihailenco:
I like this - thanks!

vmihailenco:
Released in v4.1.3

aep:
tried this, but oldProcess is always nil for me, so the code posted by jrmarkle  crashes. is this supposed to be used differently?

jrmarkle:
`oldProcess` is just the name for `Client.cmdable.process` passed back to you via `WrapProcess`. I'm pretty sure it should never be nil, so I suspect your `Client` is not initialized properly. Post a code snippet?

PR #354, opened 2016-08-21
Support geopos command

Hepri:
Added missing GEOPOS command.

But I'm not sure about part of code below, redis could produce "nil" as element of array response. If there are another solution tell me

```
// response may contain nil results
if err == Nil {
    positions = append(positions, nil)
    continue
}
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -806,6 +806,10 @@ type GeoLocation struct {
 	GeoHash                   int64
 }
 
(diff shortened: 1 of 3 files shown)


vmihailenco:
Overall looks good - thanks. I am going to tweak it in separate branch before releasing.

vmihailenco:
Released v4.1.4

Issue #353, opened 2016-08-20
Return an interface instead of StringCmd, StatusCmd, etc.


domodwyer:
First off, fantastic library; thanks very much for your hard work - it's used a lot in our production systems and never lets me down!

If go-redis methods were to return an interface exporting the same methods `StringCmd`, `StatusCmd`, `IntCmd` (etc), then users of the library would be able to mock out the responses from redis relatively easily.

For example:
```Go
type redisInterface interface {
	Get(key string) *redis.StringCmd
	Set(key string, value interface{}, expiration time.Duration) *redis.StatusCmd
	Del(keys ...string) *redis.IntCmd
}
```
The above code requires a return value of `*redis.StringCmd` etc, which does not export it's fields, making it impossible to write unit tests (i.e. no way to return an error from `Err()` in a mock).

If it was to return:
```Go
type redisInterface interface {
	Get(key string) redis.StringCmdInterface
	Set(key string, value interface{}, expiration time.Duration) redis.StatusCmdInterface
	Del(keys ...string) redis.IntCmdInterface
}
```
then client code could use the `redisInterface` in place of the `redis.Client` type, and in tests a mock implementation of `redisInterface` could be used returning mock responses (i.e a `mockStringCmd` could be set up to return a given value when `Err()` is called).

I've seen issue https://github.com/go-redis/redis/issues/332 where writing a wrapper around go-redis, and then mocking that was suggested, and while it works I feel like an extra a level of abstraction is not the best solution when the required code change is relatively straightforward.

I am **more than happy** to create a PR changing methods to return an interface, but just wanted to check if it's something that would be considered.

vmihailenco:
Thanks for the detailed proposal - I really appreciate it.

Changing function return signature from`*redis.StatusCmd` to `redis.StatusCmdInterface` is possible, but requires releasing new major version (v5) so we do not break existing user code. Besides that you still need to create type that implements `redis.StatusCmdInterface`.

It looks much more practical to export new constructor for `*redis.StatusCmd`, e.g.:

```
func NewStatusResult(status string, err error) *StatusCmd {
	var cmd StatusCmd
	cmd.val = status
	cmd.err = err
	return &cmd
}
```

What do you think?

domodwyer:
This is a fantastic workaround!

I'm happy to send in a PR (it'd be nice to give something back) if you'd like?

Dom

vmihailenco:
>I'm happy to send in a PR (it'd be nice to give something back) if you'd like?

[...]

PR #356, opened 2016-08-22
Add Cmd constructors for testing/mock purposes.

domodwyer:
As per issue https://github.com/go-redis/redis/issues/353

I've not included tests because they're relatively simple, not for production use, and if I'm honest, I'm unfamiliar with the testing framework used.

Let me know if this is a problem and I'll write some up.

### DIFF
--- a/result.go
+++ b/result.go
@@ -1,0 +1,140 @@
+package redis
+
+import "time"
+
+// NewCmdResult returns a Cmd initalised with val and err for testing
+func NewCmdResult(val interface{}, err error) *Cmd {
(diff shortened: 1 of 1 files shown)


domodwyer:
Looking at the failed [build](https://travis-ci.org/go-redis/redis/jobs/154264969) (1 of 2) it doesn't seem to be at all related to code I've committed.

Not really sure what's going on here?

vmihailenco:
Thanks - merging!

>Not really sure what's going on here?

Flaky tests - nothing more.

PR #361, opened 2016-09-08
fix iterator across empty pages

vkd:
Iterator may return empty page with non zero cursor.
I fix early return from iterator by "for" loop.

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -46,25 +46,30 @@ func (it *ScanIterator) Next() bool {
 		return true
 	}
 
-	// Return if there is more data to fetch.
-	if it.ScanCmd.cursor == 0 {
-		return false
-	}
+	for {
+		// Return if there is more data to fetch.
+		if it.ScanCmd.cursor == 0 {
+			return false
+		}
 
-	// Fetch next page.
-	if it.ScanCmd._args[0] == "scan" {
-		it.ScanCmd._args[1] = it.ScanCmd.cursor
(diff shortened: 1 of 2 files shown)


iterator.go
```diff
@@ -46,25 +46,30 @@ func (it *ScanIterator) Next() bool {
 		return true
 	}
 
-	// Return if there is more data to fetch.
-	if it.ScanCmd.cursor == 0 {
-		return false
-	}
+	for {
+		// Return if there is more data to fetch.
```
vmihailenco: Should not this say "Return if there is no more data to fetch."?

iterator.go
```diff
@@ -46,25 +46,30 @@ func (it *ScanIterator) Next() bool {
...
+		it.ScanCmd.reset()
+		it.client.process(it.ScanCmd)
+		if it.ScanCmd.Err() != nil {
+			return false
+		}
 
-	it.pos = 1
-	return len(it.ScanCmd.page) > 0
+		it.pos = 1
+		if len(it.ScanCmd.page) > 0 {
```
vmihailenco: Please add some comment like "Redis can occasionally return empty page."

vmihailenco:
Looks good - just 2 minor comments.

vmihailenco:
Thanks - merging.

vmihailenco:
Released v4.1.6 with your fix.

PR #363, opened 2016-09-09
pool.go:close conn when remove conn

lijunfei:
hi editor:
     when origin code run some time, redis server has many client link, but redis client log show that the own link is few.
     I find when remove conn from pool, the conn has not been closed.
     So I suggest this pull request.
     wish editor to give feedback.   
                                                                                                                           from lijunfei

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -222,6 +222,7 @@ func (p *ConnPool) Remove(cn *Conn, reason error) error {
 	p.connsMu.Lock()
 	p.remove(cn, reason)
 	p.connsMu.Unlock()
+	cn.Close()
 	p.queue <- struct{}{}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


internal/pool/pool.go
```diff
@@ -222,6 +222,7 @@ func (p *ConnPool) Remove(cn *Conn, reason error) error {
 	p.connsMu.Lock()
 	p.remove(cn, reason)
 	p.connsMu.Unlock()
+	cn.Close()
```
vmihailenco: Connection is already closed on line 221.

vmihailenco:
@lijunfei thank you for this PR, but I am afraid proper fixing requires more broader refactoring I did in https://github.com/go-redis/redis/pull/364/files. But thanks for your work anyway and sorry for discarding it...

vmihailenco:
@lijunfei I think I fixed this issue in v4.1.7. Please try it.

lijunfei:
@vmihailenco thank you for you reply, I will try it.

PR #376, opened 2016-09-26
Disable idle checks for cluster connection

evalphobia:
This PR disables reap connection for cluster.

`ClusterOptions.clientOptions()` does not copy `IdleCheckFrequency` field, but `Options.init()` sets default parameter (1min).
So `ConnPool.reaper()` is enabled in `NewConnPool()`.

I'm not sure the concept to disable idle check for cluster connection is right decision but I follow the comment in `ClusterOptions.clientOptions()`

```go
    // IdleCheckFrequency is not copied to disable reaper
```

If not, I'll create another PR to reflect IdleCheckFrequency from cluster options.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -627,6 +627,8 @@ func (opt *ClusterOptions) init() {
 }
 
 func (opt *ClusterOptions) clientOptions() *Options {
+	const disableIdleCheck = -1
+
 	return &Options{
 		Password: opt.Password,
 		ReadOnly: opt.ReadOnly,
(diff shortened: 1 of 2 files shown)


vmihailenco:
@evalphobia you are right - we have 1 global connection reaper for the whole cluster (https://github.com/go-redis/redis/blob/v4/cluster.go#L439-L466) so we disable individual reapers. But  can you please remove `DisableIdleCheck` and instead set `IdleCheckFrequency = -1`? I think it will have the same effect and we don't need to add additional option.

evalphobia:
@vmihailenco  Alright, I fixed it 🌻

vmihailenco:
Thanks!

PR #379, opened 2016-09-29
PubSub. Use appendIfNotExists instead of append for channels and patterns

Hepri:
Fixed bug when connection loss to server leads to exponential grow of channels and patterns slice in PubSub in every reconnect


vmihailenco:
Thanks!

PR #381, opened 2016-09-30
Add new parameter to the RingOptions to override CommandInfo first key

Unix4ever:
There is problem with `eval` and `evalsha` commands.
`COMMAND INFO eval` returns first key position `0`.
After that, redis ring chooses `eval` as a value for sharding.
Then, if you try to delete created value, ring may choose another shard
and delete won't work.
    
Eval command should be parsed, to be sharded properly, according to
redis specs: http://redis.io/commands/command .
    
I've introduced a new flag in the `RingOptions`, which will enable new
behavior: `RouteByEvalKeys `.
    
If it is enabled, `cmdFirstKey` will check if function is  `eval` or
`evalsha` commands and will check if number of params -- `cmd.arg(2) != "0"`.
If it has parameters, it will return `3`, otherwise it will return `0`.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -138,6 +138,7 @@ func redisRingOptions() *redis.RingOptions {
 		PoolTimeout:        30 * time.Second,
 		IdleTimeout:        500 * time.Millisecond,
 		IdleCheckFrequency: 500 * time.Millisecond,
+		RouteByEvalKeys:    true,
 	}
 }
 
--- a/ring.go
+++ b/ring.go
@@ -40,6 +40,9 @@ type RingOptions struct {
 	PoolTimeout        time.Duration
 	IdleTimeout        time.Duration
 	IdleCheckFrequency time.Duration
+
+	// RouteByEvalKeys flag to enable eval and evalsha key position parsing for sharding
+	RouteByEvalKeys bool
 }
 
 func (opt *RingOptions) init() {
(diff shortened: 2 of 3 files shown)


Unix4ever:
Not sure about this one:
> * do all the work in `cmdFirstKey`
Script command has variable number of parameters, and it can have no parameters at all. 
So there will be no `cmd.args(3)`. That's why I've implemented this first argument position getter in the cmd interface.

If you want to remove this method, I can do a bit different way:
`cmd.args(2)` always means number of keys for the `eval` and `evalsha` commands.
This argument is string, so I'll have to parse it back to int and it will be a bit less slower.

Tell me what you prefer, please

vmihailenco:
>But if you think that we can ignore that, I can remove this parameter, of course.

Okay, let's keep it since enabling it by default means that users will miss their data. But can you please rename it to something like `RouteByEvalKeys bool`? In v5 we can enable it by default.

>Tell me what you prefer, please

Let's do as you suggested - check `cmd.arg(2)`. I think you don't need to actually parse it - just check that `cmd.arg(2) != "0"` and then use `cmd.arg(3)`.

Unix4ever:
Ok, fixed :)

vmihailenco:
Thanks and sorry for delay!

vmihailenco:
After some of thinking I decided to remove `RouteByEvalKeys` and release it in v5. Otherwise people will continue using v4 without enabling `RouteByEvalKeys`...

PR #385, opened 2016-10-02
Enable reaper on ClusterClient and add tests.

vmihailenco:


vmihailenco:
Fixes https://github.com/go-redis/redis/pull/383

PR #387, opened 2016-10-05
Remove Addr field from clusterNode

yyoshiki41:
To avoid a duplicate definition,
remove `Addr` field from `clusterNode`.

thanks!


vmihailenco:
Thank you!

PR #390, opened 2016-10-09
if readonly is open, read from master when slave is loading data

lijunfei:
I meet a problem in cluster; When slave failover, other slave will resync data from new master, if readonly is open, service can't get data in some minute. So I suggest to get data from new master in the sync duration.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -300,6 +300,17 @@ func (c *ClusterClient) Process(cmd Cmder) error {
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks. Though we can probably do better and remeber that node is loading and remove it from the list fro a minute or so. Going to do it in separate PR.

lijunfei:
Thanks for your advise.

Issue #395, opened 2016-10-12
Discussion: remove custom import


dim:
How do you feel about removing the custom `// import "gopkg.in/redis.v5"` and allowing people to import the package as `github.com/go-redis/redis`? Two reasons:

1. Most people use some sort of vendor tools these days and they support semver, e.g. https://github.com/Masterminds/glide. You are using tags already!
2. The gopkg.in servers are unfortunately not as stable and robust as I wish they were. We have experienced several failures on builds, CI runs, etc because of non-200 responses.

Thoughts?
dim

vmihailenco:
No problem at all. Can you please send PR with some bold warning in the readme that you are on your own if you import `github.com/go-redis/redis`? Also we probably need to resurrect master branch?

dim:
Closing in favour of #397

PR #397, opened 2016-10-12
Remove static import

dim:
Two reasons:

1. Most people use some sort of vendor tools these days and they support semver, e.g. https://github.com/Masterminds/glide. You are using tags already!
2. The gopkg.in servers are unfortunately not as stable and robust as I wish they were. We have experienced several failures on builds, CI runs, etc because of non-200 responses.

Replaces #395

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,31 @@ Examples: http://godoc.org/gopkg.in/redis.v5#pkg-examples.
 
 Install:
 
-    go get gopkg.in/redis.v5
+```shell
+go get gopkg.in/redis.v5
+```
+
+Import:
+
+```go
+import "gopkg.in/redis.v5"
+```
+
+## Vendoring
+
+If you are using a vendoring tool with support for semantic versioning
+e.g. [glide](https://github.com/Masterminds/glide), you can import this
+package via its GitHub URL:
+
+```yaml
+- package: github.com/go-redis/redis
+  version: ^5.0.0
+```
+
+WARNING: please note that by importing `github.com/go-redis/redis`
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks!

vmihailenco:
But does glide handles imports like https://github.com/go-redis/redis/blob/v4/cluster.go#L9-L12 ?

dim:
It does, although it's a little inconvenient

vmihailenco:
Okay, we could probably use relative import if that helps, e.g. `./internal/errors`

dim:
Yep, that would be preferable

vmihailenco:
Unfortunately Go does not allow relative imports of internal packages - https://github.com/golang/go/issues/12282

ewang:
@dim Wondering what glide configuration you're using to get/vendor the package. When I do glide get, it's downloading the gopkg.in/redis.v5 version of the package as well since all the source code still reference that url as the import path. Additionally, trying to run the code gives me:
```
imports github.com/go-redis/redis
imports gopkg.in/redis.v5/internal/hashtag: use of internal package not allowed
```

dim:
Yeah, same problem here, importing the package via its github URL doesn't
work for the very same reason.

I doubt there is much sense in this PR as long as internal packages cannot
be imported via their relative paths. We will probably have to live with
gopkg.in for a while longer

dim:
@vmihailenco any chance of a non-gopkg v6 release any time soon?

vmihailenco:
May be in a week or two? There are some PRs that I'd like to review & merge before v6.

PR #398, opened 2016-10-13
fix broken cmdable interface of redis.Tx

yzprofile:


vmihailenco:
Hi,

Thanks for contribution, but I have a [PR](https://github.com/go-redis/redis/pull/403) that changes `Tx` to implement `Cmdable` interface. Though it is only available in v5. Therefore I am inclined to not merge this PR so:
- we don't introduce new `BaseCmdable` interface that is removed in v5
- users have motivation to upgrade to new version :)

yzprofile:
yep, I agree with that.

But this patch doesn't break any interface of v4, people will not aware of this change. 

And I notice that v5 is still in developing, before it finished, I think we can fix this broken in v4.

vmihailenco:
It does not break anything, but users will start relying on `BaseCmdable` which then is removed in v5.

That said - let's merge it. It is small and I assume it blocks you.

PS I just released v5.

PR #404, opened 2016-10-13
Use RWMutex in pool

yyoshiki41:
Replace `Mutex` to `RWMutex`.
Use `RLock()` to spend waiting for locks with minimal time.

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/go-redis/redis/404)

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -58,10 +58,10 @@ type ConnPool struct {
 
 	queue chan struct{}
 
-	connsMu sync.Mutex
+	connsMu sync.RWMutex
 	conns   []*Conn
 
-	freeConnsMu sync.Mutex
+	freeConnsMu sync.RWMutex
 	freeConns   []*Conn
 
 	stats Stats
@@ -238,17 +238,17 @@ func (p *ConnPool) remove(cn *Conn, reason error) {
 
 // Len returns total number of connections.
 func (p *ConnPool) Len() int {
-	p.connsMu.Lock()
+	p.connsMu.RLock()
 	l := len(p.conns)
-	p.connsMu.Unlock()
+	p.connsMu.RUnlock()
 	return l
 }
 
 // FreeLen returns number of free connections.
 func (p *ConnPool) FreeLen() int {
-	p.freeConnsMu.Lock()
+	p.freeConnsMu.RLock()
 	l := len(p.freeConns)
-	p.freeConnsMu.Unlock()
+	p.freeConnsMu.RUnlock()
 	return l
 }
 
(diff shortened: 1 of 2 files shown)


dim:
:+1:

vmihailenco:
This makes benchmarks 20% slower

```
benchmark                          old ns/op     new ns/op     delta
BenchmarkPoolGetPut10Conns-4       360           437           +21.39%
BenchmarkPoolGetPut100Conns-4      374           456           +21.93%
BenchmarkPoolGetPut1000Conns-4     345           440           +27.54%
```

And I doubt that code like

```
	p.connsMu.Lock()
 	l := len(p.conns)
	p.connsMu.Unlock()
```

can benefit from using `RWMutex`.

If we are to micro-optimize this I would
1. use atomics to manage Len and FreeLen
2. remove synchronization from Len and FreeLen since we don't rely on these number to be super accurate

But again I doubt that will speedup real programs...

dim:
I can only see this being a benefit if used in the contexts of a large,
public web app, with potentially thousands of active goroutines. In that
case, you could avoid the threads blocking each others reads, but that
doesn't really apply for the majority of use cases.

yyoshiki41:
@vmihailenco @dim 

That benchmark may be correct.
In terms of implementation, `RWMutex.Lock()` is totally different from `Mutex.Lock()` :thinking: 

- [RWMutex.Lock](https://golang.org/src/sync/rwmutex.go?s=2399:2424#L70)
- [Mutex.Lock](https://golang.org/src/sync/mutex.go?s=1143:1165#L34)

This benchmark example shows same results.
(`RWMutex.Lock` is slower than `Mutex.Lock()`.)

```bash
$ go test -bench .
BenchmarkMutex-4     	   10000	    104281 ns/op
BenchmarkRWMutex-4   	   10000	    150461 ns/op
```

<details>
 <summary>Code</summary>
```go
package main

import (
	"sync"
	"testing"
)

var (
	mu    sync.Mutex
	cnt   int
	rwMu  sync.RWMutex
	rwCnt int
)

func BenchmarkMutex(b *testing.B) {
	b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			for i := 0; i < 1000; i++ {
				mu.Lock()
				cnt++
				mu.Unlock()
			}
		}
	})
}

func BenchmarkRWMutex(b *testing.B) {
	b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			for i := 0; i < 1000; i++ {
				rwMu.Lock()
				rwCnt++
				rwMu.Unlock()
			}
		}
	})
}
```
</details>

We should take the frequency of `RLcok` into consideration.
I think that the number of calling `Len()` and `FreeLen()` are small(actually zero in benchmark),
so we can't benefit from using RWMutex.

Thanks for your help!!

yyoshiki41:
I'll close this PR 🍣

vmihailenco:
@yyoshiki41 thank you anyway!

PR #409, opened 2016-10-21
Fix "invalid expire time in set" for SetXX with expiration = 0

bpiddubnyi:


commands.go
```diff
@@ -788,10 +788,14 @@ func (c *cmdable) SetNX(key string, value interface{}, expiration time.Duration)
 // Zero expiration means the key has no expiration time.
 func (c *cmdable) SetXX(key string, value interface{}, expiration time.Duration) *BoolCmd {
 	var cmd *BoolCmd
-	if usePrecise(expiration) {
-		cmd = NewBoolCmd("set", key, value, "px", formatMs(expiration), "xx")
+	if expiration == 0 {
+		cmd = NewBoolCmd("set", key, value, "xx")
 	} else {
-		cmd = NewBoolCmd("set", key, value, "ex", formatSec(expiration), "xx")
+		if usePrecise(expiration) {
```
vmihailenco: You can fold it into previous if:

```
} else if usePrecise(expiration) {
```

vmihailenco:
Could you please add a test to https://github.com/go-redis/redis/blob/v5/commands_test.go#L998-L1013 ?

vmihailenco:
Thanks - merging.

PR #420, opened 2016-11-14
Add support for parsing redis:// and rediss:// URLs

freeformz:
This includes setting up a default dialer that handles the ssl
handshake.

We (Heroku) would like to move secure redis connections forward, but
have this chicken and egg problem with some drivers. Drivers for other
languages already support rediss://

### DIFF
--- a/options.go
+++ b/options.go
@@ -64,7 +64,7 @@ type Options struct {
 	// Enables read only queries on slave nodes.
 	ReadOnly bool
 
-	// Config to use when connecting via TLS
+	// TLS Config to use. When set TLS will be negotiated.
 	TLSConfig *tls.Config
 }
 
@@ -74,7 +74,12 @@ func (opt *Options) init() {
 	}
 	if opt.Dialer == nil {
 		opt.Dialer = func() (net.Conn, error) {
-			return net.DialTimeout(opt.Network, opt.Addr, opt.DialTimeout)
+			conn, err := net.DialTimeout(opt.Network, opt.Addr, opt.DialTimeout)
+			if opt.TLSConfig == nil || err != nil {
+				return conn, err
+			}
+			t := tls.Client(conn, opt.TLSConfig)
+			return t, t.Handshake()
 		}
 	}
 	if opt.PoolSize == 0 {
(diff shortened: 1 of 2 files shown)


options.go
```diff
@@ -92,6 +100,70 @@ func (opt *Options) init() {
...
+			conn, err := net.DialTimeout(o.Network, o.Addr, o.DialTimeout)
+			if err != nil {
+				return nil, err
+			}
+			if o.TLSConfig == nil {
+				o.TLSConfig = &tls.Config{InsecureSkipVerify: true}
+			}
+			t := tls.Client(conn, o.TLSConfig)
+			return t, t.Handshake()
+		}
```
vmihailenco: I think this should be moved to https://github.com/go-redis/redis/blob/v5/options.go#L70

options.go
```diff
@@ -92,6 +100,70 @@ func (opt *Options) init() {
...
+		h = "localhost"
+	}
+	if p == "" {
+		p = "6379"
+	}
+	o.Addr = net.JoinHostPort(h, p)
+
+	f := strings.FieldsFunc(u.Path, func(r rune) bool {
+		return r == '/'
+	})
```
vmihailenco: Is not this the same as `strings.Split(u.Path, "/")`? Just asking :)

options_test.go
```diff
@@ -0,0 +1,94 @@
...
+			"",
+			0, false, errors.New("Invalid redis database number: strconv.ParseInt: parsing \"iamadatabase\": invalid syntax"),
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.u, func(t *testing.T) {
+			o, err := ParseURL(c.u)
+			if c.err == nil && err != nil {
+				t.Fatalf("Expected err to be nil, but got: '%q'", err)
```
vmihailenco: nitpics:
- start with lower case
- got goes first, want (expected) goes second
- no need to quote `%q` (at least I don't see why)

For example:
- unexpected error: %q
- got %q, want %q

vmihailenco:
This looks good - thanks! I've left some minor comments, but they are not blockers.

I wonder if it is possible to connect to Redis Cluster using `redis://` URLs? I spent few minutes on it but can't google anything...

options.go
freeformz: Nope https://play.golang.org/p/OAOF51zR-a

freeformz:
@vmihailenco Pushed changes to address feedback.

freeformz:
WRT redis cluster, we could address that in another PR as well.

vmihailenco:
Looks good - thanks

PR #428, opened 2016-11-27
Fixed Readme typo ZRangeBy

nathan-jones:
Correct me if I'm wrong, but it should by ZrangeBy not ZRangeByScore?


vmihailenco:
Yeah, thanks

PR #436, opened 2016-12-05
Add MGET support for Redis Ring

sergeylanzman:
Add MGET support with Redis Ring. Mget for each shard.
Currently this synchrony without goroutines. I don't know if this ok open SHARD_COUNT goroutines for each MGET request with 10K rps?

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -427,3 +427,39 @@ func (c *Ring) pipelineExec(cmds []Cmder) (firstErr error) {
 
 	return firstErr
 }
+
+func (c *Ring) MGet(keys ...string) *SliceCmd {
+	allValues := make([]interface{}, len(keys))
+	keysByShard := map[*ringShard][]string{}
+	shards := []*ringShard{}
+	for _, key := range keys {
+		shard, _ := c.shardByKey(key)
+		if _, exist := keysByShard[shard]; !exist {
+			shards = append(shards, shard)
+		}
+		keysByShard[shard] = append(keysByShard[shard], key)
+	}
+	k := 0
+	for _, shard := range shards {
+		keysForShard := keysByShard[shard]
+		args := make([]interface{}, 1+len(keysForShard))
+		args[0] = "mget"
+		for i, key := range keysForShard {
+			args[1+i] = key
+		}
+		cmd := NewSliceCmd(args...)
+		shard.Client.Process(cmd)
+		for _, val := range cmd.val {
+			allValues[k] = val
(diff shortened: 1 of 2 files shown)


vmihailenco:
There are (too) many Redis commands that accept several keys. If we decide to merge this PR then we will need to also support all other multi-keys Redis commands, which is too much work spread across all commands. Instead I suggest to use `MGet` only if all keys end on the same shard/node and use `Pipeline` in all other cases. AFAIK that is consistent with Redis Cluster.

sergeylanzman:
Yep Redis have many commands start MXXX. We can support for all. Now MGET work wrong, if I change redis client to redis ring, I not receive every error and i can receive value from redis by MGET, but this wrong. I think we have to options:
1. Support all MXXX commands
2. Return error with all MXXX with ring.

vmihailenco:
Supporting all commands is a lot of work which I am not ready to do. Therefore returning an error sounds like a much better option to me, but the check should be general enough so we don't have to copy it for every 100+ Redis commands...

sergeylanzman:
@vmihailenco Ok I want implement MGET for ring in my code. I need shardByKey to be public and ringShard public. If this possible?

sergeylanzman:
@vmihailenco Why 100+ commands? MGET/MSET ?

vmihailenco:
>Why 100+ commands? MGET/MSET ?

What about `Del(keys ...string)`? You can expect commands.go for a list of commands that accept several keys.

>I need shardByKey to be public and ringShard public. If this possible?

It does not sound too bad, but it is not enough. You will also need `cmd.name()` to get command name, `cmd.arg(4)` to get command argument(s), `cmdInfo` to get access to command info and probably something else. That is a lot.

sergeylanzman:
let's start from MGET support and  future add other commands. This better solution then we have today. If some peoples need solution for other commands send pr or fetcher request.

eyalpost:
I don't think it makes sense to implement result aggregation (like for MGET) for every command that accepts multiple keys. DEL, EXISTS make sense. But stuff like SDIFF and BLPOP - too complicated.
I think it makes sense to add a function which checks if a command uses different shards and returns an error in that case and then call it in all functions that use multiple keys and slowly add support for some of the commands.

(Btw: Seems like ObjectIdleTime gets multiple keys but I don't think Redis supports that)

sergeylanzman:
better implementations  https://github.com/go-redis/redis/pull/493

PR #438, opened 2016-12-09
fix cmdInfo nil pointer panic

rayzyar:
fix panic when cmdInfo is nil

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1,6 +1,7 @@
 package redis
 
 import (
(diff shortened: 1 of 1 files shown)


rayzyar:
I'm not sure how to add a test

vmihailenco:
I think once https://github.com/go-redis/redis/pull/439 is merged this PR is not needed. But thanks anyway.

PR #445, opened 2016-12-13
Another attempt at removing gopkg.in dependency

dim:
As discussed before, this would mean that we would have to maintain a separate master branch with pure-github package references and this would be the new master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,3 @@ install:
   - go get gopkg.in/bsm/ratelimit.v1
   - go get github.com/onsi/ginkgo
   - go get github.com/onsi/gomega
-  - mkdir -p $HOME/gopath/src/gopkg.in
-  - mv `pwd` $HOME/gopath/src/gopkg.in/redis.v5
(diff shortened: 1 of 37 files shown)


vmihailenco:
I guess you don't want to actually merge this to v5 branch?

>this would be the new master branch.

Can you please elaborate? We can't just create master branch, because that would mean that you can only use latest version (master)...

vmihailenco:
Probably the easiest way forward is to just release v6 without gopkg.in...

dim:
No, I don't want to merge it into v5, there is just no other way to create a PR and raise your awareness. 

Releasing a v6 is certainly an option. I - personally - am not sure how useful gopkg now is, in the era of vendoring (since Go1.5) but using it was your conscious decision so I don't want to "bully" you into dropping it :)

vmihailenco:
Okay, let's create branch v6 without gopkg.in. But I'd like to have some time to check if there are any breaking changes queued for v6 before tagging v6.0.0 release...

vmihailenco:
Closing, there is master branch with this changes.

PR #452, opened 2016-12-19
Setting node latency can take a while

minjatJ:
When resolving node latency, each nodes is being ping a couple of times (10 right now).

In case of large cluster spread across the world initialisation can take a while as each node is ping one after the other. Doing each node in parallel will decrease the general initialisation time.

NB: Doing it against V4 right now as my projects are still using this version but I will try to push one against V5 soon.


vmihailenco:
Sounds good - thanks. In v5 we can use `ForEachNode`.

Issue #453, opened 2016-12-20
Why HSET(key stirng,failed string,value string) doesn't support "value" as interface{}


szyhf:
Why HSET(key stirng,failed string,value string) doesn't support "value" as interface{} ?

I don't know why HSET doesn't support value as interface{}, I tried make a test function as : 

```
func (c *cmdable) HSet_Interface(key, field, value interface{}) *BoolCmd {
	cmd := NewBoolCmd("hset", key, field, value)
	c.process(cmd)
	return cmd
}

redis.RedisHandler.HSet_Interface("Hello", "World", some_struct)
redis.RedisHandler.HGet("Hello","World")
```

It works well, is it any other reason ?

Thanks .

vmihailenco:
AFAIK it is an overlook. Do you want to sent a PR changing the signature to

```
HSet(key, field string, value interface{})
```

?

szyhf:
Sure, I've already made one, let me try ~

szyhf:
@vmihailenco Should I pull to another branch ?

vmihailenco:
Do you mean v4? No, we don't backport changes to older versions.

szyhf:
Oh, nope, some project has rules that new pull request should not push to the master branch, I was wandering if my PR is push to the right branch yesterday ~ lol

vmihailenco:
There is no master branch - only v5, v4, v3... So all is good - thanks!

PR #455, opened 2016-12-20
PR try for issue #453

szyhf:
Allow HSET \ HSETNX to set value as interface{}.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -109,8 +109,8 @@ type Cmdable interface {
 	HLen(key string) *IntCmd
 	HMGet(key string, fields ...string) *SliceCmd
 	HMSet(key string, fields map[string]string) *StatusCmd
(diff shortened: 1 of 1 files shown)


commands.go
```diff
@@ -109,8 +109,8 @@ type Cmdable interface {
 	HLen(key string) *IntCmd
 	HMGet(key string, fields ...string) *SliceCmd
 	HMSet(key string, fields map[string]string) *StatusCmd
-	HSet(key, field, value string) *BoolCmd
-	HSetNX(key, field, value string) *BoolCmd
+	HSet(key, field, value interface{}) *BoolCmd
```
vmihailenco: Can you please leave key, field to be a string? This is consistent with other commands.

szyhf: Sure, I forgot it ...

PR #463, opened 2016-12-23
Support TLS on Sentinel and Cluster Client

linxGnu:
Add TLS config to sentinel and cluster client options

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1,6 +1,7 @@
 package redis
 
 import (
+	"crypto/tls"
 	"fmt"
 	"math/rand"
 	"sync"
@@ -18,6 +19,10 @@ var errClusterNoNodes = internal.RedisError("redis: cluster has no nodes")
 // ClusterOptions are used to configure a cluster client and should be
 // passed to NewClusterClient.
 type ClusterOptions struct {
+	// The network type, either tcp or unix.
+	// Default is tcp.
+	Network string
+
 	// A seed list of host:port addresses of cluster nodes.
 	Addrs []string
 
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -18,6 +19,10 @@ var errClusterNoNodes = internal.RedisError("redis: cluster has no nodes")
 // ClusterOptions are used to configure a cluster client and should be
 // passed to NewClusterClient.
 type ClusterOptions struct {
+	// The network type, either tcp or unix.
+	// Default is tcp.
+	Network string
```
vmihailenco: I don't think unix sockets are used with Redis Cluster, because you need TCP/IP anyway for Redis Cluster to work.

cluster.go
```diff
@@ -45,6 +50,12 @@ type ClusterOptions struct {
 	PoolTimeout        time.Duration
 	IdleTimeout        time.Duration
 	IdleCheckFrequency time.Duration
+
+	// Enables read only queries on slave nodes.
+	ReadOnly bool
```
vmihailenco: I don't think this belongs here.

sentinel.go
```diff
@@ -17,6 +18,10 @@ import (
 // FailoverOptions are used to configure a failover client and should
 // be passed to NewFailoverClient.
 type FailoverOptions struct {
+	// The network type, either tcp or unix.
+	// Default is tcp.
+	Network string
```
vmihailenco: Same - unix sockets don't make much sense if you want to use automatic failover.

sentinel.go
```diff
@@ -37,6 +42,12 @@ type FailoverOptions struct {
 	PoolTimeout        time.Duration
 	IdleTimeout        time.Duration
 	IdleCheckFrequency time.Duration
+
+	// Enables read only queries on slave nodes.
+	ReadOnly bool
```
vmihailenco: Same.

vmihailenco:
Closing since does not look good as is. Please reopen with fixes if you are still interested.

PR #465, opened 2016-12-30
 WrapProcess is broken in v5

smacker:
Failed test added. Works for me in v4.


vmihailenco:
Thanks. Going to check why it does not work.

vmihailenco:
Fixed in v5.1.7

PR #473, opened 2017-01-10
Add "ScanSlice" to StringSliceCmd to allow scan into custom slice.

szyhf:
It can help StringSliceCmd using in Set\ZSet and so on to quickly turn string response into custom slice.

Not only struct but also normal built-in value like "[]string", "[]int",... can work well ( []chan may not work well ).

Example: 
```golang
//FeedComment is a simple struct implement "encoding.BinaryMarshaler" and "encoding.BinaryUnmarshaler"

redisClient := ...

key := "TestStringSliceScan"
intCmd := redisClient.SAdd(key, &FeedComment{ID: 1}, &FeedComment{ID: 2}, &FeedComment{ID: 3})

stringSliceCmd := redisClient.SMembers(key)
var container []FeedComment
if err := stringSliceCmd.ScanSlice(&container); err == nil {
        fmt.Printf("%v", container)
} else {
        fmt.Printf("%v", err)
}
```

The container can be []YourStruct or []*YourStruct, but don't use []interface{}.

### DIFF
--- a/internal/proto/scan.go
+++ b/internal/proto/scan.go
@@ -4,7 +4,8 @@ import (
 	"encoding"
 	"fmt"
 	"strconv"
-
+	"reflect"
+	
 	"gopkg.in/redis.v5/internal"
 )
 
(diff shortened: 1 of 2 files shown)


vmihailenco:
I think code can be cleaned up a bit :) I will take a look a bit later.

szyhf:
Sure, I'd think for the 

```
if isElemPtr {

}else{

}
```

for a long time, using "func block" or put it in loop, what do you think of it ?

internal/proto/scan.go
```diff
@@ -105,3 +106,63 @@ func Scan(b []byte, v interface{}) error {
...
 	}
 }
+
+
+// Scan a string slice into a custom container
+// Example:
+// var container []YourStruct; ScanSlice([]string{""},&container)
+// var container []*YourStruct; ScanSlice([]string{""},&container)
+func ScanSlice(sSlice []string, container interface{}) error {
+	val := reflect.ValueOf(container)
```
vmihailenco: You can/should immediately check if value is valid here. I think you can use [this code](https://github.com/go-pg/pg/blob/v5/types/scan.go#L52-L62) I wrote for another project. Then you should be able to get rid of `errTyp` flag.

internal/proto/scan.go
```diff
@@ -105,3 +106,63 @@ func Scan(b []byte, v interface{}) error {
...
+	errTyp := true
+	isElemPtr := true
+	if val.Kind() == reflect.Ptr {
+		if ind.Kind() == reflect.Slice {
+			errTyp = false
+			isElemPtr = elemType.Kind() == reflect.Ptr
+		}
+	}
+	if errTyp {
+		return fmt.Errorf("wrong object type `%s` for rows scan, need *[]*Type or *[]Type", val.Type())
```
vmihailenco: Go convention is to use `redis: ScanSlice(non-pointer %s)`. Similar error message is returned by json package.

internal/proto/scan.go
```diff
@@ -105,3 +106,63 @@ func Scan(b []byte, v interface{}) error {
...
+			elem := reflect.New(elemType)
+			mind := reflect.Indirect(elem)
+
+			if err := Scan(s, mind.Addr().Interface()); err != nil {
+				return fmt.Errorf("scan slice failed at index of %d: %s", index, err.Error())
+			}
+			// If elem is ptr, append with value
+			slice = reflect.Append(slice, mind)
+		}
+	}
```
vmihailenco: I suggest to rewrite this using function from https://github.com/go-pg/pg/blob/v5/internal/util.go#L5-L21. It:
- reuses slice elements if slice has any elements
- should avoid extra allocation when slice contains values, not pointers

It can be a bit slower, because `isElemPtr` check is done on every loop. But I think we can optimize it later.

vmihailenco:
Thanks for the PR and sorry for long delay. I left some comments about the code and it would be nice to have at least one test. But otherwise this approach looks good.

internal/proto/scan.go
szyhf: I'll check immediately, but I'm not sure what is "errTyp" flag, can you show me an example? My native language is not English and some description may has another meaning for me.
Thx.

szyhf: I can understand "error" "type" "flag", but I'm not sure what does it means when combines together.

szyhf: Oh....I just see my code =。=   

errTyp ...

internal/proto/scan.go
szyhf: Got it, I'm on my vacation and the network is not very well, I'll rewrite as soon as I can.

vmihailenco: Cool, thanks

internal/proto/scan.go
vmihailenco: :)

Issue #471, opened 2017-01-05
WrapProcess doesn't work for cloned client


smacker:
I track redis calls for every request. (I need to know how much time every endpoint consumed and how: go-itself/db/redis/external-http-requests)

In v4 I did this:
```
func WrapRedisClient(ctx context.Context, c *redis.Client) *redis.Client {
	txn := GetTnxFromContext(ctx)
	if txn == nil {
		return c
	}

	cCopy := *c
	cCopyP := &cCopy

	cCopyP.WrapProcess(RedisWrapper(txn))

	return cCopyP
}
```

So it measured time for redis for every transaction separately.

It doesn't work for v5 anymore because you set `client.cmdable.process = client.Process` in init, so `cmdable.process` points on original client. Do you think you can fix it? Or my case is out of your scope and I have to downgrade/fork?

smacker:
if it will support context - it will be even better for my purposes. Thanks a lot!

smacker:
fixed in #474

PR #474, opened 2017-01-11
Clone and WithContext #471

smacker:
`client.Copy()` method which will properly set `client.cmdable.process = client.Process`
And `client.WithContext(c)` that will do the same + set context.

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -1,6 +1,7 @@
 package redis // import "gopkg.in/redis.v5"
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"time"
@@ -23,6 +24,8 @@ type baseClient struct {
 
 	process func(Cmder) error
 	onClose func() error // hook called when client is closed
+
+	ctx context.Context
 }
 
 func (c *baseClient) String() string {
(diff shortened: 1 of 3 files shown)


redis_go1.7.go
```diff
@@ -0,0 +1,35 @@
+// +build go1.7
```
vmihailenco: Add go1.8 too and rename to `redis_context.go`? go1.8 should be released really soon.

redis_legacy.go
```diff
@@ -0,0 +1,15 @@
+// +build !go1.7
```
vmihailenco: Probably rename to `redis_no_context.go`.

redis.go
```diff
@@ -309,6 +301,13 @@ func NewClient(opt *Options) *Client {
 	return newClient(opt, newConnPool(opt))
 }
 
+func (c *Client) Clone() *Client {
```
vmihailenco: Can we unexport this?

Please rename `Clone` to `Copy`. I think this is more idiomatic name :)

vmihailenco:
Overall looks good. Left some minor comments to consider.

redis_go1.7.go
smacker: `"go1.7", from Go version 1.7 onward`
https://golang.org/pkg/go/build/

As I understand this build tag covers all version >=1.7

smacker:
Comments are fixed.

vmihailenco:
Thanks! Merging.

PR #480, opened 2017-01-26
ObjectIdleTime accepts one key.

vmihailenco:
Updates https://github.com/go-redis/redis/pull/436


eyalpost:
This is of course relevant for all Object* commands:
ObjectRefCount and ObjectEncoding should be fixed too..

PR #493, opened 2017-02-13
Add multi-key operation support for Redis ring

eyalpost:
This PR basically allows to perform operations such as MGET or MSET on a ring client and let the client partition and dispatch the commands to multiple shards and aggregate the results.
How this works? All Redis commands are categorized to four types: 

- No key (Operations like info, keys, scan etc.)
- Single key (Operations like GET, SET etc.)
- Multiple keys with partition\aggregate support (MGET, MSET, DEL, EXISTS and UNLINK)
- Multiple keys with single shard support (all the rest: e.g. MSETNX, BLPOP etc.)

How do we handle each type of command?
- For "No key" and "Single key", we continue to work as previous version. if a key is present we find the appropriate shard and send the command to that shard.
- For command that support aggregation, we use a partitioner\aggregator to create several new commands from the original command, send the commands to the appropriate shard, get the results and aggregate them. For DEL, EXISTS we aggregate by summing the results. For MGET, we combine the results to a new slice ordered according to the original command order
- For commands that support only single shard, we check that all keys in that command belong to the same shard and if so allow it to run. If not, we return an error. This is better than the current situation where inconsistencies can occur because the command will simply run on the first shard but some of the keys do not belong to it.

Notes: Most of the logic is in the ring and the command_paritioner files. The implementation is generic in that:
1. It's not specific for ring and can later be adapted to cluster clients
2. If new commands are added to Redis that can support aggregation, it is easy to support them since the implementation uses commandInfo metadata to find keys in commands and is not specific to MGET, MSET etc.

I added several tests that show the previous implementation didn't work correctly while the new one does.


vmihailenco:
Overall I think that this is right approach, but code requires some refactoring before it can be merged. So ff you are interested to move this forward
- be prepared that review will take some time :)
- please split this PR into several small PRs, because otherwise this PR will stuck in review for month(s)
- give me opportunity to leave some feedback before writing so much code

I think good first step would be to extend `Cmd` interface to clone & aggregate commands, e.g.:

```
keys := []string{"key1", "key2", "key3"}
cmd := NewIntCmd()
for _, key := range keys {
    cp := cmd.copy() // returns a zero copy
    processKey(key, cp)
    cmd.aggregate(cp) // aggregates result, e.g. cmd.value += cp.value
}
```

eyalpost:
I have no problem doing that step by step, but let's first agree on the concepts:
- In which branch should this be developed (now that a v6 branch was opened)?
- Do you agree with the command categorization I suggested and how to handle each category?

Then, let's agree on design and how this will developed in multiple PRs (Not sure how this works?)

eyalpost:
2 more comments: 
- I think it would be good to separate command splitting and aggregation to a different class rather than implementing it within the command itself. 
- It will be better to implement aggregate such that it receives all the cloned command rather than one by one as you suggested above. In the case of the MGET command, aggregating one by one is less efficient because we need to keep the original sort order.

vmihailenco:
>In which branch should this be developed (now that a v6 branch was opened)?

Let's do it against master (ex v6) so we don't have to back/forward-port it.

>Do you agree with the command categorization I suggested and how to handle each category?

I am not sure, e.g. I'd prefer not to have "single key" category at all so commands with single key don't have performance impact. But overall you spent on this much more time than I did and I hope that if we do it step-by-step I will catch up with you.

>I think it would be good to separate command splitting and aggregation to a different class rather than implementing it within the command itself.

But why? My point is that each command (e.g. IntCmd and StringCmd) will have aggregation specific to those commands and I don't see the benefit of having IntCmd and IntAggregation. Also we have full access to cmd private state.

>In the case of the MGET command, aggregating one by one is less efficient because we need to keep the original sort order.

I think there is big difference. In first case you will have to sort string values.  In second case you will have to sort commands and pass them sorted to aggregator.

eyalpost:
> Let's do it against master (ex v6) so we don't have to back/forward-port it.

No problem.

> I am not sure, e.g. I'd prefer not to have "single key" category at all so commands with single key don't have performance impact

The reason i'm doing this categorization at first, was to make sure single key command don't have any performance impact because they work in the regular way (i.e. no splitting\aggregation)


>My point is that each command (e.g. IntCmd and StringCmd) will have aggregation specific to those commands and I don't see the benefit of having IntCmd and IntAggregation

The problem is that aggregation is not a feature of the command **type** but rather a feature of the **command**. For example, if we later support PFCount aggregation, then we can create a different aggregator for it because it's different from the DEL command aggregator (even though both are IntCmd). If you think it's better to implement inside the command itself then I'll go for it, but for me it feels a bit weird that a command aggregates other commands. Sort of like why String.Join('x','y','z') feels better than 'x'.join('y','z') 
 
>I think there is big difference. In first case you will have to sort string values. In second case you will have to sort commands and pass them sorted to aggregator.

Not sure what you mean by that. Let me try to explain why MGET is a bit more complex. Imagine you're doing `MGET key1 key2 key3`
key1 and key3 are in shard1 
key2 is in shard2
to be efficient, we want to convert this to 2 commands:
`MGET key1 key3` and `MGET key2`
now when we get the results, if we aggregate them one by one, most likely the result of the original command will be `val1 val3 val2` (By the way, this is a bug that exists in #436 - results are not returned in the requested order). so what we have to do is get the results from **all** commands and place them correctly according to the original sort order.

vmihailenco:
>but for me it feels a bit weird that a command aggregates other commands

Okay, let's do it separately but please keep as simple as possible, i.e. instead of passing some general interfaces let's pass `IntCmd` explicitly, e.g.:

```
func sumIntCmd(cmd *IntCmd, part *IntCmd)
```

>all commands and place them correctly according to the original sort order.

You still can do sorting if you have access to keys:

```
func agg(cmd *KeysCmd, part *KeysCmd) {
    // Find position for part.key.
    for i, key := range cmd.keys {
        if key == part.key {
            cmd.values[i] = part.value
            return
        }
    }
    panic("unknown key " + part.key)
}
```

Probably I over-simplified it, but I think it should work.

eyalpost:
>`func sumIntCmd(cmd *IntCmd, part *IntCmd)`

Hmm.. Not sure how this will work. If you look at the PR, you'll see the code that calls command splitting and aggregation is located in the `func (c *Ring) Process(..)` method. here's a snippet:
```
partitioner := cmd.(cmdPartitionable).createPartitioner(cmdInfo)
cmds, err := partitioner.partition(...
...
c.runCommandsByShard(cmds);
return partitioner.aggregate(cmds, cmd)
```
so in this location, the cmd is Cmder and not IntCmd.. How do you see this work?

>`func agg(cmd *KeysCmd, part *KeysCmd)`

I think you assume that a command which has 30 keys will be splitted to 30 commands. But this isn't the case. the splitting is done by shards

yuekui2:
We need this feature for our work as well. When this is expected to be released? Thanks. -Kui

sergeylanzman:
@vmihailenco Whats new about multi-key for ring?

vmihailenco:
`MGET` across shards is not supported in Redis Cluster so that is one good reason why it is not supported in Ring too. I guess the actual reason is that such commands should work atomically, which can't be guaranteed with multiple keys spread across multiple shards.

PR #496, opened 2017-02-15
ensure wait.done to avoid deadlock in test

eyalpost:
Seems like in Go 1.4 sometimes the error returned from pubsub.ReceiveMessage is "use of closed network connection" error instead of "redis: client is closed". 
This cause the go routine to fail before it marks the waitGroup which causes the test to wait forever 
This commit doesn't fix the test but at least the build will fail sooner.


vmihailenco:
Awesome, thanks!

PR #498, opened 2017-02-16
Added implementation for WAIT command

natethebosch:
Reference: https://redis.io/commands/wait

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -271,9 +271,9 @@ func (c *cmdable) Ping() *StatusCmd {
 	return cmd
 }
 
-func (c *cmdable) Wait(n_slaves uint, n_seconds uint) *IntCmd {
+func (c *cmdable) Wait(numSlaves int, timeout time.Duration) *IntCmd {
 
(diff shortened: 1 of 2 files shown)


natethebosch:
Not sure if this command was purposely left out of the commands list... but any rate here's an implementation.

commands.go
```diff
@@ -271,6 +271,13 @@ func (c *cmdable) Ping() *StatusCmd {
 	return cmd
 }
 
+func (c *cmdable) Wait(n_slaves uint, n_seconds uint) *IntCmd {
```
vmihailenco: Go convention is to use camelCased var names. So it should be `numSlaves, numSeconds`. Overall I suggest to change types to `numSlaves int, timeout time.Duration` so it reads better:

```
client.Wait(1, time.Minute)
```

natethebosch: Good suggestions, I've made those changes

vmihailenco:
Thanks - merging.

PR #499, opened 2017-02-17
Added support for multi-client

dim:
I've been maintaining https://github.com/bsm/multiredis for a while as it simplifies many use cases and testing scenarios. It has been really useful and I would like to include it into the main repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Redis client for Golang [![Build Status](https://travis-ci.org/go-redis/redis.png?branch=v5)](https://travis-ci.org/go-redis/redis)
+# Redis client for Golang [![Build Status](https://travis-ci.org/go-redis/redis.png?branch=master)](https://travis-ci.org/go-redis/redis)
 
 Supports:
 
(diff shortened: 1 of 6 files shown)


multi.go
```diff
@@ -0,0 +1,133 @@
+package redis
+
+import "time"
+
+// MultiOptions information is required by MultiClient to establish
+// connections. There are To connect to a cluster
```
vmihailenco: `There are To connect` seems to be truncated.

vmihailenco:
I think name `Multi` is misleading and confusing, because only one client is created. And comments/docs don't say multi or multiple too which is a bad sign. May be we can find a better name? OpaqueClient? SomeClient? :)

Overall this looks rather specific, but since there are docs and comments I don't mind merging it into master.

dim:
How about `UniversalClient`?

vmihailenco:
Sounds good - please rebase this against master to be merged into master branch.

dim:
Updated, tests on Travis now pass too

PR #512, opened 2017-02-24
the timeout of WAIT command is in milliseconds.

shogo82148:
I found that the timeout of WAIT command is too short.

[The document](https://redis.io/commands/wait) says that the timeout is in milliseconds,

> If the timeout, specified in milliseconds, is reached, the command returns even if the specified number of slaves were not yet reached.

but go-redis treats it as seconds.

``` go
cmd := NewIntCmd("wait", numSlaves, int(timeout/time.Second))
```

I tried WAIT command actually(in redis 3.2.8).
`wait 1 1000` timed out in 1000ms = 1s.

``` plain
127.0.0.1:6379> role
1) "master"
2) (integer) 0
3) (empty list or set)
127.0.0.1:6379> wait 1 1000
(integer) 0
(1.07s)
```

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -275,7 +275,7 @@ func (c *cmdable) Ping() *StatusCmd {
 
 func (c *cmdable) Wait(numSlaves int, timeout time.Duration) *IntCmd {
 
-	cmd := NewIntCmd("wait", numSlaves, int(timeout/time.Second))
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks, merging.

shogo82148:
Thank you for your quick reply.

Issue #521, opened 2017-03-03
Using FailoverClient, the service could become frozen after master switches


yuekui2:
We find there is a bug in the pool.go when using FailoverClient. After one or several master switches, the service could become frozen. After some debugging, we found that it is due to timer.Reset() in PopFree() and Get(). According to the godoc, the return value of Reset() is just for backward compatibility, and it is NOT usable:

https://golang.org/pkg/time/#Timer.Reset
> Note that it is not possible to use Reset's return value correctly, as there is a race condition between draining the channel and the new timer expiring. Reset should always be invoked on stopped or expired channels, as described above. The return value exists to preserve compatibility with existing programs.

That is, it is possible the timer's channel had been drained and the drain inside the if timer.Reset() block will block forever.

As the knowledge to drain the timer channel or not is only known at the put() back to pool time, the timer should be tried to stop and drain as necessary.  The following shows the correct code snippet for Get().

```
func (p *ConnPool) Get() (*Conn, bool, error) {
...
        timer := timers.Get().(*time.Timer)
	timer.Reset(p.poolTimeout)

	select {
	case p.queue <- struct{}{}:
		if !timer.Stop() {
			<-timer.C
		}
		timers.Put(timer)
	case <-timer.C:
		timers.Put(timer)
		atomic.AddUint32(&p.stats.Timeouts, 1)
...
```

I tried to fork and submitted a fix, but ran into the trouble that all gopkg.in imports become github.com/go-redis. Not sure if this is a real issue. If it is, it is appreciated that someone could tell me how to resolve it.

Thanks.
-Kui

vmihailenco:
Hi,

Please submit a fix to v5 branch that supports gopkg.in. Then we can port it to master.

vmihailenco:
Thanks for the report and investigation!

PR #522, opened 2017-03-03
Fix wrong usage of timer Reset(), causing service frozen during master switch.

yuekui2:
Fix for issue #521.


vmihailenco:
Thanks. I created with build fixed https://github.com/go-redis/redis/pull/523.

Issue #526, opened 2017-03-13
ring mode: clean cache when a shard is recovered to avoid stale cache data


yuekui2:
In ring mode, a recovered shard could contain stale cache data. Should have logic to clean the cache before adding the shard back.

PR #528, opened 2017-03-16
Optionally reset a recovered before adding back to ring

yuekui2:
ring mode: clean cache when a shard is recovered to avoid stale cache data #526

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -65,6 +65,23 @@ func ExampleNewRing() {
 	client.Ping()
 }
 
+func ExampleShardResetCallback(c *redis.Client) (bool, error) {
+	return redis.FlushShardSafe(c, 0, 15, "ShardResetMarkKey", time.Minute)
+}
+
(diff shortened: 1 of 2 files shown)


vmihailenco:
Sorry for loooong delay, but I think this is too fragile even for Redis Ring which does not provide any guarantees about data consistency. E.g. this PR will work poorly in situation when one of the shards is not available for 5-10 seconds from only one of the client servers (due to network issues).

Instead I recommend to:
- set expiration on all keys, e.g. https://github.com/go-redis/cache sets default expiration to 24 hours
- disable persistence in Redis Server so when server crashes it starts with no keys/data

yuekui2:
it makes sense. thanks for your comments.

vmihailenco:
Thanks for your patience.

I think we could provide some general callbacks like `OnShardDown` and `OnShardUp` and then let clients to do whatever they like, e.g. check if shard was unavailable from multiple clients and do something about. Does it sound interesting?

PR #540, opened 2017-04-02
gofmt

poloskin:


vmihailenco:
Thanks, I am surprised this passes on Go 1.4 :)

PR #544, opened 2017-04-11
Simplify PubSub API

vmihailenco:


gimler:
How can we check if the redis subscribe was succefull or not?

old version
```
pubsub, err := client.Subscribe(RedisConfig.redisChannel)
if err != nil {
    log.Error(nil, "Cannot subscribe to Channel", err.Error())
    panic(err)
}
```

vmihailenco:
New version:

```
pubsub := client.Subscibe()
err := pubsub.Subscribe(RedisConfig.redisChannel)
```

This is done this way so you can ignore error returned by subscribe and let `PubSub` to reconnect when Redis becomes available.

PR #547, opened 2017-04-17
doc: mismatched IdleTimeout default value

flisky:


options.go
```diff
@@ -55,7 +55,7 @@ type Options struct {
 	PoolTimeout time.Duration
 	// Amount of time after which client closes idle connections.
 	// Should be less than server's timeout.
-	// Default is to not close idle connections.
+	// Default is 5 minutes.
```
vmihailenco: Where did you get this value?

vmihailenco: Nvn, I was looking at wrong file...

PR #552, opened 2017-04-21
Fix typo in ClusterClient.TxPipelined()

paulzhol:
It should be using ClusterClient.TxPipeline() and not ClusterClient.Pipeline().


vmihailenco:
Thanks :)

Issue #555, opened 2017-04-26
Redis Pipeline Interface


rodrigo-brito:
Hello,

I'm having problems with Redis Pipeline for tests. Actually, the Pipeline method from RedisClient is:
```go
type Cmdable interface {
	Pipeline() *Pipeline
```
If exists a Pipeline Interface for this return will help a lot.
To override it in a external Mock will be more easy.
This is possible?

felipejfc:
also suffering with that :/

felipejfc:
we can close that now! @rodrigo-brito see https://github.com/go-redis/redis/pull/557

rodrigo-brito:
Boa! @felipejfc.

PR #557, opened 2017-05-01
pipeline now has its own interface "Pipeliner"

felipejfc:
@vmihailenco can you take a look on it?
I'm not sure with what I did here https://github.com/felipejfc/redis/blob/6fca4d5ad05fdb809fb3ffafcb2b30c6be8b41f7/pipeline.go#L101-L107
To make possible for Pipeline to implement Cmdable thus not having to create yet another Cmdable-like interface.

This must solve  https://github.com/go-redis/redis/issues/555

Regards

### DIFF
--- a/pipeline.go
+++ b/pipeline.go
@@ -9,7 +9,7 @@ import (
 
 type pipelineExecer func([]Cmder) error
 
-type Pipelineable interface {
+type Pipeliner interface {
 	Cmdable
 	StatefulCmdable
 	Process(cmd Cmder) error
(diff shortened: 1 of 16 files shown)


pipeline.go
```diff
@@ -9,6 +9,17 @@ import (
 
 type pipelineExecer func([]Cmder) error
 
+type Pipelineable interface {
```
vmihailenco: Can you please rename this to `Pipeliner`? I think it makes some sense and is more idiomatic.

`Cmdable` name was chosen because `Cmder` means something completely different...

vmihailenco: Can you also add runtime type assertion under `Pipeline` definition: `var _ Pipeliner = (*Pipeline)(nil)`

vmihailenco:
Looks good. I don't like the fact that we have to create yet another interface but it looks harmless enough. I will merge this tomorrow if none objects.

felipejfc:
just renamed it @vmihailenco

vmihailenco:
Thanks!

vmihailenco:
Tagged v6.3.0

PR #558, opened 2017-05-03
Client TxPipeline method should also return a Pipeliner

felipejfc:
@vmihailenco Sorry, I've left one method returning the struct instead of the new interface, my bad...
Can you merge this and retag?

Thanks!


vmihailenco:
Sure, tagged v6.3.1

PR #560, opened 2017-05-07
Add test case for Getting from empty Hash

cnu:
In consistenthash, getting from an empty hash (before keys are added)
trying to Get a value should return an empty string.

Now has 100% coverage on consistenthash package

### DIFF
--- a/internal/consistenthash/consistenthash_test.go
+++ b/internal/consistenthash/consistenthash_test.go
@@ -34,6 +34,13 @@ func TestHashing(t *testing.T) {
 		return uint32(i)
 	})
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
This is a copy of package from https://github.com/camlistore/camlistore. So we don't modify its source code and/or tests.

Can you instead try to replace `internal/consistenthash` with https://godoc.org/github.com/golang/groupcache/consistenthash?

cnu:
I realised it was a copy from groupcache. I am working on a PR on the groupcache's repo to add the tests there. 
Closing this PR.

PR #570, opened 2017-05-25
Backoff retry 

### LINKED ISSUE #562: Backoff Retry During Network Errors
Would it be possible to have Go-Redis retry connections when maxretries is set with a backoff sleep interval? (https://github.com/jpillora/backoff) 

During intervals when my network is overloaded, it is not using backoff and making the overloaded network connection even worst by retrying immediately and producing other errors.

I'd be happy to submit a PR if this is a feature the community/maintainer wants to see in this library.

jonmchan:
Fixes #562 .

Notes:

* Made ```max number of clients reached``` a retryable error
* Backoff Max Time is configurable via config options

Please let me know if I need to adjust anything to make this PR mergable... Thanks!

### DIFF
--- a/internal/errors.go
+++ b/internal/errors.go
@@ -18,7 +18,7 @@ type RedisError string
 func (e RedisError) Error() string { return string(e) }
 
 func IsRetryableError(err error) bool {
-	return IsNetworkError(err) || err == ErrPoolTimeout || err.Error() == "ERR max number of clients reached"
+	return IsNetworkError(err) || err.Error() == "ERR max number of clients reached"
 }
 
 func IsInternalError(err error) bool {
--- a/internal/errors.go
+++ b/internal/errors.go
@@ -1,10 +1,15 @@
 package internal
 
 import (
+	"errors"
 	"io"
 	"net"
 	"strings"
 )
+
+
+var ErrClosed = errors.New("redis: client is closed")
+var ErrPoolTimeout = errors.New("redis: connection pool timeout")
 
 const Nil = RedisError("redis: nil")
 
@@ -13,7 +18,7 @@ type RedisError string
 func (e RedisError) Error() string { return string(e) }
 
 func IsRetryableError(err error) bool {
-	return IsNetworkError(err)
+	return IsNetworkError(err) || err == ErrPoolTimeout || err == RedisError("ERR max number of clients reached")
 }
 
 func IsInternalError(err error) bool {
--- a/internal/errors.go
+++ b/internal/errors.go
@@ -18,7 +18,7 @@ type RedisError string
 func (e RedisError) Error() string { return string(e) }
 
 func IsRetryableError(err error) bool {
-	return IsNetworkError(err) || err == ErrPoolTimeout || err == RedisError("ERR max number of clients reached")
+	return IsNetworkError(err) || err == ErrPoolTimeout || err.Error() == "ERR max number of clients reached"
 }
 
 func IsInternalError(err error) bool {
--- a/options.go
+++ b/options.go
@@ -35,8 +35,8 @@ type Options struct {
 	MaxRetries int
 
 	// Retry using exponential backoff wait algorithm between each retry
-	// Default is 512 milliseconds; set to -1 to disable
-	MaxBackoffRetryTimeout time.Duration
+	// Default is 512 milliseconds; set to -1 to disable any backoff sleep
+	MaxRetryBackoff time.Duration
 
 	// Dial timeout for establishing new connections.
 	// Default is 5 seconds.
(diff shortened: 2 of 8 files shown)


vmihailenco:
There is no point in retrying `ErrPoolTimeout`. If you are willing to wait more for a connection then you can just increase `PoolTimeout` * number of retries.

internal/errors.go
```diff
@@ -1,19 +1,24 @@
...
+
 const Nil = RedisError("redis: nil")
 
 type RedisError string
 
 func (e RedisError) Error() string { return string(e) }
 
 func IsRetryableError(err error) bool {
-	return IsNetworkError(err)
+	return IsNetworkError(err) || err == ErrPoolTimeout || err == RedisError("ERR max number of clients reached")
```
vmihailenco: `err.Error() == "ERR max number of clients reached"` reads better and does not allocate.

options.go
```diff
@@ -34,6 +34,10 @@ type Options struct {
 	// Default is to not retry failed commands.
 	MaxRetries int
 
+	// Retry using exponential backoff wait algorithm between each retry
+	// Default is 512 milliseconds; set to -1 to disable
+	MaxBackoffRetryTimeout time.Duration
```
vmihailenco: I think this should be called just `MaxRetryBackoff`.

options.go
```diff
@@ -108,6 +112,11 @@ func (opt *Options) init() {
 	if opt.IdleCheckFrequency == 0 {
 		opt.IdleCheckFrequency = time.Minute
 	}
+	if opt.MaxBackoffRetryTimeout == 0 {
```
vmihailenco:
```
switch opt.MaxRetryBackoff:
case -1:
case 0:
```

redis.go
```diff
@@ -96,9 +96,17 @@ func (c *baseClient) WrapProcess(fn func(oldProcess func(cmd Cmder) error) func(
...
 	for i := 0; i <= c.opt.MaxRetries; i++ {
+		if i > 0 {
+			time.Sleep(internal.RetryBackoff(i, c.opt.MaxBackoffRetryTimeout))
+		}
+
 		cn, _, err := c.conn()
 		if err != nil {
 			cmd.setErr(err)
+
+			if err != nil && internal.IsRetryableError(err) {
```
vmihailenco: Remove `err != nil && `, because we check for nil 2 lines above.

redis_test.go
```diff
@@ -156,6 +156,51 @@ var _ = Describe("Client", func() {
...
 
+
+	It("should retry with backoff", func() {
+		Expect(client.Close()).NotTo(HaveOccurred())
+
+		// use up all the available connections to force a fail
+		connectionHogClient := redis.NewClient(&redis.Options{
+			Addr:       redisAddr,
+			MaxRetries: 1,
+		})
```
vmihailenco: defer connectionHogClient.Close()

redis_test.go
```diff
@@ -156,6 +156,51 @@ var _ = Describe("Client", func() {
...
+		startRetry := time.Now()
+		err = clientRetry.Ping().Err()
+		Expect(err).To(HaveOccurred())
+		elapseRetry := time.Since(startRetry)
+
+		Expect(elapseRetry > elapseNoRetry).To(BeTrue())
+
+		clientRetry.Close()
+		clientNoRetry.Close()
+		connectionHogClient.Close()
```
vmihailenco: this should be replaced with `defer clientRetry.Close()` etc

vmihailenco:
Overall looks fine, but please fix the build

options.go
jonmchan: Do you want me to change https://github.com/Malwarebytes/redis/blob/a3e42eccb641f8f4bcee2cc1a71ab23593de9564/options.go#L101 and https://github.com/Malwarebytes/redis/blob/a3e42eccb641f8f4bcee2cc1a71ab23593de9564/options.go#L96 too? I was just following the current style of the file. This syntax is used for the other timeouts as well.

redis.go
jonmchan: Should the ```err != nil &&``` check be removed from https://github.com/go-redis/redis/pull/570/files#diff-bbc64daf788a83d9a6daf7056d49d2a8R117 as well? Also, I noticed that ```err = readReply(cn)``` (https://github.com/go-redis/redis/pull/570/files#diff-bbc64daf788a83d9a6daf7056d49d2a8R124) is missing a corresponding ```cmd.setErr(err)```. Is that necessary to run that function when readReply returns error?

internal/errors.go
```diff
@@ -1,19 +1,24 @@
...
 import (
+	"errors"
 	"io"
 	"net"
 	"strings"
 )
 
+
+var ErrClosed = errors.New("redis: client is closed")
+var ErrPoolTimeout = errors.New("redis: connection pool timeout")
```
vmihailenco: remove these 2 vars?

vmihailenco:
I will finish this after merging. Thanks for contributing this!

vmihailenco:
>Also, I noticed that err = readReply(cn) (https://github.com/go-redis/redis/pull/570/files#diff-bbc64daf788a83d9a6daf7056d49d2a8R124) is missing a corresponding cmd.setErr(err). Is that necessary to run that function when readReply returns error?

It is confusing and requires cleanup, but readReply is reponsible to set `cmd.value` and `cmd.err`.

PR #584, opened 2017-06-27
Vendor go4.org/syncutil/once

kassiansun:
As `go4.org` [official document](https://github.com/bradfitz/go4#details) stated:
> If you want to use go4, vendor it.

We only use `go4.org/syncutil/once.go`, so it should be trivial to vendor it into our source tree.

Skip `once_test.go` to cut back dependencies.

### DIFF
--- a/vendor/go4.org/syncutil/once.go
+++ b/vendor/go4.org/syncutil/once.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2014 The Camlistore Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


vmihailenco:
Can you please also remove it from https://github.com/go-redis/redis/blob/master/.travis.yml#L19 to make sure everything works?

kassiansun:
We need to test it on machines without `go vendor`. If user can't use `vendor`(lower version `go` or they just don't use `vendor` to manage dependencies), they can use `go get` to handle this dependency, or they can use our vendored one silently (if supported by the `go` toolchain).

`go test` on 1.8 is actually using our vendored `syncutil`, the `go get` one is overridden.

vmihailenco:
What about just moving the package to internal and using `github.com/go-redis/redis/internal/syncutil` so it works for everyone?

kassiansun:
But this is an external dependency, `go` has its own standard way to handle it. `syncutil` is not part of this repository, so I don't think we should put it at `internal` directory.

vmihailenco:
The only difference I see between `vendor` and `internal` folders is that internal works for every Go version. So why not to make it work?

PR #591, opened 2017-06-30
WithContext method added to Cmdable interface

### LINKED ISSUE #590: Cmdable interface is missing WithContext method
Issue title speaks for itself.

kop:
Fixes #590.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -238,6 +238,7 @@ type Cmdable interface {
 	GeoDist(key string, member1, member2, unit string) *FloatCmd
 	GeoHash(key string, members ...string) *StringSliceCmd
 	Command() *CommandsInfoCmd
+	WithContext(ctx context.Context) Cmdable
 }
 
 type StatefulCmdable interface {


kop:
Hmm, after the tests I see this is not that easy as I thought. Is it safe to add WithContext method to types like `Tx`, `Ring`, `ClusterClient`, `Pipeline`?

vmihailenco:
Perhaps, but could describe your use case so I better understand how context is used?

kop:
@vmihailenco, context is very simple - testing :)

I'd like to replace my current `*redis.Client` with `Cmdable` for testing purposes.
Speaking about the way I use contexts, well I'm passing it to every Redis command I'm issuing.

I did not really checked does this package respects context deadlines or not, but in any case that's the behaviour that is expected.

vmihailenco:
So looks like we need to:
- at least use context.Deadline() to set ReadDealine and WriteDeadline on the TCP connection
- add context to Tx, Ring, ClusterClient, Pipeline
- only support Go 1.8 & 1.9 so we don't add too many build tags (context is only available in Go 1.8)

I don't think there is real need to support cancelation signals, but that would be nice too.

kop:
I think it should not be hard to support context cancellation too.
We can always use this trick:

```go

	ctx := context.TODO()
	tcpResult := make(chan interface{})
	defer close(tcpResult)
	
	go func() {
		// Do TCP stuff
		tcpResult <- "TCP result"
	}()
	
	select {
	case result := <-tcpResult:
		// horay!
	case <-ctx.Done():
		// damn it
	}
```

vmihailenco:
Closing as stale.

Issue #603, opened 2017-07-24
ReadOnly option should only be used by cluster client.


v2e4lisp:
According to the doc https://redis.io/commands/readonly READONLY is a cluster command so the corresponding `ReadOnly` option should only be used by a cluster client.  However, in the current implementation, `ReadOnly` can be set for single-node client, which leads to a `This instance has cluster support disabled` error when `baseClient.initConn` is called.

It would be a easy fix to just rename `ReadOnly` to `readOnly`.

vmihailenco:
Sounds good. Do you want to send a PR?

PR #604, opened 2017-07-25
Limit Readonly Option to cluster client only

v2e4lisp:
As described in https://github.com/go-redis/redis/issues/603, this PR rename `ReadOnly` to `readOnly` so that it can only be used by a cluster client.


vmihailenco:
Thanks

PR #635, opened 2017-09-11
Add Cluster.DBSize

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/632


tanxiaolong:
result of DBSize() is different from DbSize()??

vmihailenco:
DbSize is deprecated in favor of DBSize.

Issue #641, opened 2017-09-23
ReceiveMessage buffer full


xtuc:
We are streaming some big JSON objects through the pub/sub mechanism of Redis. Sometimes we hit the internal limit of the ReadLine buffer and the Redis client returns a `bufio: buffer full` error instead of something.

Note that the default size of the buffer is at `4096` bytes.

It would be great to either handle the `isPrefix` returned by ReadLine or allow us to configure the buffer passed to ReadLine at initialization.

See

https://github.com/go-redis/redis/blob/74d5147d860804ea34633071d0c91a7f4b0b5c3a/internal/proto/reader.go#L62-L64

If you agree with that, I'll be happy to make a PR since I've already the code to handle that somewhere.

vmihailenco:
Can we please start with a test that reproduces the issue?

xtuc:
https://github.com/go-redis/redis/pull/642

xtuc:
Any news from this? Can I help you with something? Do we agree on https://github.com/go-redis/redis/pull/642#issuecomment-331711647 now?

vmihailenco:
You can help by writing an app that uses `ReceiveMessage` and reproduces this issue.

xtuc:
As far as I understand, that's what I explained in https://github.com/go-redis/redis/pull/642#issuecomment-331734202. 

`PubSub.ReceiveMessage` uses your ReadLine wrapper under the hood. I was asking if you agree with that. I just followed the function calls with "go to definition".

netgusto:
Seems pretty clear from the golang documentation that `isPrefix` is not an semantically an error, but a flag informing the consumer that there are more bytes to read on the line.

https://golang.org/pkg/bufio/#Reader.ReadLine

This could (should?) be handled in this code as proposed by @xtuc 
https://github.com/go-redis/redis/blob/74d5147d860804ea34633071d0c91a7f4b0b5c3a/internal/proto/reader.go#L57-L72

vmihailenco:
I still did not see a program that reproduces this issue.

netgusto:
Damn, too bad.

eelcocramer:
I get this issue when retrieving large blobs from [redux]() which is API compliant with redis.

I created a test that reproduces that error. It happens when reading from a byte array which length is larger than 4096.

https://github.com/eelcocramer/redis/blob/46c6ac7debcb631c52554a9a990a1f7d6295a71f/internal/proto/reader_test.go#L30-L41

[...]

PR #642, opened 2017-09-24
Add failing test for ReceiveMessage buffer full

xtuc:
Failing test for https://github.com/go-redis/redis/issues/641

### DIFF
--- a/internal/proto/reader_test.go
+++ b/internal/proto/reader_test.go
@@ -1,6 +1,7 @@
 package proto_test
 
 import (
+	"bufio"
 	"bytes"
 	"strings"
 	"testing"
@@ -9,6 +10,10 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+)
+
+const (
+	bufioDefaultBufSize = 4096
 )
 
 var _ = Describe("Reader", func() {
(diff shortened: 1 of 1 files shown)


internal/proto/reader_test.go
```diff
@@ -11,6 +12,10 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+const (
+	bufioDefaultBufSize = 4096
```
xtuc: Bufio doesn't export this value, I had to copy it. I hope it won't change often.

internal/proto/reader_test.go
```diff
@@ -36,6 +41,26 @@ var _ = Describe("Reader", func() {
...
+		bytes := make([]byte, bufioDefaultBufSize+1)
+		for i := range bytes {
+			bytes[i] = 'x'
+		}
+
+		str := string(bytes)
+
+		p := proto.NewReader(strings.NewReader(str + "\r\n"))
+
+		data, err := p.ReadLine()
```
xtuc: Note that even if the string is split over multiple `Reader.ReadLine` calls, it's still only one line since there is not line breaker. I would expect the `p.ReadLine()` to return the entire string at once.

vmihailenco:
Well, this is a test for bufio/proto package, not go-redis. Can you write an app that uses go-redis (PubSub?) and demonstrates the problem you reported?

xtuc:
The `ReadLine` function in the proto package (and the tests here) is a wrapper around the standard `Reader.ReadLine`, see
https://github.com/go-redis/redis/blob/74d5147d860804ea34633071d0c91a7f4b0b5c3a/internal/proto/reader.go#L57-L72

and this is the only occurrence of `bufio.ErrBufferFull` in go-redis: [search result](https://github.com/go-redis/redis/search?utf8=%E2%9C%93&q=ErrBufferFull&type=)

Issue #666, opened 2017-11-14
Add ability to retrieve result of SMEMBERS as map[string]struct{}


laouji:
Despite not having a key/value structure, in a lot of ways redis sets are more like maps than they are slices (no order, no duplicates, etc).
In order to make it easier to implement caches that can also mock the functionality of `SISMEMBER` using in memory data, it would be useful to also have a version of `SMEMBERS` that returns the data as map[string]struct{} so that the caller doesn't have to remap the data themselves when fetching the data from redis.

I would be happy to make a PR if there is any interest, but I thought I'd bring it up and see what kind of reaction this would get.

vmihailenco:
I would not mind to have `SMembersMap` if you are willing to create PR :)

PR #667, opened 2017-11-19
Add SMembersMap command

laouji:
I added a SMembersMap command that returns sets as map[string]struct{} as discussed in go-redis/redis#666


vmihailenco:
Looks good, thanks

PR #668, opened 2017-11-20
Revert "Add SMembersMap command"

totoyotonet:


vmihailenco:
Why do you want to revert that change?

PR #670, opened 2017-11-25
[redis-streams support]

nicktylah:
- Adds methods to support [redis-streams](http://antirez.com/news/114) functionality (should probably leave in PR until that spec is finalized)
    - `XAdd`
    - `XAddExt`
    - `XLen`
    - `XRange`
    - `XRangeN`
    - `XRevRange`
    - `XRevRangeN`
    - `XRead`
    - `XReadN`

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -154,12 +154,12 @@ type Cmdable interface {
 	SUnion(keys ...string) *StringSliceCmd
 	SUnionStore(destination string, keys ...string) *IntCmd
 	XAdd(key string, id *XMessageID, els map[string]interface{}) *XMessageIDCmd
-	XAddExt(key string, id XMessageID, opts XAdd, els map[string]interface{}) *XMessageIDCmd
+	XAddExt(key string, id *XMessageID, opt XAdd, els map[string]interface{}) *XMessageIDCmd
 	XLen(key string) *IntCmd
 	XRange(key string, start, stop string) *XMessageSliceCmd
 	XRangeN(key string, start, stop string, count int64) *XMessageSliceCmd
-	XRead(firstKey string, firstCheckpoint XMessageID, block time.Duration, extraStreams ...XStreamCheckpoint) *XStreamSliceCmd
-	XReadN(firstKey string, firstCheckpoint XMessageID, block time.Duration, count int64, extraStreams ...XStreamCheckpoint) *XStreamSliceCmd
+	XRead(firstKey string, firstCheckpoint *XMessageID, block time.Duration, extraStreams ...XStreamCheckpoint) *XStreamSliceCmd
+	XReadN(firstKey string, firstCheckpoint *XMessageID, block time.Duration, count int64, extraStreams ...XStreamCheckpoint) *XStreamSliceCmd
 	ZAdd(key string, members ...Z) *IntCmd
 	ZAddNX(key string, members ...Z) *IntCmd
 	ZAddXX(key string, members ...Z) *IntCmd
@@ -1266,7 +1266,7 @@ type XStream struct {
 // XStreamCheckpoint represents a query into a stream
 type XStreamCheckpoint struct {
 	Stream string
-	Checkpoint XMessageID
+	Checkpoint *XMessageID
 }
 
 // XMessage represents messages within the stream response
@@ -1280,14 +1280,13 @@ type XMessage struct {
 // and a sequence ID. E.g. 1511833543298-0
 type XMessageID struct {
 	ID int64
-	SeqID int
+	Seq int64
 	End bool
 }
 
 // XAdd represents additional options to the XAdd command
 type XAdd struct {
 	Approximate bool
-	AutoAssignID bool
 	MaxLen int64
 }
 
@@ -1312,29 +1311,29 @@ func (c *cmdable) XAdd(key string, id *XMessageID, els map[string]interface{}) *
 	if id == nil {
 		a[2] = "*"
 	} else {
-		a[2] = fmt.Sprintf("%d-%d", id.ID, id.SeqID)
+		a[2] = fmt.Sprintf("%d-%d", id.ID, id.Seq)
 	}
 	return c.xAdd(a, n, els)
 }
 
-func (c *cmdable) XAddExt(key string, id XMessageID, opts XAdd, els map[string]interface{}) *XMessageIDCmd {
+func (c *cmdable) XAddExt(key string, id *XMessageID, opt XAdd, els map[string]interface{}) *XMessageIDCmd {
 	n := 6
 	a := make([]interface{}, n+2*len(els))
 	a[0] = "xadd"
 	a[1] = key
 	a[2] = "maxlen"
 	a[3] = "~"
-	a[4] = opts.MaxLen
-	if opts.AutoAssignID {
+	a[4] = opt.MaxLen
+	if id == nil {
 		a[2] = "*"
 	} else {
-		a[2] = fmt.Sprintf("%d-%d", id.ID, id.SeqID)
+		a[2] = fmt.Sprintf("%d-%d", id.ID, id.Seq)
 	}
 
-	if opts.MaxLen == 0 {
+	if opt.MaxLen == 0 {
 		a = append(a[:2], a[:5])
 		n = 3
-	} else if !opts.Approximate {
+	} else if !opt.Approximate {
 		a = append(a[:3], a[:4])
 		n = 5
 	}
@@ -1379,10 +1378,10 @@ func (c *cmdable) xRead(a []interface{}, n int, streams []XStreamCheckpoint) *XS
 	numStreams := len(streams)
 	for i, s := range streams {
 		a[n+i] = s.Stream
-		if s.Checkpoint.End {
+		if s.Checkpoint == nil {
 			a[n+numStreams+i] = "$"
 		} else {
-			a[n+numStreams+i] = fmt.Sprintf("%d-%d", s.Checkpoint.ID, s.Checkpoint.SeqID)
+			a[n+numStreams+i] = fmt.Sprintf("%d-%d", s.Checkpoint.ID, s.Checkpoint.Seq)
 		}
 	}
 
@@ -1393,7 +1392,7 @@ func (c *cmdable) xRead(a []interface{}, n int, streams []XStreamCheckpoint) *XS
 
 func (c *cmdable) XRead(
 	firstKey string,
-	firstCheckpoint XMessageID,
+	firstCheckpoint *XMessageID,
 	block time.Duration,
 	extraStreams ...XStreamCheckpoint,
 ) *XStreamSliceCmd {
@@ -1409,7 +1408,7 @@ func (c *cmdable) XRead(
 
 func (c *cmdable) XReadN(
 	firstKey string,
-	firstCheckpoint XMessageID,
+	firstCheckpoint *XMessageID,
 	block time.Duration,
 	count int64,
 	extraStreams ...XStreamCheckpoint,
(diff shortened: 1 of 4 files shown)


commands.go
```diff
@@ -152,6 +152,13 @@ type Cmdable interface {
 	SRem(key string, members ...interface{}) *IntCmd
 	SUnion(keys ...string) *StringSliceCmd
 	SUnionStore(destination string, keys ...string) *IntCmd
+	XAdd(key string, id string, els ...interface{}) *StringCmd
+	XAddMaxLen(key string, id string, maxLen int64, approximate bool, els ...interface{}) *StringCmd
+	XLen(key string) *IntCmd
+	XRange(key string, start, stop string) *XMessageSliceCmd
+	XRangeN(key string, start, stop string, count int64) *XMessageSliceCmd
+	XRead(block int64, streams []XStreamCheckpoint) *XStreamSliceCmd
+	XReadN(block, count int64, streams []XStreamCheckpoint) *XStreamSliceCmd
```
vmihailenco: We almost certainly want `key string` to be the first argument so we can retrieve for in Redis Cluster in order to shard data - see https://github.com/go-redis/redis/blob/master/cluster.go#L543-L545. Having key in `XStreamCheckpoint` will complicate that code a lot for little/no reason.

commands.go
```diff
@@ -152,6 +152,13 @@ type Cmdable interface {
 	SRem(key string, members ...interface{}) *IntCmd
 	SUnion(keys ...string) *StringSliceCmd
 	SUnionStore(destination string, keys ...string) *IntCmd
+	XAdd(key string, id string, els ...interface{}) *StringCmd
+	XAddMaxLen(key string, id string, maxLen int64, approximate bool, els ...interface{}) *StringCmd
```
vmihailenco: I am almost sure that `XAdd` will be extended later, .e.g. with MaxAge option https://github.com/antirez/redis/issues/4450. Also `redis.XAddMaxLen("stream", "12345.0", 100, false, "foo", "bar")` is not very readable. Can we change this to:

```
redis.XAddExt("stream", "12345.1", &redis.XAdd{
    MaxLen: 100,
    Approximate: true,
    MaxAge: time.Hour, // can be added later
}, "foo", "bar")
```

commands.go
```diff
@@ -1249,6 +1256,139 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
 
 //------------------------------------------------------------------------------
 
+// X represents stream message response.
+type X struct {
+	Key  string
```
vmihailenco: Let's rename `Key` to `Stream` everywhere?

commands.go
```diff
@@ -1249,6 +1256,139 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
 
 //------------------------------------------------------------------------------
 
+// X represents stream message response.
+type X struct {
```
vmihailenco: X -> XMessage please

commands.go
```diff
@@ -1249,6 +1256,139 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
...
+// XStreamCheckpoint represents a query into a stream
+type XStreamCheckpoint struct {
+	Key string
+	Checkpoint string
+}
+
+// XMessage represents messages within the stream response
+type XMessage struct {
+	ID string
+	Values map[string]interface{}
```
vmihailenco: It looks inconsistent that we add `els ...interface{}`, but retrieve `Values map[string]interface{}`. If order of the message fields does not matter I would prefer to have `values map[string]interface{}` everywhere.

parser.go
```diff
@@ -111,6 +111,108 @@ func stringStructMapParser(rd *proto.Reader, n int64) (interface{}, error) {
 	return m, nil
 }
 
+// Implements proto.MultiBulkParse
+func xSliceParser(rd *proto.Reader, n int64) (interface{}, error) {
```
vmihailenco: xSliceParser -> xStreamSliceParser?

parser.go
```diff
@@ -111,6 +111,108 @@ func stringStructMapParser(rd *proto.Reader, n int64) (interface{}, error) {
...
+	xMap := make(map[string]interface{})
+	for i := int64(0); i < n; i += 2 {
+		var err error
+
+		key, err := rd.ReadStringReply()
+		if err != nil {
+			return nil, err
+		}
+
+		value, err := rd.ReadStringReply()
```
vmihailenco: Why not `map[string]string` if value is always string?

vmihailenco:
Hi,

Thanks for your contribution - code looks good, but there are some questions about API.

Also did you consider using `float64` for message id? It looks like wer are expected to do some math on the id so I wonder what is the best way to represent them

```
A final important thing to note about XRANGE is that, given that we receive the IDs in the reply, and the immediately successive ID is trivially obtained just incrementing the sequence part of the ID, it is possible to use XRANGE to incrementally iterate the whole stream, receiving for every call the specified number of elements. After the *SCAN family of commands in Redis, that allowed iteration of Redis data structures *despite* the fact they were not designed for being iterated, I avoided to make the same error again.
```

commands.go
```diff
@@ -152,6 +152,13 @@ type Cmdable interface {
 	SRem(key string, members ...interface{}) *IntCmd
 	SUnion(keys ...string) *StringSliceCmd
 	SUnionStore(destination string, keys ...string) *IntCmd
+	XAdd(key string, id string, els ...interface{}) *StringCmd
+	XAddMaxLen(key string, id string, maxLen int64, approximate bool, els ...interface{}) *StringCmd
+	XLen(key string) *IntCmd
+	XRange(key string, start, stop string) *XMessageSliceCmd
+	XRangeN(key string, start, stop string, count int64) *XMessageSliceCmd
+	XRead(block int64, streams []XStreamCheckpoint) *XStreamSliceCmd
```
vmihailenco: Can we change to `block time.Duration`? That will make API cleaner.

commands.go
nicktylah: Solid point -- strangely though, it seems like the `XREAD` command is intended to allow reads from multiple streams at once, e.g. 
```
> XREAD BLOCK 5000 STREAMS mystream otherstream $ 1506935385635.0
```
Will the cluster support in this library allow this kind of multiplexing?

commands.go
nicktylah: 👍
What are your thoughts about `BLOCK 0` (block forever until new message(s) appear in the stream)? Currently this will wait as long as the client's `ReadTimeout` allows, then give an `i/o timeout` error. 

Should `XRead` be immune to these since it's intended functionality?

commands.go
vmihailenco: It expects that if there are multiple keys then all keys must be on the same shard.

commands.go
vmihailenco: Yeah, library already allows you to set read timeout https://github.com/go-redis/redis/blob/master/commands.go#L947. Zero timeout is also supported.

parser.go
nicktylah: It should be an `interface{}`, but I couldn't find a method that would return the response as such

nicktylah:
Thanks for the comments, I've updated some stuff accordingly.
I also chose to represent message IDs as their own type: 
``` type XMessageID struct {
  ID int64
  SeqID int
}
```
I think this will be a useful way to interact with them. Updated the affected parsers/commands to work with them

nicois:
Sequence ID is also 64 bit.
And don't forget XREVRANGE

commands.go
```diff
@@ -1249,6 +1257,177 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
...
+	ID XMessageID
+	Values map[string]interface{}
+}
+
+// XMessageID represents a message's ID within a stream.
+// It is composed of an ID (current timestamp if auto-assigned)
+// and a sequence ID. E.g. 1511833543298-0
+type XMessageID struct {
+	ID int64
+	SeqID int
```
vmihailenco: ID -> Utime or UnixTime
SeqID -> Seq

commands.go
```diff
@@ -1249,6 +1257,177 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
...
+
+func (c *cmdable) XAddExt(key string, id XMessageID, opts XAdd, els map[string]interface{}) *XMessageIDCmd {
+	n := 6
+	a := make([]interface{}, n+2*len(els))
+	a[0] = "xadd"
+	a[1] = key
+	a[2] = "maxlen"
+	a[3] = "~"
+	a[4] = opts.MaxLen
+	if opts.AutoAssignID {
```
vmihailenco: This could be `id != nil` and AutoAssignID can be removed

commands.go
nicktylah: You're not forced to make this a timestamp; e.g. `XADD stream 1 foo bar` is a valid command. For that reason, I don't think it makes sense to force it to be a time type

commands.go
```diff
@@ -1249,6 +1257,177 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
...
+	Values map[string]interface{}
+}
+
+// XMessageID represents a message's ID within a stream.
+// It is composed of an ID (current timestamp if auto-assigned)
+// and a sequence ID. E.g. 1511833543298-0
+type XMessageID struct {
+	ID int64
+	SeqID int
+	End bool
```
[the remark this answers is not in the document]
nicktylah: I was using it to indicate `$`, but I'll use its absence as indicative of the end instead

nicktylah:
@nicois I can't run an `XREVRANGE` command against my `redis-streams` branch `redis-server`. Can you link to an example?

nicois:
@nicktylah the streams branch has been merged into unstable so pulling it won't give you the latest code any more.
Look at https://github.com/antirez/redis/commit/9bb18e54380250f3fb931028952379c3ab2dec29

nicktylah:
Pinging this PR for a re-review. Mergeable?

commands.go
```diff
@@ -1249,6 +1259,211 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
...
+	Values map[string]interface{}
+}
+
+// XMessageID represents a message's ID within a stream.
+// It is composed of an ID (current timestamp if auto-assigned)
+// and a sequence ID. E.g. 1511833543298-0
+type XMessageID struct {
+	ID int64
+	Seq int64
+}
```
vmihailenco: I think API can be simplified a lot if `XMessageID` is defined as:

```
type XMessageID string

func (XMessageID) UnixSeq() (unix int64, seq int64)
```

commands.go
```diff
@@ -1249,6 +1259,211 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
...
+	count int64,
+) *XMessageSliceCmd {
+	args := make([]interface{}, 6)
+	args[0] = "xrange"
+	args[1] = key
+	args[2] = start
+	args[3] = stop
+	args[4] = "count"
+	args[5] = count
+	cmd := NewXMessageSliceCmd(args...)
```
vmihailenco: This (and some other commands) can be written as:

```
cmd := NewXMessageSliceCmd(xrange, key, start, stop, "count", count)
```

vmihailenco:
I would like to make some changes, but what really concerns me is that there are no docs/beta releases at redis.io. Do you use unstable Redis Server? Can we wait until Redis team at least publishes beta release?

kassiansun:
Ouch, never seen this before... Should we merge the two MR, or do something based on either?

nicktylah:
Mmm good question. There's enough difference between the two PRs to extract some usefulness from both I think

skyrocknroll:
The docs are here https://redis.io/topics/streams-intro and RC is released now

vmihailenco:
Replaced by https://github.com/go-redis/redis/pull/795

PR #680, opened 2017-12-11
Update example in the README.md

veggiemonk:
Let me know if this should be the default. It seems like a reasonable choice since running a redis sentinel cluster locally for testing is not that straightforward.


vmihailenco:
I don't understand where "a redis sentinel cluster" comes from, but I want to keep this example as simple as possible. Thanks for trying to improve readme anyway.

PR #691, opened 2018-01-05
Add Process() to Cmdable interface

nicholascapo:
* Both Client and ClusterClient now implement this interface


vmihailenco:
The change looks good, but v5 branch is deprecated and does not receive updates (except security fixes). So if you are still interested in this change you need to upgrade to master branch and re-send this PR.

PR #699, opened 2018-01-29
Add WrapProcess and WrapPipelineExecer to all clients for better instrument support

hyfrey:
Currently, there is only WrapProcess method in Client, not in other client. Besides, there is no method to instrument pipeline. This pull request implement WrapPipelineExecer method to instrument pipeline, and add WrapProcess and WrapPlinelineExecer to all clients

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -445,6 +445,9 @@ type ClusterClient struct {
 	cmdsInfoOnce internal.Once
 	cmdsInfo     map[string]*CommandInfo
 
+	process            func(Cmder) error
+	wrapPipelineExecer func(func([]Cmder) error) func([]Cmder) error
+
 	// Reports whether slots reloading is in progress.
 	reloading uint32
 }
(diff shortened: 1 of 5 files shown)


cluster.go
```diff
@@ -908,9 +922,17 @@ func (c *ClusterClient) reaper(idleCheckFrequency time.Duration) {
 	}
 }
 
+func (c *ClusterClient) WrapPipelineExecer(fn func(func([]Cmder) error) func([]Cmder) error) {
```
vmihailenco: This should be called `WrapPipelineProcess` to mirror `WrapProcess`.

cluster.go
```diff
@@ -908,9 +922,17 @@ func (c *ClusterClient) reaper(idleCheckFrequency time.Duration) {
...
+	c.wrapPipelineExecer = fn
+}
+
 func (c *ClusterClient) Pipeline() Pipeliner {
+	pipelineExec := c.pipelineExec
+	if c.wrapPipelineExecer != nil {
+		pipelineExec = c.wrapPipelineExecer(pipelineExec)
+	}
```
vmihailenco: Please change the code to follow existing pattern:

```
	if c.process != nil {
		return c.process(cmd)
	}
	return c.defaultProcess(cmd)
```

hyfrey: Pipeline() is different from Process(), Pipeline() return a pipeline object which can be execed at later time, Process() exec command immediately. Did you mean something like
```
	if c.wrapPipelineProcess != nil {
		pipe := Pipeline{
			exec: c.wrapPipelineProcess(c.pipelineExec),
		}
		pipe.setProcessor(pipe.Process)
		return &pipe
	}
	pipe := Pipeline{
		exec: c.pipelineExec,
	}
	pipe.setProcessor(pipe.Process)
	return &pipe
```
or something like
```
	pipelineExec := func() func([]Cmder) error {
		if c.wrapPipelineProcess != nil {
			return c.wrapPipelineProcess(c.pipelineExec)
		}
		return c.pipelineExec
	}()

	pipe := Pipeline{
		exec: pipelineExec,
	}
	pipe.setProcessor(pipe.Process)
	return &pipe
```

PR #702, opened 2018-01-31
cluster: add support for context.Context to ClusterClient and Ring

### LINKED ISSUE #700: ClusterClient and Ring do not support Context
Would it be possible to add `Context` and `WithContext` methods to these two? Just like we have for `Client`?

I'm happy to submit a PR.

gbbr:
This change adds the `Context` and `WithContext` methods to
`ClusterClient` and `Ring`, as we have in the regular `Client`

In these two cases, `WithContext` returns the same instance of the client, and not a copy. Copying is tricky and difficult in this case because the two contain mutexes (in clusterNodes).

Fixes #700

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1,6 +1,7 @@
 package redis
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"net"
@@ -438,6 +439,8 @@ func (c *clusterState) slotNodes(slot int) []*clusterNode {
 type ClusterClient struct {
 	cmdable
 
+	ctx context.Context
+
 	opt    *ClusterOptions
 	nodes  *clusterNodes
 	_state atomic.Value
@@ -568,6 +571,23 @@ func (c *ClusterClient) cmdSlotAndNode(state *clusterState, cmd Cmder) (int, *cl
 
 	node, err := state.slotMasterNode(slot)
 	return slot, node, err
+}
+
+func (c *ClusterClient) Context() context.Context {
+	if c.ctx != nil {
+		return c.ctx
+	}
+	return context.Background()
+}
+
+// WithContext changes the ClusterClient's context to ctx and returns
+// a reference to the ClusterClient.
+func (c *ClusterClient) WithContext(ctx context.Context) *ClusterClient {
+	if ctx == nil {
+		panic("nil context")
+	}
+	c.ctx = ctx
+	return c
 }
 
 func (c *ClusterClient) Watch(fn func(*Tx) error, keys ...string) error {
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -570,6 +573,29 @@ func (c *ClusterClient) cmdSlotAndNode(state *clusterState, cmd Cmder) (int, *cl
...
+	}
+	c2 := c.copy()
+	c2.ctx = ctx
+	return c2
+}
+
+func (c *ClusterClient) copy() *ClusterClient {
+	c2 := new(ClusterClient)
+	*c2 = *c
+	c2.cmdable.setProcessor(c2.Process)
```
gbbr: Copied this from `Client` - is it really needed?

gbbr:
Working on the copy problem from CI...

cluster.go
```diff
@@ -570,6 +573,23 @@ func (c *ClusterClient) cmdSlotAndNode(state *clusterState, cmd Cmder) (int, *cl
...
+	return context.Background()
+}
+
+// WithContext changes the ClusterClient's context to ctx and returns
+// a reference to the ClusterClient.
+func (c *ClusterClient) WithContext(ctx context.Context) *ClusterClient {
+	if ctx == nil {
+		panic("nil context")
+	}
+	c.ctx = ctx
```
gbbr: Please note that this method doesn't return a copy anymore, it now returns the same instance. There is a problem with copying the ClusterClient and Ring because they contain `sync.Mutex`. I'm tempted to suggest to remove the copying from the regular `Client` too, but I'm afraid of how that would potentially affect users.

gbbr: It's true that some libraries return a copy (e.g. [golang.org/x/sync/errgroup](https://github.com/golang/dep/blob/73e9c7562b3447bc130f9fe7b2bbe6eeae6d173c/vendor/golang.org/x/sync/errgroup/errgroup.go#L33-L36)), while others return a reference to the same object (e.g. [github.com/gocql/gocql](https://github.com/gocql/gocql/blob/697e7c57f99bb127606c4d4f975c2ff4f34eca1c/session.go#L767-L770)). So I guess we could potentially go for both options.

gbbr:
Done :)

gbbr:
@vmihailenco any thoughts on this one? Would be great to have.

vmihailenco:
Sorry for the long delay. Recently I've refactored Cluster and Ring to (re)move mutexes&atomics that did not allow to copy those structs. Can you please rebase this branch on master and fix `WithContext` to return a copy of the Cluster/Ring? I think it is important to preserve this behavior.

gbbr:
This was getting complicated after the rebase, re-created the PR here https://github.com/go-redis/redis/pull/737

Issue #701, opened 2018-01-31
(ClusterClient).ForEachNode does not cover all nodes


gbbr:
In a cluster with 3 masters (ports 7000-7002) and 3 slaves (ports 7003-7005) connected as:
```go
c := redis.NewClusterClient(&redis.ClusterOptions{
	Addrs: []string{":7000", ":7001", ":7002"},
})
```
Using `WrapProcess` inside `ForEachNode` does not wrap each client, leaving gaps. `ForEachNode` uses a `for` loop to go through each master (`state.masters`) and slave (`state.slaves`), but omits some nodes (`state.nodes.nodes`).

In the above setup, the `state` field contains:
```
state.masters:
172.17.0.2:7002
172.17.0.2:7000
172.17.0.2:7001

state.slaves:
172.17.0.2:7005
172.17.0.2:7003
172.17.0.2:7004

state.nodes.nodes:
172.17.0.2:7002
172.17.0.2:7005
172.17.0.2:7000
172.17.0.2:7003
172.17.0.2:7004
:7000
:7001
:7002
172.17.0.2:7001
```
Even though `ForEachNode` goes as expected through each _unique_ node, it does not offer the possibility to tap into every client. One possible fix could be:
```diff
diff --git a/cluster.go b/cluster.go
index a2c18b3..8761000 100644
--- a/cluster.go
+++ b/cluster.go
@@ -816,11 +816,7 @@ func (c *ClusterClient) ForEachNode(fn func(client *Client) error) error {
 		}
 	}
 
-	for _, node := range state.masters {
-		wg.Add(1)
-		go worker(node)
-	}
-	for _, node := range state.slaves {
+	for _, node := range state.nodes.nodes {
 		wg.Add(1)
 		go worker(node)
 	}
```
Would this change or a similar one make sense? Or am I doing something stupid?

vmihailenco:
Why do you want to tap every client? AFAIR go-redis only uses nodes in `state.masters` and `state.slaves`. Unused nodes should be eventually removed by `clusterNodes.GC`.

Probably `:7000 :7001 :7002` should not be added to the nodes in first place, but I don't see what harm it causes...

gbbr:
[...]

PR #703, opened 2018-01-31
cluster: make ForEachNode iterate through each node

gbbr:
ForEachNode was iterating through the `masters` and `slaves` fields,
which are the result of the addresses advertised by the cluster, but do
not contain the addresses used to create the cluster (in
ClusterOptions). This was problematic in some situations. This PR
addresses that issue.

This might make more sense because it is the `nodes` property that is used
when selecting nodes, not `masters` and `slaves`. (see [here](https://github.com/gbbr/redis/blob/a0ddd1ed404fdf248caf6abd8caf55459713c123/cluster.go#L412)).

This addresses #701. If you decide against making this change, please
feel free to close the PR.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -816,11 +816,7 @@ func (c *ClusterClient) ForEachNode(fn func(client *Client) error) error {
 		}
 	}
 
-	for _, node := range state.masters {
-		wg.Add(1)
-		go worker(node)
-	}
-	for _, node := range state.slaves {
+	for _, node := range state.nodes.nodes {
 		wg.Add(1)
 		go worker(node)
 	}


vmihailenco:
I need to think about this, but I am not comfortable that with this change ForEachMaster+ForEachSlave != ForEachNode.

gbbr:
No problem, take your time. You have the context knowledge.

The way I thought of it was that `ForEachMaster` and `ForEachSlave` would iterate through the cluster advertised masters and slaves, while `ForEachNode` would iterate through all nodes including the once used at startup time.

Either way, if you decide against it, I can work around it but in some situations it could be problematic given our scenario.

Thanks for looking.

vmihailenco:
After some testing I think it should be `ForEachNode = ForEachMaster + ForEachSlave`. Otherwise it would require go-redis to keep connection/client to all nodes that ever existed in the cluster which sounds dubious.

gbbr:
Sure. I can definitely understand your point of view. And I'm ok with it 👍 I think my case is somewhat of an edge-case that should normally not happen in a real life scenario. I suspect my case was one of a misconfigured cluster, where the nodes addresses used with go-redis were different than the ones advertised by the cluster.

PR #710, opened 2018-02-14
Correctly parse EVAL response containing customized error

huandu:
When EVAL response contains a customized error, current parser will stop at this error and drop all remaining bytes.

Sample command.

    eval 'return {12, {err="error"}, 34}' 0

In this sample, result of `client.Eval` doesn't include `34`.

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -2,7 +2,6 @@ package redis_test
 
 import (
 	"encoding/json"
-	"errors"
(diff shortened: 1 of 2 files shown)


vmihailenco:
Is it possible to add a test for this behavior?

huandu:
Sure. I'll add a test case.

PR #719, opened 2018-02-26
Change Makefile to use sed syntax that will work for both GNU and BSD sed

tthyer:
I could not run `make testdeps` on a mac. The change I have made to the inline-editing option syntax will allow the sed command to work with both GNU sed and BSD sed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.rdb
 testdata/*/
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 *.rdb
 testdata/*/
-.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 *.rdb
 testdata/*/
+.idea/
```
vmihailenco: Please revert - we don't add every editor to gitignore file. The better alternative is to use global gitignore - https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

tthyer: No problem, reverted

vmihailenco:
Thanks

PR #725, opened 2018-03-02
add hook called when cmd is processing to client

tumiao:


vmihailenco:
This looks very specific to your specific needs - it is unlikely it will be useful to anyone else. Besides v3 is old and we only fix security issues for v3.

PR #729, opened 2018-03-05
Add option to balance load between master node and replica nodes

tthyer:
This change just adds the option to balance any reads between a master and its replicas. It will be overridden if RouteByLatency is also set.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -33,6 +33,8 @@ type ClusterOptions struct {
 	ReadOnly bool
 	// Allows routing read-only commands to the closest master or slave node.
 	RouteByLatency bool
+	// Allows balanced routing of read-only commands to master and slave nodes.
+	LoadBalanceNodes bool
 
 	// Following options are copied from Options struct.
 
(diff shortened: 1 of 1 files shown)


cluster.go
```diff
@@ -33,6 +33,8 @@ type ClusterOptions struct {
 	ReadOnly bool
 	// Allows routing read-only commands to the closest master or slave node.
 	RouteByLatency bool
+	// Allows balanced routing of read-only commands to master and slave nodes.
+	LoadBalanceNodes bool
```
vmihailenco: Please change this to:

```
// Allows routing read-only commands to the random master or slave node.
RouteRandomly bool
```

So it matches `RouteByLatency` more closely.

cluster.go
```diff
@@ -468,6 +470,12 @@ func (c *clusterState) slotClosestNode(slot int) (*clusterNode, error) {
 	return node, nil
 }
 
+func (c *clusterState) slotRandomEligibleNode(slot int) *clusterNode {
```
vmihailenco: slotRandomEligibleNode -> slotRandomNode, because I don't understand what sense `Eligible` tries to convey here

cluster.go
tthyer: done in 22d7121

cluster.go
tthyer: The idea was that it's not a random selection of any cluster node, but just the ones that make sense for that slot. It wasn't a great name, so I've implemented your suggestion in 22d7121.

vmihailenco:
Thanks.

PR #737, opened 2018-03-07
cluster: add support for context.Context to ClusterClient and Ring

### LINKED ISSUE #700: ClusterClient and Ring do not support Context

gbbr:
This change adds the Context and WithContext methods to
ClusterClient and Ring, as we have in the regular Client

In these two cases, WithContext returns the same instance of the client, and not a copy. Copying is tricky and difficult in this case because the two contain mutexes (in clusterNodes).

Fixes #700

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1,6 +1,7 @@
 package redis
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks for your patience.

gbbr:
Thank _you_ :)

PR #743, opened 2018-03-13
Remove costly 'appendIfNotExists' and 'remove' call from PubSub

superkinglabs:
When there are thousands of subscriptions and the rate of subscribe and unsubscribe is high (hundreds per seconds), appendIfNotExists and remove call perform very poorly and was consuming all the CPU. Replacing the slice with the map improved the performance (CPU before was at 90% afterwards came down to 10%). I also had the pprof profile before and after this change but I am not able to find it now.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -24,8 +24,8 @@ type PubSub struct {
 
 	mu       sync.Mutex
 	cn       *pool.Conn
-	channels []string
-	patterns []string
+	channels map[string]struct{}
+	patterns map[string]struct{}
 	closed   bool
 
 	cmd *Cmd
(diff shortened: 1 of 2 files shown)


vmihailenco:
Why would you subscribe/unsubscribe hundreds times per second?

superkinglabs:
To build a presence system, whenever a user comes online, he needs to subscribe to all of his friends to receive their updates and similarly when he goes offline he will need to unsubscribe all of those.

superkinglabs:
Also even if the rate of subscribe/unsubscribe is not very high, the map approach allows subscribe/unsubscribe to be O(1) without adding any extra space requirement. With the slice it would take linear time with each subscribe/unsubscribe.

pubsub.go
```diff
@@ -67,12 +67,24 @@ func (c *PubSub) _conn(channels []string) (*pool.Conn, error) {
 func (c *PubSub) resubscribe(cn *pool.Conn) error {
 	var firstErr error
 	if len(c.channels) > 0 {
-		if err := c._subscribe(cn, "subscribe", c.channels...); err != nil && firstErr == nil {
+		channels := make([]string, len(c.channels), 0)
```
vmihailenco: You need to remove `, 0` here and below to fix the build

superkinglabs: Done. Also once you merge this, can you create a new release, I need this fix to use this package in my new project.

vmihailenco: Released v6.10.2

PR #747, opened 2018-03-28
Added OnConnect callback to UniversalOptions

olvesh:


universal.go
```diff
@@ -8,25 +8,26 @@ type UniversalOptions struct {
 	// Either a single address or a seed list of host:port addresses
 	// of cluster/sentinel nodes.
 	Addrs []string
-
+
```
vmihailenco: Can you please revert all whitespace-only changes like this?

olvesh:
Will do. My editor inserts blank spaces at EOF as default.

universal.go
olvesh: Will do. My editor inserts blank spaces at EOF as default.

Issue #775, opened 2018-05-24
ClusterClient pipeline never routes to replica nodes


pgadangi:
I'm creating a ClusterClient with ReadOnly true
```
redis.NewClusterClient(&redis.ClusterOptions{
	Addrs:       addresses,
	IdleTimeout: idleTimeout,
	PoolSize:    poolSize,
	ReadOnly:    true,
	ReadTimeout: readTimeout,
})
```
When I use that client for a command like `client.HGetAll(key)` it gets routed appropriately,
but when I use a pipeline from that client `client.Pipeline().HGetAll(key)` it always gets routed to a master node.

PR #776, opened 2018-05-24
Use cmdSlotAndNode() to route ClusterClient pipeline commands

pgadangi:
Fixes https://github.com/go-redis/redis/issues/775
Non-pipeline cluster client commands are routed using cmdSlotAndNode().
This changes pipeline cluster client commands to use that as well rather than always routing to master nodes.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1207,15 +1207,32 @@ func (c *ClusterClient) mapCmdsByNode(cmds []Cmder) (map[*clusterNode][]Cmder, e
 	}
 
 	cmdsMap := make(map[*clusterNode][]Cmder)
+	cmdsAreReadOnly := c.cmdsAreReadOnly(cmds)
 	for _, cmd := range cmds {
-		slot := c.cmdSlot(cmd)
-		node, err := state.slotMasterNode(slot)
+		var node *clusterNode
+		var err error
+		if cmdsAreReadOnly {
+			_, node, err = c.cmdSlotAndNode(cmd)
+		} else {
+			slot := c.cmdSlot(cmd)
+			node, err = state.slotMasterNode(slot)
+		}
 		if err != nil {
 			return nil, err
 		}
 		cmdsMap[node] = append(cmdsMap[node], cmd)
 	}
 	return cmdsMap, nil
+}
+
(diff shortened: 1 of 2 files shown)


pgadangi:
It doesn't seem like the build failures are related to this change.
```
    LINK redis-cli
../deps/jemalloc/lib/libjemalloc.a(background_thread.o): In function `pthread_create_fptr_init':
/home/travis/gopath/src/github.com/go-redis/redis/testdata/redis/deps/jemalloc/src/background_thread.c:815: undefined reference to `dlsym'
/home/travis/gopath/src/github.com/go-redis/redis/testdata/redis/deps/jemalloc/src/background_thread.c:815: undefined reference to `dlsym'
collect2: error: ld returned 1 exit status
```

cluster.go
```diff
@@ -1195,16 +1195,9 @@ func (c *ClusterClient) defaultProcessPipeline(cmds []Cmder) error {
 }
 
 func (c *ClusterClient) mapCmdsByNode(cmds []Cmder) (map[*clusterNode][]Cmder, error) {
-	state, err := c.state.Get()
-	if err != nil {
-		setCmdsErr(cmds, err)
-		return nil, err
-	}
```
vmihailenco: This block is here so all cmds are mapped using single cluster state. Otherwise cluster can be re-configured during mapping and commands can be mapped inconsistently.

vmihailenco:
With this change following is possible for the pipeline:

```
SET foo bar <- routed to the master
GET foo <- routed to the slave
```

So `GET foo` will return nil, because `SET foo bar` is not replicated to the slave yet. I am not sure this is worth the performance gains... How important this change is for you?

pgadangi:
Thanks for taking a look, I can see why that behavior isn't desirable. Do you think it would be reasonable to do something like the following?
```
if all pipeline commands are reads {
    route with cmdSlotAndNode()
} else {
    route to masters
}
```
Our application is very read-heavy, so we pipeline commands and try to distribute them across replicas and masters.

vmihailenco:
>Do you think it would be reasonable to do something like the following?

Yes, that would be ideal.

nziebart:
Would you mind taking another look here @vmihailenco ?

Issue #772, opened 2018-05-23
Can we introduce an option minIdleConnections ?


darkpssngr:
Issue
I'm using redis-labs. The DNS resolution time for the endpoints they give are very high in regions other than US. Currently my service does not have that much traffic due to which idle connections are getting killed and new connections are spawned everytime a new request comes in. I tried turning off the IdleConnectionFrequencCheck by setting it to zero and using a custom dialer which sends keep-alive. But for some reason these connections get closed after sometime. not sure why? upon inspection of tcp dump we found that the termination is intiated by the client and not the server.

I know this problem needs to be identified and fixed but was also wondering if we could have an option which lets us specify minimum idle connections so that we always have some connections on standby when there are spikes. Or is there any way we can avoid this problem?

vmihailenco:
`IdleCheckFrequency` must be set to -1 to disable it. Zero value is the default value.

Overall I think `minIdleConnections` is a good idea and I'd like to see it supported too.

darkpssngr:
Sorry. I've set the IdleCheckFrequency to -1 only.

vmihailenco:
The code is at https://github.com/go-redis/redis/blob/master/internal/pool/pool.go#L91-L93 if you want to check yourself. Connection is also closed by the client in case of timeout / bad state.

vmihailenco:
See https://github.com/go-redis/redis/pull/778/files - it is WIP, but feel free to try it.

darkpssngr:
So the thing is I'm using my custom dialer which creates connection with keep-alive. I see keep alive msgs being sent and its being ACKed as well by the server. Out of the blue a FIN req is sent by the client and the connection is closed. Its a proper close, not initiated by the server or not something because of a bad state. Not sure whats happening.

vmihailenco:
You should also know that PubSub closes the connection when subscription is closed.

ginuerzh:
does it mean that the `IdleTimeout` option also should be set to `-1` to avoid closing idle connections?

darkpssngr:
Yes. If its not set, it defaults to 5mins. So, even though the reaper does not kill it, the connection will be discarding when asking for a new connection.
https://github.com/go-redis/redis/blob/master/internal/pool/pool.go#L185

vmihailenco:
Master branch has `MinIdleConns` option that should address this problem.

xianglinghui:
[...]

PR #778, opened 2018-05-28
Add MinIdleConns

vmihailenco:
Updates https://github.com/go-redis/redis/issues/772


darkpssngr:
In the reapStaleConn we aren't actually if the free connections are good connections? I think it will be better we do a ping with each connection to see if it's really a good connection.

PR #791, opened 2018-06-15
Handling of isPrefix in Reader.ReadLine.

dtucker2:
I've added a commit to resolve an issue where data greater than 4096 bytes could not be read without generating an error.

### DIFF
--- a/internal/proto/reader.go
+++ b/internal/proto/reader.go
@@ -65,12 +65,15 @@ func (r *Reader) ReadN(n int) ([]byte, error) {
 }
 
 func (r *Reader) ReadLine() ([]byte, error) {
-	line, isPrefix, err := r.src.ReadLine()
-	if err != nil {
-		return nil, err
-	}
-	if isPrefix {
-		return nil, bufio.ErrBufferFull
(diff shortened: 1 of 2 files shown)


vmihailenco:
What real use case this fixes? `SET` and `GET` already works for big strings.

dtucker2:
I've encountered problems whilst receiving large messages on subscribed channels.

vmihailenco:
Can you add a test that reproduces the problem?

dtucker2:
If you mean through the PubSub code, probably. I've already written one for the Reader.

vmihailenco:
Yes, for some public API. `Reader.ReadLine` is an internal API and it is not supposed to be used to read large strings.

dtucker2:
Leave it with me, I'll see what I can do.

dtucker2:
I implemented a test on the Subscribe method, however I believe I have been looking in the wrong place. It seems the problem may be the size of the net.Conn ReadBuffer, however I have still not tracked down where this is set.

vmihailenco:
Closing - please reopen if you find the root cause of this.

PR #779, opened 2018-05-29
add support for xadd/xrange/xrevrange/xread

kassiansun:
Redis 5 will add a new data structure [STREAM](https://redis.io/topics/streams-intro), this patch will add support for it.

Current issues:
- [x] add support for `xadd`
- [x] add support for `xrange` `xrevrange`
- [x] add support for `xread`
- [x] add support for `xgroup`, `xreadgroup`, `xack`
- [x] add support for `xpending` `xclaim`
- [x] add support for `xlen`
- [x] add support for `xtrim`
- [x] add support for `xdel`

PR #795, opened 2018-06-23
Add basic redis streams support

vmihailenco:
This is based on https://github.com/go-redis/redis/pull/670 with some simplifications and API changes. I've chosen https://github.com/go-redis/redis/pull/670, because it came first and is much smaller compared to https://github.com/go-redis/redis/pull/779. The plan is to merge this PR first and continue in https://github.com/go-redis/redis/pull/779 to add missing APIs.

@nicktylah @kassiansun please take a look

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -171,6 +171,16 @@ type Cmdable interface {
 	SRem(key string, members ...interface{}) *IntCmd
 	SUnion(keys ...string) *StringSliceCmd
 	SUnionStore(destination string, keys ...string) *IntCmd
+	XAdd(stream, id string, els map[string]interface{}) *StringCmd
+	XAddExt(opt *XAddExt) *StringCmd
+	XLen(key string) *IntCmd
+	XRange(stream, start, stop string) *XMessageSliceCmd
+	XRangeN(stream, start, stop string, count int64) *XMessageSliceCmd
+	XRevRange(stream string, start, stop string) *XMessageSliceCmd
+	XRevRangeN(stream string, start, stop string, count int64) *XMessageSliceCmd
+	XRead(streams ...string) *XStreamSliceCmd
+	XReadN(count int64, streams ...string) *XStreamSliceCmd
+	XReadExt(opt *XReadExt) *XStreamSliceCmd
 	ZAdd(key string, members ...Z) *IntCmd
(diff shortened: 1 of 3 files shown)


kassiansun:
Sounds good for me

commands.go
```diff
@@ -1282,6 +1292,127 @@ func (c *cmdable) SUnionStore(destination string, keys ...string) *IntCmd {
...
+			a = append(a, opt.Count)
+		}
+		if opt.Block > 0 {
+			a = append(a, "block")
+			a = append(a, int64(opt.Block/time.Millisecond))
+		}
+	}
+	a = append(a, "streams")
+	for _, s := range opt.Streams {
+		a = append(a, s)
```
nicktylah: This will fail with the error `ERR Unbalanced XREAD list of streams: for each stream key an ID or '$' must be specified.` -- you'll need to pass a list of IDs per stream

kassiansun: There's a problem that the old patch is quite old, redis' API has changed quite a bit since that. Actually it changed a lot during my patch (about one week or two).

vmihailenco: Perhaps I don't fully understand, but the "idea" was that client must specify stream+id, e.g. `[]string{stream1, stream2, 0, 0}`. That is not very type safe, but the alternative `[]StreamID{{"stream1", "0"}, StreamID{"stream2", "0"}}` is much longer and does not provide much value too...

Anyway if you have ideas how to fix/improve this code - please send small PRs and let's discuss it. I've tried to keep this code simple to use, but Redis API is not particularly nice and likely there are better ways...

vmihailenco: Yeah, that is what is used here in simplified version `XRead(streams ...string)` of `XReadExt`

kassiansun: But `XReadExt` seems not match `go-redis` style, I used `XReadBlock` `XReadBlockN` for my pr.

vmihailenco: We still can add `XReadNBlock` if you think it is used often enough.

chuzui: Hi, I notice that the code only append "block" when ```opt.Block > 0```, so I cann't set block time to 0 (that means never timeout). Or there is another way to block the xread until next message?

vmihailenco: It is an overlook - I've sent https://github.com/go-redis/redis/pull/810

PR #798, opened 2018-06-28
Establishing TLS connections respect DialTimeout

josh-tepper:
see the issue: https://github.com/go-redis/redis/issues/799

Currently when redis.Options.TLSConfig is set, the redis.Options.Dialer function can hang indefinitely!

This change causes connection negotiation + tls handshaking to together be governed by the redis.Options.DialTimeout.

This seems to be the expected behavior, per the go standard library.  See for example:
https://golang.org/pkg/crypto/tls/#DialWithDialer

### DIFF
--- a/options.go
+++ b/options.go
@@ -85,12 +85,12 @@ func (opt *Options) init() {
 	}
 	if opt.Dialer == nil {
 		opt.Dialer = func() (net.Conn, error) {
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks for the fix and detailed explanation.

PR #812, opened 2018-07-19
Hash function and nreplicas in consistent hash can be set in RingOptions

hyfrey:
We currently use redis ring with 10 nodes, however we encounter uneven load between different nodes, which can be as high as 40%. After some experiment, we found that the default configuration of consistent has used in ring, 100 nreplicas and crc32 hash function is not optimal for our use case. In our case, use 1000 nreplicas and xxhash hash function provide almost even distribution of keys between nodes. This pull request implements the functionality to set nreplicas and hash function in RingOptions to customize the behavior of consistent hash in ring.

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -16,7 +16,8 @@ import (
 	"github.com/go-redis/redis/internal/pool"
 )
 
-const nreplicas = 100
+// Hash is type of hash function used in consistent hash
+type Hash consistenthash.Hash
 
 var errRingShardsDown = errors.New("redis: all ring shards are down")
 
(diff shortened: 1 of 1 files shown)


ring.go
```diff
@@ -30,6 +31,24 @@ type RingOptions struct {
 	// Shard is considered down after 3 subsequent failed checks.
 	HeartbeatFrequency time.Duration
 
+	// Hash function used in consistent hash, will use crc32.ChecksumIEEE
+	// from hash/crc32 package if not specified
+	HashFn Hash
```
vmihailenco: Can you please rename this HashFn -> Hash? It will be consistent with type `consistenthash.Hash`

ring.go
```diff
@@ -30,6 +31,24 @@ type RingOptions struct {
...
+	//
+	// Following is deviation for common nreplicas:
+	//  --------------------------------------------------------
+	//  | nreplicas | standard error | 99% confidence interval |
+	//  |     10    |     0.3152     |      (0.37, 1.98)       |
+	//  |    100    |     0.0997     |      (0.76, 1.28)       |
+	//  |   1000    |     0.0316     |      (0.92, 1.09)       |
+	//  --------------------------------------------------------
+	//  See https://arxiv.org/abs/1406.2294 for reference
+	Nreplicas int
```
vmihailenco: consistenthash calls this just `replicas`. What about `HashReplicas` to be more specific?

vmihailenco:
I have few minor notes on naming, but other than that this looks good - thanks.

vmihailenco:
Merging - I will updates names myself.

PR #821, opened 2018-07-23
Supports new style syntax of client kill command

git-hulk:
go-redis supports old-style syntax of client kill command, and new syntax wasn't supported. This pr add new style syntax without breaking the old API by adding a new function `ClientKillByOption`.

old style syntax:
```
CLIENT KILL ip:port
```
new style syntax:
```
CLIENT KILL <option> [value] ... <option> [value]
```

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -220,6 +220,7 @@ type Cmdable interface {
 	BgRewriteAOF() *StatusCmd
 	BgSave() *StatusCmd
 	ClientKill(ipPort string) *StatusCmd
+	ClientKillByFilter(keys ...string) *IntCmd
 	ClientList() *StringCmd
 	ClientPause(dur time.Duration) *BoolCmd
 	ConfigGet(parameter string) *SliceCmd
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -220,6 +220,7 @@ type Cmdable interface {
 	BgRewriteAOF() *StatusCmd
 	BgSave() *StatusCmd
 	ClientKill(ipPort string) *StatusCmd
+	ClientKillByOption(keys ...string) *IntCmd
```
vmihailenco: I'd rather remove existing `ClientKill` and rename your version `ClientKillByOption` -> `ClientKill`. Could you do that?

vmihailenco: No, it is not - I did not notice it. Let's keep this as is, but rename `ClientKillByOption` to `ClientKillByFilter` since that is what Redis calls it `CLIENT KILL <filter> <value> ... ... <filter> <value>`. Thanks.

git-hulk:
@vmihailenco reamed and rebased, thanks for your review

vmihailenco:
Thank you.

PR #837, opened 2018-08-12
Add MaxConnAge

vmihailenco:


rafaeleyng:
Hi @vmihailenco , I'm investigating some configs and was wondering how the `MaxConnAge` config could impact on my setup. Could you please elaborate on this?

vmihailenco:
Originally this was requested for https://github.com/go-pg/pg so go-pg closes connections after some time so proxy servers like haproxy can be cleanly restarted. As I understand during restart old haproxy instance is killed only after all incoming connections are closed.

For go-redis it is probably not as relevant.

PR #849, opened 2018-08-26
Introduce Go Modules

marwan-at-work:
This PR is generated by a [tool I wrote](https://github.com/marwan-at-work/mod) that helps migrate Go packages that are tagged +2 to Semantic Import Versioning. 

Note that this change should be backwards compatible and can still be imported without the `/vX` suffix when Go Modules is off.


vmihailenco:
Thanks for the PR but I *think* it makes sense to wait more time until it is clear that `go mod` is adopted by community. It is not clear what this PR solves/improves but it:
- requires learning about go mod
- requires latest Go version(s) with go.mod support
- drops support for Go <= 1.8

bench_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/go-redis/redis"
+	"github.com/go-redis/redis/v6"
```
vmihailenco: Do we really must specify version `v6` in tests? It is a bit strange that module should "versionify" itself.

go.mod
```diff
@@ -0,0 +1,16 @@
...
+	github.com/onsi/ginkgo v1.6.0
+	github.com/onsi/gomega v1.4.1
+	golang.org/x/net v0.0.0-20180826012351-8a410e7b638d // indirect
+	golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f // indirect
+	golang.org/x/sys v0.0.0-20180824143301-4910a1d54f87 // indirect
+	golang.org/x/text v0.3.0 // indirect
+	gopkg.in/fsnotify.v1 v1.4.7 // indirect
+	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
+	gopkg.in/yaml.v2 v2.2.1 // indirect
+)
```
vmihailenco: All of these packages are test dependencies. I know this file is autogenerated but still I am asking - should we specify test dependencies?

tschellenbach:
Using Go 1.11's module system here. Struggling to install go-redis. Not quite sure yet why it doesn't work.

```
go get -u github.com/go-redis/redis
import (
    "github.com/go-redis/redis"
)
->
unknown import path "github.com/go-redis/redis": cannot find module providing package github.com/go-redis/redis
```

I didn't have this issue with other packages... Will update once I find out more about why this is broken.

vmihailenco:
Closing as stale.

PR #853, opened 2018-08-31
Add ZPopMin and ZPopMax

ElPicador:


commands.go
```diff
@@ -1694,6 +1696,28 @@ func (c *cmdable) ZInterStore(destination string, store ZStore, keys ...string)
 	return cmd
 }
 
+func (c *cmdable) ZPopMax(key string, count int64) *ZSliceCmd {
```
vmihailenco: It looks like the signature is `ZPOPMAX key [count]`. Could you change Go signature to `key string, count ...int64` and add count only if `len(count) == 1`? And in case of `len(count) > 1` we should use panic, e.g. see https://github.com/go-redis/redis/blob/045b250b08775328846227c81b39a21bd4b145e3/commands.go#L2474

ElPicador: changed.

I didn't manage to make the test for panic work, it panics but gomega doesn't intercept the panic
```
Expect(client.ZPopMin("zset", 1, 2)).To(Panic())
```

vmihailenco:
Thanks!

ElPicador:
When do you expect to release a new version with this in it?

vmihailenco:
Just tagged v6.14.1

ElPicador:
Thanks 😄

PR #860, opened 2018-09-07
support once error

airfdl:
Change-Id: Icde9abdaede8b8bff60a4a3865b4d4f757661cce

### DIFF
--- a/internal/once.go
+++ b/internal/once.go
@@ -27,6 +27,7 @@ import (
 // and is re-armed on failure.
 type Once struct {
 	m    sync.Mutex
+	err  error
 	done uint32
 }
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
Previously this code retried `fn` until success. Now it is executed once, stores the error, and returns that error on each call. I think go-pg wants previous behavior.

airfdl:
once meaning only executed  one time, store error in local instance, and each call will get a Idempotent error when fn call again.

vmihailenco:
https://github.com/go4org/go4/blob/master/syncutil/once.go#L28 disagrees with you, but anyway - go-redis relies on existing behavior.

PR #874, opened 2018-09-30
Modify circuit breaker

xianglinghui:
1. Add package flowtoken, which can control request traffic and adjust flow threshold automatically based on success rate per second.
2. Modify Redis client and Cluster client circuit breaker, change it to work with flowtoken.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"github.com/go-redis/redis/flowtoken"
 	"math"
 	"math/rand"
 	"net"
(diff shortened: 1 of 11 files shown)


cluster.go
```diff
@@ -1164,6 +1170,32 @@ func (c *ClusterClient) PoolStats() *PoolStats {
 	return &acc
 }
 
+// GetFlowtokenStates returns all flowtoken status.
+func (c *ClusterClient) GetFlowtokenStates() []*flowtoken.Snapshot {
```
vmihailenco: Looks like this func is unused?

vmihailenco:
Can you please rebase this on master and fix build? Specifically please replace convey with https://github.com/onsi/ginkgo - that is what we use for testing go-redis.

cluster.go
xianglinghui: User can trace flowtoken states by calling this func periodly.

xianglinghui:
@vmihailenco Sorry for the late reply to you, I am on vacation these days.

I have replaced convey with ginkgo, and rebase my branch on master. When I do "rebase" for the latest code, there have some weird conflicts, so I aborted the latest rebase. Please check whether I need to do rebase again.

PR #884, opened 2018-10-14
Add Limiter interface

vmihailenco:
This is an alternative to https://github.com/go-redis/redis/pull/874. Basically it defines rate limiter interface which allows to implement different limiting strategies in separate packages.

@xianglinghui what do you think? Is provided API enough to cover your needs? I am aware that code like https://github.com/go-redis/redis/blob/master/ring.go#L618-L621 requires some work in go-redis, but other than that it seems to be enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ services:
   - redis-server
 
 go:
-  - 1.7.x
-  - 1.8.x
   - 1.9.x
   - 1.10.x
   - 1.11.x
--- a/cluster.go
+++ b/cluster.go
@@ -50,6 +50,9 @@ type ClusterOptions struct {
 	// and Cluster.ReloadState to manually trigger state reloading.
 	ClusterSlots func() ([]ClusterSlot, error)
 
+	// Optional hook that is called when a new node is created.
+	OnNewNode func(*Client)
+
 	// Following options are copied from Options struct.
 
 	OnConnect func(*Conn) error
@@ -164,6 +167,10 @@ func newClusterNode(clOpt *ClusterOptions, addr string) *clusterNode {
 	node.latency = math.MaxUint32
 	if clOpt.RouteByLatency {
 		go node.updateLatency()
+	}
+
+	if clOpt.OnNewNode != nil {
+		clOpt.OnNewNode(node.Client)
 	}
 
 	return &node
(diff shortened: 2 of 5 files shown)


xianglinghui:
@vmihailenco Yes, this way can meet my needs. The rate limiter is implemented by myself and specified at Options.

The Limiter API may need node address param, because rate limit is for problem node, not for the whole cluster, like this:
type Limiter interface {
	Take(addr string) error
	Put(addr string, err error)
}

vmihailenco:
Can't you create a separate Limiter for each node?

xianglinghui:
@vmihailenco Ah, I misunderstood, I thought there is just one instance of  Limiter for a cluster client before.

When a problem node is limited by the Limiter, we may need print the address of it to log, then we can locate problem faster.

So Limiter may need an API to set the info of node which the limiter has stand for.

xianglinghui:
@vmihailenco yes，I think this way can work. The api "redis_limiter.New" just pass an address of the node, I think will be more better.

vmihailenco:
Okay, updated the PR

xianglinghui:
@vmihailenco Will you add Limiter to ClusterClient? The PR seems just support Client.

xianglinghui:
@vmihailenco yes，I think it is great.

vmihailenco:
Okay, added `OnNewNode` option.

xianglinghui:
@vmihailenco It is great, thanks very much!

vmihailenco:
Thank you! Do you think we can/should merge this PR?

xianglinghui:
@vmihailenco yes.

xianglinghui:
@vmihailenco Does this feature will merge into master branch?

vmihailenco:
Yes, sorry - I am just not quite happy with interface. Probably following is better:

```
type Limiter interface {
    // Allow return a nil if operation is allowed or an error otherwise.
    // If operation is allowed client must report the result of operation
    // whether is a success or a failure.
    Allow() error
    // ReportResult reports the result of previously allowed operation.
    // nil error indicates a success, non-nil error indicates a failure.
    ReportResult(err error)
}
```

xianglinghui:
@vmihailenco I have run a test for Limiter in my project，there have a bug in redis.go.

type baseClient struct {
	limiter  Limiter
	...
}
type Client struct {
	baseClient
	limiter Limiter
}
func (c *Client) SetLimiter(l Limiter) *Client {
	c.limiter = l
	return c
}

baseClient and Client both have "limiter" field, the c.limiter in SetLimiter will not point to baseClient's limiter，so SetLimiter will not set value to baseClient.limiter.

vmihailenco:
Should be fixed thanks.

Issue #575, opened 2017-06-07
Pub/Sub example no longer working


holladayio:
The latest updates changed pub/sub functionality. The example no longer works. The following will block waiting for messages and not pick up the first message.

```
package main

import (
"fmt"
"github.com/go-redis/redis"
)

func main() {
	client := redis.NewClient(&redis.Options{
		Addr:     "127.0.0.1:6379",
		})
	defer client.Close()

	pubsub := client.Subscribe("mychannel1")
	defer pubsub.Close()

	err := client.Publish("mychannel1", "hello").Err()
	if err != nil {
		panic(err)
	}

	msg, err := pubsub.ReceiveMessage()
	if err != nil {
		panic(err)
	}

	fmt.Println(msg.Channel, msg.Payload)
}
```

The subscriber must call `ReceiveMessage()` before publishing the message. This will produce the expected output.

```
package main

import (
"fmt"
"github.com/go-redis/redis"
)

func main() {
	client := redis.NewClient(&redis.Options{
		Addr:     "127.0.0.1:6379",
		})
	defer client.Close()

	pubsub := client.Subscribe("mychannel1")
	defer pubsub.Close()

	go func() {
		msg, err := pubsub.ReceiveMessage()
		if err != nil {
			panic(err)
		}

		fmt.Println(msg.Channel, msg.Payload)
	}()
	err := client.Publish("mychannel1", "hello").Err()
	if err != nil {
		panic(err)
	}
}
```

If this is the desired behavior the docs should be updated to reflect the changes

holladayio:
Also, does this change mean that messages can be missed while acting on the current message? (i.e. if thread isn't waiting on `ReceiveMessage()`, messages are dropped.)

vmihailenco:
The first example you posted works fine for me. And AFAIK this behavior of PubSub did not change recently.

I guess your example may not work if you configure Linux to drop data that is not immediately read by the client. But again, there were no changes in go-redis regarding this behavior. It always worked like this.

holladayio:
Hmm.. You're right. I reinstalled the library and the example works. 

However I still have an issue and I narrowed it down. I am calling Ping() to make sure I can contact Redis before performing Pub/sub. This causes the program to hang at ReceiveMessage() until I send another message. 

```
package main

import (
"fmt"
"github.com/go-redis/redis"
)

func main() {
	client := redis.NewClient(&redis.Options{
		Addr:     "127.0.0.1:6379",
		})
	defer client.Close()

[...]

PR #915, opened 2018-11-27
Add godoc about Receive call after Subscribe

robmccoll:
I had a race in my integration tests that ultimately boiled down to a subscription that wasn't always completing before the first publication. The debugging took a bit of spelunking and/or running across https://github.com/go-redis/redis/issues/575 to figure out. The behavior was surprising to me, but makes sense, so I figured a little documentation wouldn't hurt.

Haven't looked at the other client configuration Subscribe() methods. Would they benefit from a note as well?

Also, can it be expected that calling Receive() right after Subscribe() should always return a *Subscription?

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -460,6 +460,30 @@ func (c *Client) pubSub() *PubSub {
 
 // Subscribe subscribes the client to the specified channels.
 // Channels can be omitted to create empty subscription.
+// Note that this method does not wait on a response from Redis, so the
+// subscription may not be active immediately. To force the connection to wait,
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks - merging.

>Haven't looked at the other client configuration Subscribe() methods. Would they benefit from a note as well?

Yes, but we definitely do not want to copy/paste this into several places. Probably this note should be placed on PubSub.

>Also, can it be expected that calling Receive() right after Subscribe() should always return a *Subscription?

As your note says - if there are no other activities.

PR #924, opened 2018-12-07
xgroup: Add XGroupCreateMkStream

rvolosatovs:
- https://redis.io/commands/xgroup `CREATE` supports an undocumented(https://github.com/antirez/redis-doc/issues/1007) `MKSTREAM` option.
Added support for it as a boolean parameter to the `XGroupCreate` method.
- ~Added `redis.BusyGroup` const, which represents the error returned by `XGROUP CREATE` if the group already exists(see https://redis.io/commands/xgroup docs)~

Note, that `mkStream` is a breaking API change, so maybe adding a `XGroupCreateMkStream` is a better option, which is also more consistent with `ZAddNX` and similar.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -175,6 +175,7 @@ type Cmdable interface {
 	XRead(a *XReadArgs) *XStreamSliceCmd
 	XReadStreams(streams ...string) *XStreamSliceCmd
 	XGroupCreate(stream, group, start string) *StatusCmd
+	XGroupCreateMkStream(stream, group, start string) *StatusCmd
 	XGroupSetID(stream, group, start string) *StatusCmd
 	XGroupDestroy(stream, group string) *IntCmd
 	XGroupDelConsumer(stream, group, consumer string) *IntCmd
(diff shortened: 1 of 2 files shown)


internal/proto/reader.go
```diff
@@ -21,6 +21,8 @@ const (
 
 const Nil = RedisError("redis: nil")
 
+const BusyGroup = RedisError("BUSYGROUP Consumer Group name already exists")
```
vmihailenco: I don't see this returned from anywhere and TBH I don't think this deserves a separate variable. If you think otherwise please add it in a separate PR.

vmihailenco:
Yes, let's add XGroupCreateMkStream - I think it is better API and is backwards compatible.

rvolosatovs:
Lack of `BusyGroup` variable means that the clients need to do
```go
if err.Error() == "BUSYGROUP Consumer Group name already exists" {
	// handle
}
```
which is ugly

vmihailenco:
Yes, but
- you still can have that var in your code
- go-redis does not try to maintain the list of errors return by Redis

PR #928, opened 2018-12-10
slowlog command support

winjeg:
need to use in my own projects and it is a common requirement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.rdb
 testdata/*/
+*.idea/
--- a/command.go
+++ b/command.go
@@ -690,10 +690,7 @@ func commandSlowLogParser(rd *proto.Reader, n int64) (interface{}, error) {
 }
 
 func slowLogParse(rd *proto.Reader, n int64) (interface{}, error) {
-
 	var slowLog SlowLog
-	var err error
-
 	idStr, err := rd.ReadString()
 	id, err := strconv.ParseInt(idStr, 10, 64)
 	if err != nil {
--- a/command.go
+++ b/command.go
@@ -691,22 +691,19 @@ func commandSlowLogParser(rd *proto.Reader, n int64) (interface{}, error) {
 
 func slowLogParse(rd *proto.Reader, n int64) (interface{}, error) {
 	var slowLog SlowLog
-	idStr, err := rd.ReadString()
-	id, err := strconv.ParseInt(idStr, 10, 64)
+	id, err := readString2Int(rd)
 	if err != nil {
 		return nil, err
 	}
 	slowLog.Id = id
 
-	timeStr, err := rd.ReadString()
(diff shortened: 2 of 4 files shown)


command.go
```diff
@@ -640,6 +640,95 @@ func (cmd *FloatCmd) readReply(rd *proto.Reader) error {
 	cmd.val, cmd.err = rd.ReadFloatReply()
 	return cmd.err
 }
+//------------------------------------------------------------------------------
+type SlowLog struct {
+	Id       int64
+	ExecTime int64
+	Time     int64
```
vmihailenco: Can this be more Go friendly

```
Time time.Time
Duration time.Duration
```

.gitignore
```diff
@@ -1,2 +1,3 @@
 *.rdb
 testdata/*/
+*.idea/
```
vmihailenco: Please revert. These files should be added to global gitignore file.

command.go
```diff
@@ -640,6 +640,95 @@ func (cmd *FloatCmd) readReply(rd *proto.Reader) error {
...
+		}
+		vv := v.(*SlowLog)
+		logs =  append(logs, vv)
+	}
+	return logs, nil
+}
+
+func slowLogParse(rd *proto.Reader, n int64) (interface{}, error) {
+	var slowLog SlowLog
+	id, err := readString2Int(rd)
```
vmihailenco: Is it possible to use rd.ReadInt?

winjeg: At first,  I thought so. but when i use `rd.ReadInt` an error occured
```
redis: can't parse string reply: ":1544754674"
```

so I changed to `rd.ReadString`

command.go
winjeg: Hey, can you do me a favor, merge it  and then change it for me?
I deleted the original repository, and can't append the change here now.


I've already prepared the code here:

```go
type SlowLog struct {
	Id       int64
	Time     time.Time
	Duration time.Duration
	Args     []string
}

func slowLogParse(rd *proto.Reader, n int64) (interface{}, error) {
	var slowLog SlowLog
	id, err := readString2Int(rd)
	if err != nil {
		return nil, err
	}
	slowLog.Id = id

	t, err := readString2Int(rd)
	if err != nil {
		return nil, err
	}
	slowLog.Time = time.Unix(int64(t), 0)

	execTime, err := readString2Int(rd)
	if err != nil {
		return nil, err
	}
	slowLog.Duration = time.Millisecond * time.Duration(execTime)

	args, err := rd.ReadArrayReply(stringSliceParser)
	if err != nil {
		return nil, err
	}
	slowLog.Args = args.([]string)
	return &slowLog, nil
}

```

command.go
vmihailenco: Then you shoud use ReadIntReply https://github.com/go-redis/redis/blob/master/internal/proto/reader.go#L90

Issue #920, opened 2018-12-06
Add CLIENT ID


theodesp:
https://redis.io/commands/client-id
I will be willing to submit a PR if its OK.

vmihailenco:
Sure, please send a PR.

PR #929, opened 2018-12-11
Add Client Id

theodesp:
Closes https://github.com/go-redis/redis/issues/920


vmihailenco:
I have changed Id to ID because that is what we already use in other places.

PR #931, opened 2018-12-11
Retry cluster pipeline read errors

### LINKED ISSUE #930: cluster client doesn't retry on pipeline read timeouts
Here is an example:

```go
package main

import (
	"fmt"
	"time"

	"github.com/go-redis/redis"
)

func main() {
	fmt.Println("start")
	options := &redis.ClusterOptions{
		Addrs:              []string{"0.0.0.0:7000"},
		MaxRedirects:       8,
		MaxRetries:         8,
		MinRetryBackoff:    1 * time.Second,
		MaxRetryBackoff:    1 * time.Minute,
		DialTimeout:        100 * time.Millisecond,
		ReadTimeout:        100 * time.Millisecond,
		WriteTimeout:       100 * time.Millisecond,
	}
	c := redis.NewClusterClient(options)
	fmt.Printf("%+v\n", c.Options())
	defer c.Close()
	fmt.Println("ping", c.Ping())
	options.ReadTimeout = time.Millisecond
	pipe := c.Pipeline()
	for i := 0; i < 128; i++ {
		z := redis.Z{
			Score:  float64(0),
			Member: fmt.Sprintf("test%d", i),
		}
		pipe.ZAdd("z", z)
	}
	_, err := pipe.Exec()
	fmt.Println(err)
}
```

- `pipelineReadCmds` doesn't update the `failedCmds` map https://github.com/go-redis/redis/blob/master/cluster.go#L1352.
- `defaultProcessPipeline` doesn't retry because `failedCmds` is empty https://github.com/go-redis/redis/blob/master/cluster.go#L1275.

akrylysov:
Fixes #930.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1344,13 +1344,13 @@ func (c *ClusterClient) pipelineProcessCmds(
 	}
 
 	err = cn.WithReader(c.opt.ReadTimeout, func(rd *proto.Reader) error {
-		return c.pipelineReadCmds(rd, cmds, failedCmds)
+		return c.pipelineReadCmds(node, rd, cmds, failedCmds)
 	})
 	return err
 }
 
 func (c *ClusterClient) pipelineReadCmds(
-	rd *proto.Reader, cmds []Cmder, failedCmds *cmdsMap,
+	node *clusterNode, rd *proto.Reader, cmds []Cmder, failedCmds *cmdsMap,
 ) error {
 	for _, cmd := range cmds {
 		err := cmd.readReply(rd)
@@ -1366,9 +1366,12 @@ func (c *ClusterClient) pipelineReadCmds(
 			continue
 		}
 
-		return err
+		cmd.setErr(err)
+		failedCmds.mu.Lock()
+		failedCmds.m[node] = append(failedCmds.m[node], cmd)
+		failedCmds.mu.Unlock()
 	}
-	return nil
+	return cmdsFirstErr(cmds)
 }
 
 func (c *ClusterClient) checkMovedErr(
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -683,6 +683,40 @@ var _ = Describe("ClusterClient", func() {
 			Expect(len(keys)).To(BeNumerically("~", nkeys, nkeys/10))
 		})
 
+		It("retries pipeline network errors", func() {
+			optNoRetry := redisClusterOptions()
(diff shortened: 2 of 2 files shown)


cluster.go
```diff
@@ -1366,9 +1366,12 @@ func (c *ClusterClient) pipelineReadCmds(
 			continue
 		}
 
-		return err
+		cmd.setErr(err)
```
vmihailenco: this is not needed - cmd already has an error after readReply

cluster.go
```diff
@@ -1366,9 +1366,12 @@ func (c *ClusterClient) pipelineReadCmds(
...
+		failedCmds.mu.Lock()
+		failedCmds.m[node] = append(failedCmds.m[node], cmd)
+		failedCmds.mu.Unlock()
 	}
-	return nil
+	return cmdsFirstErr(cmds)
```
vmihailenco: If there is a `redis.Nil` it will be returned here. I think this is wrong.

cluster_test.go
```diff
@@ -683,6 +683,40 @@ var _ = Describe("ClusterClient", func() {
...
+
+			startRetry := time.Now()
+			_, err = clientRetry.Pipelined(func(pipe redis.Pipeliner) error {
+				pipe.Ping()
+				return nil
+			})
+			Expect(err).To(HaveOccurred())
+			elapseRetry := time.Since(startRetry)
+
+			Expect(elapseRetry).To(BeNumerically(">", elapseNoRetry, 10*time.Millisecond))
```
vmihailenco: Well, this is not very reliable. Yes, clientRetry does more work than clientNoRetry but that is the only thing we test. No guarantees it does anything useful...

akrylysov: I followed the pattern used in the `should retry with backoff` test (https://github.com/go-redis/redis/blob/4b568cdf1adcfe5cab06ad6929ad5286b3db7138/redis_test.go#L158). Can you suggest a better way of testing it?

akrylysov:
@vmihailenco feel free to close my PR and merge yours, thanks!

cluster_test.go
vmihailenco: No, not really - sorry.

vmihailenco:
I've merged https://github.com/go-redis/redis/pull/936/files - hopefully it will not cause any problems.

Thanks for this PR anyway. It was very helpful.

Issue #921, opened 2018-12-06
Add CLIENT UNBLOCK


theodesp:
https://redis.io/commands/client-unblock
I will be willing to submit a PR if its OK.

vmihailenco:
Please send a PR. It will be better if you send one PR for CLIENT ID and CLIENT UNBLOCK making sure it addresses your use case.

One possible complication is that normally Redis client uses random connection from the pool and each connection will have different id.

PR #932, opened 2018-12-11
Add Client Unblock

theodesp:
https://github.com/go-redis/redis/issues/921

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -2068,12 +2068,14 @@ func (c *cmdable) ClientID() *IntCmd {
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -2068,6 +2068,16 @@ func (c *cmdable) ClientID() *IntCmd {
 	return cmd
 }
 
+func (c *cmdable) ClientUnblock(id int64, isError bool) *IntCmd {
```
vmihailenco: Can you please add 2 commands instead: `ClientUnblock` and `ClientUnblockWithError`?

theodesp: PR Updated

PR #939, opened 2018-12-19
FIX: WATCH example stack protection (with retry count)

pascaldekloe:
Includes some comments on the transaction steps.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,8 +1,8 @@
 package redis_test
 
 import (
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


vmihailenco:
Why do we need to retry? Current code seems to work fine.

pascaldekloe:
That's how WATCH works @vmihailenco. The current code also reties by calling the function again, risking a stackoverflow. Bad example.

vmihailenco:
Okay, fair point.

PR #943, opened 2018-12-29
Fix: xreadgroup command

karimsa:
The current version does not accept message IDs in its input and subsequently does not write them out as part of the command - but that is a requirement for the `xreadgroup` command. 

It's still possible with the current API, but feels like more of a hack then a feature. You could pass in `[]string{ "streamA", "streamB", ">", ">" }` - but the field is "streams" so it seems odd to pass it IDs as well. Whereas a map for the streams field makes sense because it maps from the stream to the message ID.

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -3564,7 +3564,7 @@ var _ = Describe("Commands", func() {
 				res, err := client.XReadGroup(&redis.XReadGroupArgs{
 					Group:    "group",
 					Consumer: "consumer",
-					Streams:  []string{"stream", ">"},
+					Streams:  map[string]string{"stream": ">"},
 				}).Result()
 				Expect(err).NotTo(HaveOccurred())
 				Expect(res).To(Equal([]redis.XStream{{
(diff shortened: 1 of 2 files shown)


commands_test.go
```diff
@@ -3564,7 +3564,7 @@ var _ = Describe("Commands", func() {
 				res, err := client.XReadGroup(&redis.XReadGroupArgs{
 					Group:    "group",
 					Consumer: "consumer",
-					Streams:  []string{"stream", ">"},
+					Streams:  map[string]string{"stream": ">"},
```
vmihailenco: AFAIR I've chosen `[]string{"stream", ">"}` over `map[string]string{"stream": ">"}`, because maps are unordered.

vmihailenco:
It indeed feels like a hack, but it was done deliberately because maps are unordered. I think we should not merge this.

commands_test.go
karimsa: Could we at least add some doc examples because the current API is not very intuitive?

karimsa: Or perhaps having a separate ordered field for the IDs might be most intuitive

PR #946, opened 2019-01-08
Fix race condition when creating the cluster node

allenwq:
After upgrading to v6, we see a lot of errors of `redis: client is closed` in our system. 

It turns out that the cluster node returned in `GetOrCreate` is already closed, this is mainly due to the code piece below:
```
       // nodeCreateGroup is a singleflight.Group
        v, err := c.nodeCreateGroup.Do(xxx)

       node, ok := c.allNodes[addr]
       if ok {
		_ = v.(*clusterNode).Close()	
        // here `v` actually is `node` in some cases due to the using of singleflight.Group
		return node, err
	}
```

In specs of `singleflight.Group.Do`:
> Do executes and returns the results of the given function, making sure that only one execution is in-flight for a given key at a time.
> If a duplicate comes in, the duplicate caller waits for the original to complete and receives the same results.

Here `nodeCreateGroup.Do(xxx)` returns the **same** cluster node to different callers and only one caller gets the mutex and set the `c.allNodes[addr] = node`, other callers will do `node.Close()` which closes the node.  

This issue was originally introduced in https://github.com/go-redis/redis/pull/712/files

The fix is to remove `singleflight` group and make the creation of cluster node inside the mutex. I am not sure why `singleflight` was introduced originally. 

Comments are most welcome, I might missed out certain things.


vmihailenco:
It was done this way because previously `node.updateLatency()` was called synchronously and potentially that operation can be slow. It is not the case now though.

allenwq:
Thanks @vmihailenco for your super quick response! Hope this can be released soon.

vmihailenco:
Thank you! This should be already released as https://github.com/go-redis/redis/tree/v6.15.1

PR #950, opened 2019-01-09
cluster should refresh the state before a new pubsub connection

regeda:
Hey!

I have an issue with subscriptions (SUBSCRIBE to a channel) on the cluster setup. The client doesn't resubscribe to a promoted master when the previous master failed. The current implementation freezes a master node on the subscription setup and never update until the predefined master back alive.

Now, pubsub state refreshes the cluster state before a new connection returning.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -397,6 +397,10 @@ type clusterSlot struct {
 	nodes      []*clusterNode
 }
 
+func (c *clusterSlot) between(slot int) bool {
+	return slot >= c.start && slot <= c.end
+}
+
 type clusterSlotSlice []*clusterSlot
 
 func (p clusterSlotSlice) Len() int {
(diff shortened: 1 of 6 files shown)


regeda:
Tests failed because unstable Redis returned `(nil)` for the empty slices instead of `(empty list or set)`.

regeda:
Hey @vmihailenco 

Do you have any suggestion corresponding to PR? The problem is actual for my prod environment and I want to use the upstream instead of a fork.

vmihailenco:
Hi,

I've sent https://github.com/go-redis/redis/pull/955 with fixes from this PR. What do you think about it?

regeda:
@vmihailenco Let's drive the PR. Do you have any concerns corresponding to the implementation?

vmihailenco:
The problem with this PR is that there are many unrelated changes I am not sure I want to merge.

PR #959, opened 2019-02-01
Cluster commands respecting keyslots from command input

greggjs:
Currently, the cluster command "GETKEYSINSLOT" passes in a keyslot as a parameter. This parameter needs to be recognized as the slot to use when executing the command. Currently, it ignores it and picks a random node. This PR fixes this issue.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -775,6 +775,11 @@ func cmdSlot(cmd Cmder, pos int) int {
 }
 
 func (c *ClusterClient) cmdSlot(cmd Cmder) int {
+	args := cmd.Args()
+	if args[0] == "cluster" && args[1] == "getkeysinslot" {
+		return args[2].(int)
+	}
+
 	cmdInfo := c.cmdInfo(cmd.Name())
 	return cmdSlot(cmd, cmdFirstKeyPos(cmd, cmdInfo))
 }
--- a/cluster.go
+++ b/cluster.go
@@ -767,8 +767,11 @@ func cmdSlot(cmd Cmder, pos int) int {
 	if pos == 0 {
 		return hashtag.RandomSlot()
 	}
-	firstKey := cmd.stringArg(pos)
-	return hashtag.Slot(firstKey)
+	val, ok := cmd.Args()[pos].(int)
+	if ok {
+		return val
+	}
+	return hashtag.Slot(cmd.stringArg(pos))
 }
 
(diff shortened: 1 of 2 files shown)


vmihailenco:
Could you instead change https://github.com/go-redis/redis/blob/dba9cb31f566f15c82162dbde38ed8939d8b8e5a/cluster.go#L777-L780 like this:

```
func (c *ClusterClient) cmdSlot(cmd Cmder) int {
        args := cmd.Args()
	if args[0] == "cluster" && args[1] == "getkeysinslot" {
		return args[2].(int)
	}

	cmdInfo := c.cmdInfo(cmd.Name())
	return cmdSlot(cmd, cmdFirstKeyPos(cmd, cmdInfo))
}
```

greggjs:
Yea sure no prob man

cluster.go
```diff
@@ -767,11 +767,19 @@ func cmdSlot(cmd Cmder, pos int) int {
 	if pos == 0 {
 		return hashtag.RandomSlot()
 	}
-	firstKey := cmd.stringArg(pos)
-	return hashtag.Slot(firstKey)
+	val, ok := cmd.Args()[pos].(int)
+	if ok {
+		return val
+	}
+	return hashtag.Slot(cmd.stringArg(pos))
```
vmihailenco: Can we revert this change?

command.go
```diff
@@ -74,6 +74,11 @@ func cmdString(cmd Cmder, val interface{}) string {
 
 func cmdFirstKeyPos(cmd Cmder, info *CommandInfo) int {
 	switch cmd.Name() {
+	case "cluster":
+		switch cmd.stringArg(1) {
+		case "getkeysinslot":
+			return 2
+		}
```
vmihailenco: can we revert this change as well?

cluster.go
greggjs: So if you look at this line of code, https://github.com/go-redis/redis/blob/f3804cd0b74c6515ef206dabe7b314f2765fc788/cluster.go#L794

We reference the code that I changed. Would you suggest I change this method call to the cluster client specific method?

command.go
greggjs: Sure

cluster.go
greggjs: I updated the code to reflect the change I was referring to.

vmihailenco:
Thanks

PR #960, opened 2019-02-01
Add go.mod for Go 1.11 Modules

wmh:


vmihailenco:
Thanks, but go mod works fine without this change

wmh:
nah, i can't find definition without this file.

PR #964, opened 2019-02-04
Fix ClusterClient.WrapProcessPipeline

yansal:


vmihailenco:
What do you think about adding `WrapProcessTxPipeline` instead? It should be more accurate...

yansal:
I'm OK with not merging that change: I don't need it anymore and I'm not 100% it's a good idea to merge it.

Issue #962, opened 2019-02-03
How to write custom commands in pipeline?


mikefaraponov:
P.S. I need to access renamed CONFIG in MULTI EXEC block

mikefaraponov:
Do method still don't available in Pipeliner interface

mikefaraponov:
Nice work!

mikefaraponov:
Maybe Do should be placed to StatefulCmder, what do you think?

vmihailenco:
`StatefulCmder` is an interface - it will not help much.

PR #965, opened 2019-02-08
Support custom command in pipeline

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/962


mikefaraponov:
Do method still don't available in Pipeliner interface

PR #967, opened 2019-02-12
Add Pipeliner.Do

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/962#issuecomment-462442503


mikefaraponov:
Maybe Do should be placed into Cmdable interface?

PR #975, opened 2019-02-20
Export Redis ring shards to be able to work with them directly

vkuptcov:
As the PR about MGet/MSet commands for Redis ring support was declined ( https://github.com/go-redis/redis/pull/493 ), layers above `go-redis/redis` should have the possibility to understand, which shard is responsible for which set of keys.
The PR just makes the `Ring.Shards` field exportable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 *.rdb
 testdata/*/
+
+# IntelliJ Idea files
+.idea
+redis.iml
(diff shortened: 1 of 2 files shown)


vmihailenco:
I don't like the idea of exporting library guts - it makes refactoring harder/impossible without breaking backwards compatibility.

As for the MGET/MSET support - I recommend using pipelines instead. It provides similar performance.

vkuptcov:
Agree. Before I didn't realize that for pipelining it must be a redis client responsibility to send a request to the appropriate node for each key.
Thank you!

Issue #433, opened 2016-12-01
Use Redis replica without cluster mode


pranas:
We have Redis master + replica configuration without using cluster mode. We would like to route reads to both based on latency. I see go-redis has this feature but it seems it's only available with cluster mode. Any plans to support it on non cluster client?

vmihailenco:
No, I am not aware of any way to achieve this. And since this is essentially a Redis Cluster with 2 nodes (master+slave) I don't see much value in supporting this in standard client...

jpza:
It's quite common for people to have Redis read-replicas without Redis being in cluster-mode, at least for now since it guarantees strong consistency and is operationally simpler. 

Another issue, [https://github.com/go-redis/redis/issues/673], is also asking for this feature. 

If you don't need master sharding and you just want to scale out reads than this non-cluster mode setup of read-replicas without cluster-mode is more performant. Since the demand for this feature is present and it has been implemented by https://github.com/luin/ioredis I am bumping this.

This difficulty I see here is how to iterate over read-replicas.

pranas:
It would be great if you allow to choose different routing strategies. For our use case we want routing by latency (which you already have in cluster client) while others might want round robin routing.

ramonberrutti:
@Kuqd I have the same problem. I solve now having two connections. redisClient and redisClientRead

pranas:
We also have a similar workaround. Since we have multiple read replicas we initialise multiple go-redis clients, put them in a slice and then randomly pick one for read operations.

Kuqd:
@pranas @ramonberrutti This is good workaround. Thanks.

Though for my use case it requires a bit of refactor since I'm reading and writing a lot. I'll think about it about thanks.

vmihailenco:
See https://godoc.org/github.com/go-redis/redis#example-NewClusterClient--ManualSetup. It uses 2 master nodes, but using single master should work too.

FarhanSajid1:
> We also have a similar workaround. Since we have multiple read replicas we initialise multiple go-redis clients, put them in a slice and then randomly pick one for read operations.

Has go-redis added support, or is this the solution at the moment?

michal-mazurkiewicz:
Up! 

Is there any out of the box solution to use redis with read-replica without cluster mode?

Av1shay:
Any updates on this?
It is really common to use redis with read replica without cluster mode

I70l0teN4ik:
Another bump here. 
[...]

PR #997, opened 2019-03-27
add FailoverReadOnlyClient

eafzali:
This is related to https://github.com/go-redis/redis/issues/433 my was to make `UnversalClient` to route ReadOnly Commands to Slaves under sentinel mode if `ReadOnly` is `true` in options.
This was my expectation when I started using this library but after deployment I realized it's only using slaves for Failover regardless of the `ReadOnly` option.

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -85,10 +85,26 @@ func NewFailoverClient(failoverOpt *FailoverOptions) *Client {
 		opt: opt,
 	}
 
+	opt.Dialer = func() (net.Conn, error) {
+		addr, err := failover.MasterAddr()
+		if err != nil {
+			return nil, err
+		}
+		return net.DialTimeout("tcp", addr, opt.DialTimeout)
+	}
+
+	connPool := newConnPool(opt)
+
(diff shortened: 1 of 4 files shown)


eafzali:
I think I need some help on tests. there's no output for it it just timed out for some reason.

firstguo:
this is a good feature

firstguo:
this is a good feature

leosunmo:
I would love this functionality as well. Right now the Helm chart for Redis doesn't support proper clustering, so I am stuck using Sentinel setup for now. I am doing a massive amount of reading from cache and not much setting, so reading from 3 nodes (master + 2 slaves) would be a lot better than hammering 1 master.

nextsux:
I would love this too!
Unfortunately this code can not work. You're using ClusterClient which is not designed to work with sentinel setup. You're stuck in infinite loop due "ERR This instance has cluster support disabled"

vmihailenco:
I believe I've fixed tests to pass in https://github.com/go-redis/redis/pull/1473. Please let me know if anything does not work as expected.

PR #1000, opened 2019-03-29
Add go.mod and go.sum

kassiansun:
Basically, it's done by `go mod init` and `go mod tidy`, this will leave out `github.com/onsi/ginkgo v1.8.0 // indirect` `github.com/onsi/gomega v1.5.0 // indirect` from every projects using `go-redis` but not using ginkgo and gomega


kassiansun:
ginkgo and gomega has migrated to go module, so the entries in `go.sum` is automatically shadowed from ginkgo and gomega's go.sum, by `go mod tidy`

kassiansun:
I'm not sure whether this will infect `go mod vendor`, close it for now

kassiansun:
Tested on local machine, `go mod vendor` won't copy test dependencies to `vendor`

joseph-norman:
Hi @kassiansun, I've recently looked into doing this as I'd really want `go-redis` to be able to use Go modules. Let me know if I can be of assistance with this.

kassiansun:
Of course you can, the original code is at `master` of https://github.com/kassiansun/redis

joseph-norman:
@kassiansun which test dependencies were missing when you ran `go mod vendor`? After checking each `_test.go` file I can see `github.com/onsi/ginkgo` and `github.com/onsi/gomega` imported and also copied into the vendor directory that's created.

Is there any additional ways for me to verify this is / isn't working as expected with this lib? Before I look into it further.

PR #1008, opened 2019-04-09
fix accuracy problems when float32 convert to flaot64

sam-ke:
**my problem:**
------------------------------------
    var score float32
    myscore = 66.7

	formatScore := cast.ToFloat32(fmt.Sprintf("%0.2f", myscore))
	member := "sam123"
	mycmd := []interface{}{"Zadd", "testzsetkey", formatScore, member}
	cmd := redis.NewCmd(mycmd...)

		Addr:        "127.0.0.1",
		Password:    "", // no password set
		DB:          0,       // use default DB
		PoolSize:    1,
		IdleTimeout: 500,
	})
	
	client.Process(cmd)
	//.... blabla
-----------------------------
"formatScore"  which its type will be convert to float64, then the data stored in Redis is not accurate! 
like this:
QQ截图20190409171103

**Result：**
then get the score of "sam123" from redis, it value is : 66.669998168945312,  
**_but my expected result is: 66.7_**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,3 @@ matrix:
 install:
   - go get github.com/onsi/ginkgo
   - go get github.com/onsi/gomega
-  - go get github.com/spf13/cast
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,4 @@ matrix:
 install:
   - go get github.com/onsi/ginkgo
   - go get github.com/onsi/gomega
+  - go get github.com/spf13/cast
--- a/command_test.go
+++ b/command_test.go
@@ -58,13 +58,26 @@ var _ = Describe("Cmd", func() {
 	})
 
 	It("safe float32 convert to float64", func() {
-		var f float32 = 66.97
+		var f float32
+		f = 66.97
+
 		set := client.Set("float_key", f, 0)
 		Expect(set.Err()).NotTo(HaveOccurred())
 
 		val, err := client.Get("float_key").Float64()
 		Expect(err).NotTo(HaveOccurred())
 		Expect(val).To(Equal(float64(66.97)))
+
+		member := "sam123"
+		mycmd := []interface{}{"Zadd", "zset_float_key", f, member}
+		cmd := redis.NewStatusCmd(mycmd...)
+
+		client.Process(cmd)
+		Expect(cmd.Err()).NotTo(HaveOccurred())
+
+		fscore := client.ZScore("zset_float_key", member)
+		Expect(fscore.Err()).NotTo(HaveOccurred())
+		Expect(fscore.Val()).To(Equal(float64(66.97)))
 	})
 
 })
--- a/command_test.go
+++ b/command_test.go
@@ -57,4 +57,14 @@ var _ = Describe("Cmd", func() {
 		Expect(f).To(Equal(float64(10)))
 	})
 
+	It("safe float32 convert to float64", func() {
+		var f float32 = 66.97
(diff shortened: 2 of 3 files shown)


vmihailenco:
Please add a test to the https://github.com/go-redis/redis/blob/master/commands_test.go.

Could we fix this without a dependency on `cast` package? Perhaps we could add `w.float32(n float32)` and `w.float64(n float64)` that will write a float32 without a cast to float64?

sam-ke:
> Please add a test to the https://github.com/go-redis/redis/blob/master/commands_test.go.
> 
> Could we fix this without a dependency on `cast` package? Perhaps we could add `w.float32(n float32)` and `w.float64(n float64)` that will write a float32 without a cast to float64?

Ok, I will deal with it as soon as possible.

vmihailenco:
I've proposed alternative in https://github.com/go-redis/redis/pull/1009

Overall it looks like existing code is working fine - you just need to convert float64 back to float32

```go
package main

import (
	"fmt"

	"github.com/go-redis/redis"
)

func main() {
	red := redis.NewClient(&redis.Options{})

	f := float32(66.97)

	err := red.Set("float_key", f, 0).Err()
	if err != nil {
		panic(err)
	}

	n, err := red.Get("float_key").Float64()
	if err != nil {
		panic(err)
	}

	fmt.Println(n, float32(n))
	// 66.97000122070312 66.97
}
```

`Float32` helper should reduce confusion a bit.

sam-ke:
Your plan is really flexible，Good
But it didn't solve my problem， Because I store it in redis as a float32 type by formatted "%0.2f" and then provide data support for other applications, eg: the PHP web system... not go luanguage system. however these system does not call Float32(), other than this, PHP web system does not know whether the data is float32 or float64.  Then there will be an accuracy problem. 

Although we can use the format function to reformat in the php system, but the cost is a bit large

My request is just to get the same data as I wrote it.  I hope you will consider our usage scenarios. 
thanks!

sam-ke:
> I've proposed alternative in #1009
> 
> Overall it looks like existing code is working fine - you just need to convert float64 back to float32
> 
> ```go
> package main
> 
> import (
> 	"fmt"
> 
> 	"github.com/go-redis/redis"
> )
> 
> func main() {
> 	red := redis.NewClient(&redis.Options{})
> 
> 	f := float32(66.97)
> 
> 	err := red.Set("float_key", f, 0).Err()
> 	if err != nil {
> 		panic(err)
> 	}
> 
> 	n, err := red.Get("float_key").Float64()
> 	if err != nil {
> 		panic(err)
> 	}
> 
> 	fmt.Println(n, float32(n))
> 	// 66.97000122070312 66.97
> }
> ```
> 
> `Float32` helper should reduce confusion a bit.

Your plan is really flexible，Good
But it didn't solve my problem， Because I store it in redis as a float32 type by formatted "%0.2f" and then provide data support for other applications, eg: the PHP web system... not go luanguage system. however these system does not call Float32(), other than this, PHP web system does not know whether the data is float32 or float64. Then there will be an accuracy problem.

Although we can use the format function to reformat in the php system, but the cost is a bit large

My request is just to get the same data as I wrote it. I hope you will consider our usage scenarios.
thanks!

sam-ke:
@vmihailenco

sam-ke:
Can you see my reply?  @vmihailenco

vmihailenco:
Just try it - all should be working as expected. `float32(66.97)` is stored correctly in Redis. You can verify it using redis-cli.

Issue #990, opened 2019-03-08
Pipeline is not concurrency-safe with streams


kak-tus:
I use Cluster client with pipelining.
Then start 4 goroutines, each goroutine get pipeline, and do 100000 xadd command in pipeline.
After that I check stream length and got not exactly 400000 messages in it, but 402384 (or near this).

How to reproduce.

Start cluster
```
docker run --rm -it --network host redis:alpine --cluster-enabled yes --cluster-config-file nodes.conf --port 7001
docker run --rm -it --network host redis:alpine --cluster-enabled yes --cluster-config-file nodes.conf --port 7002
docker run --rm -it --network host redis:alpine --cluster-enabled yes --cluster-config-file nodes.conf --port 7003
docker run --rm -it --network host redis:alpine --cluster-enabled yes --cluster-config-file nodes.conf --port 7004
docker run --rm -it --network host redis:alpine --cluster-enabled yes --cluster-config-file nodes.conf --port 7005
docker run --rm -it --network host redis:alpine --cluster-enabled yes --cluster-config-file nodes.conf --port 7006
```

Init cluster
```
docker run --rm -it --network host redis:alpine redis-cli --cluster create 172.17.0.1:7001 172.17.0.1:7002 172.17.0.1:7003 172.17.0.1:7004 172.17.0.1:7005 172.17.0.1:7006 --cluster-replicas 1
```

Run script
```
package main

import (
	"sync"

	"github.com/go-redis/redis"
)

func main() {
	rd := redis.NewClusterClient(&redis.ClusterOptions{
		Addrs:    []string{"172.17.0.1:7001", "172.17.0.1:7002"},
	})

	wg := sync.WaitGroup{}

	for j := 0; j < 4; j++ {
		wg.Add(1)
		go func() {
			pipe := rd.Pipeline()

			for i := 0; i < 100000; i++ {
				pipe.XAdd(&redis.XAddArgs{
					Stream: "test-stream",
					ID:     "*",
					Values: map[string]interface{}{"m": "{}"},
				})
			}

			_, err := pipe.Exec()
			if err != nil {
				println(err.Error())
			}

			wg.Done()
		}()
	}

	wg.Wait()
}
```

Check stream length
```
docker run --rm -it --network host redis:alpine redis-cli -c -h 172.17.0.1 -p 7003 xinfo stream test-stream
```

If you got correct stream length - please try to set greater concurrency.

kak-tus:
[...]

PR #1010, opened 2019-04-12
Add comments about pipeline technique.

kak-tus:
Notes about it's usage in case of big pipelines and low read/write timeouts.

Closes https://github.com/go-redis/redis/issues/990

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1182,6 +1182,19 @@ func (c *ClusterClient) reaper(idleCheckFrequency time.Duration) {
 	}
 }
 
+// Pipeline returns mechanism to realise Redis Pipeline technique.
+//
+// Pipelining is a technique to extremely speed up processing by packing
+// operations to batches, send them at once to Redis and read a replies in a
+// singe step.
+// See https://redis.io/topics/pipelining
+//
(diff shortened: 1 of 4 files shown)


cluster.go
```diff
@@ -1182,6 +1182,19 @@ func (c *ClusterClient) reaper(idleCheckFrequency time.Duration) {
...
+// singe step.
+// See https://redis.io/topics/pipelining
+//
+// Pay attention, that Pipeline is not a transaction, so you can get unexpected
+// results in case of big pipelines and small read/write timeouts.
+// Redis client has retransmission logic in case of timeouts, pipeline
+// can be retransmitted and commands can be executed more then once.
+// To avoid this: it is good idea to use reasonable bigger read/write timeouts
+// depends of your batch size and/or use TxPipeline.
 func (c *ClusterClient) Pipeline() Pipeliner {
```
vmihailenco: Could you please move this comment to the `Pipeliner` interface so we have 1 copy of this text?

kak-tus: Done.
But idea to place this comment with Pipeline function base on that: mostly users call this function to get Pipeliner, but not got directly Pipeliner interface.
So it is logically closer to Pipeline function.

Issue #1004, opened 2019-04-04
UniversalClient missing WrapProcessPipeline and missing interface for Ring


aspacca:
Hello,

I'm working on a Elastic APM agent for this library (github.com/aspacca/apmgoredis). As you can see since `UniversalClient` is missing `WrapProcessPipeline` I have to cast explicitly to `Client` or `ClusterClient`. Is there any specific reason why the method was not added to the interface?

As for `Ring`: it seems to implement `Cmdble` but not `UniversalClient` since it misses `Watch` method, what's the reason for this?

thanks

PR #1011, opened 2019-04-14
Added WrapProcessPipeline to UniversalClient, wrap c.processTxPipelin…

aspacca:
…e in cluster client

I'm not sure why those two implementations were missing and if there is any ratio for that. Not having had yet an answer on #1004 I decided to just push some of them.

Feel free to reject if I missed any requirement why these should not be implemented


vmihailenco:
Looks good - thanks

PR #1014, opened 2019-04-18
fix pubsub panic on redis reload

nnimo:
On reload redis cluster, pubSub do panic.
```
			cn, err := node.Client.newConn()
			if err != nil {
				// in this case cn still nil, but node not nil, 
				// and next call newConn (from pubsub._conn) do panic
				// we need set node to nil before return
				node = nil

				return nil, err
			}
```


vmihailenco:
Thanks

PR #1015, opened 2019-04-19
Ring Watch

aspacca:
Adding support for `Watch` method in `Ring` (this will make `Ring` fulfill `UniversalClient` interface)

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -653,6 +653,39 @@ func (c *Ring) Close() error {
 	return c.shards.Close()
 }
 
+func (c *Ring) Watch(fn func(*Tx) error, keys ...string) error {
+	if len(keys) == 0 {
+		return fmt.Errorf("redis: Watch requires at least one key")
+	}
+
+	var shards []*ringShard
+	for _, key := range keys {
+		if key != "" {
+			shard, err := c.shards.GetByKey(hashtag.Key(key))
+			if err != nil {
+				return err
+			}
+
+			shards = append(shards, shard)
+		}
+	}
+
+	if len(shards) == 0 {
+		return fmt.Errorf("redis: Watch requires at least one shard")
+	}
+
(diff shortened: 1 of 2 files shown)


ring.go
```diff
@@ -275,14 +275,14 @@ func (c *ringShards) Heartbeat(frequency time.Duration) {
 func (c *ringShards) rebalance() {
 	hash := newConsistentHash(c.opt)
 	var shardsNum int
+	c.mu.Lock()
```
vmihailenco: Is this necessary? `hash` is a local var and `shard.IsUp` uses atomics to it should be safe to be used in multiple goroutines.

ring.go
```diff
@@ -653,6 +653,39 @@ func (c *Ring) Close() error {
...
+		}
+	}
+
+	if len(shards) == 0 {
+		return fmt.Errorf("redis: Watch requires at least one shard")
+	}
+
+	if len(shards) > 1 {
+		for _, shard := range shards[1:] {
+			if shard.Client.String() != shards[0].Client.String() {
```
vmihailenco: Can we use `shard.Client != shards[0].Client` or `shard.Client.Options() != shards[0].Client.Options()`? Relying on `String` does not look right...

ring.go
aspacca: the problem is on `for name, shard := range c.shards`, you read from `c.shards` while `Ring.Close` can set it to `nil`, so the race condition (you can check travis build fails for the first commit)

aspacca: maybe I can only `RLock` around the range and then keep the full lock before writing to `c.hash`

ring.go
aspacca: I will use the first, anyway `String` is unique on host/port, that should be what we care only for, that's why I used it in the first place. maybe I'm missing something

PR #1017, opened 2019-04-22
Ring instrumentantions

aspacca:


vmihailenco:
Could you add a test please? I also believe the fix should be in `Ring.clone`,,,

aspacca:
done

Issue #1020, opened 2019-05-02
Unable to setup a password for sentinel in FailoverOptions


jdoucy:
Sentinel process can also be password protected.
In order to make it worked, I temporarily add a hack to set a password when instantiating the SentinelClient.

vmihailenco:
What specifically did you do?

ghostbody:
I met the same problem... What did you do to solve it? @jdoucy

jdoucy:
I made a quick and dirty fix, give me 2 secs, I'll share it

jdoucy:
@ghostbody here is what I've done, really hacky for now, should use another config field because the sentinel pass can be different
https://github.com/jdoucy/redis/commit/20578efe8c8f34211cbac7eeeab4f9646eda2c5a

ghostbody:
Thanks, and this is my solution, and it wokds. @jdoucy  https://github.com/go-redis/redis/pull/1030/files

jdoucy:
You are welcome ;-)

ramonberrutti:
Some problem, Tested that pull request and work. Thank.

PR #1030, opened 2019-05-22
Add Sentinel Password Option

ghostbody:
Add Sentinel Password Option

However, it can not support sentinels usage with different passwords, but I think this should not happen.

issue: https://github.com/go-redis/redis/issues/1020#issuecomment-494358866

### DIFF
--- a/options.go
+++ b/options.go
@@ -42,6 +42,10 @@ type Options struct {
 	// Optional password. Must match the password specified in the
 	// requirepass server configuration option.
 	Password string
+
+	// Optional sentinel password. Sentinel process can also be password protected.
+	SentinelPassword string
+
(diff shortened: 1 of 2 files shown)


options.go
```diff
@@ -42,6 +42,10 @@ type Options struct {
 	// Optional password. Must match the password specified in the
 	// requirepass server configuration option.
 	Password string
+
+	// Optional sentinel password. Sentinel process can also be password protected.
+	SentinelPassword string
```
ramonberrutti: This go in sentinel.go options and not here.

vmihailenco:
Thanks, but closing in favor of https://github.com/go-redis/redis/pull/1043

Issue #1031, opened 2019-05-24
Redis LOADING error should cause different node to be queried during pipeline reads in redis-cluster


koenoki:
When processing pipeline read commands in redis-cluster if LOADING error is encountered we retry to the same node because `failedCmds` is not updated in `pipelineReadCmds()` (see https://github.com/go-redis/redis/blob/master/cluster.go#L1332). This means that if there's an extended period of time in which a node responds with LOADING error, which can happen eg after a failover recovery event in which an entire read replica needs to be replicated from the master, no pipeline reads can succeed. We unfortunately experienced this very scenario in production recently.

Is it by design that upon LOADING error is encountered during pipeline reads go-redis doesn't attempt to look for a different node to query, and if so what is/are the rationale? One factor we considered was consistency, but in redis cluster there's already no data consistency guarantee across shards anyways. Or is this an oversight that should be fixed?

As a reference for non-pipeline reads `defaultProcess()` already handles LOADING error properly (see https://github.com/go-redis/redis/blob/master/cluster.go#L946).

vmihailenco:
@koenoki PTAL https://github.com/go-redis/redis/pull/1034

PR #1034, opened 2019-05-25
Mark node as loading in Cluster pipeline read

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/1031

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1265,7 +1265,7 @@ func (c *ClusterClient) mapCmdsByNode(cmds []Cmder, cmdsMap *cmdsMap) error {
 		return err
 	}
 
-	cmdsAreReadOnly := c.cmdsAreReadOnly(cmds)
+	cmdsAreReadOnly := c.opt.ReadOnly && c.cmdsAreReadOnly(cmds)
 	for _, cmd := range cmds {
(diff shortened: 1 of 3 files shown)


cluster.go
```diff
@@ -1312,6 +1312,10 @@ func (c *ClusterClient) pipelineProcessCmds(
 	err = cn.WithReader(c.opt.ReadTimeout, func(rd *proto.Reader) error {
 		return c.pipelineReadCmds(node, rd, cmds, failedCmds)
 	})
+	if err == nil {
```
theodesp: If you return `err` anyway why don't you return it as it is?

go.mod
```diff
@@ -4,3 +4,5 @@ require (
 	github.com/onsi/ginkgo v1.8.0
 	github.com/onsi/gomega v1.5.0
 )
+
+go 1.13
```
theodesp: We don't really need that info as nothing changed here

cluster.go
vmihailenco: Fixed

go.mod
vmihailenco: This is what go mod does automatically - not going to fight it on this matter

PR #1040, opened 2019-05-31
Feature/hook new

vmihailenco:


e-zhang:
@vmihailenco I think the change to replace `.clone()` with `*c` resulted in the `c.cmdable` still being bounded to the previous `c.Process` as a result the return from `WithContext` returns a client that on `.Process` still uses the old `ctx` or nil if it does not exist. 

Is this intended behavior?

vmihailenco:
@e-zhang You are right - should be fixed in https://github.com/go-redis/redis/pull/1080

Issue #582, opened 2017-06-21
What is the Context / WithContext methods of clients use for?


seriousben:
Hi, Just wondering if there is a way to use context when using this library?

vmihailenco:
Right now it is used only for passing some data with the client instance context. How would you like to use it?

seriousben:
I would like to context in the same way `request.WithContext` works where the context can cancel the redis request if it is canceled or timed out.
It could also help instrumentation.

With request:
```go
ctx := context.WithTimeout(...) // OR WithCancel | WithDeadline | ...
reqNoCtx, err := http.NewRequest(...)
if err != nil {
    return err
}

req := reqNoCtx.WithContext(ctx)
```


What it could look like in go-redis right now:
```
client.WithContext(ctx) // Same connection pool but with a specific context for operations
client.Get(...)
```

Or nicer with a `redis.ContextClient`:
```
ctxClient.Get(ctx, ...)
```

vmihailenco:
What is the use case for canceling Redis request? Is not `Options.ReadTimeout` enough?

seriousben:
Yes, it might be. I am just asking this question because I am used to passing context objects around to easily cancel work done when an incoming http request is cancelled or its computation is taking too long.

The difference between ReadTimeout and Context, is that the ReadTimeout is per redis operations where the context could have been created at the start of an incoming http request and passed down to the redis operation.

It might be a non issue, I was just wondering if the Context/WithContext of client would be doing what I expected them to do. And the answer is no in favor of using ReadTimeout which is a totally good solution. Thanks.

seriousben:
I'll close the issue now, thanks for the answers.

vmihailenco:
Let's keep this open for future references.

Overall go-redis could use `ctx.Deadline()` and set it on the underlying connection, e.g.

```
ctx, _ := context.WithTImeout(context.TODO(), time.Second)
client.WithContext(ctx).Get()
```

will result in something like

```
conn := pool.Get()
tm, ok := ctx.Deadline()
if ok {
    conn.SetReadDeadline(tm)
    conn.SetWriteDeadline(tm)
}
```

It should not be hard to do and can be occasionally useful...

lukescott:
There is also `context.WithCancel`. If a user aborts their http request you can use `ctx.Done()` to cancel the in-progress request.

lcosmin:
I have a scenario which could benefit from having the possibility of cancelling a Redis operation via a context:  

[...]

PR #1047, opened 2019-06-08
Use Context.Deadline to set net.Conn deadline

vmihailenco:
Updates https://github.com/go-redis/redis/issues/582

### DIFF
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -1,6 +1,7 @@
 package pool
 
 import (
+	"context"
 	"net"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


internal/pool/conn.go
```diff
@@ -93,3 +80,22 @@ func (cn *Conn) WithWriter(timeout time.Duration, fn func(wr *proto.Writer) erro
 func (cn *Conn) Close() error {
 	return cn.netConn.Close()
 }
+
+func (cn *Conn) deadline(ctx context.Context, timeout time.Duration) time.Time {
+	if ctx != nil {
+		tm, ok := ctx.Deadline()
+		if ok {
+			cn.SetUsedAt(tm)
```
abursavich: SetUsedAt should always be called with the current time.

internal/pool/conn.go
```diff
@@ -93,3 +80,22 @@ func (cn *Conn) WithWriter(timeout time.Duration, fn func(wr *proto.Writer) erro
 func (cn *Conn) Close() error {
 	return cn.netConn.Close()
 }
+
+func (cn *Conn) deadline(ctx context.Context, timeout time.Duration) time.Time {
```
abursavich: This should probably return the lesser of the context deadline and the timeout deadline, if both are set.

vmihailenco: That is debatable. I personally think that this is  low-level API and it is up for user to do whatever makes sense for them.

abursavich: It's the way the timeouts and deadlines are composed in the context package. Skimming through the standard library a bit, it's how the net Dialer's DialContext method works. I think it would would be less surprising to users.

context.WithTimeout and context.WithDeadline: https://godoc.org/context#WithDeadline
net.Dialer.DialContext: https://golang.org/src/net/dial.go#L105

vmihailenco: Thanks, that is a good point though dialer has all sorts of knobs (Timeout, Deadline, and context) so chosing the earliest deadline makes sense there (or at least they must chose between 3).

But in go-redis I *think* situation is slightly different because Options.Read/WriteTimeout exist on different level/layer than context. First is an option for all net read/writes and second is passed explicitly as an arg and is likely different for every command.

And using context's deadline is just more flexible - sometimes you may need to use higher than default timeout.

vmihailenco: After some consideration I think you are right and it makes sense to copy the default approach here - https://github.com/go-redis/redis/pull/1051

PR #1048, opened 2019-06-08
Check context.Done while waiting for a connection

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/582

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -130,8 +130,8 @@ func (p *ConnPool) NewConn() (*Conn, error) {
 	return p._NewConn(nil, false)
 }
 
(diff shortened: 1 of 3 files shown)


internal/pool/pool.go
```diff
@@ -248,15 +248,24 @@ func (p *ConnPool) getTurn() {
...
+		return ctx.Err()
 	case p.queue <- struct{}{}:
 		return nil
 	default:
 		timer := timers.Get().(*time.Timer)
 		timer.Reset(p.opt.PoolTimeout)
 
 		select {
+		case <-done:
+			return ctx.Err()
```
abursavich: On this path, the timer isn't stopped and returned to the pool.

vmihailenco: Thanks

PR #1054, opened 2019-06-14
Add more stats

cep21:
Adds stats for the direct length of the `queue` chan and for waits on an empty queue chan

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -375,7 +375,8 @@ func (p *ConnPool) Stats() *Stats {
 
 		TotalConns: uint32(p.Len()),
 		IdleConns:  uint32(idleLen),
-		QueueSize:  uint32(p.queue),
+		QueueSize:  uint32(len(p.queue)),
+		QueueWaits: atomic.LoadUint32(&p.stats.QueueWaits),
 		StaleConns: atomic.LoadUint32(&p.stats.StaleConns),
 	}
 }
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -31,6 +31,8 @@ type Stats struct {
 	TotalConns uint32 // number of total connections in the pool
 	IdleConns  uint32 // number of idle connections in the pool
 	StaleConns uint32 // number of stale connections removed from the pool
+	QueueSize  uint32 // Size of the outstanding PoolSize queue
+	QueueWaits uint32 // The times we had to sleep waiting our turn on the connection queue
 }
 
 type Pooler interface {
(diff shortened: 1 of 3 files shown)


internal/pool/pool.go
```diff
@@ -249,22 +251,23 @@ func (p *ConnPool) getTurn() {
 }
 
 func (p *ConnPool) waitTurn(ctx context.Context) error {
-	var done <-chan struct{}
-	if ctx != nil {
-		done = ctx.Done()
+	if ctx == nil {
+		// The background context's Done is nil by default.
+		ctx = context.Background()
 	}
```
vmihailenco: Please revert

vmihailenco:
I'd like to hear why this is needed.

internal/pool/pool.go
cep21: Why should this be reverted?  This version of the code is cleaner.

The existing code actually had a latent nil pointer bug due to how it was structured.

internal/pool/pool.go
```diff
@@ -249,22 +251,23 @@ func (p *ConnPool) getTurn() {
...
 
 	select {
-	case <-done:
+	case <-ctx.Done():
 		return ctx.Err()
```
cep21: The old code has a bug on this line (259).  This version fixes that bug.

cep21:
We are using go-redis in a clustered mode at at large scale (overall traffic of over 100,000 requests per second). We occasionally see large spikes of request times in our p99 latency.  We are trying to track down why this is happening.  We believe it is due to spikes in the wait queue.

internal/pool/pool.go
cep21: Sorry I should clarify this wasn't a bug, but instead a static analysis issue.  You block on 'done', but return ctx.Err, but allow ctx to be nil.  It fails linting because it's very strange to return a parameter value of something you've previously checked to allow nil, but in the current setup of the code it's fine since done is nil.  I've never seen `if ctx != nil` checks in Go code.  Usually these become context.TODO or context.Background as soon up the call stack as possible, like I'm doing here.

I can revert if you like

vmihailenco:
My concern is that QueueSize name is just mis-leading (number of queued commands?) and docs "Size of the outstanding PoolSize queue" does not help (me) to understand its meaning. And both QueueSize and QueueWait are implementation details - pool can be implemented without `queue chan struct{}` and those vars will be obsolete (which is not a big problem but still).

>We are trying to track down why this is happening. We believe it is due to spikes in the wait queue.

I am skeptical that these stats will help you to identify the problem. QueueWait of 10 nanoseconds is perfectly fine and QueueSize logged here and there does not say much.

So I suggest you to try to use this branch / your fork to identify the problem you are having. And if it turns out that I am wrong and these stats are useful please reach me and I will reconsider.

internal/pool/pool.go
vmihailenco: I agree that it is better to use context.TODO or context.Background instead of nil, but it should happen higher in the hierarchy - not here.

cep21:
What if the stat was changed to include the time inside `case p.queue <- struct{}{}:`?  We're seeing strange behaviour with the go-redis library and it's difficult to get insight into why our p99 latency occasionally spikes.

popepothead420:
Why are u in my computer an d phone jack

vmihailenco:
As was discussed it is not clear these stats are widely useful.

PR #1055, opened 2019-06-14
Inc StaleConns in ReapStaleConns call.

cep21:
The current implementation does not increment the reaped
connection counter if ReapStaleConns is called
inside func (c *ClusterClient) reaper(idleCheckFrequency time.Duration)

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -465,7 +465,7 @@ func (p *ConnPool) reaper(frequency time.Duration) {
 		if p.closed() {
 			break
 		}
-		n, err := p.ReapStaleConns()
+		_, err := p.ReapStaleConns()
 		if err != nil {
 			internal.Logf("ReapStaleConns failed: %s", err)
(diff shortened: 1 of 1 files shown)


vmihailenco:
Looking at the code at the code I don't think this change is needed. Could you add more details?

cep21:
If the `ClusterClient` calls `reaper`, it will not increment the reaped connection count.

cep21:
This line of code

`func NewClusterClient(opt *ClusterOptions) *ClusterClient {`

calls this line of code

`go c.reaper(opt.IdleCheckFrequency)`

Which calls this line of code 

`_, err := node.Client.connPool.(*pool.ConnPool).ReapStaleConns()`

Notice how `_` is assigned in the return.  This integer is never incremented anywhere, and the count of these reaped connections is throw away.

vmihailenco:
Okay, that makes sense - please fix the build.

PR #1062, opened 2019-06-21
sentinel support proxy, such as socks5

Jasonix:
current the redis client can access server over proxy, but the sentinel client can't

### DIFF
--- a/options.go
+++ b/options.go
@@ -37,6 +37,9 @@ type Options struct {
 	// Network and Addr options.
 	Dialer func(ctx context.Context, network, addr string) (net.Conn, error)
 
+	// For sentinel proxy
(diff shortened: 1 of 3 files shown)


options.go
```diff
@@ -37,6 +37,9 @@ type Options struct {
 	// Network and Addr options.
 	Dialer func(ctx context.Context, network, addr string) (net.Conn, error)
 
+	// For sentinel proxy
+	DumpDialer func(ctx context.Context, network, addr string) (net.Conn, error)
```
vmihailenco: This is wrong place for sentinel specific options. Probably you are looking for FailoverOptions.

Also I don't understand the logic behind `DumpDialer`

Issue #1071, opened 2019-06-28
Custom Dialer not taken into account in FailoverClient


MonstyB:
FailoverOptions struct contains a custom Dialer field to force (for example) a socks proxy.
It is not taken into account which to me not being able to connect because i need to use SOCKS proxy.

The problem seems to be located in sentinel.go :

`c.opt.Dialer = c.dial` 

-> this overrides the provided dialer.

MonstyB:
Hi, I pulled the updated master but it is still not working.

Custom dialer gets overridden by default dialer.

PR #1073, opened 2019-06-29
Use Dialer in Sentinel failover when available

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/1071

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -306,8 +306,9 @@ func (c *sentinelFailover) Close() error {
 
 func (c *sentinelFailover) Pool() *pool.ConnPool {
 	c.poolOnce.Do(func() {
-		c.opt.Dialer = c.dial
-		c.pool = newConnPool(c.opt)
+		opt := *c.opt
+		opt.Dialer = c.dial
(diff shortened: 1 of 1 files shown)


MonstyB:
Hi, I pulled the updated master but it is still not working.

Custom dialer gets overridden by default dialer.

This is my client declaration : 

`	client = redis.NewFailoverClient(&redis.FailoverOptions{
		MasterName:    masterName,
		SentinelAddrs: []string{clusterNode1, clusterNode2, clusterNode3},
		Password:      password,
			// This dialer is not used : must be a bug
			fmt.Println("Connecting to SOCKS5 prior to this.")
			dialer, err := proxy.SOCKS5("tcp", socksProxyHost, nil, proxy.Direct)
			if err != nil {
				fmt.Println("Oops, cannot connect to the proxy :", err)
			}
			return dialer.Dial(network, addr)
		},
	})
`

I tried to comment out : 
`opt.Dialer = c.dial` in sentinel.go and it finally went through my custom Dialer but unfortunately, "addr" is valued to "FailoverClient" so it can't connect to Redis.

PR #1074, opened 2019-07-01
Add PubSub.OnSubscription

huangjunwen:
Add a callback option to PubSub, which will be called after each subscription event (e.g. subscribe/unsubscribe/psubscribe/punsubscribe) .

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -40,6 +40,8 @@ type PubSub struct {
 	chOnce sync.Once
 	ch     chan *Message
 	ping   chan struct{}
+
+	onSubscribed func(*Subscription)
 }
 
 func (c *PubSub) String() string {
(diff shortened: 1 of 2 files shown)


vmihailenco:
Well, now it looks like `Channel` should be changed to

```
func (c *PubSub) Channel() <-chan interface{}
```

and send both `*Message` and `*Subscription`. Having 2 separate mechanisms (chan and OnSubscription) for delivering 2 types of messages does not look right.

huangjunwen:
What about adding something like:
```go
func (c *PubSub) Channels(msgChanSize, subChanSize int) (<-chan *Message, <-chan *Subscription) 
```
 This does not break compatible but use one more channel :)

vmihailenco:
I've decided to keep backwards compatibility and add `ChannelWithSubscriptions(size int) <-chan interface{}` instead.

Your suggestion about 2 channels is viable too but it loses information about message order which can be an important nuance in some rare cases.

PR #1078, opened 2019-07-03
add scan func for hset data to struct

stone-stones:
I would like to have a way to get hset direct to my own define struct


vmihailenco:
This was replaced by https://github.com/go-redis/redis/pull/1631

@stone-stones thanks for working on this, but the code here is way too simple and has no tests. It is my fault I have not tried to help you to improve it :(

PR #1079, opened 2019-07-03
feat: Add Client.SetOnClose

heavenmirror:
baseClient.onClose() can not set outside


vmihailenco:
This is private API needed because client is closed by the user. I don't think it is useful as public API because user has full control over closing the client and if needed this hook can be implemented in user space.

PR #1087, opened 2019-07-09
Add support for individual passwords for ring shards

mbenford:
This PR enhances the `Ring` client by adding a new `Passwords` field to it so that each shard can have its own password. This change doesn't introduce a breaking change since the existing `Password` field has precedence over the new one and applications already using the `Ring` client with a single password won't stop working.

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -27,6 +27,10 @@ type RingOptions struct {
 	// Map of name => host:port addresses of ring shards.
 	Addrs map[string]string
 
+	// Map of name => password of ring shards, to allow different shards to have
+	// different passwords. It will be ignored if the Password field is set.
+	Passwords map[string]string
+
 	// Frequency of PING commands sent to check shards availability.
 	// Shard is considered down after 3 subsequent failed checks.
 	HeartbeatFrequency time.Duration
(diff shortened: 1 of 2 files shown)


rafaeleyng:
This would be nice, because (depending on how your infrastructure works) sometimes you cannot set the same password to all shards.

ring.go
```diff
@@ -76,6 +80,10 @@ type RingOptions struct {
 }
 
 func (opt *RingOptions) init() {
+	if opt.Passwords == nil {
+		opt.Passwords = make(map[string]string)
```
vmihailenco: This can be removed. No need for an allocation - reading from nil map is fine.

ring_test.go
```diff
@@ -172,6 +172,34 @@ var _ = Describe("Redis Ring", func() {
...
+	Describe("shard passwords", func() {
+		It("can be initialized with a single password, used for all shards", func() {
+			ring = redis.NewRing(&redis.RingOptions{
+				Addrs: map[string]string {
+					"shard1": "host1",
+				},
+				Password: "some-password",
+			})
+
+			Expect(ring).ToNot(BeNil())
```
vmihailenco: ring can never be nil here. You should at least call `ring.Ping()` to test if this is working.

ring_test.go
```diff
@@ -172,6 +172,34 @@ var _ = Describe("Redis Ring", func() {
...
+					"shard-1": "host-1",
+					"shard-2": "host-2",
+				},
+				Passwords: map[string]string{
+					"shard-1": "pwd-1",
+					"shard-2": "pwd-2",
+				},
+			})
+
+			Expect(ring).ToNot(BeNil())
```
vmihailenco: Same here - this does not test anything.

ring.go
mbenford: Done!

ring_test.go
mbenford: :+1:

ring_test.go
mbenford: :+1:

vmihailenco:
Thanks

PR #1092, opened 2019-07-18
Making LpushX and RpushX variadic

sjindal995:


sjindal995:
Issue with username. Will create new one.

Issue #1064, opened 2019-06-24
Problems after Reconnection


billmi:
Start : 
```
client := redis.NewClient(&redis.Options{
		Addr:     c.Host + ":" + c.Port,
		Password: c.Password,
		DB:       c.Db,
		PoolSize : 10,
		MinIdleConns : 10,
	})
// listen
go func(){
	for{
		fmt.Println(client.PoolStats())
		time.Sleep(1 * time.Second)
	}
}()

```
redis # Clients:
connected_clients:12
client_longest_output_list:0
client_biggest_input_buf:0
blocked_clients:0

================
When,Restart Redis-Server..Waiting for reconnection..Re-check...
redis # Clients:
connected_clients:3
client_longest_output_list:0
client_biggest_input_buf:0
blocked_clients:0

Connection pool does not work after reconnection.

Please have a look~

PR #1105, opened 2019-07-26
protect pollSize and idleConnsLen with mutex

waterandair:
If addIdleConn() stopped due to the error, poolSize and idleConnsLen will be confusing.

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -92,9 +92,7 @@ func NewConnPool(opt *Options) *ConnPool {
 		idleConns: make([]*Conn, 0, opt.PoolSize),
 	}
 
-	for i := 0; i < opt.MinIdleConns; i++ {
-		p.checkMinIdleConns()
-	}
+	p.checkMinIdleConns()
 
 	if opt.IdleTimeout > 0 && opt.IdleCheckFrequency > 0 {
 		go p.reaper(opt.IdleCheckFrequency)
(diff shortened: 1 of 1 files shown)


internal/pool/pool.go
```diff
@@ -107,16 +105,21 @@ func (p *ConnPool) checkMinIdleConns() {
 	if p.opt.MinIdleConns == 0 {
 		return
 	}
-	if p.poolSize < p.opt.PoolSize && p.idleConnsLen < p.opt.MinIdleConns {
-		p.poolSize++
-		p.idleConnsLen++
-		go p.addIdleConn()
+	if p.poolSize < p.opt.PoolSize {
+		balanceLen := p.opt.MinIdleConns - p.idleConnsLen
+		for i := 0; i < balanceLen; i++ {
```
vmihailenco: Now it is possible that `p.poolSize >= p.opt.PoolSize`. Probably it should be:

```go
}
```

internal/pool/pool.go
```diff
@@ -107,16 +105,21 @@ func (p *ConnPool) checkMinIdleConns() {
...
+			go p.addIdleConn()
+		}
 	}
 }
 
 func (p *ConnPool) addIdleConn() {
 	cn, err := p.newConn(context.TODO(), true)
 	if err != nil {
+		p.poolSize--
+		p.idleConnsLen--
```
vmihailenco: I believe you should hold `connsMu` here. Otherwise it looks good :+1:

waterandair: Thanks for your review.I updated the code.

vmihailenco:
Should help with https://github.com/go-redis/redis/issues/1064

Issue #1107, opened 2019-07-29
 Fix: max_retries should fail when the context deadline timeouts


8treenet:
https://github.com/8treenet/redis/commit/3a11286f6894d096b03bf2028dde0c6e98e1d8ec

vmihailenco:
I don't think this is needed - `false` is returned by default.

8treenet:
I've already tested that the context timeout is not valid, so I'm still retrying.

vmihailenco:
I don't understand what that means...

8treenet:
It's time you learned Chinese.

vmihailenco:
Nah, I am strill struggling with English :)

Issue #1100, opened 2019-07-25
Retry sleeps supporting context cancellation


kollektiv:
Version: 6bc7daa5b1e86745a6976ac1c4dfe6c76ea6af1f

v7 now has context support. There are calls to `time.Sleep` but they do not respect context cancellations
https://github.com/go-redis/redis/blob/6bc7daa5b1e86745a6976ac1c4dfe6c76ea6af1f/redis.go#L247

An option is to use a context-aware sleep https://github.com/aws/aws-sdk-go/blob/bd8be1f3a00996ada22da77e8bdd22cd6c740732/aws/context_sleep.go#L12

vmihailenco:
Sounds good - could you send a PR?

Issue #1099, opened 2019-07-25
Read errors not set on Cmdable


kollektiv:
Version: 6bc7daa5b1e86745a6976ac1c4dfe6c76ea6af1f

If there is a read error that occurs here https://github.com/go-redis/redis/blob/6bc7daa5b1e86745a6976ac1c4dfe6c76ea6af1f/redis.go#L271 the error is not set on the Cmdable. Setting the read error on the Cmdable is useful to get the error in the AfterProcess hook. What are your thoughts on setting the read error on the Cmdable, like when write errors happen?

kollektiv:
My mistake, `cmd.readReply` will set the error onto the cmdable

vmihailenco:
Let's keep this open. I believe there is at least one case where `WithWriter` does not an error on the command and hopefully I can make code more obviously correct.

kollektiv:
In what case will `WithWriter` not set an error on the cmdable?

PR #1108, opened 2019-07-30
Cancel sleep when context is cancelled

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/1100
Closes https://github.com/go-redis/redis/issues/1107

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -754,7 +754,9 @@ func (c *ClusterClient) process(ctx context.Context, cmd Cmder) error {
 		var err error
 
 		if attempt > 0 {
-			time.Sleep(c.retryBackoff(attempt))
+			if err := internal.Sleep(ctx, c.retryBackoff(attempt)); err != nil {
+				return err
+			}
 		}
 
 		if node == nil {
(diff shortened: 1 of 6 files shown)


internal/error.go
```diff
@@ -9,10 +10,10 @@ import (
 )
 
 func IsRetryableError(err error, retryTimeout bool) bool {
-	if err == nil {
+	switch err {
+	case nil, context.Canceled, context.DeadlineExceeded:
```
kollektiv: Is this needed? I would expect `context.Canceled|DeadlineExceeded` to return false in the previous code

cluster.go
```diff
@@ -754,7 +754,9 @@ func (c *ClusterClient) process(ctx context.Context, cmd Cmder) error {
 		var err error
 
 		if attempt > 0 {
-			time.Sleep(c.retryBackoff(attempt))
+			if err := internal.Sleep(ctx, c.retryBackoff(attempt)); err != nil {
+				return err
```
kollektiv: Does there need to be a `cmd.setErr(err)` here and elsewhere?

internal/error.go
vmihailenco: It is not needed, but `err == context.Canceled` is fast(er) and it is also explicit.

cluster.go
vmihailenco: It should be. It is just becomes hard to track when `cmd.setErr` should be called. I hope to address this in https://github.com/go-redis/redis/issues/1099

PR #1112, opened 2019-08-01
Close single conn connection pool

patrickwhite256:
This is a bugfix for issue #1113.

This is intended to solve the issue while making as few other changes as possible. Since `SingleConnPool` is not used in any other context, this seemed reasonable.

I'm completely open to alternative ways to solve this, but the essence is that `SingleConnPool.Remove` _must_ flag the pool as no longer able to return a valid connection.

### DIFF
--- a/internal/error.go
+++ b/internal/error.go
@@ -2,12 +2,15 @@ package internal
 
 import (
 	"context"
+	"errors"
 	"io"
 	"net"
 	"strings"
 
 	"github.com/go-redis/redis/internal/proto"
 )
+
+var ErrSingleConnPoolClosed = errors.New("redis: SingleConnPool is closed")
 
 func IsRetryableError(err error, retryTimeout bool) bool {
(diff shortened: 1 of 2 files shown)


vmihailenco:
Hi,

SingleConnPool is used in [single place])(https://github.com/go-redis/redis/search?q=NewSingleConnPool&unscoped_q=NewSingleConnPool) only to initialize the connection - run commands such as `Auth` and `SelectDB`. If there is an error the connection is [removed](https://github.com/go-redis/redis/blob/master/redis.go#L172-L176) from the pool by the caller. After connection is initialized the SingleConnPool is not used anywhere else. That is why the implementation is so dumb / unfinished - it is not a real pool...

patrickwhite256:
The problem is that because `Remove` is not implemented, an error is hidden and the connection that should be discarded is in fact kept.

PR #1120, opened 2019-08-09
Pop idle connections from the front of the pool

kollektiv:
Version: 0cf98f921

Pop idle connections from the front so the pool behaves like a queue instead of a stack. go-redis performance is significantly better when the Redis cluster is under memory pressure (evicting data due to LRU). Most of the errors were read timeout errors from net.Conn.Read with a 1s deadline set, and they go away with this change. My theory is that conns are more uniformly used and the Redis server likes this better

# Testing Setup

AWS ElastiCache Redis Cluster setup:
* Engine 3.2.10 (Reproduces on 5.0.3 too)
* 3 shards
* 3 nodes per shard
* cache.r4.xlarge node type

Redis cluster client configuration:

```go
dialer := net.Dialer{
  Timeout:   1 * time.Second,
  KeepAlive: 30 * time.Second,
}

client := redis.NewClusterClient(&redis.ClusterOptions{
  Addrs:        []string{"addr"},
  MaxRedirects: 1,
  ReadOnly:     true,
  DialTimeout:  1 * time.Second,
  ReadTimeout:  1 * time.Second,
  WriteTimeout: 1 * time.Second,
  PoolSize:     30,
  Dialer: func(ctx context.Context, network string, addr string) (net.Conn, error) {
    return dialer.DialContext(ctx, network, addr)
  },
  IdleTimeout:        -1,
  IdleCheckFrequency: -1,
})
```

There are 30 Go application processes running with the above client, and cumulatively generating 80K GET/s and 15K SETNX/s. When Redis starts evicting data due to memory pressure, I see read timeouts and higher latency.

At 0:57 in the graph below is the switch to use this change
goredis_patch

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -301,9 +301,9 @@ func (p *ConnPool) popIdle() *Conn {
 		return nil
 	}
 
-	idx := len(p.idleConns) - 1
-	cn := p.idleConns[idx]
-	p.idleConns = p.idleConns[:idx]
+	// Take idle conn from front
+	cn := p.idleConns[0]
+	p.idleConns = p.idleConns[1:]
 	p.idleConnsLen--
 	p.checkMinIdleConns()
 	return cn


internal/pool/pool.go
```diff
@@ -301,9 +301,9 @@ func (p *ConnPool) popIdle() *Conn {
 		return nil
 	}
 
-	idx := len(p.idleConns) - 1
-	cn := p.idleConns[idx]
-	p.idleConns = p.idleConns[:idx]
+	// Take idle conn from front
+	cn := p.idleConns[0]
+	p.idleConns = p.idleConns[1:]
```
vmihailenco: Each time you create a new slice with first element unaccessible any more. This is a memory leak until Go allocates new underlying storage time to time. It should be `p.idleConns = append(p.idleConns[:0], p.idleConns[1:]...)` but it is more expensive...

Also with such usage pattern connections don't become stale any more.

vmihailenco:
Just to make sure - you use the same `redis.ClusterOptions` with your change? Specifically `IdleTimeout` and `IdleCheckFrequency` is `-1` in both deploys?

internal/pool/pool.go
kollektiv: I originally had `p.idleConns = append(p.idleConns[:0], p.idleConns[1:]...)` but changed it. If poolSize is 100 (as an example), then each popIdle will end up copying 100 pointers.

`p.idleConns = p.idleConns[1:]` will also reduce the capacity of the slice, and append will create a new array if the capacity is not sufficient. This is constant time in most cases. There shouldn't be a memory leak because the underlying array is eventually GCed when the array is doubled

kollektiv:
> Just to make sure - you use the same redis.ClusterOptions with your change? Specifically IdleTimeout and IdleCheckFrequency is -1 in both deploys?

Yes, both times use the same Idle settings

internal/pool/pool.go
kollektiv: > Also with such usage pattern connections don't become stale any more so ReapStaleConns is not useful any more.

Yeah cycling through the connections makes this not very useful, unless there is very low traffic to redis

kollektiv:
@vmihailenco What do you think about making this an option?

vmihailenco:
I need to reproduce your results first. Locally in benchmarks your change does not make any measurable difference. Are *you* able to reproduce this locally?

kollektiv:
I haven't been able to reproduce this locally. I can reproduce this in a AWS environment. Do you  have access to an AWS environment?

kollektiv:
@vmihailenco Bump on the above question

vmihailenco:
I just tried to run benchmarks ` go test -run=NONE -bench=BenchmarkRedis.` on AWS environment with and without your change and I can't spot any difference. That does not prove much since they only measure CPU time - not latency and each run gives slightly different results.

If you want to move this forward you have to provide instructions how to reproduce your results. E.g. some program that can be run and that demonstrates the difference.

vmihailenco:
Closing since there is no way to say whether this improves anything or not.

PR #1129, opened 2019-08-14
internal/proto: use strict ReadLine

vmihailenco:


internal/proto/reader.go
```diff
@@ -44,27 +44,40 @@ func (r *Reader) Buffered() int {
...
 
+// readLine that returns an error if:
+//   - there is a pending read error;
+//   - or line does not end with \r\n.
+func (r *Reader) readLine() ([]byte, error) {
+	b, err := r.rd.ReadSlice('\n')
+	if err != nil {
+		return nil, err
+	}
+	if len(b) <= 2 || b[len(b)-1] != '\n' || b[len(b)-2] != '\r' {
```
deff7: bytes.HasSuffix

PR #1150, opened 2019-09-03
Remove go.mod and go.sum

### LINKED ISSUE #1149: version 6.15.3+ is not usable with Go modules
I'm opening a new issue, because the problem is slightly different.

Before v6.15.2, the repository didn't use Go module, but it was usable in a project using Go modules.
Since v6.15.3, the repository contains a `go.mod` file that doesn't respect the Go module rules.
Since Go 1.13, the Go module rules are stricter.
You can't install anymore v6.15.3 and v6.15.4 with Go 1.13 (FYI it will be released very soon)
I've opened an issue about this https://github.com/golang/go/issues/34005 and they confirmed that the issue is on go-redis side.

I see a simple solution: remove entirely the Go module support from v6, and release a new tag v6.15.5.

It seems that v7 is using the Go module correctly, but it's still in beta.

WDYT ?

pierrre:
v6 branch is not compatible with Go module, and causes issue with Go 1.13.
The simplest solution is to remove the module support.

v7 will have correct module support.

Fixes #1149


pierrre:
I was waiting for the build :smile: 
https://travis-ci.org/go-redis/redis/builds/580261879

vmihailenco:
Thanks. Tagged v6.15.5

pierrre:
Thank you !

PR #1162, opened 2019-09-17
Fix "redis: discarding bad PubSub connection" stuck state

iproctor:
This addresses the issue where the client is stuck in a loop of "redis: discarding bad PubSub connection" while failing to receive subscription messages. 

The following sequence causes this stuck state:
healthy is initialized to true.
Timer event fires, a new Ping command is issued, and healthy = false.
Another timer event fires before the ping response, a new Ping command is issued but the connection is dropped by c.reconnect before a response can be received.

The client will remain in this reconnect loop since there is no way to set healthy = true. This PR sets healthy = true after the reconnect, which allows the next timer event's Ping response to arrive before another reconnect.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -467,8 +467,8 @@ func (c *PubSub) initPing() {
 						pingErr = errPingTimeout
 					}
 					c.mu.Lock()
-					healthy = true
 					c.reconnect(pingErr)
+					healthy = true
 					c.mu.Unlock()
 				}
 			case <-c.exit:
--- a/pubsub.go
+++ b/pubsub.go
@@ -467,6 +467,7 @@ func (c *PubSub) initPing() {
 						pingErr = errPingTimeout
 					}
 					c.mu.Lock()
+					healthy = true
 					c.reconnect(pingErr)
 					c.mu.Unlock()
 				}


vmihailenco:
Hi,

> loop since there is no way to set healthy = true

The idea is that reconnect should open a new connection and PubSub messages should start coming through. Any received message will send [a Go message](https://github.com/go-redis/redis/blob/master/pubsub.go#L504-L508) on `c.ping` channel which will set `healthy = true`.

But if there are no message at all then I guess you are right - connection will be dropped every 30 seconds until there is a new message.

Does that sound accurate with what you are observing?

vmihailenco:
>if there are no message at all

On the second thought this should not be possible since on reconnection we should resubscribe to the Redis channels and receive `Subscription` message which will set `healthy = true`.

Do I miss anything?

iproctor:
I think you're right. I was seeing these reconnects in my logs but it's because I have a state where I don't have any subscribed topics (my set of subscribed topics is dynamic and sometimes empty). Maybe this change would be desirable to avoid that error logging, though I'm not sure how many people are doing the same thing as me.

On the other hand, I was definitely getting into a state where messages weren't coming in and this seemed to fix it. I can't really reproduce that state though.

vmihailenco:
I think it is worth merging this - thanks!

PR #1166, opened 2019-09-24
add xinfo groups command

saitofun:


command.go
```diff
@@ -1285,6 +1285,96 @@ func (cmd *XPendingExtCmd) readReply(rd *proto.Reader) error {
 
 //------------------------------------------------------------------------------
 
+type XInfoGroupsCmd struct {
+	baseCmd
+	val []XGroup
+}
+
+type XGroup struct {
```
vmihailenco: And objections if I rename this to XInfoGroups?

Issue #1148, opened 2019-09-03
can't build the project


wrick198:
error information:
cluster.go:16:2: cannot find package "github.com/go-redis/redis/v7/internal" in any of:

In this project, there is no v7, why?

wrick198:
in cluster.go:
import (
	"context"
	"crypto/tls"
	"fmt"
	"math"
	"math/rand"
	"net"
	"runtime"
	"sort"
	"sync"
	"sync/atomic"
	"time"

	"github.com/go-redis/redis/v7/internal"
	"github.com/go-redis/redis/v7/internal/hashtag"
	"github.com/go-redis/redis/v7/internal/pool"
	"github.com/go-redis/redis/v7/internal/proto"
)
However, there is no project "github.com/go-redis/redis/v7/internal" in any of place.
why?

vmihailenco:
You need to use Go and activate Go module support - https://github.com/golang/go/wiki/Modules#how-to-install-and-activate-module-support

PR #1171, opened 2019-10-03
Dep compatibility

kollektiv:
Fixes: #1148, #1143

Currently if you try to use dep with go-redis v7, dep will place the files in `./vendor/github.com/go-redis/redis`, and then error because the imports, `github.com/go-redis/redis/v7`, are mismatched with file locations. Dep minimal module awareness issue: https://github.com/golang/dep/issues/1962

Go files have been moved to the top-level into the "v7" subdirectory. This makes it possible to use go-redis v7 with the dep dependency manager, and still compatible with go modules. dep places files on disk at "./github.com/go-redis/redis/v7".

The `forward.go` file is generated using goforward (https://go-review.googlesource.com/c/tools/+/137076/) to add aliases that point to github.com/go-redis/redis to github.com/go-redis/redis/v7. The goforward tool is mentioned in this section https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher


vmihailenco:
Sorry, but this is considerable amount of work and demand is not high. I've migrated all projects to `go mod` and don't regret it - in long term that is the best and simplest thing to do.

PR #1173, opened 2019-10-04
Sentinel goroutine leak fixed

sobolevsv:
SentinelClient is not closed due to race condition causing leak of 3 goroutines. Reproduced on high load ( thousands rps)


vmihailenco:
Thanks!

vmihailenco:
I've released v6.15.6 and ported this change to v7.

PR #1188, opened 2019-11-06
PubSub#Channel*(): allow to customize the timeout

Al2Klimov:


Al2Klimov:
@N-o-X What about this solution for our problem?

vmihailenco:
Any details why you need custom ping timeout?

Al2Klimov:
To set it as high as possible not to run into it at all not to have to allocate huge channels.

vmihailenco:
The API is already bloated and I don't believe many people need customizable ping timeout. I am fine with increasing ping timeout to 1 minute if that helps.

PR #1189, opened 2019-11-12
Set key to lowercase in cmdsInfoCache.Get()

WTIFS:
In function cmdInfo(name string), we use lower cmd name to find the cmdsInfo. 
```golang
//name is lower case
func (c *Ring) cmdInfo(name string) *CommandInfo {
	cmdsInfo, err := c.cmdsInfoCache.Get()
	if err != nil {
		return nil
	}
	info := cmdsInfo[name]
	if info == nil {
		internal.Logger.Printf("info for cmd=%s not found", name)
	}
	return info
}
```
But for extensions like bloom filter, the "command" returns names with uppercase. 

I think if we use lowercase name to find the cmdInfo, then we should turn cmd names to lower case when we fill the cmdsInfoCache.

### DIFF
--- a/command.go
+++ b/command.go
@@ -2006,6 +2006,12 @@ func (c *cmdsInfoCache) Get() (map[string]*CommandInfo, error) {
 		if err != nil {
 			return err
 		}
+		for key := range cmds {
+			lowerKey := internal.ToLower(key)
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks

PR #1199, opened 2019-11-26
FailoverClient with read-only support

nextsux:
Allow `FailoverClient` `ReadOnly` option. Client then connects to redis replicas (or master if no slave is available) for performing read-only queries. User is responsible for deciding which query can be run against it.

It aims to implement the same as #997 but without misusing ClusterClient which is not suitable for this.

### DIFF
--- a/options.go
+++ b/options.go
@@ -104,6 +104,9 @@ type Options struct {
 	// Enables read only queries on slave nodes.
 	readOnly bool
 
+	// Enables read only queries on redis replicas in sentinel mode
+	sentinelReadOnly bool
+
 	// TLS Config to use. When set TLS will be negotiated.
 	TLSConfig *tls.Config
 
--- a/sentinel.go
+++ b/sentinel.go
@@ -335,9 +335,9 @@ func (c *sentinelFailover) dial(ctx context.Context, network, _ string) (net.Con
 	var err error
 
 	if c.opt.sentinelReadOnly {
-		addr, err = c.RandomSlaveAddr()
+		addr, err = c.RandomSlaveAddr(ctx)
 	} else {
-		addr, err = c.MasterAddr()
+		addr, err = c.MasterAddr(ctx)
 	}
 
 	if err != nil {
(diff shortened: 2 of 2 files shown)


dstrop:
`sentinel slaves` returns all slaves sentinel knows about even disconected slaves (they will have flags s_down and disconnected)
this might cause dial to try connecting to dead slave and fail

nextsux:
Good point @dstrop I'll fix that

landrade:
Hi people, is there something that I can do to help you with it? I'm so interested in this feature.

takakawa:
I want this feature too! When will the commit be merged?

DrWrong:
should we listen the sentinel's `o_down` event to remove the redis conn in  Pool

fatpat:
is there any chances for this PR to come up in v8 somehow ?

dmitsh:
there is a conflicting file

dmitsh:
@vmihailenco , could you review this PR?

vmihailenco:
@nextsux sorry for ignoring this.

The code looks clean and self-contained so I don't mind merging it. But I *guess* it will not "properly" balance the load for 2 reasons:
- This PR really balances connections in the pool, not Redis commands.
- What is worse - go-redis does not evenly utilizes connections in the pool. It uses First In, First Out queue - some connections are used more often than others.

But that is all theory and in practise this PR may be good enough for some workloads. @nextsux do you have any experience using this PR under the load?

First In, First Out queue was picked so some connections are unused at all and we can close them after some time. I think it is not terribly useful and we can change pool to use other strategy if it helps with this PR.

---

But really using ClusterClient (PR #997) looks more appealing to me. It already has all the logic to properly balance the load - all we need is some flag / refactoring to disable cluster specific features...

nextsux:
no problem @vmihailenco.
You're right it balances connections, not commands which works pretty fine for me. I'm running earlier version of this patch (before rebase yesterday, https://github.com/nextsux/redis/tree/sentinel-read-only-v7) in production since I've created this PR - that is from 2020/11 and works fine:

image

red one is current master which is not receiving read traffic.

vmihailenco:
Good, let's hope it will work even better for others too. Meging.

PR #1205, opened 2019-12-05
properly point redisServerConf to redis.conf

jskelcy:
I am working on a mac and I was having a number of issues running the tests. One of them being that the default redis config seemed to be in the wrong place. I am not sure if I the config was supposed to be placed at the existing `redisServerConf` by some script, but I needed to make this change to have it work properly. 

If I am doing something wrong and I should not need to make this update let me know.


vmihailenco:
No sure what is happening, but it seems to be working either way so let's merge this. Thanks!

PR #1208, opened 2019-12-09
FIX: the BLPOP/BRPOP command may reach socket timeout when no message…

git-hulk:
While the Redis server uses cron to check where there are clients reach
timeout. The clientsCron makes some effort to process all the clients
every second and cannot be strictly guaranteed, So we simply add 1
second here.

see also Redis `clientsCron` function


vmihailenco:
This should be already handled in https://github.com/go-redis/redis/blob/master/redis.go#L303-L312 by adding `+ 10*time.Second`.

git-hulk:
@vmihailenco Sorry that I didn't check the detail, I would close this PR

Issue #1242, opened 2020-01-20
HSET does not support multiple field values 


Rajat-Goyal:
I think the the HSET should currently support multiple field-value pairs.
 
As I can see, since HMSet has been deprecated the library has started using HSet underneath HMSet. But, the HSet still does not support more than single field and value. 
We should also change/enhance the library's HSet to be able to do the same by making it possible to send multiple field value pairs. 

The library should reflect redis documentation for HSET. Or is there a reason that we have still kept it like this?

vmihailenco:
Did you see https://github.com/go-redis/redis/blob/master/commands.go#L986-L989 ? It shows how to use HMSet.

Rajat-Goyal:
Yes I saw this. What i am saying is that HMSET is deprecated and we do keep it to support folks who are still using HMSET which is a good thing. But library's HSET abstraction does not reflect the the spec from redis since we cannot pass multiple field value pairs as mentioned here https://redis.io/commands/hset
`As of Redis 4.0.0, HSET is variadic and allows for multiple field/value pairs.` 
I am just saying we should enhance the HSet abstraction to support this ^ 
HMSet abstraction can help me achieve it correct; but, I should be able to use HSET as per the specs.

vmihailenco:
Go method `HMSet` uses Redis `HSet` internally - so just continue to use `rdb.HMSet` - it uses correct Redis API.

Rajat-Goyal:
I don't think i am able to explain it correctly what my concern was 
1. HMSet there to support old codes that use HMSet [**Good**] 
2. HMSet internally uses HSet [**Good**]
3. HSet can only take a single field value [**Not Good**]

If the redis documentation says we can send multiple field value pairs for single hashKey using HSet I should be able to do that with this library's HSet also. Why should I use HMSet in a new codebase only because I want to send multiple field value pairs.

vmihailenco:
I decided to keep both `HSet` and `HMSet` Go API (not Redis API), because that preserves backwards compatibility *and* provides API that supports both single and multiple fields in sensible way. There is some confusion but no real harm in that.

If you have an idea how to make `HSet` support both single and multiple fields in a nice and usable way - please show it.

pavanyewale:
Hi guys, I have solve this issue and make a pull request please see it #1245

Rajat-Goyal:
[...]

PR #1246, opened 2020-01-23
modify HSET to support multiple fieldValue and return an integer reply

Rajat-Goyal:
- change interface to return IntCmd to support Integer reply 
- change signature to make HSet variadic to support multiple fieldValue pairs 
- Add tests for multiple fieldValue pairs in HSet 

ref: https://redis.io/commands/hset
Issue : #1242

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -131,7 +131,7 @@ type Cmdable interface {
 	HLen(key string) *IntCmd
 	HMGet(key string, fields ...string) *SliceCmd
 	HMSet(key string, values ...interface{}) *IntCmd
-	HSet(key, field string, fieldValues ...interface{}) *IntCmd
+	HSet(key string, fieldValues ...interface{}) *IntCmd
 	HSetNX(key, field string, value interface{}) *BoolCmd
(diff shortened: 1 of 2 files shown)


Rajat-Goyal:
@vmihailenco Could you please review this and let me know if something is missing. I can work on the same.

vmihailenco:
Now we have `HSet` and `HMSet` that are exactly the same. And the old `HSet` function has stricter API and does not allocate an `interface{}` - very subtle differences  that may be useful. I am not sure what to do.

Rajat-Goyal:
I understand what you mean. 
But in case I would require the previous HSET, shouldn’t I be using an older version of this library too with an older version of Redis. 

Now with using the latest Redis version and latest version for this library I am still writing a wrapper on my end to do this the right way.

Rajat-Goyal:
Redis 6.0 is already at beta and this spec changes are as per Redis 4; for which redis have already announced this year as the last year for official support. Redis themselves don’t support anything before v4

PR #1248, opened 2020-01-27
add support for context cancelation (#1196)

### LINKED ISSUE #1196: Context cancel not working on blocking operations
After eagerly testing out the newest v7 Beta, I stumbled across an odd issue, when using contexts in conjunction with blocking operations (such as BRPop).

Consider this simple example:

```go
client := redis.NewClient(&redis.Options{})

cancelCtx, cancelFn := context.WithCancel(context.Background())

wg := &sync.WaitGroup{}
wg.Add(1)

go func() {
	client.WithContext(cancelCtx).BRPop(0, "some-key")
	wg.Done()
}()

time.Sleep(2 * time.Second)

cancelFn()
wg.Wait()

// .... will never get here
```

So what happens is that the client establishes the connection, and waits indefinitely due to the `0` timeout (which is perfectly fine). However, this way it is impossible to cancel the connection, if it has been successfully established (hence, the exemplary `time.Sleep`).

Oddly tough, this might be an unexpected limitation of Go itself. To cite the documentation of the undelying [Dialer.DialContext](https://golang.org/pkg/net/#Dialer.DialContext) used, emphasis by me:

> The provided Context must be non-nil. If the context expires before the connection is complete, an error is returned. **Once successfully connected, any expiration of the context will not affect the connection.**

I have no idea how to even fix this, unfortunately. Probably some clever switch statement leveraging `ctx.Done()`, and a channel something something while waiting for the first bytes (Reader maybe?)

calebdoxsey:
Hi. These changes add support for context cancellation and should fix #1196 .

Cancellation of an underlying network connection is implemented by calling `SetDeadline` with `time.Now`.

### DIFF
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -58,35 +58,73 @@ func (cn *Conn) RemoteAddr() net.Addr {
 }
 
 func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error) error {
-	err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout))
-	if err != nil {
-		return err
-	}
-	return fn(cn.rd)
+	return cn.cancelWithSetDeadline(ctx, func() error {
+		err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout))
+		if err != nil {
+			return err
+		}
+		return fn(cn.rd)
+	})
 }
 
 func (cn *Conn) WithWriter(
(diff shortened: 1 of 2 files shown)


vmihailenco:
Looks like build fails because of the race in the code. Overall my first impression is that probably this should be done on a higher level somewhere in the `process` method. This way we only need 1 goroutine per command - your code uses 1 goroutine per read and write.

calebdoxsey:
Ok.

PR #1250, opened 2020-01-29
Make universal options functions public

furkansenharputlu:
**Why is it necessary?**

The built-in `NewUniversalClient` function logic is not compatible with our logic. Therefore, we wanted to implement our logic. The following is our code snippet:

```
if opts.MasterName != "" {
	client = redis.NewFailoverClient(opts.failover())
} else if cfg.EnableCluster {
	client = redis.NewClusterClient(opts.cluster())
} else {
	client = redis.NewClient(opts.simple())
}
```

To handle it, I copied private `failover`, `cluster` and `simple` functions to my repo redundantly. If they were public, it would be very useful. See: https://github.com/TykTechnologies/tyk/blob/c7634e7703796ea64c1cc21d663e45de3b0970cf/storage/redis_cluster.go#L174-L275

If you are ok with this change, I can add comments to those functions with your guidance.

### DIFF
--- a/universal.go
+++ b/universal.go
@@ -49,6 +49,7 @@ type UniversalOptions struct {
 	MasterName string
 }
 
+// Cluster returns cluster options created from the universal options.
 func (o *UniversalOptions) Cluster() *ClusterOptions {
 	if len(o.Addrs) == 0 {
 		o.Addrs = []string{"127.0.0.1:6379"}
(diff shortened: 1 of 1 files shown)


universal.go
```diff
@@ -49,7 +49,7 @@ type UniversalOptions struct {
 	MasterName string
 }
 
-func (o *UniversalOptions) cluster() *ClusterOptions {
+func (o *UniversalOptions) Cluster() *ClusterOptions {
```
vmihailenco: Please add something simple like `Cluster returns Redis Cluster options created from the universal options`. If you have better version - use it :)

PR #1270, opened 2020-03-06
Fix for retry logic. Do not retry command the same node

olbabushkin:


vmihailenco:
Thanks, but this is rather complicated and only checks that current and prev nodes do not match. Random is simpler and has guarantees to be random / unbiased.

PR #1273, opened 2020-03-13
Store time in RFC3339Nano to keep nanoseconds

theo-delaune-argus:
Hi,

Time package has nanosecond precision.
Actually go-redis removes this precision when time is stored or fetched with RFC3339.

```
t := time.Now() // time with nanoseconds precision
_ := redisClient.Set("my-time", t, 0).Err()

ft, _ := redisClient.Get("my-time").Time() 

// ft != t
```

To deal with these nanoseconds, I suggest RFC3339Nano.


vmihailenco:
Thanks!

PR #1281, opened 2020-03-25
Add client instantiation callback to allow more flexible configuration

dbussink:
There's currently no way to specify a TLS configuration for a Redis Ring setup. This adds configuration for this so it can be set up in a similar way as the password, with either a single one for all shards or a
configuration per shard.

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -2,7 +2,6 @@ package redis
 
 import (
 	"context"
-	"crypto/tls"
 	"errors"
 	"fmt"
 	"math/rand"
@@ -32,10 +31,6 @@ type RingOptions struct {
 	// different passwords. It will be ignored if the Password field is set.
 	Passwords map[string]string
 
-	// Map of name => tls configurations of ring shards, to allow shards to have
-	// different TLS configuration. Itwill be ignored if the TLSConfig field is set.
-	TLSConfigs map[string]*tls.Config
-
 	// Frequency of PING commands sent to check shards availability.
 	// Shard is considered down after 3 subsequent failed checks.
 	HeartbeatFrequency time.Duration
@@ -61,6 +56,9 @@ type RingOptions struct {
 	//  See https://arxiv.org/abs/1406.2294 for reference
 	HashReplicas int
 
+	// NewClient creates a shard client with provided name and options.
+	NewClient func(name string, opt *Options) *Client
+
 	// Optional hook that is called when a new shard is created.
 	OnNewShard func(*Client)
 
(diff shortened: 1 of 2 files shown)


ring.go
```diff
@@ -132,6 +139,13 @@ func (opt *RingOptions) getPassword(shard string) string {
 	return opt.Password
 }
 
+func (opt *RingOptions) getTLSConfig(shard string) *tls.Config {
+	if opt.TLSConfig != nil {
```
vmihailenco: TLSConfigs?

dbussink: @vmihailenco This is following the same kinda setup as the password one. If there's a single config at the toplevel, use that, otherwise check the shards config. There it's also called `opt.Password` vs. `opt.Passwords` for the shard list.

ring.go
```diff
@@ -31,6 +32,10 @@ type RingOptions struct {
 	// different passwords. It will be ignored if the Password field is set.
 	Passwords map[string]string
 
+	// Map of name => tls configurations of ring shards, to allow shards to have
+	// different TLS configuration. Itwill be ignored if the TLSConfig field is set.
+	TLSConfigs map[string]*tls.Config
```
vmihailenco: What about new option

```go
// NewClient creates a shard client with provided name and options.
NewClient func(name string, opt Options) *Client
```

which can replace `OnNewShard`, `passwords`, and `TLSConfigs`. What do you think?

ring.go
vmihailenco: Then it should be `opt.TLSConfig == nil`

ring.go
dbussink: @vmihailenco I was going for the smallest change possible here, but if the project consensus would be that this approach is better, this can be changed? 

Would that present a bigger breaking change though for any existing clients? Any new option should not break existing usage, but this seems like a bigger change that would have that impact?

ring.go
dbussink: @vmihailenco Of course, fixing that.

ring.go
vmihailenco: I suggest to
1. add `NewClient` option without adding `TLSConfigs`
2. deprecate `Passwords` and `OnNewShard` which are going to be removed in next major version

I can handle 2.). Does that sound good?

dbussink: @vmihailenco Yeah, I'll try to find some time for it then and make this change.

dbussink: @vmihailenco I've implemented your suggestion, what do you think?

vmihailenco:
Looks good - thanks!

PR #1298, opened 2020-04-16
add NewTimeCmdResult to allow unit testing with TimeCmd

ocallaco:
Needed this function for some unit tests


vmihailenco:
Thanks!

PR #1299, opened 2020-04-18
implement randezvous-hashing as an option for ring

rafaeleyng:
Hello and thanks for the awesome project 😄 

My team uses this lib in some very high-load services, specially the ring client. We've used rings sized as small as a couple nodes and as big as a couple of dozens of nodes, depending on traffic.

We've experienced some great variations in distributions of keys through the nodes. We've tweaked and tested changing the `replicas` param, tryng values such as 10, 100, 500, 1000, 2000 and others, but in the best scenario the node with most keys would still have around 20% more keys than the one with less keys (up to 40% in the worst scenarios).

We've decided to try the [randezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing) algorithm instead of the consistent hashing. Using this algorithm, the difference between the nodes with the least and the most keys dropped to around 2%.

To get a sense of the distribution, you could run this test (not included in the PR):

```go
func TestDistribution(t *testing.T) {
	hash := New(nil)
	hash.Add("1", "2", "3", "4", "5", "6")

	results := make(map[string]int, 10)

	for i := 0; i < 1000000; i++ {
		key := strconv.Itoa(i)

		site := hash.Get(key)
		if val, ok := results[site]; ok {
			results[site] = val + 1
		} else {
			results[site] = 1
		}
	}

	fmt.Println(results)
}
```

This yields the distribution (across 6 nodes) `map[1:167014 2:166473 3:165828 4:166104 5:167113 6:167468]`, which is very well balanced (less than 1% of difference between nodes)

This doesn't come for free, as the node resolution algorithm has complexity O(n), where n is the number of nodes. The benchmarks, though, showed that even with 512 nodes, **we are still talking about microseconds or even nanoseconds** in the difference. In a scenario like ours (couple of dozens of nodes), that difference is absolutly negligible (nanoseconds).

The usage of the randezvous-hashing as proposed in this PR is optional, the default continues to be the consistent hashing. The changes in the existing code are minimal to support both.

The only point I'm not totally satisfied with the implementation is:

```go
hasher := sha1.New()
hasher.Write([]byte(site + key))
siteWeight := m.hash(hasher.Sum(nil))
```

In my tests, that was to only way I could get a good distribution. I think that using a `sha1` there wouldn't be necessary, but I couldn't get good results without it. I've tried all the functions in the `hash` package (FNV, CRC etc), but no cigar. Actually, that was a point of confusion, the `m.hash` seems to me to be a checksum, not a hash (I'm no crypto expert). Maybe the bad distribution without the `sha1` hash has something to do with the nature of the keys I was using in the tests (sequential numbers converted to string, with or without a 0's padding).

Anyway, I would gladly improve that part if you have any suggestion.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/go-redis/redis/v7
 
 require (
+	github.com/cespare/xxhash v1.1.0
 	github.com/golang/protobuf v1.3.2 // indirect
 	github.com/kr/pretty v0.1.0 // indirect
 	github.com/onsi/ginkgo v1.10.1
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,6 @@
+github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAElWljhcU=
+github.com/cespare/xxhash v1.1.0 h1:a6HrQnmkObjyL+Gs60czilIUGqrzKutQD6XZog3p+ko=
+github.com/cespare/xxhash v1.1.0/go.mod h1:XrSqR1VqqWfGrhpAt58auRo0WTKS1nRRg3ghfAqPWnc=
 github.com/fsnotify/fsnotify v1.4.7 h1:IXs+QLmnXW2CcXuY+8Mzv/fWEsPGWxqefPtCP5CnV9I=
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
 github.com/golang/protobuf v1.2.0 h1:P3YflyNX/ehuJFLhxviNdFxQPkGK5cDcApsge1SqnvM=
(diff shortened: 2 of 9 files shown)


rafaeleyng:
Not sure why the checks failed, the logs on Travis won't show me anything

vmihailenco:
Hello!

The failure on Travis looks like this

```
golangci-lint run
internal/rendezvoushash/rendezvoushash.go:48:15: Error return value of `hasher.Write` is not checked (errcheck)
		hasher.Write([]byte(site + key))
		            ^
internal/consistenthash/consistenthash.go:28: File is not `goimports`-ed (goimports)
internal/rendezvoushash/rendezvoushash.go:11: File is not `goimports`-ed (goimports)
	hash     internal.Hash
	sites    []string
internal/rendezvoushash/rendezvoushash.go:32:2: S1011: should replace loop with `m.sites = append(m.sites, sites...)` (gosimple)
	for _, site := range sites {
	^
```

You should be able to reproduce it locally by running `golangci-lint run`.

>I've tried all the functions in the hash package (FNV, CRC etc), but no cigar

Those are not the best hash function and we should start using 64-bit hash for better results. Using xxhash64 rendezvous provides following distribution:

```
map[1:166483 2:167680 3:166116 4:166852 5:166650 6:166219]
```

While consistenthash with 1000 replicas and xxhash64:

```
map[1:169846 2:165154 3:161643 4:173815 5:164098 6:165444]
```

Benchmark now looks like this. rendezvous:

```
BenchmarkGet8-12      	 5165445	       225 ns/op	      16 B/op	       1 allocs/op
BenchmarkGet32-12     	 1892876	       640 ns/op	      26 B/op	       1 allocs/op
BenchmarkGet128-12    	  528529	      2251 ns/op	      44 B/op	       1 allocs/op
BenchmarkGet512-12    	  102086	     11106 ns/op	      35 B/op	       1 allocs/op
```

consistenthash:

```
BenchmarkGet8-12      	10102969	       123 ns/op	       8 B/op	       1 allocs/op
BenchmarkGet32-12     	 8701545	       135 ns/op	       8 B/op	       1 allocs/op
BenchmarkGet128-12    	 6834182	       180 ns/op	       9 B/op	       1 allocs/op
BenchmarkGet512-12    	 4964846	       240 ns/op	      14 B/op	       1 allocs/op
```

I've pushed my changes to https://github.com/go-redis/redis/compare/rendezvous-hashing-ring?expand=1 - let me know if that works for you. If all is good we should be able to get that in v8 branch.

vmihailenco:
There is also https://medium.com/@dgryski/consistent-hashing-algorithmic-tradeoffs-ef6b8e2fcae8 and his implementation at https://github.com/dgryski/go-rendezvous/blob/master/rdv.go which uses custom hash that should be faster.

rafaeleyng:
@vmihailenco Your changes look nice. I've fixed the tests, everything looks fine for me.

About the links you've sent about the custom hash, I wouldn't bother with that right now. The difference is still in nanoseconds for hundreds of nodes, seems overkill to try to optimize that (at least for most use cases).

rafaeleyng:
@vmihailenco Hello there, anything else to do in this MR or looks good for you?

rafaeleyng:
@vmihailenco Hello 😄 , just pinging to re-state that I'm all up for making any needed changes, or if no change is needed, can I expect this PR to be merged anytime soon?

vmihailenco:
Sorry for delay. I've something similar yet different in https://github.com/go-redis/redis/pull/1361. The main difference is that it provides more flexible interface that allows switching consistent hash algorithms and also uses faster rendezvous implementation by dgryski.

Sorry for "discarding" your work - I should have handled this better :/

rafaeleyng:
@vmihailenco no problem! I'm happy the idea got in, anyways

PR #1312, opened 2020-04-29
add a circuit breaker inside each node

YiweiZhuang:
we want to have circuit breaker for each node base client on go-redis to rate limit some connections.
Copy paste code from api_common/circuit/circuitBreaker.go


YiweiZhuang:
actually only want to apply these changes on my project not merge to go-redis

PR #1315, opened 2020-04-30
fix NewConnPool checkMinIdleConns data race

lwightmoon:


vmihailenco:
`NewConnPool` is by definition race free - only one goroutine has access to the pool while it is created.

lwightmoon:
checkMinIdleConns is not safe ,may add lock when called in NewConnPool

lwightmoon:
> `NewConnPool` is by definition race free - only one goroutine has access to the pool while it is created.

checkMinIdleConns is not safe ,may add lock when called in NewConnPool
image

theoptz:
+1

vmihailenco:
See https://github.com/go-redis/redis/pull/1315#issuecomment-621644965 - I think there can not be a race here.

PR #1316, opened 2020-05-01
allow cluster client to use limiter

YiweiZhuang:
code users can pass a limiter struct into cluster client then allow each node client in the cluster to use that limiter

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -76,6 +76,7 @@ type ClusterOptions struct {
 	IdleCheckFrequency time.Duration
 
 	TLSConfig *tls.Config
+	Limiter   Limiter
 }
 
 func (opt *ClusterOptions) init() {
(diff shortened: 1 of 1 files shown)


cluster.go
```diff
@@ -145,6 +146,7 @@ func (opt *ClusterOptions) clientOptions() *Options {
 		IdleCheckFrequency: disableIdleCheck,
 
 		TLSConfig: opt.TLSConfig,
+		Limiter:   opt.Limiter,
```
ThomasJannaud, line 149: I think we would want a limiter per node

walktall:
@vmihailenco Hi, any idea on this PR? We have the same issue that may get help from this. Thanks.

vmihailenco:
Looks good - thanks!

walktall:
@vmihailenco I believe we should pass a Limiter constructor to the option instead of a Limiter, as mentioned by @ThomasJannaud

walktall:
@vmihailenco will try make a PR for this.

vmihailenco:
@walktall indeed. Please hold on - I think I have a solution.

walktall:
> @walktall indeed. Please hold on - I think I have a solution.

Sure, thanks!

vmihailenco:
@walktall please check https://github.com/go-redis/redis/pull/1332

PR #1318, opened 2020-05-07
cmdsInfo to actually use and return the first error

manavo:
`cmdsInfo` at the moment sets the `firstErr` to the fallback option, which means line 518 would never execute, and the actual error would be masked.

(For example, I just had to track this down because I was trying to use Redis 6, and it isn't compatible)


vmihailenco:
Thanks!

PR #1320, opened 2020-05-11
Prevent panic during nil to string type assertion

### LINKED ISSUE #1319: Calling Unsubcribe twice on PubSub causes interface conversion panic
Trying to Unsubscribe from all topics, while the client isn't subscribed to any topic causes func newMessage to panic:

```
panic: interface conversion: interface {} is nil, not string

goroutine 19 [running]:
github.com/go-redis/redis/v7.(*PubSub).newMessage(0xc000120000, 0x66dd00, 0xc00000e260, 0x0, 0xc00004ce00, 0x0, 0x0)
        /home/hjr265/Projects/go-redis-bug/vendor/github.com/go-redis/redis/v7/pubsub.go:314 +0x94a
github.com/go-redis/redis/v7.(*PubSub).ReceiveTimeout(0xc000120000, 0x0, 0x67f201, 0xc00007c720, 0x0, 0x0)
        /home/hjr265/Projects/go-redis-bug/vendor/github.com/go-redis/redis/v7/pubsub.go:362 +0x145
github.com/go-redis/redis/v7.(*PubSub).Receive(...)
        /home/hjr265/Projects/go-redis-bug/vendor/github.com/go-redis/redis/v7/pubsub.go:369
github.com/go-redis/redis/v7.(*PubSub).initMsgChan.func1(0xc000120000)
        /home/hjr265/Projects/go-redis-bug/vendor/github.com/go-redis/redis/v7/pubsub.go:490 +0x8b
created by github.com/go-redis/redis/v7.(*PubSub).initMsgChan
        /home/hjr265/Projects/go-redis-bug/vendor/github.com/go-redis/redis/v7/pubsub.go:484 +0x6d
exit status 2
``` 

I could use this snippet of code to reliably reproduce this issue on my end:

``` go
cl := redis.NewClient(&redis.Options{
	Addr: ":32768",
})
ps := cl.Subscribe("x")
ps.Unsubscribe()
ps.Unsubscribe()
<-ps.Channel()
```

The issue affects both v6.15.7 and v7.2.0.

It stems from the fact that for an unsubscribe message, Redis may push a payload like:

``` go
["unsubscribe", nil, 0]
```

The same behavior can be seen with the UNSUBSCRIBE command in the CLI:

```
127.0.0.1:6379> UNSUBSCRIBE
1) "unsubscribe"
2) (nil)
3) (integer) 0
```

And, this snippet from pubsub.go:310 tries to assert the type of the second element to string:

``` go
		switch kind := reply[0].(string); kind {
		case "subscribe", "unsubscribe", "psubscribe", "punsubscribe":
			return &Subscription{
				Kind:    kind,
				Channel: reply[1].(string),
				Count:   int(reply[2].(int64)),
			}, nil
```

I discovered this issue when [my project's](https://toph.co) web server would crash periodically with a similar stack trace as shared above. I was able to workaround the issue by adding a check before the assertion to ensure that `reply[1]` isn't nil.

hjr265:
This change causes func `newMessage` to assert `reply[1].(string)` safely and prevents the function from panicking. This one closes #1319.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -309,9 +309,10 @@ func (c *PubSub) newMessage(reply interface{}) (interface{}, error) {
 	case []interface{}:
 		switch kind := reply[0].(string); kind {
 		case "subscribe", "unsubscribe", "psubscribe", "punsubscribe":
+			channel, _ := reply[1].(string)
 			return &Subscription{
 				Kind:    kind,
-				Channel: reply[1].(string),
+				Channel: channel,
 				Count:   int(reply[2].(int64)),
 			}, nil
 		case "message":


vmihailenco:
Thanks!

PR #1332, opened 2020-05-21
Add ClusterOptions.NewClient hook

vmihailenco:
Replaces https://github.com/go-redis/redis/pull/1316

Can be used like this

```
rdb := redis.NewClusterClient(&redis.ClusterOptions{
	Addrs: []string{":7000", ":7001", ":7002", ":7003", ":7004", ":7005"},

	NewClient: func(opt *redis.Options) *redis.Client {
		opt.Limiter = getLimiter(opt)
		return redis.NewClient(opt)
	},
})
```

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -68,6 +68,9 @@ type ClusterOptions struct {
 	ReadTimeout  time.Duration
 	WriteTimeout time.Duration
 
+	// NewClient creates a cluster node client with provided name and options.
+	NewClient func(opt *Options) *Client
+
 	// PoolSize applies per cluster node and not for the whole cluster.
 	PoolSize           int
 	MinIdleConns       int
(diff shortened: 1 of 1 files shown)


walktall:
@vmihailenco Thanks! LGTM!

cluster.go
```diff
@@ -166,7 +171,7 @@ func newClusterNode(clOpt *ClusterOptions, addr string) *clusterNode {
 	opt := clOpt.clientOptions()
 	opt.Addr = addr
 	node := clusterNode{
-		Client: NewClient(opt),
+		Client: clOpt.NewClient(opt),
```
ThomasJannaud, line 174: what would be nice would be to have the limiter take the address as a parameter. If one shard is down then we can have a circuit breaker that disables that shard only

ThomasJannaud:
Thanks!

for anyone needed a limiter per options:

```
opts.NewClient = func(opt *redis.Options) *redis.Client {
  opt.Limiter = _myCircuitBreaker[addr]  // you need to implement your own circuit breaker, one per address
  return redis.NewClient(opt)
}
client := redis.NewClusterClient(opts)
```

PR #1337, opened 2020-05-22
Add required context parameter in example client

davidrthorn:
Client.Ping() requires a context, so the example as shown does not work.


vmihailenco:
Thanks for the PR, but I believe this is already fixed. Closing.

PR #1341, opened 2020-05-24
use linked list to store conn,impove the performance when put the conn to pool

lwightmoon:


lwightmoon:
if the put operation get the lock spend much time, the idlepool will be empty,then great number of connection will be created

vmihailenco:
Do you have any benchmarks to share? There are already some in internal/pool folder, but you may need to add something new to demonstrate the improvement.

lwightmoon:
only the size is big like 100000 the imporement is obviously, so just close the pr

Issue #1329, opened 2020-05-14
NewClusterClient().Command()  return error:   redis: got 7 elements in COMMAND reply, wanted 6


dodo-luck:
/**********************env**********************/
go V1.14.2
redis V6.0.1
go-redis V7

/**********************code**********************/
package main

import (
	"fmt"
	"github.com/go-redis/redis/v7"
)

func main() {
	rdb := redis.NewClusterClient(&redis.ClusterOptions{
		Addrs: []string{":7000", ":7001", ":7002", ":7003", ":7004", ":7005"},
	})
	
	fmt.Println(rdb.Command())
}

/**********************************************/
 redis: got 7 elements in COMMAND reply, wanted 6

 why ?

dodo-luck:
func call:    _process()  --> cmdInfo()  cmdSlot()  -->  cmdsInfo()  -->  Command()

func (c *ClusterClient) _process(ctx context.Context, cmd Cmder) error                                                 
func (c *ClusterClient) cmdInfo(name string) *CommandInfo          
func (c *ClusterClient) cmdsInfo() (map[string]*CommandInfo, error)
func (c cmdable) Command() *CommandsInfoCmd
func cmdSlot(cmd Cmder, pos int) int {
	if pos == 0 {
		return hashtag.RandomSlot()
	}
	firstKey := cmd.stringArg(pos)
	return hashtag.Slot(firstKey)
} 

This error causes the function(_process) to be slow . because  the return value of the function(cmdInfo) is nil ,  in that case, the function(func cmdSlot(cmd Cmder, pos int) int) arg (pos) will be 0.
不晓得咋回事0.0

vmihailenco:
Should be fixed by https://github.com/go-redis/redis/pull/1357.

PR #1355, opened 2020-06-05
Fix performance degradation due to non-caching of command info in Redis 6 Cluster 

yuuki:
## What is the problem?

I observed performance degradation when using redis.ClusterClient in Redis 6.
pprof reported the following bottleneck caused by mutex lock (https://github.com/go-redis/redis/blob/a999d1ecd81303df8128547746c62358f78048f5/command.go#L2153-L2154).

ScreenShot 2020-06-04 at 9 01 38 PM

This mutex lock is for the client to cache COMMAND reply just once (https://github.com/go-redis/redis/blob/a999d1ecd81303df8128547746c62358f78048f5/cluster.go#L688). 
However, I noticed that during the debugging of this performance degradation problem, for some reason the client executed COMMAND command every time.

## What caused the problem?

I got the following error that has already reported in https://github.com/go-redis/redis/issues/1329 in Redis 6. 

> redis: got 7 elements in COMMAND reply, wanted 6

This error prevented the internal.once from setting the done flag
(https://github.com/go-redis/redis/blob/a999d1ecd81303df8128547746c62358f78048f5/internal/once.go#L54-L57). As a result, the client did not cache COMMAND reply.

This error is caused by [the change of additional command categories](https://github.com/antirez/redis/commit/c5e717c637cbb1c80e1259560ebf995fb7920628) in Redis 6. 
I show the example of the output of COMMAND and COMMAND INFO reply. 

- Redis 5.0.8 (6 elements)

```
127.0.0.1:6379> COMMAND INFO xadd
1) 1) "xadd"
   2) (integer) -5
   3) 1) write
      2) denyoom
      3) random
      4) fast
   4) (integer) 1
   5) (integer) 1
   6) (integer) 1
```

- Redis 6.0.1 (7 elements)

```
127.0.0.1:6379> COMMAND INFO xadd
1) 1) "xadd"
   2) (integer) -5
   3) 1) write
      2) denyoom
      3) random
      4) fast
   4) (integer) 1
   5) (integer) 1
   6) (integer) 1
   7) 1) @write
      2) @stream
      3) @fast
```

Thus, I have fixed go-redis to handle this 7th elements for additional ACL categories. Finally, I solved my problem of performance degradation.

### DIFF
--- a/command.go
+++ b/command.go
@@ -2072,8 +2072,8 @@ func (cmd *CommandsInfoCmd) readReply(rd *proto.Reader) error {
 }
 
 func commandInfoParser(rd *proto.Reader, n int64) (interface{}, error) {
-	if n != 7 {
-		return nil, fmt.Errorf("redis: got %d elements in COMMAND reply, wanted 7", n)
+	if n != 6 && n != 7 {
+		return nil, fmt.Errorf("redis: got %d elements in COMMAND reply, wanted 6 or 7(Redis 6 or later)", n)
 	}
 
 	var cmd CommandInfo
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks for the PR - I've merged this a part of https://github.com/go-redis/redis/pull/1357

yuuki:
Thanks for your quick response!

PR #1361, opened 2020-06-08
Use Rendezvous in Ring. Thanks @rafaeleyng for initial idea and implementation

vmihailenco:
Closes https://github.com/go-redis/redis/pull/1299


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/go-redis/redis/v8
 
 require (
+	github.com/cespare/xxhash v1.1.0
+	github.com/dgryski/go-rendezvous v0.0.0-20180401054734-3692eb46c031
```
dbussink, line 5: It looks like this package has no license defined at https://github.com/dgryski/go-rendezvous?

dgryski: Just pushed an MIT license: https://github.com/dgryski/go-rendezvous/commit/8d2f3be8b2d9f5841b0858fc9030d1e6749a5d49

dbussink: @dgryski Thanks!

PR #1368, opened 2020-06-11
add support for slowlog command

wziww:


command.go
```diff
@@ -2161,3 +2161,116 @@ func (c *cmdsInfoCache) Get() (map[string]*CommandInfo, error) {
...
+			createdAt, err := rd.ReadIntReply()
+			if err != nil {
+				return nil, err
+			}
+			createdAtTime := time.Unix(createdAt, 0)
+			costs, err := rd.ReadIntReply()
+			if err != nil {
+				return nil, err
+			}
+			costsDuration := time.Duration(costs)
```
vmihailenco, line 2236: https://redis.io/commands/slowlog says

>The amount of time needed for its execution, in microseconds

Should it be `costs * time.Microsecond`?

commands_test.go
```diff
@@ -4029,6 +4029,28 @@ var _ = Describe("Commands", func() {
...
+
+		It("returns slow query result", func() {
+
+			var (
+				CONFIGSLOWKEY string = "slowlog-log-slower-than"
+			)
+			old := client.ConfigGet(ctx, CONFIGSLOWKEY).Val()
+			client.ConfigSet(ctx, CONFIGSLOWKEY, "0")
+			defer client.ConfigSet(ctx, CONFIGSLOWKEY, old[1].(string))
+			_, e := client.Eval(ctx, "redis.call('slowlog','reset')", []string{}).Result()
```
vmihailenco, line 4042: I wonder if you use `Eval` to execute custom command? go-redis supports `client.Do(ctx, "slowlog", "reset")` to achieve the same.

vmihailenco:
@wziww this looks pretty good - sorry for the delay. I have 2 small questions if you are still around. Anyway I will merge this tomorrow.

commands_test.go
wziww: I want to keep the same code-style, which all test case in this file  use `client.Eval` to execute the command :)

command.go
wziww: Yes, fix in https://github.com/go-redis/redis/pull/1368/commits/2d31f687b3738412951a1aa6958bd937c2365ff7#diff-9c42b524cb14ca001c61267826cbefb1R2236

vmihailenco:
@wziww thanks, merged this in https://github.com/go-redis/redis/pull/1472

PR #1372, opened 2020-06-12
limit the speed to create conn

lwightmoon:
can create connection with only one goroutine?in our case,whe have dozens of service nodes with go-redis that connect the same redis cluster server ,some time the number of requests to access redis-server fast growth and then decline,this caused the redis-server listen queue of a socket overflowed and the go-redis io/timeout .although the limiter interface may relief the problem，but we don't want to limit the speed to get conn from pool,cant just limit the speed to create conn ?


vmihailenco:
Sounds like you want to reduce PoolSize - that effectively limit number of connections to Redis Server.

lwightmoon:
> Sounds like you want to reduce PoolSize - that effectively limits number of connections to Redis Server.

I want to reduce the speed to establish the connection not to reduce the PoolSize.go-redis establish conn concurrently. I want to use only one goroutine to establish conn. The problem is losts of connections are established at a short time causes the redis sever listen queue overflower.

vmihailenco:
This is first time I encounter this problem. And it looks strange that your server supports N connections but does not allow to open simultaneously those N connections. Why not to bump that limit? or lower number of connections?

8k connections to a single server is way too much. And spike from 8k to 12k is not that big. It is not clear what is the problem here.

lwightmoon:
> This is first time I encounter this problem. And it looks strange that your server supports N connections but does not allow to open simultaneously those N connections. Why not to bump that limit? or lower number of connections?
> 
> 8k connections to a single server is way too much. And spike from 8k to 12k is not that big. It is not clear what is the problem here.

The problem is the server can not establish too many connections at a short time,if the client request to establish slowly the server can establish all conns;although connections spiking from 8k to 12k is not big,they were established at a very short time this causes the redis server listen queue overflow.if the poolsize is n and the client num is m;there will be up to n*m request to establish connections simultaneously;if just use one goroutine to create conn ,there will be up to n request to estbablish connections simultaneously;

vmihailenco:
I suggest to do something simpler like:

```go
gate := syncutil.NewGate(1)

rdb := redis.NewClient(&redis.Options{
		gate.Start()
		defer gate.Done()

		netDialer := &net.Dialer{
			Timeout:   opt.DialTimeout,
			KeepAlive: 5 * time.Minute,
		}
		return netDialer.DialContext(ctx, network, addr)
    },
})
```

lwightmoon:
This looks like good

vmihailenco:
Closing since there is a reasonable alternative in https://github.com/go-redis/redis/pull/1372#issuecomment-645242567

PR #1377, opened 2020-06-15
Make cluster client get updated nodes' addrs

davidfu13:
During loading state, if all the nodes can't be connected, it's
possible that all the nodes' IP addresses have changed at the same
time. Clear the clusterAddrs so that the cluster client has chance
to get updated nodes' IP addresses by using the initial setting of
cluster addresses.


vmihailenco:
Thanks - I will update the code so clusterAddrs is properly updated.

Issue #1388, opened 2020-06-26
XAddArgs is a map and does not preserve fields order


adambabik:
From the docs about [XADD](https://redis.io/commands/xadd):

> An entry is composed of a set of field-value pairs, it is basically a small dictionary. The field-value pairs are stored in the same order they are given by the user, and commands to read the stream such as XRANGE or XREAD are guaranteed to return the fields and values exactly in the same order they were added by XADD.

This means that I can rely on the order of field-value pairs in an entry added to a stream.

However, [`Client.XAdd`](https://pkg.go.dev/github.com/go-redis/redis/v8?tab=doc#Client.XAdd) passes key-value pairs using `map` which does not preserve the order of fields. For example, this code:

```go
rdb.XAdd(context.Background(), &redis.XAddArgs{
	Stream: "my-stream",
	Values: map[string]interface{}{
		"field1": "value1",
		"field2": "value2",
		"field3": "value3",
	},
})
```

may produce arbitrary order of the key-value pairs in the stream entry.

My proposal is to replace `map` with `[]interface{}{"key1", "value1", ...}`, however, this is not backward compatible. 

Alternatively, it can work similarly to [`Client.HSet`](https://pkg.go.dev/github.com/go-redis/redis/v8?tab=doc#Client.HSet) so ofer all three ways to specify the pairs:
```
- HSet("myhash", "key1", "value1", "key2", "value2")
- HSet("myhash", []string{"key1", "value1", "key2", "value2"})
- HSet("myhash", map[string]interface{}{"key1": "value1", "key2": "value2"})
```

vmihailenco:
>Alternatively, it can work similarly to Client.HSet so ofer all three ways to specify the pairs:

This seems to be preferable - can you send a PR?

adambabik:
@vmihailenco yes, I will submit it in 1-2 days.

xpofei:
redis-py seems to have the same problem

PR #1389, opened 2020-06-28
Add support for slices in XAddArgs.Values

adambabik:
This PR solves the problem described in https://github.com/go-redis/redis/issues/1388.

Previously, `XAddArgs.Values` was of type `map[string]interface{}`. Now, it is `interface{}`. This change is backward compatible.

This change uses `appendArgs` utility function in `XAdd`. It was extended to support `[]interface{}` for the first item in the `src` slice. This might be a bit controversial as `appendArgs` is not used directly like that anywhere else. If so, I will rewrite this fragment not to use `appendArgs`.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -43,6 +43,9 @@ func appendArgs(dst, src []interface{}) []interface{} {
 				dst = append(dst, s)
 			}
 			return dst
+		case []interface{}:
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks!

Issue #1392, opened 2020-06-29
Adding metrics instrumentation for OpenTelemetry


TigerHe7:
@reggiemcdonald and I were looking into adding a Redis integration for OpenTelemetry, but noticed support for tracing has already been added to this repo.

We're wondering if we can pursue **adding metrics instrumentation for OpenTelemetry** to this repository?

Just wanted to request permission (perhaps from @vmihailenco) before getting started!

vmihailenco:
@TigerHe7 that sounds good - please go ahead! If you have a list of metrics you would like to add - please share the list. Also it would be easier if your first PR contains single metric so we have a chance to discuss high level details before adding too many metrics.

I am not overly familiar with metrics myself but these links should be helpful
- https://github.com/open-telemetry/oteps/pull/108/files
- https://github.com/open-telemetry/opentelemetry-specification/pull/657/files

reggiemcdonald:
@vmihailenco I noticed that the otel hook will only record spans if a parent exists in the context. Is this the intended behaviour?
Code for reference:
https://github.com/go-redis/redis/blob/dc52593c8c63bc2a05796ec44efd317fd3e14b64/redisext/otel.go#L18-L21

vmihailenco:
@reggiemcdonald yes, that is intended. The assumption is that there should be a parent span when OpenTelemetry is used. Usually it is a span for HTTP handler / queue handler. Does it cause any problems?

reggiemcdonald:
Great to know, thanks! 

Would you mind if I updated the docs to make this more clear for future users?

TigerHe7:
@reggiemcdonald and I thinking of adding these additional metrics to this Redis client, based on the previous OpenCencus integration: https://github.com/opencensus-integrations/redis/pull/1/files#diff-cbb25aea043ef5888a4f0283ee00649aR20-R33

We currently already track the number of write commands and new connections created. 

Do @vmihailenco or @MrAlias have any thoughts, or does this seem reasonable?

PR #1398, opened 2020-07-07
Introduce metrics capture with OpenTelemetry

TigerHe7:
Summary: **Introduces the ability to capture metrics with the OpenTelemetry library**
Addresses https://github.com/go-redis/redis/issues/1392

### Changes
* Added `Instruments` object for recording metrics
* Added example client for testing and demo purposes
* Records the number of new connections created, incremented in the `dialConn` function in `pool.go`
* Records the number of write operations performed, incremented in the `_process` function in `redis.go`

### Testing with the example client
Validated that
* trivial cases with metrics exported to CLI
* no change in behavior when metrics are *not* exported
* the number of connections recorded reflects additional connections created by performing Redis operations in goroutines and is limited by the `PoolSize` option
* the number write operations is incremented whenever `.Set`, `.Get`, and `.Del` are called.

### Note
* Failing CI tests match those failing on the latest version of master


internal/instruments.go
```diff
@@ -0,0 +1,44 @@
...
+	// Count of new connections
+	NewConnectionsCount metric.Int64Counter
+}
+
+var Instruments = initInstruments()
+
+func initInstruments() *openTelemetryInstrumentation {
+	meter := global.Meter("github.com/go-redis/redis")
+
+	writeCount, err := meter.NewInt64Counter("redis.num_writes",
```
MrAlias, line 22: I would leave out the prefix `num`. It is clear by the instrument type this is counting something and a metric name of `writes` will describe what that is.

```suggestion
	writeCount, err := meter.NewInt64Counter("redis.writes",
```

internal/instruments.go
```diff
@@ -0,0 +1,44 @@
...
+		metric.WithDescription("the number of writes initiated"),
+		metric.WithUnit(unit.Bytes),
+	)
+	if err != nil {
+		// TODO: handle errors
+		log.Printf("failed to create instrument WriteCount")
+	}
+
+	newConnectionsCount, err := meter.NewInt64Counter("redis.num_new_conn",
+		metric.WithDescription("the number of new connections created"),
```
MrAlias, line 32: Similarly here, I would drop the prefix `num` and possibly `new`.

```suggestion
	newConnectionsCount, err := meter.NewInt64Counter("redis.connections",
		metric.WithDescription("the number of connections created"),
```

internal/instruments.go
```diff
@@ -0,0 +1,44 @@
...
+	meter := global.Meter("github.com/go-redis/redis")
+
+	writeCount, err := meter.NewInt64Counter("redis.num_writes",
+		metric.WithDescription("the number of writes initiated"),
+		metric.WithUnit(unit.Bytes),
+	)
+	if err != nil {
+		// TODO: handle errors
+		log.Printf("failed to create instrument WriteCount")
+	}
```
MrAlias, line 29: Probably a better option to use the project logger here: https://github.com/go-redis/redis/blob/master/internal/log.go#L12

E.g.

```suggestion
	if err != nil {
		// TODO: handle errors
		Logger.Printf("failed to create instrument WriteCount")
	}
```

internal/instruments.go
```diff
@@ -0,0 +1,44 @@
...
+
+	newConnectionsCount, err := meter.NewInt64Counter("redis.num_new_conn",
+		metric.WithDescription("the number of new connections created"),
+	)
+	if err != nil {
+		// TODO: handle errors
+		log.Printf("failed to create instrument NumNewConnections")
```
MrAlias, line 36: similar here.

internal/pool/pool.go
```diff
@@ -179,6 +179,7 @@ func (p *ConnPool) dialConn(ctx context.Context, pooled bool) (*Conn, error) {
 		return nil, err
 	}
 
+	internal.Instruments.NewConnectionsCount.Add(ctx, 1)
```
MrAlias, line 182: Did we want to include any labels with this?

redis.go
```diff
@@ -328,6 +328,7 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder) error {
 			retryTimeout := true
 			err := c.withConn(ctx, func(ctx context.Context, cn *pool.Conn) error {
 				err := cn.WithWriter(ctx, c.opt.WriteTimeout, func(wr *proto.Writer) error {
+					internal.Instruments.WriteCount.Add(ctx, 1)
```
MrAlias, line 331: Similar here WRT labels.

internal/instruments.go
```diff
@@ -0,0 +1,42 @@
...
+var Instruments = initInstruments()
+
+func initInstruments() *openTelemetryInstrumentation {
+	meter := global.Meter("github.com/go-redis/redis")
+
+	writeCount, err := meter.NewInt64Counter("redis.writes",
+		metric.WithDescription("the number of writes initiated"),
+	)
+	if err != nil {
+		// TODO: handle errors
```
reggiemcdonald, line 24:
```suggestion
		writeCount, _ = metric.NoopProvider{}.Meter("no-op").NewInt64Counter(
		        "github.com/go-redis/redis",
		)
```

I wanted to ask - I was thinking of doing this as a fallback in the event that we encounter an error. Is this a valid solution @MrAlias ?

internal/instruments.go
```diff
@@ -0,0 +1,40 @@
...
+		Logger.Printf("failed to create instrument WriteCount")
+	}
+
+	newConnectionsCount, err := meter.NewInt64Counter("redis.connections",
+		metric.WithDescription("the number of connections created"),
+	)
+	if err != nil {
+		// TODO: handle errors
```
reggiemcdonald, line 32: Here as well

internal/pool/pool.go
reggiemcdonald: Yes I think that would be a good idea. 

Before we added any labels though, we wanted to show @vmihailenco the high-level changes to the code that would be involved when adding metrics

internal/instruments.go
MrAlias: The default in the case of an error is to pretty much do this: https://github.com/open-telemetry/opentelemetry-go/blob/master/api/metric/sync.go#L130-L135

There isn't really much need to explicitly do this.

MrAlias: The important thing is that the error is correctly bubbled up. The instrumentation should not halt normal operation of the code, but the user should have the ability to discover an error has occurred and there are no metric events being recorded.

reggiemcdonald: Oh nice I see that now 👍

internal/instruments.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type openTelemetryInstrumentation struct {
+	// Count of write commands performed
+	WriteCount metric.Int64Counter
+	// Count of new connections
+	NewConnectionsCount metric.Int64Counter
+}
+
+var Instruments = initInstruments()
```
vmihailenco, line 15: Can't this be just vars? E.g.

```
var (
    WritesCounter metric.Int64Counter
    ConnsCounter metric.Int64Counter
)
```

internal/instruments.go
```diff
@@ -0,0 +1,38 @@
...
+	)
+	if err != nil {
+		Logger.Printf("failed to create instrument NumNewConnections")
+	}
+
+	return &openTelemetryInstrumentation{
+		WriteCount:          writeCount,
+		NewConnectionsCount: newConnectionsCount,
+	}
+}
```
vmihailenco, line 38: I would do something more straighforward like

```
func init() {
    meter := metric.Must(global.Meter("github.com/go-redis/redis"))
    WritesCounter = meter.NewInt64Counter(...)
}
```

redis.go
vmihailenco, line 331: Perhaps it makes sense to move this to internal/pool/conn.go and instrument `WithWriter`

internal/instruments.go
```diff
@@ -0,0 +1,38 @@
...
+	meter := global.Meter("github.com/go-redis/redis")
+
+	writeCount, err := meter.NewInt64Counter("redis.writes",
+		metric.WithDescription("the number of writes initiated"),
+	)
+	if err != nil {
+		Logger.Printf("failed to create instrument WriteCount")
+	}
+
+	newConnectionsCount, err := meter.NewInt64Counter("redis.connections",
```
vmihailenco, line 27: Perhaps `redis.dials`? `redis.connections` would be a good name for number of active connections. Or `redis.new_connections` and `redis.active_connections`

TigerHe7: Updated it to `new_connections`, thanks!

redis.go
TigerHe7: Moved it to within `WithWriter` where the `WriteCount` will be incremented if `WithWriter` runs until completion. Please let me know if you think this is correct.

internal/instruments.go
TigerHe7: Updated, with the `initInstruments` function handling the case of `metric.Must(global.Meter("github.com/go-redis/redis"))` panicking

internal/instruments.go
TigerHe7: We thought it might be more readable to use the `Instruments` object to hold the different counters so we can call `internal.Instruments.WriteCount.Add(ctx, 1)` instead of `internal.WriteCount.Add(ctx, 1)`. If you think it'd be better to remove the `Instruments` object, we can definitely remove it. Please let us know!

TigerHe7: Updated so we use the vars directly as per your suggestion, and revalidated behavior using the example client. Thanks for your feedback!

vmihailenco:
Looks good - thanks. I would consider using measurement instead of counter, e.g. if we replace WritesCounter with WritesMeasurement (Int64ValueRecorder?) with MinMaxSumCount aggregator we get:
- number of writes (just as now)
- mean write duration
- min/max duration of write

Same logic applies to NewConnectionsCounter.

Issue #1317, opened 2020-05-04
go-redis v7 7.2.0 with redis 5.0.8 seems to hit a dead lock


shrivardhan92:
We do not see this problem with 5.0.4 redis

Sample code:
```
// create cluster client
rdb := 	redis.NewClusterClient(clusterOptions)

// spawn of 10 concurrent writers
for i := 0; i < 10; i++ {
  go func() {
    defer wg.Done()
    ci.processEvents(rdb)
  }()
}

```

**in 5.0.8 we see a dead lock. We see exactly 10 go-routines waiting on github.com/go-redis/redis/v7/internal.(*Once).Do***

[go-redis-issue.log](https://github.com/go-redis/redis/files/4576984/go-redis-issue.log)

Links to:  https://github.com/go-redis/redis/issues/1313

fishy:
I don't think this is really a deadlock, nor do I think this is related to #1313. I think this is just that in the current implementation `internal.Once` could hold the lock for excessively long period time when there are problems talking to the cluster, making it look like a deadlock.

The rest is copied from the discussion on #1404 (cc @rbg @vmihailenco ):

We recently started to use go-redis to do some high volume requests, and observed occasional error/latency spikes. We tried redispipe and never observe any error/latency spike with the same usage. We do prefer go-redis overall feature wise (redispipe works better for this use case, but for it's lacking on some other, nor generic use cases), and want to avoid the situation of using two different redis libraries for different use cases in the same code base.

The retries on the errors here, is an example of go-redis code making a bad situation worse, as the reason I described above. I can understand the rational of "we only want to cache successes, not errors". But the current implementation could result in a lock being held for excessively long period of time. I don't know what's a good fix right now. Maybe we can make the timeout in `ClusterClient.cmdsInfo` more configurable? That's just a random idea.

Also we tried to apply this patch, and still observed the error/latency spikes (but probably less frequent), so this is not the only problem we hit.

vmihailenco:
I wonder if you use `ReadTimeout` and `WriteTimeout` options? They allow to configure how much time cmdsInfo spends waiting for a response from Redis. There should not be a deadlock, but ClusterClient won't process commands until commands info is available. That seems okay because ClusterClient can't properly route commands without that info.

fishy:
[...]

PR #1404, opened 2020-07-14
Use sync.Once instead of internal.Once

fishy:
Store both result and error in cmdsInfoCache struct, so we can use
stdlib sync.Once instead of internal.Once.

This should also address the concerns raised by
https://github.com/go-redis/redis/issues/1317.
I don't believe internal.Once actually caused deadlock, but switching
back to stdlib sync.Once would eliminate that doubt.


vmihailenco:
`internal.Once` will can the `fn` again if there is an error. `sync.Once` will can the `fn` once and remember the error forever. This change does not look right.

fishy:
> `internal.Once` will can the `fn` again if there is an error. `sync.Once` will can the `fn` once and remember the error forever. This change does not look right.

Oh I missed the part that in `internal.Once` you only store done when it's successful.

I think that's what caused #1317 looks like a deadlock: when they have 10 concurrent calls to the same `internal.Once.Do`, in `sync.Once` implementation only one will be executed and others will return immediately after the one executed returns. But in `internal.Once` implementation, all the waiting ones will be executed as long as the previous one didn't success. So if the cluster is unhealthy, those attempts will wait for N*timeout. In the actual function executed, there's no context timeout either (https://github.com/go-redis/redis/blob/62469e4f420affe899ccfeadadcf25531078f94f/cluster.go#L1560), which makes the timeout to be even longer.

rbg:
@fishy that is not quite correct. @vmihailenco is correct in that internal.Once is doing the correct intended behaviour. The func() is performed exactly once iff it returns err == nil.  If func() happens to block then the mutex waiters will, well.. wait.

fishy:
Yea I understand that this PR is not the correct fix now. I'm just trying to have a discussion regarding #1317, or to some extend, the lock contention/error handling problem of go-redis library in general. If you prefer I can move this discussion to the issue instead.

We recently started to use go-redis to do some high volume requests, and observed occasional error/latency spikes. We tried redispipe and never observe any error/latency spike with the same usage. We do prefer go-redis overall feature wise (redispipe works better for this use case, but for it's lacking on some other, nor generic use cases), and want to avoid the situation of using two different redis libraries for different use cases in the same code base.

The retries on the errors here, is an example of go-redis code making a bad situation worse, as the reason I described above. I can understand the rational of "we only want to cache successes, not errors". But the current implementation could result in a lock being held for excessively long period of time. I don't know what's a good fix right now. Maybe we can make the timeout in `ClusterClient.cmdsInfo` more configurable? That's just a random idea.

Also we tried to apply this patch, and still observe the error/latency spikes (but probably less frequent), so this is not the only problem we hit.

rbg:
> I'm just trying to have a discussion regarding #1317,

that issue would be the appropriate place to discuss.

fishy:
Closing this PR as we moved the discussion to #1317

PR #1405, opened 2020-07-14
Use atomic.Value instead of lock for ConnPool.lastDialError

fishy:
This makes the reading and writing of lastDialError from the pool
faster, as atomic.Value is much more lightweight than the mutex.

Note that using error in atomic.Value directly could cause panics,
because errors could have inconsistent types. Thus wrap them with a
simple struct.


vmihailenco:
Looks good - thanks!

Issue #1409, opened 2020-07-15
v8: Heavy-weight dependencies


kyleconroy:
I've been trying to integrate go-redis into my application, but am running into dependency problems. After initializing a fresh Go module and installing v8, I was surprised by the breadth and scope of the required dependencies.

```
go mod init github.com/kyleconroy/redis-mod
go get github.com/go-redis/redis/v8
```

Here's the resulting [go.sum](https://gist.github.com/kyleconroy/9268906c2ddb4bc669bb3eb4cc6d85c4), which list 60 packages:

```
cloud.google.com/go
dmitri.shuralyov.com/gpu/mtl
github.com/BurntSushi/toml
github.com/BurntSushi/xgb
github.com/DataDog/sketches-go
github.com/OneOfOne/xxhash
github.com/benbjohnson/clock
github.com/census-instrumentation/opencensus-proto
github.com/cespare/xxhash
github.com/client9/misspell
github.com/cncf/udpa/go
github.com/davecgh/go-spew
github.com/dgryski/go-rendezvous
github.com/envoyproxy/go-control-plane
github.com/envoyproxy/protoc-gen-validate
github.com/fsnotify/fsnotify
github.com/go-gl/glfw/v3.3/glfw
github.com/go-redis/redis
github.com/go-redis/redis/v8
github.com/golang/glog
github.com/golang/mock
github.com/golang/protobuf
github.com/google/go-cmp
github.com/google/gofuzz
github.com/hpcloud/tail
github.com/kr/pretty
github.com/kr/pty
github.com/kr/text
github.com/onsi/ginkgo
github.com/onsi/gomega
github.com/opentracing/opentracing-go
github.com/pmezard/go-difflib
github.com/prometheus/client_model
github.com/spaolacci/murmur3
github.com/stretchr/objx
github.com/stretchr/testify
go.opentelemetry.io/otel
golang.org/x/crypto
golang.org/x/exp
golang.org/x/image
golang.org/x/lint
golang.org/x/mobile
golang.org/x/mod
golang.org/x/net
golang.org/x/oauth2
golang.org/x/sync
golang.org/x/sys
golang.org/x/text
golang.org/x/tools
golang.org/x/xerrors
google.golang.org/appengine
google.golang.org/genproto
google.golang.org/grpc
google.golang.org/protobuf
gopkg.in/check.v1
gopkg.in/fsnotify.v1
gopkg.in/tomb.v1
gopkg.in/yaml.v2
[...]

PR #1410, opened 2020-07-15
Replace golang.org/x/exp/rand with math/rand

kyleconroy:
This is an example change to show how replacing the `golang.org/x/exp/rand` package can fix some of the issues raised in #1409. I'm new to this code base, so I'm assuming there's a good reason to use `exp/rand` and that this PR probably introduces performance regressions or bugs. Because of those issues, I've opened this as draft. Feel free to close.


vmihailenco:
AFAIR golang.org/x/exp/rand is significantly faster for some APIs. That seems to be more important than small go.mod file. Also see my answer in https://github.com/go-redis/redis/issues/1409#issuecomment-659173389

kyleconroy:
@vmihailenco not surprising at all. I assumed there was a good reason to use `exp/rand`

PR #1419, opened 2020-07-21
feat(option): default dialtimeout take effect when dialer is nil

lsytj0413:
The default DialTimeout will not take effect when opt.DialTimeout is zero and opt.Dialer is nil.


vmihailenco:
Looks good - thanks!

PR #1422, opened 2020-07-23
Allow setting the maximum number of connections per client

natewernimont-wk:
I have noticed while using go-redis that sometimes the application will have a surge of requests to redis. In response, we encounter many dial timeouts, saturate our pool, and start spinning through a massive number of one use connections. Instead of allowing these requests to go through, we would like to be able to restrict the number of active connections per client. This PR aims to implement that.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -69,7 +69,9 @@ type ClusterOptions struct {
 	WriteTimeout time.Duration
 
 	// PoolSize applies per cluster node and not for the whole cluster.
-	PoolSize           int
+	PoolSize int
+	// MaxConns applies per cluster node and not for the whole cluster.
+	MaxConns           int
 	MinIdleConns       int
 	MaxConnAge         time.Duration
 	PoolTimeout        time.Duration
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -83,7 +83,7 @@ type ConnPool struct {
 	poolSize     int
 	idleConnsLen int
 
-	connsCountMu sync.Mutex
+	connsCountMu sync.RWMutex
 	connsCount   int
 
 	stats Stats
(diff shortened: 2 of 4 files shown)


vmihailenco:
How this is different from PoolSize?

natewernimont-wk:
> How this is different from PoolSize?

`PoolSize` reflects the maximum number of cached/pooled connections. However, the client currently allows one-off queries to be made beyond the constrains of `PoolSize`. So, if one were to have `PoolSize` at say, 200 connections, and there was a sudden flurry of 1000 requests to redis, then the first 200 connections will be utilized and cached for later re-use, but the remaining 800 connections would all be on time use and destroyed upon the completion of the query. The goal here would be to prevent the client from being able to go above a certain number of connections at any one point in time.

natewernimont-wk:
This would be similar to the behavior of https://golang.org/pkg/database/sql/#DB.SetMaxOpenConns

natewernimont-wk:
I have been digging through the code some more, and I realized that calls to `Get` on the `ConnPool` won't allow new connections above the pool size due to the `waitTurn` func. However, my application is seeing random spikes in new connections to redis while the current connection count is stable. I am not able to explain that yet. We do not use `Subscribe`, so it looks like we do not call `NewConn`.

aniketsupertramp:
@natewernimont-wk  Hey Nate , did you find the root cause for your issue? 
 I have been  facing similar issue where the latency spikes and  the new connections are spiking up  on the redis cluster.

aniketsupertramp:
Hi @natewernimont-wk  did you get the root cause of this issue?  We are facing similar issues where cluster's new connections count spikes up after an interval but current connections remains pretty much same , causing latencies.

PR #1431, opened 2020-08-05
Prevent index out ouf bunds error

mvrhov:
I'm reusing this for disque client... and get index out of bounds error because there really are just 2 items in array not 3.

BTW.. is it possible to merge this down to 6.x series as we really don't want to do an upgrade

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -44,9 +44,10 @@ func (it *ScanIterator) Next(ctx context.Context) bool {
 		}
 
 		// Fetch next page.
-		if it.cmd.args[0] == "scan" {
+		switch it.cmd.args[0] {
+		case "scan", "qscan":
 			it.cmd.args[1] = it.cmd.cursor
-		} else {
+		default:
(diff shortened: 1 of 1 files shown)


iterator.go
```diff
@@ -44,7 +44,7 @@ func (it *ScanIterator) Next(ctx context.Context) bool {
 		}
 
 		// Fetch next page.
-		if it.cmd.args[0] == "scan" {
+		if it.cmd.args[0] == "scan" || len(it.cmd.args) < 3 {
```
vmihailenco, line 47: Could you instead specify the command here? Something like

```
switch it.cmd.args[0] {
    case "scan", "foo", "bar":
       it.cmd.args[1] = it.cmd.cursor
    default:
        it.cmd.args[2] = it.cmd.cursor
}
```

It seems to be easier to maintain it this way.

vmihailenco:
Thanks. I wll make sure to update v6 too.

Issue #1416, opened 2020-07-20
v7.4.0 nil dereference


chxv:
when using  go-redis@v7.4.0, it occur panic "nil dereference" randomly.
env: macos 10.15.5, go 1.14.4
the variable `node` may be nil though `err` is nil

* debug info in goland
file: v7@v7.4.0/cluster.go
![image](https://user-images.githubusercontent.com/26496749/87927260-86f02d80-cab5-11ea-9ce6-0535b9e52d56.png)
then the panic occur.
![image](https://user-images.githubusercontent.com/26496749/87927462-d0407d00-cab5-11ea-93be-1c004f95518a.png)

Issue #1401, opened 2020-07-09
go-redis v8.0.0-beta.5 panic on Redis Cluster 6.0.5


gferreux:
Hi, here is the panic:

`panic: runtime error: invalid memory address or nil pointer dereference
 -> github.com/go-redis/redis/v8.(*ClusterClient)._process
 ->   /go/pkg/mod/github.com/go-redis/redis/v8@v8.0.0-beta.5/cluster.go:776
    github.com/go-redis/redis/v8.(*ClusterClient).process
      /go/pkg/mod/github.com/go-redis/redis/v8@v8.0.0-beta.5/cluster.go:738
    github.com/go-redis/redis/v8.hooks.process
      /go/pkg/mod/github.com/go-redis/redis/v8@v8.0.0-beta.5/redis.go:57
    github.com/go-redis/redis/v8.(*ClusterClient).Process
      /go/pkg/mod/github.com/go-redis/redis/v8@v8.0.0-beta.5/cluster.go:734
    github.com/go-redis/redis/v8.cmdable.Get
      /go/pkg/mod/github.com/go-redis/redis/v8@v8.0.0-beta.5/commands.go:671`

- go-redis: v8.0.0-beta.5
- redis-server: 6.0.5
- client config: redis://redis-shard0-0.redis.redis:6379,redis-shard0-1.redis.redis:6379,redis-shard0-2.redis.redis:6379,redis-shard1-0.redis.redis:6379,redis-shard1-1.redis.redis:6379,redis-shard1-2.redis.redis:6379,redis-shard2-0.redis.redis:6379,redis-shard2-1.redis.redis:6379,redis-shard2-2.redis.redis:6379,redis-shard3-0.redis.redis:6379,redis-shard3-1.redis.redis:6379,redis-shard3-2.redis.redis:6379,redis-shard4-0.redis.redis:6379,redis-shard4-1.redis.redis:6379,redis-shard4-2.redis.redis:6379?idletimeout=300&poolsize=50&pooltimeout=30&routebylatency=true
- server config: redis-cluster with 5 shards and 2 replicas each
- context: when a node goes down

Note: client config with `routerandomly=true` do not panic

vmihailenco:
Can you provide a minimal repro?

PR #1434, opened 2020-08-06
fix nil node dereference when use RouteByLatency in cluster

GreenHedgehog:
There is was a nil pointer dereference when all nodes are failing, It was described early in this issues #1416 and #1401


vmihailenco:
Thanks!

Issue #1393, opened 2020-07-01
No support set keepTTL?


357711589:
How should  I do?

https://redis.io/commands/set
>= 6.0: Added the KEEPTTL option.

x1nchen:
like this 

```
err := rdb.Do("set", key, value, "keepttl").Err()
if err != nil {}
```

more refer to this discussion on https://github.com/go-redis/redis/pull/1439

357711589:
> like this
> 
> ```
> err := rdb.Do("set", key, value, "keepttl").Err()
> if err != nil {}
> ```
> 
> more refer to this discussion on #1439

OK Thanks

vmihailenco:
@x1nchen what about using `-1` expiration in `Set` command as indicator for `KEEPTTL`? E.g. `Set(ctx, "foo", "bar", -1)` means `SET foo bar KEEPTTL`.

Adding separate command for KEEPTTL is problematic since we need 3 of them:
- SetKeepTTL
- SetKeepTTLNx
- SetKeepTTLEx

x1nchen:
@vmihailenco 

> Adding separate command for KEEPTTL is problematic since we need 3 of them:

> SetKeepTTL
> SetKeepTTLNx
> SetKeepTTLEx

Yes. You're right!

x1nchen:
@vmihailenco 

> what about using -1 expiration in Set command as indicator for KEEPTTL? E.g. Set(ctx, "foo", "bar", -1) means SET foo bar KEEPTTL.

Oh, this mean is too implicit for users. I prefer to add new 3 apis just like you introduced above.

vmihailenco:
It could be

```redis
const KeepTTL = -1
Set(ctx, "foo", "bar", KeepTTL)
SetNX(ctx, "foo", "bar", KeepTTL)
SetXX(ctx, "foo", "bar", KeepTTL)
```

What do you think?

x1nchen:
> It could be
> 
> ```go
> const KeepTTL = -1
> Set(ctx, "foo", "bar", KeepTTL)
> SetNX(ctx, "foo", "bar", KeepTTL)
> SetXX(ctx, "foo", "bar", KeepTTL)
> ```
> 
> What do you think?

You mean you want to implement this feature as the code shows below?

```
func (c cmdable) Set(ctx context.Context, key string, value interface{}, expiration time.Duration) *StatusCmd {
	args := make([]interface{}, 3, 5)
	args[0] = "set"
	args[1] = key
[...]

Issue #1308, opened 2020-04-25
how to KEEPTTL


codenoid:
https://github.com/antirez/redis/pull/6679

yugandharbandi2220:
Hi, any update on this support for go-redis

x1nchen:
like this 

```
err := rdb.Do("set", key, value, "keepttl").Err()
if err != nil {}
```

more refer to this discussion on https://github.com/go-redis/redis/pull/1439

PR #1439, opened 2020-08-13
feat: add support for set keepttl

x1nchen:
issue #1393

ref https://redis.io/commands/set


x1nchen:
https://github.com/go-redis/redis/issues/1308

x1nchen:
Well, I suppose this PR is unnecessary.

Just use redis.Client.Do method 

```
// Do creates a Cmd from the args and processes the cmd.
func (c *Client) Do(args ...interface{}) *Cmd {
	return c.DoContext(c.ctx, args...)
}
```

for example

```
err := rdb.Do("set", key, value, "keepttl").Err()
if err != nil {}
```

PR #1441, opened 2020-08-14
fix  the idleConnsLen may be less than MinIdleConns

tongjichao:


vmihailenco:
Could you explain what this fixed and how?

dmitsh:
> Could you explain what this fixed and how?

Looks like it fixed Travis CI job.

vmihailenco:
I've re-run CI job and now it fails just like the master :) Nevertheless I need to understand the change before merging it...

dmitsh:
> I've re-run CI job and now it fails just like the master :) Nevertheless I need to understand the change before merging it...

Hi @vmihailenco , off the topic, but do you know why Travis CI fails? I was running `go test ./...` locally on my machine, but I'm getting different errors. The ClusterClient test seems to pass, though (the one that I noticed was failing in Travis CI)

vmihailenco:
@dmitsh sorry - I've missed your comment. The build is fixed now - the problem was a flaky test that relied on some internal details.

@tongjichao closing since I don't understand what this fixes. Please re-open if this is still a problem.

PR #1448, opened 2020-08-21
Support string array in pubsub message payload

rueian:
Hi, I'd like to add the support of string array in pubsub message payload.

Since when enabling the ["Redis server-assisted client side caching"](https://redis.io/topics/client-side-caching) feature with RESP2 and Redis 6, the redis-server could send string array payload which contains keys to invalid via the pub/sub message.

This PR stores the string array payload into a new field `PayloadSlice` of the pubsub`Message` struct for not breaking the  existing code that using the existing `Payload` field.

This PR is far from a complete client side caching solution, but at least this may let user be able to use this new redis feature with this library.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -284,9 +284,10 @@ func (m *Subscription) String() string {
 
 // Message received as result of a PUBLISH command issued by another client.
 type Message struct {
-	Channel string
-	Pattern string
-	Payload string
+	Channel      string
+	Pattern      string
+	Payload      string
+	PayloadSlice []string
 }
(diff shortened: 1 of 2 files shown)


rueian:
Could someone please take a look?

vmihailenco:
@rueian sorry for the delay - was busy with other stuff. This looks pretty good.

Regarding `Payload` vs `PayloadSlice` - we could break compatibility for v8 release, but I guess it makes sense to postpone this decision until we properly support RESP2.

PR #1453, opened 2020-08-26
Remove grpc dependency

apersson:
open telemetry dropped the dependency on grpc in v0.11.0. In v0.11.0, the
`api/kv` and `api/label` packages were unified and moved to `label`.


apersson:
I see you already got this resolved. Thanks!

PR #1454, opened 2020-08-26
Trace dialer targets in the relevant spans

asuffield:
This is hugely useful for debugging subtly broken setups where one of the addresses turns out to be wrong.


vmihailenco:
Looks good - thanks!

PR #1462, opened 2020-09-03
use 3 sentinel instances in unit testing

dmitsh:
Signed-off-by: Dmitry Shmulevich <dmitry.shmulevich@sysdig.com>

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -35,13 +35,18 @@ const (
 	sentinelMasterPort = "9123"
 	sentinelSlave1Port = "9124"
 	sentinelSlave2Port = "9125"
-	sentinelPort       = "9126"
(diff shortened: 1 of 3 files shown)


vmihailenco:
Thanks! I guess you are trying to fix the build... I am not sure bug I guess it is related to recent change in SingleConnPool - I will give it a try too...

dmitsh:
Thanks @vmihailenco. Yes, I've been debugging sentinel test. I noticed that single sentinel instance constantly alternates the slaves when it tries to elect a new master. With three sentinel instances that doesn't happen.
However, I'm still getting net.OpError("use of closed network connection") when connecting to the newly elected master.
I will keep troubleshooting when I have spare cycles.

Issue #1478, opened 2020-09-10
afterhook could not excute when beforehook returns error


kaka19ace:
For example: 
https://github.com/go-redis/redisext/blob/master/otel.go.   
opentelemetry  call span.End() in afterProcess.  

there is an advice like https://docs.djangoproject.com/en/1.8/topics/http/middleware/
![](https://docs.djangoproject.com/en/1.8/_images/middleware.svg)
```golang
        // record process index
	processIndex := 0

        // process hooks as stack(lifo) sequence.
	// before, break if err occurs, and record current processIndex.   
	for ; err == nil && processIndex < len(hooks); processIndex++ {
		cmd, err = hooks[processIndex].BeforeProcess(ctx, cmd)
	}

	if err != nil {
            err = fn(ctx, cmd)
        } 

	// process response,  start from [index -> 0], not all afterHooks.
	for index := processIndex - 1; index >= 0; index-- {
		err = hooks[index].AfterProcess(ctx, cmd, err) // maybe redirect err by hook
	}
        cmd.SetErr(err)
	return err
```

vmihailenco:
@kaka19ace thanks - that makes a lot of sense. Please take a loot at https://github.com/go-redis/redis/pull/1479

PR #1479, opened 2020-09-10
Make sure to call after hook on error

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/1478

@kaka19ace please review.

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -48,83 +48,63 @@ func (hs *hooks) AddHook(hook Hook) {
 func (hs hooks) process(
 	ctx context.Context, cmd Cmder, fn func(context.Context, Cmder) error,
 ) error {
-	ctx, err := hs.beforeProcess(ctx, cmd)
-	if err != nil {
-		cmd.SetErr(err)
-		return err
+	if len(hs.hooks) == 0 {
+		return fn(ctx, cmd)
 	}
 
-	cmdErr := fn(ctx, cmd)
+	var hookIndex int
+	var retErr error
 
-	if err := hs.afterProcess(ctx, cmd); err != nil {
-		cmd.SetErr(err)
-		return err
-	}
-
-	return cmdErr
-}
-
-func (hs hooks) beforeProcess(ctx context.Context, cmd Cmder) (context.Context, error) {
-	for _, h := range hs.hooks {
-		var err error
-		ctx, err = h.BeforeProcess(ctx, cmd)
-		if err != nil {
-			return nil, err
+	for ; hookIndex < len(hs.hooks) && retErr == nil; hookIndex++ {
+		ctx, retErr = hs.hooks[hookIndex].BeforeProcess(ctx, cmd)
+		if retErr != nil {
+			cmd.SetErr(retErr)
 		}
 	}
-	return ctx, nil
-}
 
-func (hs hooks) afterProcess(ctx context.Context, cmd Cmder) error {
-	var firstErr error
-	for i := len(hs.hooks) - 1; i >= 0; i-- {
-		h := hs.hooks[i]
-		if err := h.AfterProcess(ctx, cmd); err != nil && firstErr == nil {
-			firstErr = err
+	if retErr == nil {
+		retErr = fn(ctx, cmd)
+	}
+
+	for hookIndex--; hookIndex >= 0; hookIndex-- {
+		if err := hs.hooks[hookIndex].AfterProcess(ctx, cmd); err != nil {
+			retErr = err
+			cmd.SetErr(retErr)
 		}
 	}
-	return firstErr
+
+	return retErr
 }
 
 func (hs hooks) processPipeline(
 	ctx context.Context, cmds []Cmder, fn func(context.Context, []Cmder) error,
 ) error {
-	ctx, err := hs.beforeProcessPipeline(ctx, cmds)
-	if err != nil {
-		setCmdsErr(cmds, err)
-		return err
+	if len(hs.hooks) == 0 {
+		return fn(ctx, cmds)
 	}
 
-	cmdsErr := fn(ctx, cmds)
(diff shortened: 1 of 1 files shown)


redis.go
```diff
@@ -48,83 +48,57 @@ func (hs *hooks) AddHook(hook Hook) {
...
-	if err != nil {
-		cmd.SetErr(err)
-		return err
+	var hookIndex int
+	var retErr error
+
+	for ; hookIndex < len(hs.hooks); hookIndex++ {
+		ctx, retErr = hs.hooks[hookIndex].BeforeProcess(ctx, cmd)
+		if retErr != nil {
+			cmd.SetErr(retErr)
```
kaka19ace, line 57: break will not trigger idx++

redis.go
```diff
@@ -48,83 +48,57 @@ func (hs *hooks) AddHook(hook Hook) {
...
+		}
 	}
 
-	cmdErr := fn(ctx, cmd)
-
-	if err := hs.afterProcess(ctx, cmd); err != nil {
-		cmd.SetErr(err)
-		return err
+	if retErr == nil {
+		retErr = fn(ctx, cmd)
```
kaka19ace, line 63: if retErr != nil -> cmd.SetErr(retErr)

kaka19ace: cm.SetErr occurs in fn?

redis.go
```diff
@@ -48,83 +48,57 @@ func (hs *hooks) AddHook(hook Hook) {
...
 	ctx context.Context, cmd Cmder, fn func(context.Context, Cmder) error,
 ) error {
-	ctx, err := hs.beforeProcess(ctx, cmd)
-	if err != nil {
-		cmd.SetErr(err)
-		return err
+	var hookIndex int
+	var retErr error
+
+	for ; hookIndex < len(hs.hooks); hookIndex++ {
```
kaka19ace, line 54: maybe 
```for ; retErr == nil && hookIndex < len(hs.hooks); hookIndex++```
looks nice, 
and  idx-- is not required at afterprocess starts.

File: redis.go (hunk 2 above)
Comment (vmihailenco) [reply redis.go:63]: Yes, cmd.SetErr is not needed here - `fn` sets the `cmd.err`

File: redis.go (hunk 1 above)
Comment (vmihailenco) [reply redis.go:57]: Right, but why that is a problem? I call AfterProcess for all functions except the last one that returned the error.

File: redis.go (hunk 3 above)
Comment (vmihailenco) [reply redis.go:54]: I believe with `retErr == nil` I need `idx -= 2` if I don't want to include the hook that returned the error. Am I missing something?

Comment (kaka19ace) [reply redis.go:54]: example:   
beforeProcess is stop at i'th idx (i>=0),  afterProcess should starts at hooks[i], not hooks[i-1]

Comment (kaka19ace) [reply redis.go:54]: code:
```golang
// after hook
for hookIndex--; hookIndex >= 0; hookIndex--
```
hookIndex = i
the  first `hookIndex--` will be set i-1.  it will lose the i'th hook.

vmihailenco: Yes, I intentionally skipped the AfterProcess hook that returned the error from BeforeProcess. Since I don't have strong preference I've changed the code to include it.

Please take another look.

kaka19ace: if beforehooks run all success, hookIndex is equals to len(hooks),
afterhooks will be run out of range.

redis.go
```diff
@@ -48,83 +48,57 @@ func (hs *hooks) AddHook(hook Hook) {
...
-	return cmdErr
-}
-
-func (hs hooks) beforeProcess(ctx context.Context, cmd Cmder) (context.Context, error) {
-	for _, h := range hs.hooks {
-		var err error
-		ctx, err = h.BeforeProcess(ctx, cmd)
-		if err != nil {
-			return nil, err
+	for ; hookIndex >= 0; hookIndex-- {
```
kaka19ace, line 66: another issues:
if beforehooks run all success, hookIndex is equals to len(hooks),
afterhooks will be run out of range.

that's why 
using 
```for ; err == nil && hookIndex < len(hs.hooks); hookIndex++ ```
instead of 
```for ;  hookIndex < len(hs.hooks); hookIndex++ ``` is better.

you don't need to check hookIndex is out of range. :)

File: redis.go (hunk 3 above)
Comment (kaka19ace) [reply redis.go:58]: there is example,  register a hook wicth feature is a authentification feature,  
if error is returns from this beforehook,  

the i'th afterhook  should record the error and emit it's metrics  with error msg.

Comment (kaka19ace) [reply redis.go:58]: example:  
there is a hook registered in hooks at i'th index, 
beforehook is focus on authentification, 
afterhook records log and emit metrics,  the error msg is used in afterhook.

File: redis.go (hunk 4 above)
Comment (vmihailenco) [reply redis.go]: You are right - fixed the code

Comment (kaka19ace) [reply redis.go]: pipelineHooks is also need to fixed :)

File: redis.go [hunk 5]
```diff
@@ -48,83 +48,64 @@ func (hs *hooks) AddHook(hook Hook) {
...
+	var retErr error
 
-	if err := hs.afterProcessPipeline(ctx, cmds); err != nil {
-		setCmdsErr(cmds, err)
-		return err
+	for ; hookIndex < len(hs.hooks) && retErr == nil; hookIndex++ {
+		ctx, retErr = hs.hooks[hookIndex].BeforeProcessPipeline(ctx, cmds)
+		if retErr != nil {
+			setCmdsErr(cmds, retErr)
+			break
```
kaka19ace, line 93: there is no need break here

vmihailenco: Thanks, removed

kaka19ace:
LGTM 👍

vmihailenco:
@kaka19ace thanks for the review!

PR #1480, opened 2020-09-10
Fix not applying updated ClusterClient context after calling WithContext method

GreenHedgehog:
In this method, аfter cloning `ClusterClient` instance, `cmdsInfoCache` will still use context from the original. Small fix for this.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -707,6 +707,7 @@ func (c *ClusterClient) WithContext(ctx context.Context) *ClusterClient {
 	clone.cmdable = clone.Process
 	clone.hooks.lock()
 	clone.ctx = ctx
+	clone.cmdsInfoCache = newCmdsInfoCache(clone.cmdsInfo)
 	return &clone
 }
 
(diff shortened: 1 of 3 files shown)


vmihailenco:
This change also resets `cmdsInfoCache` - previously `cmdsInfoCache` was reused between different copies. Is this really worth it to fix context?

GreenHedgehog:
> This change also resets `cmdsInfoCache` - previously `cmdsInfoCache` was reused between different copies. Is this really worth it to fix context?

If i will add coping of cache it would be okay?

vmihailenco:
>If i will add coping of cache it would be okay?

That would mean that different cluster client have separate cache.

I believe the proper fix is to explicitly pass `context.Context` to `cmdsInfoCache.fn` as an argument.

GreenHedgehog:
Yeah, youre right, its the best solution. Pushed new commit with that fix.

vmihailenco:
Thanks - merged.

PR #1487, opened 2020-09-11
Use maxRedirects only for Redirect errors and maxRetries for network errors

duke-cliff:
Use maxRedirects only for Redirect errors(moved/ask/read only/loading...)
use maxRetries for network error retries.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -31,7 +31,7 @@ type ClusterOptions struct {
 	NewClient func(opt *Options) *Client
 
 	// The maximum number of retries before giving up. Command is retried
-	// on network errors and MOVED/ASK redirects.
+	// on MOVED/ASK redirects.
 	// Default is 3 retries.
 	MaxRedirects int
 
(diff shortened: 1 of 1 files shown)


cluster.go
```diff
@@ -766,9 +768,15 @@ func (c *ClusterClient) _process(ctx context.Context, cmd Cmder) error {
 	var node *clusterNode
 	var ask bool
 	var lastErr error
-	for attempt := 0; attempt <= c.opt.MaxRedirects; attempt++ {
-		if attempt > 0 {
-			if err := internal.Sleep(ctx, c.retryBackoff(attempt)); err != nil {
+	var failureAttempt = 0
+	var redirectAttempt = 0
+	for redirectAttempt <= c.opt.MaxRedirects || failureAttempt < c.opt.MaxRetries {
```
vmihailenco, line 773: You probably want it to be `&&`, not `||`.

vmihailenco:
This only updates part of the code - pipelines still use `MaxRedirects`.

vmihailenco:
Closing since this is rather complicated and there is no progress.

shenqianjin:
this change looks good.

@vmihailenco why did we close it? is there any alternative solution about this ?

PR #1495, opened 2020-09-17
Reduce dependency chain

dim:
Importing golang.org/x/exp results in a pretty heavy set of dependencies which is not nice. To avoid lock contention on `math/rand`, I have added a dedicated internal package with its own mutex.


example_instrumentation_test.go
```diff
@@ -57,7 +57,7 @@ func ExamplePipeline_instrumentation() {
 	// pipeline finished processing: [ping: PONG ping: PONG]
 }
 
-func ExampleWatch_instrumentation() {
+func ExampleClient_Watch_instrumentation() {
```
dim, line 60: `go vet` was complaining

vmihailenco:
>To avoid lock contention

It is not about lock contention. AFAIR code like `rand.New(rand.NewSource(seed))` is several times faster with exp/rand and I just use `exp/rand` everywhere by default (because why not :). I believe go-redis does not use `rand.New` so this PR should be fine.

Thanks!

PR #1496, opened 2020-09-17
Fixes #1386; pipeline.Exec() sometimes returns spurious `StatusCmd`

### LINKED ISSUE #1386: v7 transaction pipeline returns incorrect results sometimes
Trying to upgrade to v7.4.0 from v6.15.2 but I am running into issues with the new transaction pipeline code. 

When executing the transaction pipeline, I am sometimes getting the MULTI StatusCmd back in our result set now. Before, I would always just get back the BoolCmds. Now it seems that in certain situations I lose one of the BoolCmds I should be getting back and instead get a StatusCmd back of the MULTI command.  

I am using the Client in redis.go and initializing the transactional pipeline by calling the func TxPipeline() on that client. 
Using that pipeline, I then add each item. See example code below. 

```
pipe := client.TxPipeline()

// for each item I issue the SetNX
pipe.SetNX(key, value, expiration)

response, err := pipe.Exec()
```

If no error is returned, I then loop through the response array of []redis.Cmder's. This array is where I am having the issue. Before it would always just return BoolCmd's where as now it seems to sometimes return the first item as the StatusCmd from the MULTI cmd. So, when I had 6 items to set in redis, I got back 6 items in the response array but the first would be StatusCmd and the remaining 5 would be BoolCmds. I'd be missing one BoolCmd and have an extra StatusCmd that I don't expect.

It doesn't seem to always happen though... another time when I had 7 items to set in redis, I correctly got back all 7 items and they were all BoolCmds. 

Additionally, from my testing I believe this might be the problem commit: 218b17f0fc567ce62f7d6d0607caff1d8100354f 
When I test with commit 709e543f16bf20d4f6df33bc7d24fb4b6a0991bc I don't see this issue.

sgp:
This is caused by function `wrapMultiExec` sometimes clobbering the caller's
cmd slice, corrupting future requests.

### DIFF
--- a/pipeline_test.go
+++ b/pipeline_test.go
@@ -1,6 +1,8 @@
 package redis_test
 
 import (
+	"strconv"
+
 	"github.com/go-redis/redis/v8"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks.

PR #1493, opened 2020-09-17
Feature/ctx cancel

vmihailenco:
Closes https://github.com/go-redis/redis/pull/1490

PR #1497, opened 2020-09-17
remove ctx fields, as they serve no purpose anymore

kernle32dll:
While migrating one of my projects to v8, I realized that there are still `context` fields left, which don't server any purpose anymore after https://github.com/go-redis/redis/pull/1493.

So I propose to remove them altogether.

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -208,36 +208,6 @@ func BenchmarkZAdd(b *testing.B) {
 	})
 }
 
-var clientSink *redis.Client
-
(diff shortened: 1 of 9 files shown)


vmihailenco:
Thanks, I will keep this for v9.

dittrichlucas (approved):

kernle32dll:
Wow, I completely forgot about this :D I will rebase onto master this evening.

vmihailenco:
Looks like a similar PR was merged some time ago and this is no longer needed. Thanks anyway.

Issue #1501, opened 2020-09-21
Wrong cmd slot of xread in cluster


chengjing1181122598:
Issue tracker is used for reporting bugs and discussing new features. Please use
[stackoverflow](https://stackoverflow.com) for supporting issues.

## Expected Behavior

Xread should not occur MOVED error in cluster.
## Current Behavior

Xread always occur MOVED error in cluster.
## Steps to Reproduce

1. github.com/go-redis/redis/v7 v7.4.0
2. In cluster mode
3. XRead(&redis.XReadArgs{Streams: []string{"test_stream_pub_sub", "$"}, Count:   100, Block:   0})
![image](https://user-images.githubusercontent.com/23028470/93732289-3dca7000-fc03-11ea-83ad-4653e236b499.png)
4. Wrong cmd args position used to slot, should be "test_stream_pub_sub" instead of "count".
5. Incidentally please check other cmd about stream.

vmihailenco:
go-redis uses command information from Redis

```
command info xread
1) 1) "xread"
   2) (integer) -4
   3) 1) readonly
      2) movablekeys
   4) (integer) 1
   5) (integer) 1
   6) (integer) 1
   7) 1) @read
      2) @stream
      3) @slow
      4) @blocking
```

Here `4) (integer) 1` is the position (`1`) of the key argument. So it uses `args[1]` which is `count`. So everything works as expected (but incorrectly).

chengjing1181122598:
> go-redis uses [command information](https://redis.io/commands/command#nested-result-array) from Redis
> 
> ```
> command info xread
> 1) 1) "xread"
>    2) (integer) -4
>    3) 1) readonly
>       2) movablekeys
>    4) (integer) 1
>    5) (integer) 1
>    6) (integer) 1
>    7) 1) @read
>       2) @stream
>       3) @slow
>       4) @blocking
> ```
> 
> Here `4) (integer) 1` is the position (`1`) of the key argument. So it uses `args[1]` which is `count`. So everything works as expected (but incorrectly).

chengjing1181122598:
> go-redis uses [command information](https://redis.io/commands/command#nested-result-array) from Redis
> 
> ```
> command info xread
> 1) 1) "xread"
>    2) (integer) -4
>    3) 1) readonly
>       2) movablekeys
>    4) (integer) 1
>    5) (integer) 1
>    6) (integer) 1
>    7) 1) @read
>       2) @stream
>       3) @slow
>       4) @blocking
[...]

PR #1502, opened 2020-09-22
fix: firstKeyPos cmdSlot

jamsonzan:
#1501 

only **xread xreadgroup zunionstore zinterstore** need to fix

### DIFF
--- a/command.go
+++ b/command.go
@@ -18,6 +18,7 @@ type Cmder interface {
 	Args() []interface{}
 	String() string
 	stringArg(int) string
+	firstKeyOffset() int
 
 	readTimeout() *time.Duration
 	readReply(rd *proto.Reader) error
@@ -106,6 +107,7 @@ type baseCmd struct {
 	ctx  context.Context
 	args []interface{}
 	err  error
+	keyOffset int
 
 	_readTimeout *time.Duration
 }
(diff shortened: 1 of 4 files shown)


command.go
```diff
@@ -1976,14 +1987,14 @@ func (cmd *GeoPosCmd) readReply(rd *proto.Reader) error {
...
-	StepCount   int8
-	ReadOnly    bool
+	Name         string
+	Arity        int8
+	Flags        []string
+	ACLFlags     []string
+	FirstKeyBase int8
+	LastKeyBase  int8
+	StepCount    int8
+	ReadOnly     bool
```
vmihailenco, line 1997: Please revert - there is no need to change this to fix the issue.

command.go
```diff
@@ -106,6 +108,7 @@ type baseCmd struct {
 	ctx  context.Context
 	args []interface{}
 	err  error
+	keyOffset int
```
vmihailenco, line 111: Using `keyBase+keyOffset` looks a bit complicated. Can't we rename this to `keyPos` and use it when `cmd.keyPos != 0` and fallback to `info.FirstKeyPos`?

command.go
```diff
@@ -76,7 +78,7 @@ func cmdFirstKeyPos(cmd Cmder, info *CommandInfo) int {
 	if info == nil {
 		return 0
 	}
-	return int(info.FirstKeyPos)
+	return int(info.FirstKeyBase) + cmd.firstKeyOffset()
```
vmihailenco, line 81: E.g.

```
if pos := cmd.firstKeyPos(); pos != 0 {
    return pos
}
return info.FirstKeyPos
```

jamsonzan: good!

vmihailenco:
Thanks!

PR #1503, opened 2020-09-22
Add XINFO STREAM

icholy:
This adds support for the `XINFO STREAM` command.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1081,34 +1081,43 @@ func (cmd *XMessageSliceCmd) readReply(rd *proto.Reader) error {
 	return nil
 }
 
+func readXMessage(rd *proto.Reader) (XMessage, error) {
+	n, err := rd.ReadArrayLen()
+	if err != nil {
+		return XMessage{}, err
+	}
+	if n != 2 {
+		return XMessage{}, fmt.Errorf("got %d, wanted 2", n)
+	}
+
+	id, err := rd.ReadString()
+	if err != nil {
+		return XMessage{}, err
+	}
+
+	var values map[string]interface{}
+
+	v, err := rd.ReadArrayReply(stringInterfaceMapParser)
+	if err != nil {
+		if err != proto.Nil {
+			return XMessage{}, err
+		}
+	} else {
+		values = v.(map[string]interface{})
+	}
+
+	return XMessage{
+		ID:     id,
+		Values: values,
+	}, nil
+}
+
 // xMessageSliceParser implements proto.MultiBulkParse.
 func xMessageSliceParser(rd *proto.Reader, n int64) (interface{}, error) {
 	msgs := make([]XMessage, n)
-	for i := 0; i < len(msgs); i++ {
-		i := i
-		_, err := rd.ReadArrayReply(func(rd *proto.Reader, n int64) (interface{}, error) {
-			id, err := rd.ReadString()
-			if err != nil {
-				return nil, err
-			}
-
-			var values map[string]interface{}
-
-			v, err := rd.ReadArrayReply(stringInterfaceMapParser)
-			if err != nil {
-				if err != proto.Nil {
-					return nil, err
-				}
-			} else {
-				values = v.(map[string]interface{})
-			}
-
-			msgs[i] = XMessage{
-				ID:     id,
-				Values: values,
-			}
-			return nil, nil
-		})
+	for i := int64(0); i < n; i++ {
+		var err error
+		msgs[i], err = readXMessage(rd)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


icholy:
@vmihailenco this is ready for review

command.go
```diff
@@ -1066,37 +1066,39 @@ func (cmd *XMessageSliceCmd) readReply(rd *proto.Reader) error {
...
-				ID:     id,
-				Values: values,
-			}
-			return nil, nil
-		})
+		msg, err := rd.ReadArrayReply(xMessageParser)
 		if err != nil {
 			return nil, err
 		}
+		msgs[i] = msg.(XMessage)
```
vmihailenco, line 1101: It was written this way to avoid interface{} allocation and `XMessage` copying. It is important in this loop since we could read hundreds of messages in one request.

I suggest to refactor this code as:

```go
_, err := rd.ReadArrayReply(func(rd *proto.Reader, n int64) (interface{}, error) {
    var err error
    msgs[i], err = readMessage(rd)
    if err != nil {
        return nil, err
    }
    return nil, nil
})
```

command.go
```diff
@@ -1470,6 +1472,100 @@ func xGroupInfoParser(rd *proto.Reader, n int64) (interface{}, error) {
...
+			args: []interface{}{"xinfo", "stream", stream},
+		},
+	}
+}
+
+func (cmd *XInfoStreamCmd) Val() XInfoStream {
+	return cmd.val
+}
+
+func (cmd *XInfoStreamCmd) Result() (XInfoStream, error) {
```
vmihailenco, line 1505: `Val` and `Result` should return `*XInfoStream`. `XInfoStream` is too big for returning a copy.

vmihailenco:
Thanks for the PR. I suggest to replace `ReadArrayReply` with `ReadArrayLen` as shown in https://github.com/go-redis/redis/pull/1505/files. Otherwise we are doing unnecessary allocations and copying in a loop.

icholy:
@vmihailenco  what are you your thoughts on adding a helper to `proto.Reader` for expected array lengths? Something like this:

``` go
func (r *Reader) ExpectArrayLen(n int) error {
	got, err := r.ReadArrayLen()
	if err != nil {
		return err
	}
	if n != got {
		return fmt.Errorf("redis: unexpected array length; want %d, got %d", n, got)
	}
	return nil
}
```

An alternative name could be `ReadFixedArrayLen`.

icholy:
@vmihailenco I've updated the PR

icholy:
Looks like a flakey test

```
redis: 2020/09/23 15:35:58 ring.go:321: ring shard state changed: Redis<:6390 db:0> is down
redis: 2020/09/23 15:35:58 ring.go:321: ring shard state changed: Redis<:6391 db:0> is down
redis: 2020/09/23 15:35:59 sentinel.go:536: sentinel: GetMasterAddrByName name="mymaster" failed: dial tcp :9126: connect: connection refused
redis: 2020/09/23 15:35:59 sentinel.go:536: sentinel: GetMasterAddrByName name="mymaster" failed: dial tcp :9126: connect: connection refused
redis: 2020/09/23 15:35:59 sentinel.go:471: sentinel: GetMasterAddrByName master="mymaster" failed: dial tcp :9126: connect: connection refused
redis: 2020/09/23 15:35:59 sentinel.go:471: sentinel: GetMasterAddrByName master="mymaster" failed: dial tcp :9127: connect: connection refused
redis: 2020/09/23 15:35:59 sentinel.go:471: sentinel: GetMasterAddrByName master="mymaster" failed: dial tcp :9128: connect: connection refused
```

vmihailenco:
Thanks!

vmihailenco:
>what are you your thoughts on adding a helper to proto.Reader for expected array lengths? Something like this:

Good idea, though we need to add some hint to the error message as to where the error comes from, e.g.

```go
func (r *Reader) ExpectArrayLen(caller string, n int) error {
	got, err := r.ReadArrayLen()
	if err != nil {
		return err
	}
	if n != got {
		return fmt.Errorf("redis: %s: unexpected array length; want %d, got %d", caller, n, got)
	}
	return nil
}
```

Filed https://github.com/go-redis/redis/issues/1508

meysampg:
@vmihailenco This is not a minor change (as it versioned from `8.2.1` to `8.2.2`). It breaks BC because it changes the `Cmdable` interface by adding `XInfoStream(ctx context.Context, key string) *XInfoStreamCmd` to it and all packages which are locked on `8.2.*` will be broken by updating to this version. 
Does `go-redis/redis` follow semver or it uses its rules? For the second case, is there any documentation about it? I checked https://redis.uptrace.dev/ and I didn't find anything about version policy.

icholy:
Same with #1505

PR #1511, opened 2020-09-27
Add unix socket connection by url stirng.

tty2:
There is only one way if you choose to connect to redis by url string: connect by tcp socket.
This PR offers the way to connect to unix socket using url string as well.

### DIFF
--- a/options.go
+++ b/options.go
@@ -29,6 +29,7 @@ type Limiter interface {
 	ReportResult(result error)
 }
 
+// Options keeps the settings to setup redis connection.
 type Options struct {
 	// The network type, either tcp or unix.
 	// Default is tcp.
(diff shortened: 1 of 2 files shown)


options.go
```diff
@@ -232,12 +247,43 @@ func ParseURL(redisURL string) (*Options, error) {
...
+	o := &Options{
+		Network: "unix",
+	}
+
+	if strings.TrimSpace(u.Path) == "" { // path is required with unix connection
+		return nil, errors.New("empty redis unix socket path")
+	}
+	o.Addr = u.Path
+
+	if u.User != nil {
```
vmihailenco, line 270: It looks like the rest of the func should be the same for `setupTCPConn` and `setupUnixConn`. Could you extract it to a separate func and use it from `setupTCPConn` and `setupUnixConn`?

tty2: Not really. The same `user/password` part only.
But `Network` field we have to setup from u.Path because unix socket is a path to a file.
And we have to get database number different way: from u.Query.
I can move `user/password` logic to separate func but we'll have to write the same amount of code on call this func and it looks like redundant func from the first glance.
WDYT?

tty2: It turned out that you were right.
I've moved `user/password` logic to separate func.
Take a look.
Thanks.

vmihailenco:
Thanks!

PR #1512, opened 2020-09-28
 add query telemetry span

git-hulk:
add query span which contains remote addr as a label, it can help us find out which Redis nodes were failing/slowing more quickly


vmihailenco:
Thanks for the PR but I have something simpler in https://github.com/go-redis/redis/pull/1513

PR #1516, opened 2020-10-04
Feature/config check socket file

tty2:
Instead of check if u.Path is not empty only, check if file we want too use in unix socket connection exists.
Fail fast, before trying to ping.

### DIFF
--- a/options.go
+++ b/options.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


vmihailenco:
I wonder why "Fail fast, before trying to ping"? This function parses the config - it is surprising that it also tried to connect / check if file exists etc.

tty2:
Yes, It's hard to follow the thoughts I want to say. 

I mean we call first `ParseURL, NewClient` and then `Ping`, and we fail before calling `NewClient` and `Ping` if file does not exist.
Sorry for the confusing message.

vmihailenco:
Closing since `ParseURL` should not be concerned whether socket exists or not. You still can `ParseURL` and check socket with `os.Stat` before calling `NewClient` - if you need it.

tty2:
> You still can `ParseURL` and check socket with `os.Stat` before calling `NewClient` - if you need it.

It makes sense. I guess you are right.

PR #1523, opened 2020-10-09
Fixes bufio buffer full error when reading from an array larger than 4096 bytes

eelcocramer:
I get this issue when retrieving large blobs from [redix](https://github.com/alash3al/redix) which is API compliant with redis.

I created a test that reproduces that error. It happens when reading from a byte array which length is larger than 4096.

See also #641

## Expected Behavior

The blob should be read in full.

## Current Behavior

An error is returned.

## Possible Solution

See, patch in this PR.

## Steps to Reproduce

See the added test case in this PR which (when ran without the fix added) would result in the following:

```bash
❯ go test internal/proto/reader.go internal/proto/reader_test.go 
--- FAIL: TestReader_ReadLine (0.00s)
    reader_test.go:51: Should be able to read the full buffer: bufio: buffer full
    reader_test.go:55: Values must be equal: ""
FAIL
FAIL	command-line-arguments	0.272s
FAIL
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,8 +1,5 @@
 blank_issues_enabled: false
 contact_links:
   - name: Discord
-    url: https://discord.gg/ZQJDR3R
+    url: https://discord.gg/rWtp5Aj
     about: Ask a question at Discord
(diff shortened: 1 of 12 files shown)


eelcocramer:
Closing this PR, will create another one. My clone is a mess :-(

PR #1524, opened 2020-10-09
Fixes bufio buffer full error when reading from an array larger than 4096 bytes

eelcocramer:
I get this issue when retrieving large blobs from [redix](https://github.com/alash3al/redix) which is API compliant with redis.

I created a test that reproduces that error. It happens when reading from a byte array which length is larger than 4096.

See also #641

## Expected Behavior

The blob should be read in full.

## Current Behavior

An error is returned.

## Possible Solution

See, patch in this PR.

## Steps to Reproduce

See the added test case in this PR which (when ran without the fix added) would result in the following:

```bash
❯ go test internal/proto/reader.go internal/proto/reader_test.go 
--- FAIL: TestReader_ReadLine (0.00s)
    reader_test.go:51: Should be able to read the full buffer: bufio: buffer full
    reader_test.go:55: Values must be equal: ""
FAIL
FAIL	command-line-arguments	0.272s
FAIL
```

### DIFF
--- a/internal/proto/reader.go
+++ b/internal/proto/reader.go
@@ -55,12 +55,9 @@ func (r *Reader) Reset(rd io.Reader) {
 }
 
 func (r *Reader) ReadLine() ([]byte, error) {
-	line, err := r.rd.ReadBytes('\n')
-	if err != nil && err != io.EOF {
+	line, err := r.readLine()
+	if err != nil {
 		return nil, err
-	}
-	if len(line) == 0 {
-		return nil, fmt.Errorf("redis: reply is empty")
 	}
 	if isNilReply(line) {
 		return nil, Nil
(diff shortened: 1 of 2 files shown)


internal/proto/reader.go
```diff
@@ -55,10 +55,13 @@ func (r *Reader) Reset(rd io.Reader) {
 }
 
 func (r *Reader) ReadLine() ([]byte, error) {
-	line, err := r.readLine()
-	if err != nil {
+	line, err := r.rd.ReadBytes('\n')
```
vmihailenco, line 58: Supposedly this allocates a new byte slice for each read line. If so, then it is a significant performance penalty for something fast like Redis client.

vmihailenco: Instead you probably need to modify `readLine` to properly handle `bufio.ErrBufferFull`.

eelcocramer: Thanks. I will look into it further.

eelcocramer:
Changed the implementation.

Original benchmark was:

```
goos: darwin
goarch: amd64
BenchmarkReader_ParseReply_Status-8   	17320110	       66.5 ns/op
BenchmarkReader_ParseReply_Int-8      	33233448	       33.9 ns/op
BenchmarkReader_ParseReply_Error-8    	16776595	       77.0 ns/op
BenchmarkReader_ParseReply_String-8   	13546156	       85.4 ns/op
BenchmarkReader_ParseReply_Slice-8    	3824382	      317 ns/op
PASS
ok  	command-line-arguments	8.036s
```

After the change the benchmark is a bit worse:

````
oos: darwin
goarch: amd64
BenchmarkReader_ParseReply_Status-8   	15872990	       79.6 ns/op
BenchmarkReader_ParseReply_Int-8      	31321155	       37.4 ns/op
BenchmarkReader_ParseReply_Error-8    	16365578	       84.5 ns/op
BenchmarkReader_ParseReply_String-8   	13537149	       92.0 ns/op
BenchmarkReader_ParseReply_Slice-8    	3660882	      331 ns/op
PASS
ok  	command-line-arguments	8.543s
```

PR #1529, opened 2020-10-11
Fix/issue 1384

### LINKED ISSUE #1384: Time.UnmarshalBinary: unsupported version
I guess this issue related with writer logic.

When I set new value, `time.Time` object will be saved as string, but in scan function `time.Time` handling by `UnmarshalBinary` method.

https://github.com/go-redis/redis/blob/master/internal/proto/writer.go#L99
https://github.com/go-redis/redis/blob/master/internal/proto/scan.go#L103

Thanks

tty2:
fix #1384 
Add time.Time type to Scan method.

PAY ATTENTION: there is a questionable part. I've added function level `nolint: gocyclo` cause with +1 branch limit is broken.
I can split the switch inside to 2 parts and move them to separate functions. And remove `nolint` of course. But IMHO the function will lose current readability.
WDYT?

### DIFF
--- a/internal/proto/scan.go
+++ b/internal/proto/scan.go
@@ -9,7 +9,7 @@ import (
 	"github.com/go-redis/redis/v8/internal/util"
 )
 
-// Scan parses bytes `b` to 'v' with appropriate type.
+// Scan parses bytes `b` to `v` with appropriate type.
 func Scan(b []byte, v interface{}) error {
 	switch v := v.(type) {
 	case nil:
--- a/internal/proto/scan.go
+++ b/internal/proto/scan.go
@@ -104,7 +104,7 @@ func Scan(b []byte, v interface{}) error {
 		return nil
 	case *time.Time:
 		var err error
-		*v, err = time.Parse(time.RFC3339, util.BytesToString(b))
+		*v, err = time.Parse(time.RFC3339Nano, util.BytesToString(b))
 		return err
 	case encoding.BinaryUnmarshaler:
 		return v.UnmarshalBinary(b)
--- a/internal/proto/scan.go
+++ b/internal/proto/scan.go
@@ -9,6 +9,8 @@ import (
 	"github.com/go-redis/redis/v8/internal/util"
 )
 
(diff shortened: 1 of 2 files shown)


internal/proto/scan.go
```diff
@@ -99,6 +102,10 @@ func Scan(b []byte, v interface{}) error {
 	case *bool:
 		*v = len(b) == 1 && b[0] == '1'
 		return nil
+	case *time.Time:
+		var err error
+		*v, err = time.Parse(time.RFC3339, util.BytesToString(b))
```
vmihailenco, line 107: Thanks, but why not `RFC3339Nano`?

internal/proto/scan_test.go
```diff
@@ -46,3 +50,37 @@ var _ = Describe("ScanSlice", func() {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if !toWrite.Equal(toScan) {
+			t.Fatal(errors.New("toWrite and toScan is not equal"))
+		}
+	})
+
+}
```
vmihailenco, line 86: This is a rather complicated. Instead you could add a test like:

```
tm := time.Now()
rdb.Set(ctx, "foo", tm, 0)

var tm2 time.Time
rdb.Get(ctx, "foo").Scan(&tm2)

// check tm2 equals tm
```

internal/proto/scan.go
tty2: My bad!
I don't know why but I thought that in Write RRC3339.

vmihailenco:
>And remove nolint of course. But IMHO the function will lose current readability. WDYT?

I am fine with nolint.

internal/proto/scan_test.go
tty2: I had to move package to proto_test. Otherwise I had circular imports.

vmihailenco:
Thanks, merging.

>I had to move package to proto_test. Otherwise I had circular imports.

I will move the test to redis_test.go.

Issue #1534, opened 2020-10-13
cannot find package "go.opentelemetry.io/otel/api/trace"


vmihailenco:
To fix `cannot find package "go.opentelemetry.io/otel/api/trace"`:

1. Make sure to initialize a Go [module](https://blog.golang.org/using-go-modules): `go mod init github.com/my/repo`

2. Make sure to use correct import path with *v8* in the end: `github.com/go-redis/redis/v8`

gonzalochief:
go get -u go.opentelemetry.io/otel/apils

I got this error when initializing the module and downloading all dependencies:
go get go.opentelemetry.io/otel/apils: module go.opentelemetry.io/otel@upgrade found (v0.13.0), but does not contain package go.opentelemetry.io/otel/apils

vmihailenco:
@vmihailenco you don't need to install go.opentelemetry.io/otel/apils. And you have typo in the name.

gonzalochief:
Thanks for answering in such a short time... I appreciate that
just to make my previous post clear, the typo was a copy paste issue for this post (my bad). The thing is that go says 
go get go.opentelemetry.io/otel/api: module go.opentelemetry.io/otel@upgrade found (v0.13.0), but does not contain package go.opentelemetry.io/otel/api
the same goes with the /otel/api/trace

vmihailenco:
I don't understand what you are trying to do. To install go-redis you don't need opentelemetry:

```bash
mkdir /tmp/redis-test
cd /tmp/redis-test
go mod init redis-test
go get github.com/go-redis/redis/v8
```

gonzalochief:
Sorry if ai'm not clear:
I've been using go-redis for months now. yesterday I tried to build my code and the compiler showed me the error I sent on my initial post. Go keeps trying to get the package go.opentelemetry.io/otel/api/trace. If I do a go build, the error is:
/home/gonzalo/golib/src/github.com/go-redis/redis/internal/util.go:10:2: cannot find package "go.opentelemetry.io/otel/api/trace" in any of:
/usr/local/go/src/go.opentelemetry.io/otel/api/trace (from $GOROOT)
/home/gonzalo/golib/src/go.opentelemetry.io/otel/api/trace (from $GOPATH)
/home/gonzalo/go/src/go.opentelemetry.io/otel/api/trace

Praesyx:
Yes, I also get this error when installing packages.
[...]

PR #1542, opened 2020-10-16
Update opentelemetry import - fix trace error

gonzalochief:
I fixed the import that was causing an error on clean (newly created) development environments.
The original code was importing a package that does not exists anymore as it was merged into the Otel package.
The code is working on my development machine (Linux). The fix should go to the master repo and branch so build actions do not fail on builds that pull the source directly from GitHub.

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -1147,7 +1147,7 @@ var _ = Describe("Commands", func() {
 
 		It("should Set with keepttl", func() {
 			// set with ttl
-			set := client.Set(ctx, "key", "hello", 5 * time.Second)
+			set := client.Set(ctx, "key", "hello", 5*time.Second)
 			Expect(set.Err()).NotTo(HaveOccurred())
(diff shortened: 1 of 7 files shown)


vmihailenco:
This PR is similar to https://github.com/go-redis/redis/pull/1541 . The proper fix is described at https://github.com/go-redis/redis/issues/1534

As you should know go-redis uses Go modules to manage dependencies. You *must* use Go modules or use some other mechanism to pin dependencies to some version. It is not optional.

This PR does not provide any reliable way to build go-redis and therefore does not fix anything. go-redis will be "broken" again as soon as there is a change in dev version of any other go-redis dependency. And it is completely fine - it is your workflow that is broken / unsupported.

At the same time this PR makes impossible to update go-redis for majority of users that properly use Go modules. Because after this change master branch requires unreleased and unstable OpenTelemetry version which is not meant for production.

PR #1545, opened 2020-10-21
Feature/gofumpt

tty2:
Up `golang-ci` version.

Bring back `gofumpt`. Looks like it was disabled because on 1.28.* `golang-ci` version, it behaves strange.

Small fix. Remove empty line in import (gofumpt lint-ed).

Disable new linter [nlreturn](https://github.com/ssgreg/nlreturn), which look a bit strange (only imho).


vmihailenco:
Thanks!

PR #1546, opened 2020-10-22
Add SetEX command

TwinProduction:
I'm aware that, in essence, the `Set` function can do everything that `SetEX` function can do, however, I'm using this library to test a Redis-compatible server implementation, and due to the fact that there's no way to send the `SETEX` command using go-redis/redis, I'm unable to fully test the aforementioned Redis-compatible server.

Assuming this reasoning is fine with you, one of my concerns is that the order in which `SETEX` takes an argument is the following:
```
SETEX key seconds value
```

But in order to be consistent with other functions, I opted to use the following function signature:
```
SetEX(ctx context.Context, key string, value interface{}, expiration time.Duration)
```
What do you think?

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -124,6 +124,7 @@ type Cmdable interface {
 	MSet(ctx context.Context, values ...interface{}) *StatusCmd
 	MSetNX(ctx context.Context, values ...interface{}) *BoolCmd
 	Set(ctx context.Context, key string, value interface{}, expiration time.Duration) *StatusCmd
+	SetEX(ctx context.Context, key string, value interface{}, expiration time.Duration) *StatusCmd
 	SetNX(ctx context.Context, key string, value interface{}, expiration time.Duration) *BoolCmd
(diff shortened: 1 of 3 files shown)


vmihailenco:
You can use `rdb.Do(ctx, "key", 1234, "value").Result()` to execute arbitrary command. It is less convenient but does not clutter the API for everyone...

TwinProduction:
I personally think that having an API that matches Redis' API shouldn't count as clutter, since ultimately, an API wrapper should cover the entire API, but since there is a way to do it, then I'm fine with the lack of a dedicated command for SETEX.

That said, although probably not very significant, I bet that the SETEX command is slightly faster than the SET command, as it has no optional arguments to take into consideration/process (NX, XX).

I actually just created a quick benchmark comparing the Set function with the SETEX function using client.Do, as you suggested), and while the difference is small, SETEX has an edge.

vmihailenco:
Thanks!

PR #1551, opened 2020-10-31
ADD: RetryConditionFunc to options which allows the user to decide when to retry

git-hulk:


vmihailenco:
What are you trying to achieve? What commands/errors do you want to retry?

git-hulk:
> What are you trying to achieve? What commands/errors do you want to retry?

This feature would allow us to auto-retry on some idempotent commands(read-only commands like get/mget/... etc) instead of retrying all commands when timeout

git-hulk:
@vmihailenco any thoughts about this PR?

vmihailenco:
@git-hulk the presence of `dummyWatchCommand` suggests that the API is not quite right. E.g. it is not clear what to do about pipelines (this API accepts single cmd).

Overall this looks like a rate problem without an elegant solution. Which it adds some maintenance burden. I don't think such PRs should be merged.

git-hulk:
okay, close the PR

PR #1556, opened 2020-11-09
Add LPOS command

Yanisky:
Because I started using this library recently, I'm not sure if the functions I've created align with the rest of the library's API.

I've decided to create two functions `LPos` and `LPosCount` that both accept Maxlen as the last parameter, I did not create any structs because it didn't seem necessary.

It did cross my mind to create 4 Commands (`LPos`, `LPosMaxlen`, `LPosCount` and `LPosCountMaxlen`) because maxlen is only used in very special cases.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,50 @@
+name: Go
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    services:
+      # Label used to access the service container
+      redis:
+        # Docker Hub image
+        image: redis
+        # Set health checks to wait until redis has started
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # Maps port 6379 on service container to the host
+          - 6379:6379
+    steps:
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.14
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Get dependencies
+        run: |
+          go get -v -t -d ./...
+          if [ -f Gopkg.toml ]; then
+              curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+              dep ensure
+          fi
+      - name: Install golangci
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.31.0
+      - name: Test
+        run: make
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,7 +2,7 @@ name: Go
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, feature/lpos-command ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,7 +2,7 @@ name: Go
 
 on:
   push:
-    branches: [ master, feature/lpos-command ]
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 
--- a/commands.go
+++ b/commands.go
@@ -168,6 +168,7 @@ type Cmdable interface {
 	LInsertAfter(ctx context.Context, key string, pivot, value interface{}) *IntCmd
 	LLen(ctx context.Context, key string) *IntCmd
 	LPop(ctx context.Context, key string) *StringCmd
+	LPos(ctx context.Context, key string, value string, rank int64) *IntCmd
 	LPush(ctx context.Context, key string, values ...interface{}) *IntCmd
 	LPushX(ctx context.Context, key string, values ...interface{}) *IntCmd
 	LRange(ctx context.Context, key string, start, stop int64) *StringSliceCmd
@@ -1184,6 +1185,12 @@ func (c cmdable) LLen(ctx context.Context, key string) *IntCmd {
 
 func (c cmdable) LPop(ctx context.Context, key string) *StringCmd {
 	cmd := NewStringCmd(ctx, "lpop", key)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+func (c cmdable) LPos(ctx context.Context, key string, value string, rank int64) *IntCmd {
+	cmd := NewIntCmd(ctx, "lpos", key, value, "rank", rank)
 	_ = c(ctx, cmd)
 	return cmd
 }
--- a/commands.go
+++ b/commands.go
@@ -168,8 +168,8 @@ type Cmdable interface {
 	LInsertAfter(ctx context.Context, key string, pivot, value interface{}) *IntCmd
 	LLen(ctx context.Context, key string) *IntCmd
 	LPop(ctx context.Context, key string) *StringCmd
-	LPos(ctx context.Context, key string, value string, args *LPosArgs) *IntCmd
-	LPosCount(ctx context.Context, key string, value string, count int64, args *LPosArgs) *IntSliceCmd
+	LPos(ctx context.Context, key string, value string, args LPosArgs) *IntCmd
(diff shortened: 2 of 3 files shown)


commands.go
```diff
@@ -1188,6 +1190,18 @@ func (c cmdable) LPop(ctx context.Context, key string) *StringCmd {
...
+	cmd := NewIntCmd(ctx, "lpos", key, value, "rank", rank, "maxlen", maxlen)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+func (c cmdable) LPosCount(ctx context.Context, key string, value string, rank int64, count int64, maxlen int64) *IntSliceCmd {
+	cmd := NewIntSliceCmd(ctx, "lpos", key, value, "rank", rank, "count", count, "maxlen", maxlen)
+	_ = c(ctx, cmd)
+	return cmd
+}
```
vmihailenco, line 1203: What if we replace these 2 commands with one:

```go
type LPosArgs struct {
    Rank int64
    Count int64
    MaxLen int64
}

LPos(ctx context.Context, key, value string, args LPosArgs)
```

Yanisky: To get you more context in case you missed it:
LPos with Count **NOT** set returns an integer or nil if not found.
LPos with Count set returns an array, empty array if none found.

As you suggested we can have only one command that returns an array. I do prefer having only one command than two.

Both Count and MaxLen default to 0 in Redis, but Rank's default is 1. Setting Rank to 0 results in a failure. 

We could be helpful and set good default ourselves.

From a user perspective and in the simplest case possible (get the first pos of an item) which of the following options would you prefer?
```go
// 1 -  We'll check that Rank is 0 and set it to 1
client.Lpos(context.Background(), "list", "item", LPosArgs{}) 

// 2 - We'll check for nil and do the defaults ourselves, Rank=1, Count=0, MaxLen = 0
client.Lpos(context.Background(), "list", "item", nil)

// 3 - User needs to always set rank to 1, just LPosArgs{} would fail.
client.Lpos(context.Background(), "list", "item", LPosArgs{ Rank: 1 }) 
```
I prefer option 2. We set good defaults and if the user wants more control they'll need to provide a struct with correct settings.

What are your thoughts?

.github/workflows/go.yml
```diff
@@ -0,0 +1,50 @@
+name: Go
```
Yanisky, line 1: @vmihailenco I'm developing on a Windows machine and this was the quickest way to run the tests. 

Do you want me to remove this github action file?

commands.go
vmihailenco: Thanks for the details! What about

```go
type LPosArgs struct {
    Rank int64
    MaxLen int64
}

LPos(ctx context.Context, key, value string, args LPosArgs) *IntCmd
LPosCount(ctx context.Context, key, value string, count int64, args LPosArgs) *IntSliceCmd
```

It is slightly more readable than just using positional arguments.

.github/workflows/go.yml
vmihailenco: I don't mind to try Github workflow. Do I have to do anything to enable it?

.github/workflows/go.yml
```diff
@@ -0,0 +1,50 @@
...
+
+on:
+  push:
+    branches: [ master, feature/lpos-command ]
+  pull_request:
+    branches: [ master ]
+
```
Yanisky, line 8: @vmihailenco You don't need to do anything to enable it.

The lines above is where you set the conditions that trigger this action. 

In this case I've set it to whenever there's a `git push` to branches `master` or `feature/lpos-command` . Or when there's a pull request created that targets `master`.

I will remove `feature/lpos-command` from it as it only made sense to my use case. 

I think these are sensible defaults (pull request to master or commits to master) and they are probably the same triggers as travisci.

vmihailenco: Sounds good - thanks.

commands.go
```diff
@@ -1190,14 +1190,28 @@ func (c cmdable) LPop(ctx context.Context, key string) *StringCmd {
...
+type LPosArgs struct {
+	Rank, MaxLen int64
+}
+
+func (c cmdable) LPos(ctx context.Context, key string, value string, args *LPosArgs) *IntCmd {
+	var rank, maxLen int64 = 1, 0
+	if args != nil {
+		rank = args.Rank
+		maxLen = args.MaxLen
+	}
```
Yanisky, line 1202: @vmihailenco What do you think of this? Setting sensible defaults.

Signature:
```go 
LPosCount(ctx context.Context, key string, value string, count int64, args *LPosArgs) 
```
Simplest use case looks like this:
```go
client.LPos(ctx, "list", "b", nil)
```
If the user needs more control he/she would need to pass in a LPosArgs struct

```go
client.LPos(ctx, "list", "b", &redis.LPosArgs{Rank: 2, MaxLen: 1})
```

.github/workflows/go.yml
Yanisky: `feature/lpos-command` is out! 👍

commands.go
vmihailenco: Perhaps it is convenient for your use case but it is counter-intuitive that 

```go
client.LPos(ctx, "list", "b", nil) // rank=1
client.LPos(ctx, "list", "b", &redis.LPosArgs{MaxLen: 1}) // rank=0
```

I believe it should be

```go
client.LPos(ctx, "list", "b", redis.LPosArgs{}) // lpos key value
client.LPos(ctx, "list", "b", redis.LPosArgs{MaxLen: 1}) // lpos key value maxlen 1
client.LPos(ctx, "list", "b", redis.LPosArgs{Rank: 1}) // lpos key value rank 1
client.LPos(ctx, "list", "b", redis.LPosArgs{Rank: 1, MaxLen: 1}) // lpos key value rank 1 maxlen 1
```

Yanisky: Yeah... you're right, that does look odd. 

My reasoning was that if you send `&redis.LPosArgs{MaxLen: 1}` you should know that you're also sending `Rank:0` because of how Go "zero values" work and 0 is an invalid value for `Rank`. Sending `nil` would state more clearly that you don't want to set neither `Rank` nor `MaxLen`. In Redis not setting `Rank` is the same as having `Rank=1`, in other words, we could say that the default value for `Rank` in redis is `1`.

In your example, what do you think it's the best option to make `client.LPos(ctx, "list", "b", redis.LPosArgs{})` not fail. Check for `Rank == 0` and set it to `1` or check for `Rank == 0` and not set it?

vmihailenco: If `Rank = 0` is not useful/allowed then I expect that `redis.LPosArgs{}` does not set the rank. That applies even if we use `*redis.LPosArgs` and allow nil. E.g. `nil` and `&redis.LPosArgs{}` should both omit rank if `Rank = 0` does not make sense.

>Sending nil would state more clearly that you don't want to set neither Rank nor MaxLen

I am fine with allowing `nil` if you are more comfortable with it. But I believe the API should behave the same since AFAIK `Rank=0` and `MaxLen=0` does not make any sense

```go
client.LPos(ctx, "list", "b", nil) // lpos key value
client.LPos(ctx, "list", "b", &redis.LPosArgs{}) // lpos key value
client.LPos(ctx, "list", "b", &redis.LPosArgs{MaxLen: 1}) // lpos key value maxlen 1
client.LPos(ctx, "list", "b", &redis.LPosArgs{Rank: 1}) // lpos key value rank 1
client.LPos(ctx, "list", "b", &redis.LPosArgs{Rank: 1, MaxLen: 1}) // lpos key value rank 1 maxlen 1
```

I am leaning towards `redis.LPosArgs{}` since `nil` value does not have special meaning here. But it is probably not important enough.

>check for Rank == 0 and not set it?

This!

Yanisky: Thanks! 

I'll try to sort it out later today.

commands.go
```diff
@@ -1188,6 +1190,37 @@ func (c cmdable) LPop(ctx context.Context, key string) *StringCmd {
...
+		args = append(args, "rank", a.Rank)
+	}
+	if a.MaxLen != 0 {
+		args = append(args, "maxlen", a.MaxLen)
+	}
+
+	cmd := NewIntCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
```
Yanisky, line 1209: @vmihailenco I've made the changes for `LPos` and `LPosCount` and removed the option for `nil`.

vmihailenco:
Thanks!

PR #1557, opened 2020-11-12
CLUSTER_REDIR_UNSTABLE error should be retry

wziww:
We should also try again when the error is CLUSTER_REDIR_UNSTABLE.
https://github.com/redis/redis/blob/380f6048e0bbc762f12fa50b57b73cf29049f967/src/cluster.c#L5838-L5840


vmihailenco:
Thanks

PR #1562, opened 2020-11-17
Update golangci-lint to v1.32

johejo:
Disable some new linters.
And removed unnecessary intenal util function.


vmihailenco:
Thanks

PR #1569, opened 2020-11-25
change the serialization interface naming to avoid repeated calls with gob serialization into an endless loop

Licoy:
Add a custom serialization interface to avoid conflicts with native gob serialization and cause repeated calls to gob, enter an endless loop, and eventually cause stack overflow.

### DIFF
--- a/internal/proto/scan.go
+++ b/internal/proto/scan.go
@@ -1,7 +1,6 @@
 package proto
 
 import (
-	"encoding"
 	"fmt"
 	"reflect"
 	"time"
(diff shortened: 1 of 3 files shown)


vmihailenco:
There is no conflict - `encoding.BinaryUnmarshaler` is not specific to gob.

Licoy:
@vmihailenco For example, using Gob for serialization in `BinaryMarshaler` will cause repeated calls

Licoy:
@vmihailenco Because gob will also determine whether there is implementation of `encoding.BinaryUnmarshaler`, if so, it will directly call the implementation method of `encoding.BinaryUnmarshaler`, which enters an infinite call

vmihailenco:
See https://msgpack.uptrace.dev/faq/#fatal-error-stack-overflow for a workaround.

Licoy:
@vmihailenco thanks

Issue #1568, opened 2020-11-24
Race condition in context cancellation for cluster commands


dbussink:
When updating to the latest go-redis driver, we hit a problem in our CI test suite where it showed up with a data race. We run our CI suite with the race detector enabled to catch problems. It triggered after updating the go-redis driver (v8.0.0-beta.7 to v8.4.0). It looks like it is a race bug inside the Redis driver and not in our code. We also were able to reproduce the race with a test script independent from our project.

## Expected Behavior

```go
func TestXReadRace(t *testing.T) {
	client := redis.NewIntegrationRedis(t)
	ctx, cancel := context.WithCancel(context.Background())

	for i := 0; i < 100; i++ {
		stream := fmt.Sprintf("foo.test.%d", i)
		go func() {
			_, err := client.XRead(ctx, &goredis.XReadArgs{
				Streams: []string{stream, "$"},
				Block:   1 * time.Second,
			}).Result()
			if err != context.Canceled {
				panic(err)
			}
		}()
	}

	time.Sleep(10 * time.Millisecond)
	cancel()
}
```

We'd expect that this operation wouldn't trigger the race detector, but that here `err` returned from `XRead` has the message that the context was cancelled. 

## Current Behavior

Right now sometimes the race detector triggers when this happens. It's not consistent, so multiple runs of the test script might be needed and the number of goroutines spawned might also need to be tweaked to reproduce this on a specific machine (too low didn't trigger for me on my laptop, but too high didn't trigger either). 

```
dirkjan@dbussink ~/code/go-redis-race $ go test -v -count=1 -race
=== RUN   TestXReadRace
--- PASS: TestXReadRace (0.03s)
PASS
==================
WARNING: DATA RACE
Write at 0x00c000249008 by goroutine 212:
  github.com/go-redis/redis/v8.(*ClusterSlotsCmd).readReply.func1()
      /Users/dirkjan/go/pkg/mod/github.com/go-redis/redis/v8@v8.4.0/command.go:1796 +0xb0
  github.com/go-redis/redis/v8/internal/proto.(*Reader).ReadArrayReply()
      /Users/dirkjan/go/pkg/mod/github.com/go-redis/redis/v8@v8.4.0/internal/proto/reader.go:190 +0x2c1
  github.com/go-redis/redis/v8.(*ClusterSlotsCmd).readReply()
      /Users/dirkjan/go/pkg/mod/github.com/go-redis/redis/v8@v8.4.0/command.go:1795 +0x64
[...]

PR #1573, opened 2020-12-06
Add wrapper for net.Conn to support Context cancellation properly

roy2220:
Hi, I just encountered the same issue as #1568, and after the review, I don't think this issue is resolved by #1572.

Let's figure out what's going on:

1. Since the Read/Write on `net.Conn` doesn't support `Context`, and there are no methods provided such as SetReadContext/SetWriteContext, `net.Conn` is unaware of the cancellation of `Context`.

2. In order to avoid the sustained blocking on `net.Conn` IO after the cancellation of `Context`, we spawn a new goroutine for `net.Conn` IO, and wait for the cancellation in the current goroutine ([code](https://github.com/go-redis/redis/blob/02ccf05ed073709d0c4a3b933a50b13a2b87c0b6/redis.go#L317)).

3. Once the current goroutine receives the cancellation of `Context`, we let the newly spawn goroutine alone, save the error to `cmd` and return it immediately ([code](https://github.com/go-redis/redis/blob/02ccf05ed073709d0c4a3b933a50b13a2b87c0b6/redis.go#L60)). 

4. Since the process of `cmd` is completed, the user try **retrieving the result** and error from `cmd` at once by `cmd.Result()` ([code](https://github.com/go-redis/redis/blob/02ccf05ed073709d0c4a3b933a50b13a2b87c0b6/command.go#L206))

5. At this timing, the newly spawn goroutine just finished the IO, and try **saving the result** to  `cmd` simultaneously ([code](https://github.com/go-redis/redis/blob/02ccf05ed073709d0c4a3b933a50b13a2b87c0b6/command.go#L318)). 

6. The data race takes place.

The root cause of this bug is that `net.Conn` doesn't support `Context` and we work it around, so things get complicated.

Finally, I made this PR implementing a wrapper for `net.Conn` to simulate  SetReadContext/SetWriteContext properly, feel free to edit or make any changes.

### DIFF
--- a/error.go
+++ b/error.go
@@ -68,7 +68,7 @@ func isBadConn(err error, allowTimeout bool) bool {
 	switch err {
 	case nil:
 		return false
-	case context.Canceled, context.DeadlineExceeded:
+	case context.Canceled:
 		return true
 	}
 
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -16,7 +16,7 @@ var noDeadline = time.Time{}
 
 type Conn struct {
 	usedAt  int64 // atomic
-	netConn net.Conn
+	netConn *netConn
 
 	rd *proto.Reader
 	bw *bufio.Writer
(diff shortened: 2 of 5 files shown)


vmihailenco:
>At this timing, the newly spawn goroutine just finished the IO, and try saving the result to cmd simultaneously (code).

@roy2220 #1572 removes this step by closing the connection so there can't be any pending IO. Do you think this is still a problem?

>I don't think this issue is resolved by #1572.

Would be great if you can provide a reproducer.

roy2220:
> > At this timing, the newly spawn goroutine just finished the IO, and try saving the result to cmd simultaneously (code).
> 
> @roy2220 #1572 removes this step by closing the connection so there can't be any pending IO. Do you think this is still a problem?
> 
> > I don't think this issue is resolved by #1572.
> 
> Would be great if you can provide a reproducer.

Consider the tricky timing, while closing the connection, the newly spawn goroutine is on the way to save the result, closing the connection can't stop it.

vmihailenco:
>closing the connection can't stop it.

Indeed. I've sent https://github.com/go-redis/redis/pull/1574 which waits for the goroutine to finish.

PR #1574, opened 2020-12-06
Wait for the goroutine to finish

vmihailenco:
Fixes https://github.com/go-redis/redis/pull/1573

@roy2220 PTAL

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -316,6 +316,8 @@ func (c *baseClient) withConn(
 		select {
 		case <-done:
 			_ = cn.Close()
+			// Wait for the goroutine to finish and send something.
+			<-errc
 
 			err = ctx.Err()
 			return err


roy2220:
@vmihailenco Looks good to me.

How do you consider my solution that [implements SetReadContext/SetWriteContext on `net.Conn`](https://github.com/go-redis/redis/pull/1573/files)? I think it could fix things in a straight way, I have written tests for it, although it failed other tests, I will fix them all if you agree on it.

vmihailenco:
>How do you consider my solution that

It looks good but complex :) That is ~200 lines of synchronization code that is rather hard to read and maintain. This fix is much smaller.

roy2220:
> That is ~200 lines of synchronization code that is rather hard to read and maintain
OK, it's a bit complicated indeed.

PR #1579, opened 2020-12-15
Update README on branch v8

lee0ray:


vmihailenco:
Thanks for the effort, but v8 branch was merged into the master some time ago. We need to remove it instead of updating the readme.

PR #1585, opened 2020-12-20
Added test setup document in README.md

2014BDuck:
Hi maintainers.

I am new to golang. And I tried to figure out how to run the test for go-redis project. It seems there is no document yet for this. So I would like to add few lines to describe how I run the test suite.

Please correct me if it's wrong. Thanks for your patience.


vmihailenco:
Thanks

PR #1589, opened 2020-12-28
add disable/-1 comment to max retries

williamkinney:
Since 0 will not disable max retries (defaults to 3), add an explicit comment regarding how to disable.


vmihailenco:
Thanks

Issue #869, opened 2018-09-20
Should "manual cluster" (ClusterSlots) require redis to run in cluster mode?


krzysztofdrys:
I am confused about the `ClusterSlots` property in `NewClusterClient`. The documentation says that:
```
	// Optional function that returns cluster slots information.
	// It is useful to manually create cluster of standalone Redis servers
	// and load-balance read/write operations between master and slaves.
	// It can use service like ZooKeeper to maintain configuration information
	// and Cluster.ReloadState to manually trigger state reloading.
```
I assume that it means, that target redis does not need to run in a cluster mode. So I decided to give it a try:
```
import (
	"github.com/go-redis/redis"
)

func main() {
	clusterSlots := func() ([]redis.ClusterSlot, error) {
		slots := []redis.ClusterSlot{
			// First node with 1 master and 1 slave.
			{
				Start: 0,
				End:   16383,
				Nodes: []redis.ClusterNode{{
					Addr: ":16380", // master
				}, {
					Addr: ":26380", // 1st slave
				}},
			},
		}
		return slots, nil
	}

	redisdb := redis.NewClusterClient(&redis.ClusterOptions{
		ClusterSlots:  clusterSlots,
		RouteRandomly: true,
	})
	err := redisdb.ReloadState()
	if err != nil {
		panic(err)
	}
	if err := redisdb.Ping(); err != nil {
		panic(err)
	}
}
```
This exists with:
```
panic: ping: ERR This instance has cluster support disabled

goroutine 1 [running]:
main.main()
	/Users/krzysztofdrys/go/src/tmp/rediscli/main.go:33 +0xce

Process finished with exit code 2
```

What am I doing wrong? Am reading the documentation in a wrong way? Or am I using `ClusterSlots` in a wrong way?

krzysztofdrys:
After some debugging, I think the problem is here:
https://github.com/go-redis/redis/blob/master/cluster.go#L130

I think that this line should read:
```
		readOnly:        opt.ReadOnly && opt.ClusterSlots != nil
```
So `readonly` should not be send, in "manual cluster" mode. 

Though, I am not sure if this is a bug or an intended behaviour. :)

vmihailenco:
Thanks for the report - this should be fixed in master.

krzysztofdrys:
Thanks for the prompt reply!

PR #1591, opened 2021-01-06
ClusterSlots Config and ReadOnly

keitwb:
You have to set the ReadOnly flag on the ClusterOptions to make it fall back to
non-master nodes on reads. For non-clustered situations where you have multiple
nodes in a single ClusterSlot, it made the non-master node useless because it would
never route requests to it and if you set ReadOnly=true it would issue a READONLY
command to it, which would fail.

See comment in code for more details.

I believe this is a better fix for https://github.com/go-redis/redis/issues/869 than what was previously done in that it allows better use of redis instances.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -86,7 +86,7 @@ func (opt *ClusterOptions) init() {
 		opt.MaxRedirects = 3
 	}
 
-	if (opt.RouteByLatency || opt.RouteRandomly) && opt.ClusterSlots == nil {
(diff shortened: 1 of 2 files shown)


keitwb:
I don't think the test failure is due to my change. I cannot reproduce locally.  Can someone confirm?

vmihailenco:
Thanks. Glad to see someone is using this functionality :)

PR #1595, opened 2021-01-08
Add ScanType command to Scan with 'type' option

forain:
As of version 6.0 you can use this 'type' option to ask SCAN to only
return objects that match a given type, allowing you to
iterate through the database looking for keys of a specific type.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -141,6 +141,7 @@ type Cmdable interface {
 	BitField(ctx context.Context, key string, args ...interface{}) *IntSliceCmd
 
 	Scan(ctx context.Context, cursor uint64, match string, count int64) *ScanCmd
+	ScanType(ctx context.Context, cursor uint64, match string, count int64, keyType string) *ScanCmd
 	SScan(ctx context.Context, key string, cursor uint64, match string, count int64) *ScanCmd
 	HScan(ctx context.Context, key string, cursor uint64, match string, count int64) *ScanCmd
 	ZScan(ctx context.Context, key string, cursor uint64, match string, count int64) *ScanCmd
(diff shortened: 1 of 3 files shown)


rbg:
would it be more friendly to perhaps add a TScan command ?

forain:
> would it be more friendly to perhaps add a TScan command ?

Good idea! Is "TScan" the best name for it? We already have SScan for REDIS sscan command and HScan for REDIS hscan command. Having TScan for REDIS "scan command with type option". I'm afraid it might cause confusion.

rbg:
yeah -- TScan was just a suggestion, @vmihailenco i am sure will have a great suggestion :)

rbg:
your change as it stands breaks the build so will need some lov; whihc is why i was suggesting an alternative command..

forain:
> your change as it stands breaks the build so will need some lov; which is why i was suggesting an alternative command..

Fixed the build by updating tests accordingly. Let's wait for @vmihailenco to check if it'd better adding it to existing Scan as I did or creating a new one with a different name (TScan? TypeScan? ScanWithType?).

blaquezzz (approved):

commands.go
```diff
@@ -140,7 +140,7 @@ type Cmdable interface {
 	BitPos(ctx context.Context, key string, bit int64, pos ...int64) *IntCmd
 	BitField(ctx context.Context, key string, args ...interface{}) *IntSliceCmd
 
-	Scan(ctx context.Context, cursor uint64, match string, count int64) *ScanCmd
+	Scan(ctx context.Context, cursor uint64, match string, keyType string, count int64) *ScanCmd
```
vmihailenco, line 143: It looks like in Redis the `type` is the last arg. Besides this is a breaking change.

Can we instead add `ScanType(ctx, cursor, match, count, keyType)`? That is backwards compatible and more explicit.

rbg: That would be a better solution

forain:
@rbg @vmihailenco I followed your instructions and reimplemented it as a new "ScanType" command. Also added command and example tests. Build tests passed. Let me know if I missed something!

vmihailenco:
:+1: thanks

PR #1601, opened 2021-01-16
Fix broken otel example module, tidy

zikaeroh:
The otel example module has been receiving module updates with the rest of `go-redis`, but doesn't actually build as the API for `otel` has changed. Update it to have the same behavior as previously, but with the current packages. Also fix the module name and add a missing replacement directive, all of which make the Go tooling happy again.

Additionally, run `go mod tidy` on the modules in the repo. It's best to keep these tidy, as the contents of `go.sum` are dumped into modules that depend on `go-redis` (which means big redundant blocks of sums in other projects).

It'd be nice for a CI check to verify everything builds and that the modules are kept tidy, but that's another story.


vmihailenco:
Thanks

PR #1615, opened 2021-01-21
Ensure Watch closes the transaction on panic

adriansmares:
This PR ensures that `Watch` will close the transaction if the user provided function `panic`s. 

Panics which occur inside `fn` but are recover outside of it render the client unusable (the connection pool never recovers since `tx.Close` is never called).


vmihailenco:
Thanks

Issue #1603, opened 2021-01-18
PR propsal: Scanning results and maps to struct


knadh:
Hi @vmihailenco,

Here's a [simple solution](https://gist.github.com/knadh/63862e5713c7ed296e0a14b888ca86aa) that scans the results of type `[]interface{}` from `NewSliceCmd` and `NewStringSliceCmd` to a struct using reflection. The comment on the function shows example usage. Useful for scanning the results of HMGET, HGETALL etc. directly into a struct.

What do you think of merging this feature into the lib in one of the following ways? If yes, I can send a PR.
- As `goredis.ScanToStruct(vals, target)`
- As `utils.ScanToStruct(vals, target)`
- As `SliceCmd.ScanToStruct(target)`

Related: https://github.com/go-redis/redis/issues/672 https://github.com/go-redis/redis/issues/348 https://github.com/go-redis/redis/issues/970 https://github.com/go-redis/redis/issues/16

vmihailenco:
Hi @knadh,

Two obvious things that are missing are tests and docs/examples. I won't insist much on docs, but tests are mandatory.

And I guess code requires more work too and I don't have much time to spend on this. I would say that this issue requires 10-20 hours of work to finish. Do you have that time? :)

knadh:
@vmihailenco the gist is just a proof of concept. Of course, it'll need tests + examples before it can go into the lib.

> And I guess code requires more work too

Could you elaborate on what exactly?

vmihailenco:
> Could you elaborate on what exactly?

Take this only as suggestions since I have not spend much time on this but I have some experience with https://github.com/vmihailenco/msgpack.

I think it makes sense to start with with [decoder](https://github.com/vmihailenco/msgpack/blob/v5/decode_value.go#L19-L45) func. So for each type you have a func that scans a string into a value:

```go
type decoderFunc func(v reflect.Value, s string) error

var valueDecoders = []decoderFunc{
	reflect.Bool:        decodeBoolValue,
	reflect.Int:           decodeInt64Value,
	..
}
```

Then for each struct you can construct a map of decoders once and reuse it:

```go
type field struct {
    name     string // field name
    decoder decoderFunc
}

type structInfo struct {
    typ reflect.Type // struct type
    m   map[string]*field
}

func (s *structInfo) Decode(v reflect.Value, key, value string) error {
    return s.m[key].decoder(v, value)
}
```

Then your function will look like this:

```go
[...]

PR #1631, opened 2021-01-28
Add redis.Scan() to scan results from redis maps into structs.

knadh:
The package uses reflection to decode default types (int, string etc.) from Redis map results (key-value pair sequences) into
struct fields where the fields are matched to Redis keys by tags.

Similar to how `encoding/json` allows custom decoders using`UnmarshalJSON()`, the package supports decoding of arbitrary
types into struct fields by defining a `Decode(string) error`function on types.

The field/type spec of every struct that's passed to Scan() is cached in the package so that subsequent scans avoid iteration
and reflection of the struct's fields.

Issue: https://github.com/go-redis/redis/issues/1603

### DIFF
--- a/command.go
+++ b/command.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/go-redis/redis/v8/internal"
+	"github.com/go-redis/redis/v8/internal/hscan"
 	"github.com/go-redis/redis/v8/internal/proto"
 	"github.com/go-redis/redis/v8/internal/util"
 )
@@ -369,6 +370,13 @@ func (cmd *SliceCmd) Result() ([]interface{}, error) {
 
 func (cmd *SliceCmd) String() string {
 	return cmdString(cmd, cmd.val)
+}
+
+// Scan scans the results from a key-value Redis map result set ([]interface{})
+// like HMGET and HGETALL to a destination struct.
+// The Redis keys are matched to the struct's field with the `redis` tag.
+func (cmd *SliceCmd) Scan(val interface{}) error {
+	return hscan.Scan(cmd.val, val)
 }
 
 func (cmd *SliceCmd) readReply(rd *proto.Reader) error {
--- a/command.go
+++ b/command.go
@@ -374,7 +374,7 @@ func (cmd *SliceCmd) String() string {
 
 // Scan scans the results from the map into a destination struct. The map keys
 // are matched in the Redis struct fields by the `redis:"field"` tag.
-func (cmd *SliceCmd) Scan(dest interface{}) error {
+func (cmd *SliceCmd) Scan(dst interface{}) error {
 	if cmd.err != nil {
 		return cmd.err
 	}
@@ -389,7 +389,7 @@ func (cmd *SliceCmd) Scan(dest interface{}) error {
 		args = cmd.args[1:]
 	}
 
-	return hscan.Scan(args, cmd.val, dest)
+	return hscan.Scan(dst, args, cmd.val)
 }
 
 func (cmd *SliceCmd) readReply(rd *proto.Reader) error {
@@ -940,23 +940,23 @@ func (cmd *StringStringMapCmd) String() string {
 
 // Scan scans the results from the map into a destination struct. The map keys
 // are matched in the Redis struct fields by the `redis:"field"` tag.
-func (cmd *StringStringMapCmd) Scan(dest interface{}) error {
+func (cmd *StringStringMapCmd) Scan(dst interface{}) error {
 	if cmd.err != nil {
 		return cmd.err
 	}
 
-	// Pass the list of keys and values. Skip the first to args (command, key),
-	// eg: HGETALL map.
-	var (
-		keys = make([]interface{}, 0, len(cmd.val))
-		vals = make([]interface{}, 0, len(cmd.val))
-	)
-	for k, v := range cmd.val {
-		keys = append(keys, k)
-		vals = append(vals, v)
+	strct, err := hscan.Struct(dst)
+	if err != nil {
+		return err
 	}
 
-	return hscan.Scan(keys, vals, dest)
+	for k, v := range cmd.val {
+		if err := strct.Scan(k, v); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func (cmd *StringStringMapCmd) readReply(rd *proto.Reader) error {
--- a/command.go
+++ b/command.go
@@ -375,9 +375,21 @@ func (cmd *SliceCmd) String() string {
 // Scan scans the results from the map into a destination struct. The map keys
 // are matched in the Redis struct fields by the `redis:"field"` tag.
 func (cmd *SliceCmd) Scan(dest interface{}) error {
-	// Pass the list of keys and values. Skip the first to args (command, key),
-	// eg: HMGET map.
-	return hscan.Scan(cmd.args[2:], cmd.val, dest)
+	if cmd.err != nil {
+		return cmd.err
+	}
+
+	// Pass the list of keys and values.
+	// Skip the first two args for: HMGET key
+	var args []interface{}
+	if cmd.args[0] == "hmget" {
+		args = cmd.args[2:]
+	} else {
+		// Otherwise, it's: MGET field field ...
+		args = cmd.args[1:]
+	}
+
+	return hscan.Scan(args, cmd.val, dest)
 }
 
 func (cmd *SliceCmd) readReply(rd *proto.Reader) error {
@@ -929,6 +941,10 @@ func (cmd *StringStringMapCmd) String() string {
 // Scan scans the results from the map into a destination struct. The map keys
 // are matched in the Redis struct fields by the `redis:"field"` tag.
 func (cmd *StringStringMapCmd) Scan(dest interface{}) error {
+	if cmd.err != nil {
+		return cmd.err
+	}
+
 	// Pass the list of keys and values. Skip the first to args (command, key),
 	// eg: HGETALL map.
 	var (
--- a/command.go
+++ b/command.go
@@ -372,11 +372,12 @@ func (cmd *SliceCmd) String() string {
 	return cmdString(cmd, cmd.val)
 }
 
-// Scan scans the results from a key-value Redis map result set ([]interface{})
-// like HMGET and HGETALL to a destination struct.
-// The Redis keys are matched to the struct's field with the `redis` tag.
-func (cmd *SliceCmd) Scan(val interface{}) error {
-	return hscan.Scan(cmd.val, val)
+// Scan scans the results from the map into a destination struct. The map keys
+// are matched in the Redis struct fields by the `redis:"field"` tag.
+func (cmd *SliceCmd) Scan(dest interface{}) error {
+	// Pass the list of keys and values. Skip the first to args (command, key),
+	// eg: HMGET map.
+	return hscan.Scan(cmd.args[2:], cmd.val, dest)
 }
 
 func (cmd *SliceCmd) readReply(rd *proto.Reader) error {
@@ -923,6 +924,23 @@ func (cmd *StringStringMapCmd) Result() (map[string]string, error) {
 
 func (cmd *StringStringMapCmd) String() string {
 	return cmdString(cmd, cmd.val)
+}
+
+// Scan scans the results from the map into a destination struct. The map keys
+// are matched in the Redis struct fields by the `redis:"field"` tag.
+func (cmd *StringStringMapCmd) Scan(dest interface{}) error {
+	// Pass the list of keys and values. Skip the first to args (command, key),
+	// eg: HGETALL map.
+	var (
+		keys = make([]interface{}, 0, len(cmd.val))
+		vals = make([]interface{}, 0, len(cmd.val))
+	)
+	for k, v := range cmd.val {
+		keys = append(keys, k)
+		vals = append(vals, v)
+	}
+
+	return hscan.Scan(keys, vals, dest)
 }
 
 func (cmd *StringStringMapCmd) readReply(rd *proto.Reader) error {
--- a/internal/hscan/hscan_test.go
+++ b/internal/hscan/hscan_test.go
@@ -19,6 +19,8 @@ type data struct {
 	Bool   bool    `redis:"bool"`
 }
 
+type i []interface{}
+
 func TestGinkgoSuite(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "hscan")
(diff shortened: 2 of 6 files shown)


decoder.go
```diff
@@ -0,0 +1,211 @@
...
+	"strings"
+	"sync"
+)
+
+// Decoder represents a type with a custom Decode() function that can
+// take a string and parse it to set a desired value on a struct field,
+// similar to how encoding/json implements UnmarshalJSON().
+type Decoder interface {
+	Decode(string) error
+}
```
vmihailenco, line 16: I suggest to drop this interface for now for simplicity. We can add something simimlar later.

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+
+// defDecoder represents decoding functions for default built-in types.
+type defDecoder func(reflect.Value, string) error
```
vmihailenco, line 19: defDecoder -> decoderFunc

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+
+// structField represents a single field in a target struct.
+type structField struct {
+	index int
+
+	// Decoder.Decode() or defDecoder()
+	fn interface{}
```
vmihailenco, line 26: interface{} -> decoderFunc. To support `Decoder` interface we can later add:

```go
func decodeInterface(v reflect.Value, s string) error {
    return v.Interface().(Decoder).Decode(s)
}
```

But let's not worry about that for now.

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+var (
+	defDecoders = map[reflect.Kind]defDecoder{
```
vmihailenco, line 30: defDecoders -> decoders
map[reflect.Kind]defDecoder -> []defDecoder

Also these types are usually called standard / buit-in / primitives, not default.

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+		reflect.Slice:   decodeStringSlice,
+		reflect.String:  decodeString,
+	}
+
+	// Global map of struct field specs that is populated once for every new
+	// struct type. This caches the field types and the corresponding decoder
+	// functions to avoid iterating through struct fields on every scan.
+	// {type: { structFieldTag:  }}
+	structSpecs = make(map[reflect.Type]map[string]*structField)
+	mut         sync.RWMutex
```
vmihailenco, line 53: There is a convention that mutex is defined before the vars it protects. I would also create a few helper types to avoid having such big map type, e.g.

```go
type StructMap struct {
    mu sync.Mutex
    m   map[reflect.Type]StructFields
}

type StructFields struct {
    m map[string]*StructField
}

type StructField struct { ... }

var structMap = NewStructMap()
```

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+// Scan scans the results from a key-value Redis map result set ([]interface{})
+// to a destination struct. The Redis keys are matched to the struct's field
+// tags defined in the struct indifcated by fieldTag.
+// eg: fieldTag=redis where Name string `redis:"name"`
+func Scan(vals []interface{}, dest interface{}, fieldTag string) error {
```
vmihailenco, line 60: Do we need to support custom `fieldTag` here? I would drop it and use `redis`, because otherwise you need a separate struct cache for each different `fieldTag`.

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+	if len(vals)%2 != 0 {
+		return errors.New("args should have an even number of items (key-val)")
+	}
+
+	// The destination to scan into should be a struct pointer.
+	ob := reflect.ValueOf(dest)
```
vmihailenco, line 66: This is minor, but this is usually called just `v` (ob -> v).

I usually associate `ob` with observer. `obj` with object. And `v` or `value` with `reflect.Value`.

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+	if ob.Kind() != reflect.Ptr || ob.IsNil() {
+		return errors.New("dest must be a non-nil pointer to a struct")
```
vmihailenco, line 68: An error message that most Go developers recognize would be `fmt.Errorf("redis.Scan(non-pointer %T)", dest)`

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	ob = ob.Elem()
+
+	if ob.Kind() != reflect.Struct {
+		return errors.New("dest must be a non-nil pointer to a struct")
```
vmihailenco, line 73: `fmt.Errorf("redis.Scan(unsupported %T)", dest)`
or
`fmt.Errorf("redis.Scan(non-struct %T)", dest)`

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+			continue
+		}
+
+		tag = strings.Split(tag, ",")[0]
+		if tag == "" {
+			continue
+		}
+
+		// If the field has a custom Decode() (implements Decoder) defined on it, use that.
+		if df, ok := f.Addr().Interface().(Decoder); ok {
```
vmihailenco, line 149: Let's drop this for now

decoder.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+
+	return out
+}
+
+func decodeBool(f reflect.Value, s string) error {
+	b, err := strconv.ParseBool(s)
+	if err != nil {
+		return errors.New("expected boolean")
```
vmihailenco, line 167: Please don't hide the original error here and below. It should be just `return err`.

vmihailenco:
:+1: I've left some comments - hopefully it is not too much :) Also please set "Allow edits from maintainers" on the PR so I can add commits...

Let's move the code to `internal/hscan` package for now so all types are private. And the public API could look like `err := rdb.HGetAll(ctx, "key").Scan(&strct)`.

knadh:
Incorporated all the changes you suggested and moved the package to `internal/hscan`. 

PS: Did a force push because it was mostly hygiene fixes.

internal/hscan/structmap.go
```diff
@@ -0,0 +1,87 @@
...
+		f := ob.Field(i)
+		if !f.IsValid() || !f.CanSet() {
+			continue
+		}
+
+		tag := ob.Type().Field(i).Tag.Get(fieldTag)
+		if tag == "" || tag == "-" {
+			continue
+		}
+
```
monkey92t, line 76: Will there be an accident? reflect.StructTag.Get documentation: "If the tag does not have the conventional format, the value returned by Get is unspecified".
so...should we think that tags are all conventional format?

knadh: Should be fine. If the tag is empty or if it's malformed (hence, not `redis`), the field will simply be ignored.

monkey92t: i think reflect.StructTag.Get() means that if it is not "conventional format", it will return an ambiguous value, not empty.

vmihailenco:
@knadh I've pushed few small changes, but this looks good already.

Could you add `SliceCmd.Scan` and a test (or two :)?

internal/hscan/structmap.go
```diff
@@ -0,0 +1,82 @@
...
+		out = newStructFields()
+	)
+
+	for i := 0; i < num; i++ {
+		f := ob.Field(i)
+		if !f.IsValid() || !f.CanSet() {
+			continue
+		}
+
+		tag := ob.Type().Field(i).Tag.Get(fieldTag)
```
vmihailenco, line 67: This could be slightly simpler - `f.Type().Tag...`

knadh:
> Could you add `SliceCmd.Scan` and a test (or two :)?

Yep, will push this.

internal/hscan/structmap.go
vmihailenco: Out of experience with msgpack/go-pg - Go just returns some string. Nothing bad happens. I would not put too many efforts into validating that string... at least not at this stage :)

knadh:
> Could you add SliceCmd.Scan and a test (or two :)?

Added here: https://github.com/go-redis/redis/pull/1631/commits/a4144ea98eb00abe3eb5f78621d93c1c8c229493

internal/hscan/structmap.go
```diff
@@ -0,0 +1,79 @@
...
+		out = newStructFields()
+	)
+
+	for i := 0; i < num; i++ {
+		f := t.Field(i)
+
+		tag := f.Tag.Get(fieldTag)
+		if tag == "" || tag == "-" {
+			continue
+		}
```
vmihailenco, line 67: It looks we are ignoring fields that don't have `redis:"something"` tag. Is that intentional?

Usually we are supposed to generate default name from Go field name and only ignore the field when `redis:"-"`. What do you think?

https://pkg.go.dev/github.com/codemodus/kace could be used to generate default name.

vmihailenco:
Could you add one end-to-end test or maybe even an example? That will help people to get started

```go
rdb.HSet(..)
rdb.HSet(...)

rdb.HGetAll(ctx, "key").Scan(&dst)
```

internal/hscan/structmap.go
monkey92t: this parser is too simple....can only do the most basic work,  it needs to continue to improve its functions,
i can continue to improve it this week,,,if needed

knadh: > It looks we are ignoring fields that don't have redis:"something" tag. Is that intentional?

Yeah. Explicitly defining what's to be scanned is better (fewer gotchas) in my opinion. Opting in rather than opting out with `"-"`.

> this parser is too simple....can only do the most basic work, it needs to continue to improve its functions,

It only handles primitive types. The right way to extend it with the `Decode()` interface which was in the original PR. I hope @vmihailenco has plans to introduce that functionality soon!

knadh:
I just realised, `StringStringMapCmd` will also need to support `Scan()` for `HGetAll()` to work. Okay to add?

command.go
```diff
@@ -371,6 +372,14 @@ func (cmd *SliceCmd) String() string {
 	return cmdString(cmd, cmd.val)
 }
 
+// Scan scans the results from the map into a destination struct. The map keys
+// are matched in the Redis struct fields by the `redis:"field"` tag.
+func (cmd *SliceCmd) Scan(dest interface{}) error {
+	// Pass the list of keys and values. Skip the first to args (command, key),
+	// eg: HMGET map.
+	return hscan.Scan(cmd.args[2:], cmd.val, dest)
```
monkey92t, line 380: @knadh here you ignore cmd.err (and *StringStringMapCmd), 
mget command will encounter an error, it should run normally, here is: cmd.args[?:]
mget: mget key1 key2 key3 ...
hmget hmget key field1 field2 field3

knadh: Good spot @monkey92t. Fixed here https://github.com/knadh/redis/commit/077e145f6a26a8a441fa45d2dadb6526935beca3

vmihailenco:
@knadh I've pushed a change with `StructValue` so `StringStringMapCmd.Scan`  can be slightly more efficient. I think this PR is ready to be merged if you are happy with that change. It will be part of v8 but marked unstable so we can change it based on the feedback.

internal/hscan/hscan_test.go
```diff
@@ -0,0 +1,117 @@
...
+		// the struct spec maps don't conflict.
+		type data2 struct {
+			String string  `redis:"string"`
+			Bytes  []byte  `redis:"byte"`
+			Int    int     `redis:"int"`
+			Uint   uint    `redis:"uint"`
+			Float  float32 `redis:"float"`
+			Bool   bool    `redis:"bool"`
+		}
+		var d2 data2
```
vmihailenco, line 75: Just FYI this can be written as

```go
var d2 struct {
			String string  `redis:"string"`
			Bytes  []byte  `redis:"byte"`
			Int    int     `redis:"int"`
			Uint   uint    `redis:"uint"`
			Float  float32 `redis:"float"`
			Bool   bool    `redis:"bool"`
}
```

monkey92t:
missing test for `mget` command, we should complete it.
and...if it is 386 platform, there may be ERROR: "strconv.ParseInt: parsing "1234567890123456789": value out of range". 
reflect.Int64、reflect.Uint64、reflect.Float64,
but I’m not sure if we still need to be compatible with the 386 platform


`
redis 127.0.0.1:6379>> set key 1234567890123456789

type Temp struct {
	Key int64 `redis:"key"`
}
`
it should run normally.

knadh:
> missing test for `mget` command, we should complete it.

Added test https://github.com/knadh/redis/commit/f8a546b4820de9e52fe0d244e24dca155688abef. @monkey92t I guess you could add platform specific parsing.

> @knadh I've pushed a change with StructValue so StringStringMapCmd.Scan can be slightly more efficient. I think this PR is ready to be merged if you are happy with that change. It will be part of v8 but marked unstable so we can change it based on the feedback.

@vmihailenco looks fine. Please go ahead.

vmihailenco:
:partying_face:

mdelclaro:
Hello,

What if in this case I have a slice like `[]interface{}` instead of an interface only? I'm getting this `redis.Scan(non-struct *[]models.Price)`

I see that in the code you check for `if v.Kind() != reflect.Struct`

I need it to be a slice since the result I'm getting from `MGet` is a slice.

Any ideias on what I'm doing wrong?

monkey92t:
> Hello,
> 
> What if in this case I have a slice like `[]interface{}` instead of an interface only? I'm getting this `redis.Scan(non-struct *[]models.Price)`
> 
> I see that in the code you check for `if v.Kind() != reflect.Struct`
> 
> I need it to be a slice since the result I'm getting from `MGet` is a slice.
> 
> Any ideias on what I'm doing wrong?

under normal circumstances, the data format of redis is as follows:

```go
127.0.0.1:6379>>set id 100
127.0.0.1:6379>>set name monkey
127.0.0.1:6379>>set age 20

127.0.0.1:6379>>mget id name age
//["id": 100, "name": "monkey", "age": 20]

//golang
type Per struct {
	ID   int    `redis:"id"`
	Name string `redis:"name"`
	Age  int    `redis:"age"`
}
p := &Per{}
err := redis.Mget(ctx, "id", "name", "age").Scan(p)

// p.ID = 100
// p.Name = monkey
// p.Age = 20
```

i don’t know what you mean by `[]interface`, if i don’t understand correctly, you can add an example to describe your problem in detail.

mdelclaro:
> > Hello,
> > What if in this case I have a slice like `[]interface{}` instead of an interface only? I'm getting this `redis.Scan(non-struct *[]models.Price)`
> > I see that in the code you check for `if v.Kind() != reflect.Struct`
> > I need it to be a slice since the result I'm getting from `MGet` is a slice.
> > Any ideias on what I'm doing wrong?
> 
> under normal circumstances, the data format of redis is as follows:
> 
> ```go
> 127.0.0.1:6379>>set id 100
> 127.0.0.1:6379>>set name monkey
> 127.0.0.1:6379>>set age 20
> 
> 127.0.0.1:6379>>mget id name age
> //{"id": 100, "name": "monkey", "age": 20}
> 
> //golang
> type Per struct {
> 	ID   int    `redis:"id"`
> 	Name string `redis:"name"`
> 	Age  int    `redis:"age"`
> }
> p := &Per{}
> err := redis.Mget(ctx, "id", "name", "age").Scan(p)
> 
> // p.ID = 100
> // p.Name = monkey
> // p.Age = 20
> ```
> 
> i don’t know what you mean by `[]interface`, if i don’t understand correctly, you can add an example to describe your problem in detail.

I have hundreds of keys like `price/tokenX-tokenY`, so for what I see I should create a struct with all the keys, which would not be possible. And each value is a JSON structure. 

The response from MGet is something like this:

`
[ {"id":1,"base":"X","quote":"Y", "price": 1} ]
`

I'm a little lost here on what I have to do :')

mdelclaro:
> > > Hello,
> > > What if in this case I have a slice like `[]interface{}` instead of an interface only? I'm getting this `redis.Scan(non-struct *[]models.Price)`
> > > I see that in the code you check for `if v.Kind() != reflect.Struct`
> > > I need it to be a slice since the result I'm getting from `MGet` is a slice.
> > > Any ideias on what I'm doing wrong?
> > 
> > 
> > under normal circumstances, the data format of redis is as follows:
> > ```go
> > 127.0.0.1:6379>>set id 100
> > 127.0.0.1:6379>>set name monkey
> > 127.0.0.1:6379>>set age 20
> > 
> > 127.0.0.1:6379>>mget id name age
> > //{"id": 100, "name": "monkey", "age": 20}
> > 
> > //golang
> > type Per struct {
> > 	ID   int    `redis:"id"`
> > 	Name string `redis:"name"`
> > 	Age  int    `redis:"age"`
> > }
> > p := &Per{}
> > err := redis.Mget(ctx, "id", "name", "age").Scan(p)
> > 
> > // p.ID = 100
> > // p.Name = monkey
> > // p.Age = 20
> > ```
> > 
> > 
> > i don’t know what you mean by `[]interface`, if i don’t understand correctly, you can add an example to describe your problem in detail.
> 
> I have hundreds of keys like `price/tokenX-tokenY`, so for what I see I should create a struct with all the keys, which would not be possible. And each value is a JSON structure.
> 
> The response from MGet is something like this:
> 
> `[ {"id":1,"base":"X","quote":"Y", "price": 1} ]`
> 
> I'm a little lost here on what I have to do :')

I resolved it by doing the following: 

```golang
res := Cli.MGet(ctx, keys...)

var prices []models.Price

str := fmt.Sprintf("%+v", res.Val())
rawIn := json.RawMessage(str)

bytes, err := rawIn.MarshalJSON()

json.Unmarshal(bytes, &prices)
```

Thanks!

knadh:
@mdelclaro The `Scan()` function is specifically for scanning Redis values by their keys into structs. This is useful for MGET, HGETALL etc. In your example, you are pulling JSON strings as values from Redis and scanning keys of JSON strings, not keys from Redis. So, you indeed have to use json.Unmarshal.

csdenboer:
Scan doesn’t seem to work on nested structures: https://github.com/go-redis/redis/issues/1806. Based on the comments in this PR, it seems that “the package supports decoding of arbitrary
types into struct fields by defining a Decode(string) error function on types” was not actually implemented. Is this correct? If yes, will it be implemented anytime soon?

ysweid:
Decoding `time.Time` could be a big win.

haijianyang:
```go
cannot scan redis.result 2021-12-28T11:31:45.611441+08:00 into struct field UpdatedAt of type time.Time, error-redis.Scan(unsupported time.Time)
```
Hi guys, when scan will support **time.Time**?

knadh:
The issue with `time.Time` is the format to parse. Since `2021-12-28T11:31:45.611441+08:00` is the default format that's stringified, maybe we can hardcode scanning for just this one format and document that. @vmihailenco what do you think?

monkey92t:
See #2317

PR #1649, opened 2021-02-03
support "XINFO CONSUMERS"

kkkbird:
add function "XInfoConsumers" as "XInfoGroups", tested on redis 6.0.10

btw: refer to [xinfo](https://redis.io/commands/xinfo) document, these codes may not workable in future redis release

> Note that you should not rely on the fields exact position, nor on the number of fields, new fields may be added in the future. So a well behaving client should fetch the whole list, and report it to the user, for example, as a dictionary data structure. Low level clients such as C clients where the items will likely be reported back in a linear array should document that the order is undefined.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1403,6 +1403,103 @@ func (cmd *XPendingExtCmd) readReply(rd *proto.Reader) error {
 
 //------------------------------------------------------------------------------
 
+type XInfoConsumersCmd struct {
+	baseCmd
+	val []XInfoConsumer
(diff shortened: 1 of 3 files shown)


vmihailenco:
@kkkbird thanks for the contribution - code looks good. Is it possible to add a test?

kkkbird:
test is added

vmihailenco:
@kkkbird thank you :+1:

PR #1654, opened 2021-02-05
Fixed broken fuzzer

AdamKorcz:
As the fuzzer was moved, a few small changes are necessary to make it run.

### DIFF
--- a/fuzz/fuzz.go
+++ b/fuzz/fuzz.go
@@ -4,16 +4,17 @@ package fuzz
 
 import (
 	"context"
+	"github.com/go-redis/redis"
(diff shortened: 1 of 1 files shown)


fuzz/fuzz.go
```diff
@@ -4,16 +4,17 @@ package fuzz
 
 import (
 	"context"
+	"github.com/go-redis/redis"
```
vmihailenco, line 7: v8 suffix - I will add it myself.

vmihailenco:
:+1:  Sorry for breaking it

PR #1655, opened 2021-02-07
Make FailoverClient(sentinel mode) more robust

wangrzneu:
Make FailoverClient(sentinel mode) more robust
1. add failover option AllowDisconnectedSlaves
2. shuffle the sentinel addresses
3. query other sentinels when get empty slave addresses


sentinel.go
```diff
@@ -433,10 +437,26 @@ func (c *sentinelFailover) closeSentinel() error {
...
-	if err != nil {
-		return "", err
+	if c.opt == nil {
+		return "", errors.New("opt is nil")
 	}
+
+	var addresses []string
+	var err error
+
+	for _, allowDisconnected := range []bool{false, true} {
```
vmihailenco, line 447: Can we simplify this? It is getting late here and I am having troubles trying to understand what this does...

If it is not possible - please add a comment explaining what is happening.

sentinel.go
```diff
@@ -553,9 +591,9 @@ func (c *sentinelFailover) getSlaveAddrs(ctx context.Context, sentinel *Sentinel
 	return parseSlaveAddrs(addrs)
 }
 
-func parseSlaveAddrs(addrs []interface{}) []string {
+func parseSlaveAddrs(addrs []interface{}, allowedFlags ...string) []string {
```
vmihailenco, line 594: If I read this code correctly `allowedFlags` should be replaced with `keepDisconnected bool`.

sentinel.go
```diff
@@ -576,8 +614,10 @@ func parseSlaveAddrs(addrs []interface{}) []string {
 		}
 
 		for _, flag := range flags {
-			switch flag {
-			case "s_down", "o_down", "disconnected":
```
vmihailenco, line 580: Please keep this switch but add a separate clause like:

```
case "disconnected":
    if !keepDisconnected {
        isDown = true
    }
```

sentinel.go
```diff
@@ -509,8 +529,20 @@ func (c *sentinelFailover) slaveAddrs(ctx context.Context) ([]string, error) {
 			return addrs, nil
 		}
 		_ = c.closeSentinel()
+	} else {
+		// shuffle the sentinelAddrs to make the load more balanced
+		rand.Shuffle(len(c.sentinelAddrs), func(i, j int) {
```
vmihailenco, line 534: Should this code really be in `else` clause?

sentinel.go
```diff
@@ -36,6 +36,10 @@ type FailoverOptions struct {
 	// Route all commands to slave read-only nodes.
 	SlaveOnly bool
 
+	// Allow to access slaves disconnected with master when cannot get connected slaves
+	// Now, this option only works in RandomSlaveAddr function.
+	AllowDisconnectedSlaves bool
```
vmihailenco, line 41: Allow is a bit vague here. I would rename to `UseDisconnectedSlaves`.

Also perhaps disconnected slaves should be used only when there are no healthy slaves?

sentinel.go
wangrzneu: > Should this code really be in `else` clause?

I add `QuerySentinelRandomly` option to decide whether to shuffle.

sentinel.go
wangrzneu: I fix it.

sentinel.go
wangrzneu: It is renamed to UseDisconnectedSlaves.

wangrzneu:
@vmihailenco PTAL

sentinel.go
vmihailenco: I am not sure this is an essential option so I probably going to remove it. Let me know if you think it **is** essential.

I guess your original code was correct and we only want to shuffle this list once.

vmihailenco:
Looks good. I am going to tweak code a bit myself but but nothing big.

PR #1658, opened 2021-02-11
Adds support for IDLE to XPendingExt command

rhyselsmore:
As per subject, adds support for the IDLE Option to `xpending`


rhyselsmore:
Disregard - closing this out until the option lands in stable.

PR #1662, opened 2021-02-16
Add SetArgs command

ynden:
SetArgs supports all the SET command options. 
It is the alternative to the Set function when the user wants more control over the SET options.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -797,9 +797,9 @@ func (c cmdable) Set(ctx context.Context, key string, value interface{}, expirat
 //
 // When Get is true, the command returns the old value stored at key, or nil when key did not exist.
 type SetArgs struct {
-	Mode     string
-	ExpireAt time.Duration
-	Get      bool
+	Mode string
+	TTL  time.Duration
+	Get  bool
 }
 
 // SetArgs provides a way to call the SET command with SetArgs arguments.
@@ -809,13 +809,13 @@ func (c cmdable) SetArgs(ctx context.Context, key string, value interface{}, a *
 	// We set a rule to only use EX & PX options for expire time.
 	// We only need to support one of the two format (EX, PX) OR (EXAT, PXAT)
 	// because it is transparent to the user what we use here.
-	if a.ExpireAt > 0 {
-		if usePrecise(a.ExpireAt) {
-			args = append(args, "px", formatMs(ctx, a.ExpireAt))
+	if a.TTL > 0 {
+		if usePrecise(a.TTL) {
+			args = append(args, "px", formatMs(ctx, a.TTL))
 		} else {
-			args = append(args, "ex", formatSec(ctx, a.ExpireAt))
+			args = append(args, "ex", formatSec(ctx, a.TTL))
 		}
-	} else if a.ExpireAt == KeepTTL {
+	} else if a.TTL == KeepTTL {
 		args = append(args, "keepttl")
 	}
 
--- a/commands.go
+++ b/commands.go
@@ -810,16 +810,18 @@ type SetArgs struct {
 func (c cmdable) SetArgs(ctx context.Context, key string, value interface{}, a *SetArgs) *StatusCmd {
 	args := []interface{}{"set", key, value}
 
-	if !a.ExpireAt.IsZero() {
+	if a.KeepTTL {
+		args = append(args, "keepttl")
+	}
+
+	if !a.ExpireAt.IsZero() && !a.KeepTTL {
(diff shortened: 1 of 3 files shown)


commands_test.go
```diff
@@ -1169,7 +1374,7 @@ var _ = Describe("Commands", func() {
 			Expect(val).To(Equal("hello"))
 
 			Eventually(func() error {
-				return client.Get(ctx, "foo").Err()
+				return client.Get(ctx, "key").Err()
```
ynden, line 1377: I think it should be "key" and not "foo", since we want to test that "key" expired correctly.
The test passed because "foo" did not exist here, so it was alway nil.

commands_test.go
```diff
@@ -247,7 +247,7 @@ var _ = Describe("Commands", func() {
 		It("should Command", func() {
 			cmds, err := client.Command(ctx).Result()
 			Expect(err).NotTo(HaveOccurred())
-			Expect(len(cmds)).To(BeNumerically("~", 200, 20))
+			Expect(len(cmds)).To(BeNumerically("~", 200, 25))
```
ynden, line 250: I increased the threshold since we are testing against redis 6.2-rc3 version, so there are more commands.

commands.go
```diff
@@ -788,6 +788,50 @@ func (c cmdable) Set(ctx context.Context, key string, value interface{}, expirat
...
+//
+// When Get is true, the command returns the old value stored at key, or nil when key did not exist.
+type SetArgs struct {
+	Mode     string
+	ExpireAt time.Duration
+	Get      bool
+}
+
+// SetWithArgs provides a way to call the SET command with SetArgs arguments.
+func (c cmdable) SetWithArgs(ctx context.Context, key string, value interface{}, a *SetArgs) *StatusCmd {
```
vmihailenco, line 806: Let's call this `SetArgs` with we already have such name and it is better to follow one convention.

commands.go
```diff
@@ -788,6 +788,50 @@ func (c cmdable) Set(ctx context.Context, key string, value interface{}, expirat
 	return cmd
 }
 
+// SetArgs provides arguments for the SetWithArgs command.
+//
+// Mode can be `NX` or `XX` or empty.
```
vmihailenco, line 793: This should be a comment on `Mode string` field. And similar for Zero and Get.

commands.go
```diff
@@ -788,6 +788,50 @@ func (c cmdable) Set(ctx context.Context, key string, value interface{}, expirat
...
+//
+// Zero expiration means the key has no expiration time.
+// KeepTTL(-1) expiration is a Redis KEEPTTL option to keep existing TTL.
+//
+// When Get is true, the command returns the old value stored at key, or nil when key did not exist.
+type SetArgs struct {
+	Mode     string
+	ExpireAt time.Duration
```
vmihailenco, line 801: I believe ExpireAt should be called TTL.

And there is no support for `EXAT timestamp|PXAT milliseconds-timestamp`. II is fine if you decide not to support it in this PR, but I believe we need:

```go
TTL time.Duration // for px/ex
ExpireAt time.Time // for pxat/exat
```

commands.go
```diff
@@ -788,6 +788,50 @@ func (c cmdable) Set(ctx context.Context, key string, value interface{}, expirat
...
+	// We set a rule to only use EX & PX options for expire time.
+	// We only need to support one of the two format (EX, PX) OR (EXAT, PXAT)
+	// because it is transparent to the user what we use here.
+	if a.ExpireAt > 0 {
+		if usePrecise(a.ExpireAt) {
+			args = append(args, "px", formatMs(ctx, a.ExpireAt))
+		} else {
+			args = append(args, "ex", formatSec(ctx, a.ExpireAt))
+		}
+	} else if a.ExpireAt == KeepTTL {
```
vmihailenco, line 818: I would prefer this to be `KeepTTL bool` field.

```
if a.KeepTTL { ... }
```

It is more explicit and we have both `TTL` and `ExplireAt` which can be confusing.

commands_test.go
vmihailenco: Thanks for fixing this :+1:

vmihailenco:
Thanks. I've left few comments but overall this is looking good. Test coverage is impressing :+1:

ynden:
Thank you for you review! I'll look at this.

commands.go
ynden: I improved the readability

vmihailenco:
Thanks :+1: I will experiment with API a bit too...

Issue #1673, opened 2021-02-25
Ping too much in PubSub


PeerXu:
Hi, everyone.

I wrote a program with redis over 4G network.

So, traffic is limited.

When I am using `PSubscribe` command to subscribe message,

there are many ping packets over the network.

## Expected Behavior

`PSubscribe` succeed and ping channel for 1 minutes once.

## Current Behavior

Ping every seconds!

## Steps to Reproduce

```go
package main

import (
	"context"
	"fmt"

	"github.com/go-redis/redis/v8"
)

func main() {
	var opts redis.Options
	opts.Addr = "127.0.0.1:6379"
	client := redis.NewClient(&opts)
	ps := client.PSubscribe(context.Background(), "hello")
	ch := ps.Channel()
	for {
		select {
		case _, ok := <-ch:
			if !ok {
				return
			}
			fmt.Println("receive message")
		}
	}
}

```
```bash
$ sudo tcpdump -i lo
# catch ping pong packets every seconds
```

vmihailenco:
Redis Server is supposed to serve thousands of commands per second. It is hard to believe that a ping once a second causes any problems.

PeerXu:
@vmihailenco Yes, Redis Server is good enough to handle thousands of commands per second.
But ping once need to cost about 250~ Bytes(Ping packet, Pong packet, TCP ACK packet),
then 648 MBytes per month.
In some IoT situation, data traffic is very limited.

I read the code and found ping period is 1 minutes in old version,
but change to ping every second for faster renew the subscription.
Could we ping faster when disconnected, and ping slower when connected?

vmihailenco:
>Could we ping faster when disconnected, and ping slower when connected?

Yes, this sounds much better. But it still should not be smaller than 5 seconds...

rbg:
sounds like it should be externalized with sensible defaults for the normal use-case. That could allow IOT'ers the option to alter to accommodate Cell/LoRa

PeerXu:
@rbg, I change #1674 to expose the ping strategy interface if set.
Default ping frequency is 1 second as normal use-case.

monkey92t:
redis-server does not resend data on dropped connections, so we need to know that the connection is dropped and reconnect faster. 

I agree with @vmihailenco . you may consider increasing the time, but it should not be too long. If the ping cycle is controlled externally, there will be more unreliability.

PeerXu:
[...]

PR #1674, opened 2021-02-25
reduce ping times in PubSub

PeerXu:
To solve [issue1673](https://github.com/go-redis/redis/issues/1673)

### DIFF
--- a/options.go
+++ b/options.go
@@ -196,10 +196,6 @@ func (opt *Options) init() {
 	case 0:
 		opt.MaxRetryBackoff = 512 * time.Millisecond
 	}
-
-	if opt.SubscriptionPingStrategy == nil {
-		opt.SubscriptionPingStrategy = &defaultSubscriptionPingStrategy{}
-	}
 }
 
 func (opt *Options) clone() *Options {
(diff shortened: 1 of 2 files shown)


PeerXu:
@rbg Could you review this PR, please?

monkey92t:
I think ping is only a guarantee function of pubsub, and an overly complicated scheduling strategy should not be set, which may affect the reliability of pubsub itself.

We can increase the parameter to allow the setting of the ping period, but it should not be too large. This is caused by the pubsub feature of redis. It is not reliable. If the connection is dropped, redis-server will not reissue the lost data after recovery, so it is necessary Find the disconnection and reconnect in the shortest possible time.

i think this PR should be closed, it is not suitable to be merged.

vmihailenco:
@PeerXu sorry for ignoring this.

I agree with @monkey92t that the proposed solution is too complex for such task (which at least looks simple on the surface). Even though it is completely optional (thank you for this).

I've created https://github.com/go-redis/redis/issues/1700 with possible API to solve this in a more elegant manner. Let me know what do you think.

PR #1675, opened 2021-02-25
fix unit test errors that may appear in pool_test.go

monkey92t:
possible errors：

```go
/redis/pool_test.go:13
  reuses connections [It]
  /redis/pool_test.go:119

  Expected
      <uint32>: 98
  to equal
      <uint32>: 99

  /redis/pool_test.go:148
```

Code:
```go
It("reuses connections", func() {
		for i := 0; i < 100; i++ {
			val, err := client.Ping(ctx).Result()
			Expect(val).To(Equal("PONG"))
		}

		pool := client.Pool()
		Expect(pool.Len()).To(Equal(1))
		Expect(pool.IdleLen()).To(Equal(1))

		stats := pool.Stats()
		Expect(stats.Hits).To(Equal(uint32(99)))
		Expect(stats.Misses).To(Equal(uint32(1)))
		Expect(stats.Timeouts).To(Equal(uint32(0)))
})
```

because `opt.IdleTimeout = 1 * time.Second` and `conn.UsedAt` are unix timestamps.
if 100 times `client.Ping()` exceeds 1 second, `conn.deadline()` will be triggered to drop the connection, `conn.Get()` will create a new conn, and the number of `stats.Misses` will change.

### DIFF
--- a/pool_test.go
+++ b/pool_test.go
@@ -17,7 +17,7 @@ var _ = Describe("pool", func() {
 		opt := redisOptions()
 		opt.MinIdleConns = 0
 		opt.MaxConnAge = 0
-		opt.IdleTimeout = 2 * time.Second
+		opt.IdleTimeout = time.Second
 		client = redis.NewClient(opt)
 	})
 
@@ -105,6 +105,13 @@ var _ = Describe("pool", func() {
 	})
 
 	It("reuses connections", func() {
+		// explain: https://github.com/go-redis/redis/pull/1675
+		opt := redisOptions()
+		opt.MinIdleConns = 0
+		opt.MaxConnAge = 0
+		opt.IdleTimeout = 2 * time.Second
+		client = redis.NewClient(opt)
+
 		for i := 0; i < 100; i++ {
 			val, err := client.Ping(ctx).Result()
 			Expect(err).NotTo(HaveOccurred())
@@ -135,7 +142,7 @@ var _ = Describe("pool", func() {
 			StaleConns: 0,
 		}))
 
-		time.Sleep(3 * time.Second)
+		time.Sleep(2 * time.Second)
 
 		stats = client.PoolStats()
 		Expect(stats).To(Equal(&redis.PoolStats{
--- a/pool_test.go
+++ b/pool_test.go
@@ -17,7 +17,7 @@ var _ = Describe("pool", func() {
 		opt := redisOptions()
 		opt.MinIdleConns = 0
 		opt.MaxConnAge = 0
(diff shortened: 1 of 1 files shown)


monkey92t:
i encountered the data race error again, i will continue to track them.

vmihailenco:
@monkey92t thanks for working on this. Failing builds can be annoying.

I am a bit uneasy with this, because further bumping `IdleTimeout` is to some extent like completely disabling it. And it was enabled globally to make sure `IdleTimeout` does not affect the end result.

What about disabling `IdleTimeout = -1` just for this test? I.e. creating a client just to run the test and closing it in the end.

monkey92t:
> @monkey92t thanks for working on this. Failing builds can be annoying.
> 
> I am a bit uneasy with this, because further bumping `IdleTimeout` is to some extent like completely disabling it. And it was enabled globally to make sure `IdleTimeout` does not affect the end result.
> 
> What about disabling `IdleTimeout = -1` just for this test? I.e. creating a client just to run the test and closing it in the end.

it is not to increase `IdleTimeout` to avoid errors, because the unit of `conn.UsedAt` is seconds. if `IdleTimeout` is set to 1 second, we can easily trigger `conn.deadline()` and lose the connection.

```go
// set now = 2021-01-01 00:00:00:999  =  unix timestamp 1609430400
redis.Ping()  // call "cn.SetUsedAt(now)", set conn.usedAt = 1609430400

//the time after 1 millisecond is 2021-01-01 00:00:01:000000
time.Sleep(1 * Millisecond)

redis.Ping()
//the next time the redis command is executed, `pool.Get` will be called, `pool.Get()` will determine whether the connection needs to be dropped.

//pool.go
if p.isStaleConn(cn) {
	_ = p.CloseConn(cn)
	continue
}

//func pool.isStaleConn, opt.IdleTimeout = 1 * time.Second
// now = 2021-01-01 00:00:01:000000
// cn.UsedAt() = time.Unix(1609430400, 0) = 2021-01-01 00:00:00
if p.opt.IdleTimeout > 0 && now.Sub(cn.UsedAt()) >= p.opt.IdleTimeout {
	return true
}
```

in the example, only 1 millisecond has passed, but the pool dropped the connection.
after idletimeout is set to 2 seconds, the minimum interval between two requests is 1.0000...001 seconds to trigger `conn.deadline()`.

monkey92t:
if the idletimeout setting is not good globally, you can set it separately for `It("reuses connections", func() {})`?

vmihailenco:
Thansk :+1:

PR #1680, opened 2021-03-04
Make SetArgs available to Pipeliner

lhchavez:
SetArgs is amazing! It would be even more amazing to be able to use it
within a non-transactional Pipeline.


lhchavez:
huh, are those errors caused by this change? if so, are there any debugging pointers so that i can try to fix them?

vmihailenco:
Thanks. The build failure is unrelated.

PR #1685, opened 2021-03-05
Add ZMScore cmd

BenjaminPrieur:


commands.go
```diff
@@ -2010,6 +2011,14 @@ func (c cmdable) ZInterStore(ctx context.Context, destination string, store *ZSt
 	return cmd
 }
 
+func (c cmdable) ZMScore(ctx context.Context, key string, members ...interface{}) *FloatSliceCmd {
```
vmihailenco, line 2014: I see other `Z*` commands use `member string` instead of `interface{}`. Should we use `...string` here or I am missing something?

vmihailenco:
@BenjaminPrieur thanks for the PR. I have left one comment, but otherwise this looks good.

BenjaminPrieur:
@vmihailenco I did the changes you mentioned

vmihailenco:
Thanks :+1:

Issue #1657, opened 2021-02-09
Possible data race


vmihailenco:
Full build log: https://travis-ci.org/github/go-redis/redis/jobs/757558441 . Help is appreciated - so far I was not able to understand what is happening.

For example:

```
WARNING: DATA RACE
Read at 0x00c000010598 by goroutine 27:
  internal/race.ReadRange()
      /home/travis/.gimme/versions/go1.14.14.linux.amd64/src/internal/race/race.go:45 +0xab
  syscall.Write()
      /home/travis/.gimme/versions/go1.14.14.linux.amd64/src/syscall/syscall_unix.go:217 +0x8f
  internal/poll.(*FD).Write()
      /home/travis/.gimme/versions/go1.14.14.linux.amd64/src/internal/poll/fd_unix.go:268 +0x1f8
  net.(*netFD).Write()
      /home/travis/.gimme/versions/go1.14.14.linux.amd64/src/net/fd_unix.go:220 +0x65
  net.(*conn).Write()
      /home/travis/.gimme/versions/go1.14.14.linux.amd64/src/net/net.go:196 +0xa0
  net.(*TCPConn).Write()
      <autogenerated>:1 +0x69
  bufio.(*Writer).Write()
      /home/travis/.gimme/versions/go1.14.14.linux.amd64/src/bufio/bufio.go:623 +0x184
  github.com/go-redis/redis/v8/internal/proto.(*Writer).bytes()
      /home/travis/gopath/src/github.com/go-redis/redis/internal/proto/writer.go:122 +0xf7
  github.com/go-redis/redis/v8/internal/proto.(*Writer).string()
      /home/travis/gopath/src/github.com/go-redis/redis/internal/proto/writer.go:130 +0xf6e
  github.com/go-redis/redis/v8/internal/proto.(*Writer).WriteArg()
      /home/travis/gopath/src/github.com/go-redis/redis/internal/proto/writer.go:66 +0xf86
  github.com/go-redis/redis/v8/internal/proto.(*Writer).WriteArgs()
      /home/travis/gopath/src/github.com/go-redis/redis/internal/proto/writer.go:46 +0xf9
  github.com/go-redis/redis/v8.writeCmd()
      /home/travis/gopath/src/github.com/go-redis/redis/command.go:59 +0x4c
  github.com/go-redis/redis/v8.writeCmds()
      /home/travis/gopath/src/github.com/go-redis/redis/command.go:51 +0x87
[...]

PR #1696, opened 2021-03-14
fix test error

monkey92t:
fix test error:

The critical value of time may appear to cause test errors, link: https://github.com/redis/redis/blob/5b48d900498c85bbf4772c1d466c214439888115/src/object.c#L1265-L1272
```go
• Failure [0.068 seconds]
Commands
/redis/commands_test.go:17
  keys
  /redis/commands_test.go:304
    should Object [It]
    /redis/commands_test.go:462

    Expected
        <time.Duration>: 1000000000
    to equal
        <time.Duration>: 0

    /redis/commands_test.go:473
```
--------------------------------------------------------------

In addition, I tracked down more test errors: `CLUSTERDOWN The cluster is down`.
but I can't determine the cause for the time being. It may be redis-server instability or go-redis cluster problems. I will keep track of them.

I still can't reproduce https://github.com/go-redis/redis/issues/1657, maybe it's a problem with my test environment?
The go developers said that it might be a problem with pool: sync.pool{}, but they can’t confirm it. 
We can wait for the go developers’ investigation results, link: https://github.com/golang/go/issues/44672

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -457,6 +457,7 @@ var _ = Describe("Commands", func() {
 		})
 
 		It("should Object", func() {
+			start := time.Now()
 			set := client.Set(ctx, "key", "hello", 0)
 			Expect(set.Err()).NotTo(HaveOccurred())
 			Expect(set.Val()).To(Equal("OK"))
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks :+1:

vmihailenco:
>We can wait for the go developers’ investigation results, link: golang/go#44672

Let's keep discussion in https://github.com/go-redis/redis/issues/1657. Hopefully we will accumulate enough data to get this fixed.

PR #1711, opened 2021-03-27
add cmd(getex / getdel / hrandfield / zrandmember)

monkey92t:


commands.go
```diff
@@ -710,6 +721,35 @@ func (c cmdable) GetSet(ctx context.Context, key string, value interface{}) *Str
 	return cmd
 }
 
+// redis-server version >= 6.2.0.
+//
+// A value of zero means that the expiration time will not be changed.
+// Persist(-2) Remove the time to live associated with the key.
+func (c cmdable) GetEX(ctx context.Context, key string, expiration time.Duration) *StringCmd {
```
vmihailenco, line 728: I think `GETEX` means `GET` + `EX[PIRE]` which should be `GetEx` in Go. What about calling this `GetExpire`? It is unfortunate to intrudoce a different name for Redis command, but it is confusing to have the names like `SetNX` and `GetEX` which mean completely different things...

I don't think we need `redis.Persist` at, because in `Set` we use `0` to persist the key, .e.g it is `Set(key, value, 0)`, not `Set(key, value, redis.Persist)`. So passing `0` should mean `PERSIST`.

commands.go
```diff
@@ -2256,6 +2310,18 @@ func (c cmdable) ZUnionStore(ctx context.Context, dest string, store *ZStore) *I
 	return cmd
 }
 
+func (c cmdable) ZRandMember(ctx context.Context, key string, count int, withScores bool) *StringSliceCmd {
+	args := make([]interface{}, 0, 4)
+	args = append(args, "zrandmember", key, count)
```
vmihailenco, line 2315: Is it okay to use count=0 here too? If so having the comment here wouldn't hurt :)

commands_test.go
```diff
@@ -1801,6 +1832,22 @@ var _ = Describe("Commands", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			v := client.HRandField(ctx, "hash", 1, false)
+			Expect(v.Err()).NotTo(HaveOccurred())
+			Expect(v.Result()).To(Or(Equal([]string{"key1"}), Equal([]string{"key2"})))
+
+			var slice []string
+			err = client.HRandField(ctx, "hash", 1, true).ScanSlice(&slice)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(slice).To(Or(Equal([]string{"key1", "hello1"}), Equal([]string{"key2", "hello2"})))
```
vmihailenco, line 1849: Would be nice to test count=0 here in case Redis will start rejecting it some day...

monkey92t:
I am also very confused about getex, but I think it should be consistent with the redis command.

vmihailenco:
:+1: I am going to rename GetEX to `GetEx` since `Ex` is not an acronym:
- Ex - **ex**pire
- NX - **n**ot **e**xists

vmihailenco:
And we already have `SetEx` :)

PR #1716, opened 2021-04-05
Support redis.RESP3 protocol

monkey92t:


vmihailenco:
Let me know if you need any help with design. My first impression is that `pvs []*proto.Value` going to be slow since we need to allocate an intermediate value before scanning the result. Reading directly into the dest variable is much more efficient...

monkey92t:
Yes, I found the problem, I need to redesign it.

PR #1720, opened 2021-04-09
hmset support map[string]string

leizihui:
hgetall return map[string]string ,but hmset do not support map[string]string


vmihailenco:
Thanks

PR #1729, opened 2021-04-16
ScriptLoad, ScriptExists and ScriptFlush for ClusterClient

### LINKED ISSUE #1725: ScriptLoad on ClusterClient does not load on all instances
ScriptLoad function on ClusterClient is loading the script on only one of the masters.

## Expected Behavior
Load the script on all the instances.
The next invocation of `EvalSha` based on hash partioning key should execute without any error.

Example command:
`rdb.EvalSha(ctx, hashOfScriptFile, []string{"exampleKey:{1}"}, args...)`

## Current Behavior
Loading script only on one master.

Throws the following error when executing the script
`NOSCRIPT No matching script. Please use EVAL.`

This is because, based on hash key it has redirected to other instance where the script hasn't been loaded

## Possible Solution

ScriptLoad to get all the instances & load the script in all instances

## Context (Environment)
Latest version

skryukov:
Changes:
`ScriptLoad` loads script to all shards of the cluster.
`ScriptFlush` removes scripts from all shards of the cluster.
`ScriptExists` checks all shards of the cluster and returns false if any of the shards returned false.

Closes #1725

### DIFF
--- a/cluster_commands.go
+++ b/cluster_commands.go
@@ -2,6 +2,7 @@ package redis
 
 import (
 	"context"
+	"sync"
 	"sync/atomic"
 )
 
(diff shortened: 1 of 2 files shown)


cluster_commands.go
```diff
@@ -23,3 +23,57 @@ func (c *ClusterClient) DBSize(ctx context.Context) *IntCmd {
...
+
+func (c *ClusterClient) ScriptLoad(ctx context.Context, script string) *StringCmd {
+	cmd := NewStringCmd(ctx, "script", "load", script)
+	err := c.ForEachShard(ctx, func(ctx context.Context, shard *Client) error {
+		val, err := shard.ScriptLoad(ctx, script).Result()
+		if err != nil {
+			return err
+		}
+		if cmd.Val() == "" {
+			cmd.val = val
```
vmihailenco, line 35: This is racy. Just wrapping this code with `sync.Mutex` should be enough.

cluster_commands.go
```diff
@@ -23,3 +23,57 @@ func (c *ClusterClient) DBSize(ctx context.Context) *IntCmd {
...
+
+	err := c.ForEachShard(ctx, func(ctx context.Context, shard *Client) error {
+		val, err := shard.ScriptExists(ctx, hashes...).Result()
+		if err != nil {
+			return err
+		}
+
+		for i, v := range val {
+			result[i] = result[i] && v
+		}
```
vmihailenco, line 68: This is racy too.

vmihailenco:
@skryukov this looks good, but please fix the race (see the build failure).

vmihailenco:
Thanks :+1:

anooshcnayak:
> Changes:
> `ScriptLoad` loads script to all shards of the cluster.
> `ScriptFlush` removes scripts from all shards of the cluster.
> `ScriptExists` checks all shards of the cluster and returns false if any of the shards returned false.
> 
> Closes #1725

Thanks a lot for the quick implementation. @skryukov

PR #1736, opened 2021-04-22
support resp3 protocol

monkey92t:
Support RESP2 and RESP3 protocols.  On this basis, a richer`Redis server-assisted client side caching` can be achieved

But it breaks the API in v8, should it be merged into v9?

### DIFF
--- a/command.go
+++ b/command.go
@@ -316,15 +316,24 @@ func (cmd *Cmd) Bool() (bool, error) {
 }
 
 func (cmd *Cmd) readReply(rd *proto.Reader) (err error) {
-	cmd.val, err = rd.ReadReply(sliceParser)
+	cmd.val, err = rd.ReadReply(aggregateParser)
 	return err
+}
+
+// aggregateParser implements proto.AggregateBulkParse.
+func aggregateParser(rd *proto.Reader, typ byte, n int64) (interface{}, error) {
+	// map
+	if typ == proto.RespMap {
+		return mapParser(rd, n)
+	}
+	return sliceParser(rd, n)
 }
 
 // sliceParser implements proto.MultiBulkParse.
 func sliceParser(rd *proto.Reader, n int64) (interface{}, error) {
 	vals := make([]interface{}, n)
 	for i := 0; i < len(vals); i++ {
-		v, err := rd.ReadReply(sliceParser)
+		v, err := rd.ReadReply(aggregateParser)
 		if err != nil {
 			if err == Nil {
 				vals[i] = nil
@@ -339,6 +348,31 @@ func sliceParser(rd *proto.Reader, n int64) (interface{}, error) {
 		vals[i] = v
 	}
 	return vals, nil
+}
+
+// mapParser implements proto.MultiBulkParse.
+func mapParser(rd *proto.Reader, n int64) (interface{}, error) {
+	m := make(map[interface{}]interface{}, n)
+	for i := int64(0); i < n; i++ {
+		k, err := rd.ReadReply(aggregateParser)
+		if err != nil {
+			return nil, err
+		}
+		v, err := rd.ReadReply(aggregateParser)
+		if err != nil {
+			if err == Nil {
+				m[k] = Nil
+				continue
+			}
+			if err, ok := err.(proto.RedisError); ok {
+				m[k] = err
+				continue
+			}
+			return nil, err
+		}
+		m[k] = v
+	}
+	return m, nil
 }
 
 //------------------------------------------------------------------------------
@@ -473,7 +507,7 @@ func (cmd *IntCmd) String() string {
 }
 
 func (cmd *IntCmd) readReply(rd *proto.Reader) (err error) {
-	cmd.val, err = rd.ReadIntReply()
+	cmd.val, err = rd.ReadInt()
 	return err
 }
 
@@ -512,7 +546,7 @@ func (cmd *IntSliceCmd) readReply(rd *proto.Reader) error {
 	_, err := rd.ReadArrayReply(func(rd *proto.Reader, n int64) (interface{}, error) {
 		cmd.val = make([]int64, n)
 		for i := 0; i < len(cmd.val); i++ {
-			num, err := rd.ReadIntReply()
+			num, err := rd.ReadInt()
 			if err != nil {
 				return nil, err
 			}
@@ -557,7 +591,7 @@ func (cmd *DurationCmd) String() string {
 }
 
 func (cmd *DurationCmd) readReply(rd *proto.Reader) error {
-	n, err := rd.ReadIntReply()
+	n, err := rd.ReadInt()
 	if err != nil {
 		return err
 	}
@@ -811,7 +845,7 @@ func (cmd *FloatCmd) String() string {
 }
 
 func (cmd *FloatCmd) readReply(rd *proto.Reader) (err error) {
-	cmd.val, err = rd.ReadFloatReply()
+	cmd.val, err = rd.ReadFloat()
 	return err
 }
 
@@ -850,7 +884,7 @@ func (cmd *FloatSliceCmd) readReply(rd *proto.Reader) error {
 	_, err := rd.ReadArrayReply(func(rd *proto.Reader, n int64) (interface{}, error) {
 		cmd.val = make([]float64, n)
 		for i := 0; i < len(cmd.val); i++ {
-			switch num, err := rd.ReadFloatReply(); {
+			switch num, err := rd.ReadFloat(); {
 			case err == Nil:
 				cmd.val[i] = 0
 			case err != nil:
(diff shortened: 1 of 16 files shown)


command.go
```diff
@@ -919,6 +953,93 @@ func (cmd *StringSliceCmd) readReply(rd *proto.Reader) error {
 
 //------------------------------------------------------------------------------
 
+type KeyValue struct {
+	Key   string
+	Value string
+}
+
+type SliceKeyValueCmd struct {
```
vmihailenco, line 961: Should be KeyValueSliceCmd

command.go
```diff
@@ -919,6 +953,93 @@ func (cmd *StringSliceCmd) readReply(rd *proto.Reader) error {
...
+}
+
+func (cmd *SliceKeyValueCmd) readReply(rd *proto.Reader) error { // nolint: dupl
+	_, err := rd.ReadArrayReply(func(rd *proto.Reader, n int64) (interface{}, error) {
+		if n == 0 {
+			cmd.val = make([]KeyValue, 0)
+			return nil, nil
+		}
+
+		b, err := rd.NextType()
```
vmihailenco, line 997: I would expect something like `rd.ReadReplyType`. `Next` implies that this is an iterator and `Type` and is not specific enough.

command.go
```diff
@@ -1016,9 +1137,9 @@ func (cmd *StringStringMapCmd) Scan(dst interface{}) error {
 }
 
 func (cmd *StringStringMapCmd) readReply(rd *proto.Reader) error {
-	_, err := rd.ReadArrayReply(func(rd *proto.Reader, n int64) (interface{}, error) {
-		cmd.val = make(map[string]string, n/2)
-		for i := int64(0); i < n; i += 2 {
+	_, err := rd.ReadMapReply(func(rd *proto.Reader, n int64) (interface{}, error) {
```
vmihailenco, line 1140: Let's not introduce more helpers like `ReadArrayReply`, `ReadMapReply`, `ReadReply` - they don't help much and only make code look more complex. E.g. the more straightforward code is

```go
n, err := rd.ReadMapLen()
if err != nil {
    return err
}

for i := int64(0); i < n; i++ {
}
```

It is more readable and even slightly faster. What do you think?

command.go
```diff
@@ -1556,12 +1677,12 @@ func (cmd *XInfoConsumersCmd) readReply(rd *proto.Reader) error {
 func readXConsumerInfo(rd *proto.Reader) (XInfoConsumer, error) {
 	var consumer XInfoConsumer
 
-	n, err := rd.ReadArrayLen()
+	n, err := rd.ReadMapLen()
```
vmihailenco, line 1680: This is a good example of what I mean about `ReadMapReply`. We can even add a helper like `ReadFixedMapLen(3)` that will check that map has the expected number of entries.

command.go
```diff
@@ -2589,3 +2762,126 @@ func (cmd *SlowLogCmd) readReply(rd *proto.Reader) error {
...
+			return err
+		}
+		cmd.val[k] = v
+	}
+	return nil
+}
+
+//-----------------------------------------------------------------------
+
+type SliceMapStringStringCmd struct {
```
vmihailenco, line 2828: MapStringStringSliceCmd

commands.go
```diff
@@ -168,7 +168,8 @@ type Cmdable interface {
 	HMSet(ctx context.Context, key string, values ...interface{}) *BoolCmd
 	HSetNX(ctx context.Context, key, field string, value interface{}) *BoolCmd
 	HVals(ctx context.Context, key string) *StringSliceCmd
-	HRandField(ctx context.Context, key string, count int, withValues bool) *StringSliceCmd
+	HRandField(ctx context.Context, key string, count int) *StringSliceCmd
+	HRandFieldWithValues(ctx context.Context, key string, count int) *SliceKeyValueCmd
```
vmihailenco, line 172: I like this :+1:

internal/proto/reader.go
```diff
@@ -4,18 +4,36 @@ import (
...
-	ArrayReply  = '*'
+	RespStatus    = '+' // +<string>\r\n
+	RespError     = '-' // -<string>\r\n
+	RespString    = '$' // $<length>\r\n<bytes>\r\n
+	RespInteger   = ':' // :<number>\r\n
+	RespNil       = '_' // _\r\n
+	RespFloat     = ',' // ,<floating-point-number>\r\n (golang float)
+	RespBool      = '#' // true: #t\r\n false: #f\r\n
+	RespBlobError = '!' // !<length>\r\n<bytes>\r\n
+	RespVerb      = '=' // =<length>\r\nFORMAT:<bytes>\r\n
```
vmihailenco, line 23: `Verb` looks like a wrong shortcut for `verbatim`. Let's keep the long version `RespVerbatim`

internal/proto/reader.go
```diff
@@ -4,18 +4,36 @@ import (
...
 const (
-	ErrorReply  = '-'
-	StatusReply = '+'
-	IntReply    = ':'
-	StringReply = '$'
-	ArrayReply  = '*'
+	RespStatus    = '+' // +<string>\r\n
+	RespError     = '-' // -<string>\r\n
+	RespString    = '$' // $<length>\r\n<bytes>\r\n
+	RespInteger   = ':' // :<number>\r\n
```
vmihailenco, line 18: RespInteger -> RespInt since we have `ReadInt` and `RespBigInt`.

internal/proto/reader.go
```diff
@@ -28,20 +46,34 @@ func (RedisError) RedisError() {}
 
 //------------------------------------------------------------------------------
 
-type MultiBulkParse func(*Reader, int64) (interface{}, error)
+var (
+	uvinf    = math.Inf(1)
+	uvneginf = math.Inf(-1)
```
vmihailenco, line 51: Let's remove this or at least rename to `posInf` and `negInf`?

internal/proto/reader.go
```diff
@@ -28,20 +46,34 @@ func (RedisError) RedisError() {}
...
+)
+
+type (
+	AggregateBulkParse func(*Reader, byte, int64) (interface{}, error)
+	MultiBulkParse     func(*Reader, int64) (interface{}, error)
+)
```
vmihailenco, line 57: Ideally we should not introduce `AggregateBulkParse` - see my comment about `ReadMapReply`.

internal/proto/reader.go
```diff
@@ -54,15 +86,18 @@ func (r *Reader) Reset(rd io.Reader) {
...
 }
 
+// ReadLine may return unexpected attribute types, it is best to use Pathfinder.
 func (r *Reader) ReadLine() ([]byte, error) {
-	line, err := r.readLine()
+	return r.readLine()
+}
+
+// NextType get the data type of the next row.
+func (r *Reader) NextType() (byte, error) {
```
vmihailenco, line 95: PeekReplyType or PeekRespType

internal/proto/reader.go
```diff
@@ -92,240 +127,377 @@ func (r *Reader) readLine() ([]byte, error) {
...
+		return "", fmt.Errorf("redis: can't parse verbatim string reply: %q", line)
+	}
+	return s[4:], nil
+}
+
+// -------------------------------
+
+// Pathfinder find up possible errors in redis responses and discard attributes types,
+// `line` that returns the true reply.
+func (r *Reader) Pathfinder() ([]byte, error) {
```
vmihailenco, line 229: I am not sure I like the name `Pathfinder` - is this a common name that I don't recognize?

I think we should continue to call this `ReadLine` and introduce a new name later when we need to parse attributes.

internal/proto/reader.go
```diff
@@ -92,240 +127,377 @@ func (r *Reader) readLine() ([]byte, error) {
...
-		return nil, fmt.Errorf("redis: can't parse string reply: %.100q", line)
+		return 0, fmt.Errorf("redis: can't parse map reply: %.100q", line)
 	}
 }
 
-func (r *Reader) _readTmpBytesReply(line []byte) ([]byte, error) {
-	if isNilReply(line) {
-		return nil, Nil
+// Discard the data represented by line, if line is nil, read the next line.
+func (r *Reader) Discard(line []byte) (err error) {
```
vmihailenco, line 432: Let's remove the part that handles `if line is nil, read the next line.`. You can introduce a separate function like `DiscardNext()` if that is a common operation.

monkey92t:
Thanks for your suggestions, I will fix these problems.

pubsub_test.go
```diff
@@ -15,16 +14,11 @@ import (
...
-	var clientID int64
 
 	BeforeEach(func() {
 		opt := redisOptions()
 		opt.MinIdleConns = 0
 		opt.MaxConnAge = 0
-		opt.OnConnect = func(ctx context.Context, cn *redis.Conn) (err error) {
-			clientID, err = cn.ClientID(ctx).Result()
-			return err
-		}
```
vmihailenco, line 27: Any reasons this is removed?

pubsub_test.go
```diff
@@ -421,30 +415,6 @@ var _ = Describe("PubSub", func() {
...
-		err = client2.Do(ctx, "SET", "mykey", "myvalue").Err()
-		Expect(err).NotTo(HaveOccurred())
-
-		ch := pubsub.Channel()
-
-		var msg *redis.Message
-		Eventually(ch).Should(Receive(&msg))
-		Expect(msg.Channel).To(Equal("__redis__:invalidate"))
-		Expect(msg.PayloadSlice).To(Equal([]string{"mykey"}))
-	})
```
vmihailenco, line 446: Same here.

internal/proto/reader.go
```diff
@@ -28,20 +46,34 @@ func (RedisError) RedisError() {}
...
 	return &Reader{
 		rd:   bufio.NewReader(rd),
-		_buf: make([]byte, 64),
+		Resp: 2,
 	}
 }
 
+func (r *Reader) SetResp(v int) {
+	r.Resp = v
+}
```
vmihailenco, line 75: It looks like this can/should be removed.

vmihailenco:
@monkey92t I've left few comments - please take a look. Most of them are trivial but some are not. If you think it is best to split the remaining work in multiple PRs - I will totally understand.

>Support RESP2 and RESP3 protocols

Looks like this PR requires RESP3 protocol. I think it is fine but please clarify. We can consider supporting RESP2 later if possible.

>But it breaks the API in v8, should it be merged into v9?

Yes, I will create a v9 branch - let's work there.

pubsub_test.go
monkey92t: In RESP3 mode, `client.Subscribe(ctx, "__redis__:invalidate")` does not work, it works in RESP2 mode, i removed this part. 
We should use the correct `Redis server-assisted client side caching` (to be implemented soon)

monkey92t:
Close this PR and move to v9

PR #1739, opened 2021-04-26
Add support for resp3 protocol

monkey92t:
1. Support redis RESP3 protocol, can work under RESP2 and RESP3.
2. If redis-server version >=6.0, RESP3 protocol is enabled by default. Otherwise use RESP2 protocol (redis < 6.0).
3. Remove test `Subscribe __redis__:invalidate`, it only works in RESP2 mode(redis6.0+).
4. `$EOF:<40 bytes marker><CR><LF>` and `*?<CR><LF>` that have not been enabled by redis-server have not been implemented for the time being, they will be followed up in the future.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -168,7 +168,8 @@ type Cmdable interface {
 	HMSet(ctx context.Context, key string, values ...interface{}) *BoolCmd
 	HSetNX(ctx context.Context, key, field string, value interface{}) *BoolCmd
 	HVals(ctx context.Context, key string) *StringSliceCmd
-	HRandField(ctx context.Context, key string, count int, withValues bool) *StringSliceCmd
+	HRandField(ctx context.Context, key string, count int) *StringSliceCmd
+	HRandFieldWithValues(ctx context.Context, key string, count int) *SliceKeyValueCmd
 
 	BLPop(ctx context.Context, timeout time.Duration, keys ...string) *StringSliceCmd
 	BRPop(ctx context.Context, timeout time.Duration, keys ...string) *StringSliceCmd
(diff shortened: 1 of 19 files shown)


commands.go
```diff
@@ -287,7 +289,7 @@ type Cmdable interface {
 	ClientList(ctx context.Context) *StringCmd
 	ClientPause(ctx context.Context, dur time.Duration) *BoolCmd
 	ClientID(ctx context.Context) *IntCmd
-	ConfigGet(ctx context.Context, parameter string) *SliceCmd
+	ConfigGet(ctx context.Context, parameter string) *StringStringMapCmd
```
vmihailenco, line 292: MapStringStringCmd since we already have `MapStringInterfaceCmd`

vmihailenco:
>If redis-server version >=6.0, RESP3 protocol is enabled by default. Otherwise use RESP2 protocol (redis < 6.0).

It looks like the code always uses https://redis.io/commands/hello to switch to RESP3. And `HELLO` is only available since v6.0.0. So I guess v9 branch
- does not support Redis < 6.0
- always uses RESP3

Which is fine - I just want to confirm my understanding.

monkey92t:
> > If redis-server version >=6.0, RESP3 protocol is enabled by default. Otherwise use RESP2 protocol (redis < 6.0).
> 
> It looks like the code always uses https://redis.io/commands/hello to switch to RESP3. And `HELLO` is only available since v6.0.0. So I guess v9 branch
> 
> * does not support Redis < 6.0
> * always uses RESP3
> 
> Which is fine - I just want to confirm my understanding.

If conn.HELLO(ctx, 3) fails to execute (redis<6.0), there will be no error, and it will continue to use RESP2 to execute commands.
If redis-server >= 6.0, HELLO is executed successfully and RESP3 is used.

vmihailenco (approved):

vmihailenco:
Awesome, thanks.

PR #1742, opened 2021-04-28
Add command SMISMEMBER

nigelis:


commands.go
```diff
@@ -200,6 +200,7 @@ type Cmdable interface {
 	SInter(ctx context.Context, keys ...string) *StringSliceCmd
 	SInterStore(ctx context.Context, destination string, keys ...string) *IntCmd
 	SIsMember(ctx context.Context, key string, member interface{}) *BoolCmd
+	SMIsMember(ctx context.Context, key string, members ...interface{}) *BoolSliceCmd
```
vmihailenco, line 203: I believe we should change `SIsMember` to accept multiple args and use `smismember` if `len(members) > 1`. What do you think?

nigelis: I understand that these two commands are quite similar, but changing `SIsMember` to accept multiple args breaks the API compatibility. The current type of return value is `*BoolCmd`, while the desired is `*BoolSliceCmd`.

monkey92t: Can we consider modifying it in v9? they function exactly the same.

nigelis: This looks good to me.

I can modify the `SIsMember` and open another PR to branch v9 after you draw the conclusion.

vmihailenco: I haven't realized they return different results. No need to merge them then.

PR #1743, opened 2021-04-28
Use the right redis client in tests

nigelis:
The client used in some tests is mistaken.


vmihailenco:
Thanks. We probably should rename client to rdb everywhere...

PR #1746, opened 2021-05-06
Support IDLE argument on xpending method (redis >= 6.2.0)

ibenitez:


monkey92t:
I did not notice that it is the same PR as #1750, let's close it

Issue #1518, opened 2020-10-07
Sentinel, discarding bad PubSub connection: redis: ping timeout


pierrre:
This issue is not really a duplicate of https://github.com/go-redis/redis/issues/1407 , because I've got new informations.

I'm using v8.2.3.
I'm creating a failover/sentinel client.

My logs shows
```
discarding bad PubSub connection: redis: ping timeout
```
approximately every minute.

If I understand correctly, the sentinel/failover client is using pubsub in order to receive status change from the server. (I'm not really an expert here).

So I looked at this code
https://github.com/go-redis/redis/blob/7125bf611e5d7d9bb4487dd6cb80d888bad92d23/pubsub.go#L478-L512

If I understand correctly:
- if we receive pubsub messages frequently enough, we will receive a message from the channel `c.ping`, so it will be considered healty
- the `pingTimeout` is now 1 second (when the previous issue was opened, it was 30 seconds)
- if we don't receive a pubsub message, then the code sends a "ping" command
- if the pubsub was previously healthy, it is changed to not healthy, which will trigger again a ping later
- if the pubsub was previously not healty, it is ALWAYS disconnected/reconnected, and an error is logged (even if the ping command didn't fail)
- even if `pingErr` is nil, it will be defined with `errPingTimeout`

From my point of view, as a Go developer and user of this library, this code looks really suspicious.
If I understand correctly, if we don't receive any pubsub message after 1 or 2 seconds, it will disconnect/reconnect the pubsub, and log an error message.

## Expected Behavior

I should not see this error log.

## Current Behavior

I can see this error log.

## Possible Solution

The code inside `case <-timer.C` should be rewritten:
- check the `pingErr`. If it's nil, don't try to reconnect
- IMO the `healthy` variable is not useful. If we receive a pubsub message, simply reset the time. Otherwise send a "ping" command, and if it fails, disconnect/reconnect.

vmihailenco:
Looks like you are right that we can better use the result of `Ping` command. In Postgres from which I ported this code `Ping` is sent on a separate connection - so the result does not mean much. Still we can't be 100% sure that successful write to a connection means that the connection is healthy.

>when the previous issue was opened, it was 30 seconds

30 seconds can be too much in some cases...

git-hulk:
[...]

Issue #1700, opened 2021-03-17
Chan config / options


vmihailenco:
For https://github.com/go-redis/redis/issues/1673

Add new method `Chan` that accepts config to change buffer size and ping frequency, e.g.

```go
type ChanConfig {
    BufferSize         int
    PingFrequency time.Duration
}

func (c *PubSub) Chan(cfg ChanConfig) <-chan *Message {
    ...
}
```

In v9 `Chan` should completely replace `ChannelSize` and `Channel`.

---

Alternatively we could use options instead of config:

```go
func (c *PubSub) Chan(opt ...ChanOption) <-chan *Message {
    ...
}

pubsub.Chan(redis.WIthChanSize(100), redis.WithChanPingFrequency(time.Second))
```

Such API looks better / cooler, but for new users it is harder to find all available config options...

Any thoughts?

rbg:
i like this solution over the solution in #1674 

I also do not think that it is that hard to find config options.. esp. if they are documented :)

monkey92t:
we can increase the frequency of ping after disconnection to recover as soon as possiblew:
how do you feel?

```go
for {
		timer.Reset(pingTimeout)
		select {
		case <-timer.C:
			err := c.Ping(ctx).Err()
			if err != nil {
				pingTimeout = 1 * time.Second
			} else {
				pingTimeout = configTimeout
			}
		}
	}
```

PeerXu:
> ```go
> for {
> 		timer.Reset(pingTimeout)
> 		select {
> 		case <-timer.C:
> 			err := c.Ping(ctx).Err()
> 			if err != nil {
> -				pingTimeout = 1 * time.Second
> +				pingTimeout = retryTimeout
> 			} else {
> 				pingTimeout = configTimeout
> 			}
> 		}
> 	}
> ```

On ping failed, pingTimeout should be optional.

> ```go
> func (c *PubSub) Chan(opt ...ChanOption) <-chan *Message {
>    ...
> }
>
> pubsub.Chan(
> 	redis.WIthChanSize(100), 
> 	redis.WithChanPingFrequency(time.Second /* default ping freq */, time.Second /* retried ping freq */))
> ```

About WithChanPingFrequency definition

```go
func WithChanPingFrequency(timeout time.Duration, retryTimeout ...time.Duration)
```

If retryTimeout not set, should be assigned by timeout.

monkey92t:
[...]

PR #1747, opened 2021-05-07
reconnect only on fail ping

snwfog:
Addressing: https://github.com/go-redis/redis/issues/1518

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -2,6 +2,7 @@ package redis
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
@@ -16,6 +17,8 @@ const (
 	pingTimeout     = time.Second
 	chanSendTimeout = time.Minute
 )
+
+var errPingTimeout = errors.New("redis: ping timeout")
 
 // PubSub implements Pub/Sub commands as described in
 // http://redis.io/topics/pubsub. Message receiving is NOT safe
@@ -43,6 +46,9 @@ type PubSub struct {
 	msgCh  chan *Message
 	allCh  chan interface{}
 	ping   chan struct{}
+
+	// Health check timeout, default 5 second.
+	healthTimeout time.Duration
 }
 
 func (c *PubSub) String() string {
@@ -53,6 +59,9 @@ func (c *PubSub) String() string {
 
 func (c *PubSub) init() {
 	c.exit = make(chan struct{})
+	if c.healthTimeout == 0 {
+		c.healthTimeout = 5 * time.Second
+	}
 }
 
 func (c *PubSub) connWithLock(ctx context.Context) (*pool.Conn, error) {
@@ -268,6 +277,12 @@ func (c *PubSub) Ping(ctx context.Context, payload ...string) error {
 	return err
 }
 
+func (c *PubSub) SetHealthTimeout(t time.Duration) {
+	if t > 0 {
+		c.healthTimeout = t
+	}
+}
+
 // Subscription received after a successful subscription to channel.
 type Subscription struct {
 	// Can be "subscribe", "unsubscribe", "psubscribe" or "punsubscribe".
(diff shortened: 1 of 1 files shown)


rbg:
can you describe what you think your code change is doing please.

snwfog:
@rbg  I am trying to solve few issues mentioned in #1518. 

1. On the first timeout, `pingErr` is not used at all from calling `Ping`.
2. On a consecutive timeout, in the case where `pingErr` is nil, we still close current connection and reconnect. This has lead to lost of messages.
3. `healthy` is used locally only; it doesn't need to be inside of mutex.

monkey92t:
How about this way? Any suggestions?

```go
type PubSub struct {
	// ...
	healthTimeout time.Duration
}

func (c *PubSub) init() {
	//...
	if c.healthTimeout == 0 {
		c.healthTimeout = 5 * time.Second
	}
}

func (c *PubSub) SetHealthTimeout(t time.Duration) {
	if t > 0 {
		c.healthTimeout = t
	}
}

func (c *PubSub) initPing() error {
	//....
	go func() {
		timer := time.NewTimer(time.Minute)
		timer.Stop()

		healthy := true
		timeout := pingTimeout // 1 * time.Second
		for {
			timer.Reset(timeout)
			select {
			case <-c.ping:
				healthy = true
				timeout = pingTimeout
				if !timer.Stop() {
					<-timer.C
				}
			case <-timer.C:
				var healthyErr error

				if healthy {
					healthyErr = c.Ping(ctx)
					healthy = false
				} else {
					healthyErr = errPingTimeout
				}

				if healthyErr != nil {
					c.mu.Lock()
					c.reconnect(ctx, pingErr)
					c.mu.Unlock()

					healthy = true
				} else {
					timeout = c.healthTimeout
				}
			case <-c.exit:
				return
			}
		}
	}()
}
```

monkey92t:
I made changes, any suggestions?
@vmihailenco  Review?

pubsub.go
```diff
@@ -483,26 +495,35 @@ func (c *PubSub) initPing() {
...
 					c.mu.Lock()
-					c.reconnect(ctx, pingErr)
-					healthy = true
+					c.reconnect(ctx, healthyErr)
 					c.mu.Unlock()
+
+					healthy = true
+					timeout = pingTimeout
+				} else {
+					timeout = c.healthTimeout
```
rbg, line 526: does this make sense -- should we always  use the time that was explicitly set ? 

                                 if healthyErr != nil {
					c.mu.Lock()
					c.reconnect(ctx, healthyErr)
					c.mu.Unlock()
					healthy = true
				} 
				timeout = c.healthTimeout

monkey92t: It is the waiting time of the ping command.

vmihailenco:
I've recently refactored similar code for [Bun](https://github.com/uptrace/bun/blob/master/driver/pgdriver/listener.go#L224-L254). As you can see, there is a `Channel` method that accepts an arbitrary number of options. All the code that is required to support it is extracted to the `Channel` struct. It is the 2nd option described in https://github.com/go-redis/redis/issues/1700

Do you think we can do the same thing in go-redis?

pubsub.go
```diff
@@ -271,6 +277,12 @@ func (c *PubSub) Ping(ctx context.Context, payload ...string) error {
 	return err
 }
 
+func (c *PubSub) SetHealthTimeout(t time.Duration) {
+	if t > 0 {
+		c.healthTimeout = t
+	}
+}
+
```
vmihailenco, line 285: `SetHealthTimeout` only makes sense for `Channel`, but is set on `PubSub`. It is confusing.

rbg:
> I've recently refactored similar code for [Bun](https://github.com/uptrace/bun/blob/master/driver/pgdriver/listener.go#L224-L254). As you can see, there is a `Channel` method that accepts an arbitrary number of options. All the code that is required to support it is extracted to the `channel` struct. It is the 2nd option described in https://github.com/go-redis/redis/issues/1700
> 
> 
> 
> Do you think we can do the same thing in go-redis? We can't just copy-paste the code as is because PubSub in Redis works differently, but we can use the same API.

That is much clearer

monkey92t:
@vmihailenco 

I understand #1700, but there is still a problem with `ChannelSize(size int)`.
If we replace `ChannelSize(size int)` with #1700(`Chan(cfg ChanConfig)`), it will break the compatibility of the v8 version.
#1700 can work in v9.
I want to find a solution for the current v8 version (although not as good as #1700), but it can solve the current problems.

I can remove `SetHealthTimeout` and set it to a fixed value, such as 5 seconds.

vmihailenco:
>I can remove SetHealthTimeout and set it to a fixed value, such as 5 seconds.

I am fine with this and a finding "quick" temp solution in general.

>If we replace ChannelSize(size int) with #1700(Chan(cfg ChanConfig)), it will break the compatibility of the v8 version.
#1700 can work in v9.

I think we should change existing `Channel` to `Channel(opts ...ChannelOption)` (which is backwards compatible). And then change `ChannelSize` to call `Channel`:

```go
func (c *Channel) Channel(opts ...ChannelOption) <-chan *Message {}

func (c *Channel) ChannelSize(size int)  <-chan *Message {
    return c.Channel(redis.WithChannelSize(size))
}
```

All that should be backwards compatible.

monkey92t:
Fixed by #1764

PR #1751, opened 2021-05-12
Allow for custom latency and error handling

keitwb:
The standard method of simply calling the PING command does not take
into account failures of the ping method.  The added client option
`LatencyHealthFunc(c *Client) bool` allows for failures to be taken into
account, along with LOADING errors when using persistent Redis.  The
default behavior has been kept exactly the same as before.

Also adds a client option `OnErrFunc(n NodeExt, err error)` that exposes
a very restricted Node interface to a custom function that can mark the
node as failed (or choose not to) based on the error gotten.  E.g. a
connection refused error could cause the node to be marked as failed,
unlike the current client behavior.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -21,6 +21,8 @@ import (
 
 var errClusterNoNodes = fmt.Errorf("redis: cluster has no nodes")
 
+const maxLatency = math.MaxUint32
+
 // ClusterOptions are used to configure a cluster client and should be
 // passed to NewClusterClient.
 type ClusterOptions struct {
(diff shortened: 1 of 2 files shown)


keitwb:
I think the remaining lint issues are from unrelated code.

vmihailenco:
As is it is just a bunch of hooks that probably allow fixing your specific problem but are confusing/opaque for everyone else. It is not clear how to use this and even the problem you are trying to solve ("failures of the ping method" does not tell me much).

Issue #1686, opened 2021-03-05
add xautoclaim command


justinrush:
## Expected Behavior

It would be nice to have the [XAUTOCLAIM](https://redis.io/commands/xautoclaim) command to reduce the calls to claim pending messages to 1.

## Current Behavior

You need to issue an XPENDING command and then an XCLAIM command

## Possible Solution

Add a XAutoClaim command.

PR #1780, opened 2021-06-02
Add XAutoClaim command

ericmillin:
Closes https://github.com/go-redis/redis/issues/1686

Redis docs: https://redis.io/commands/xautoclaim

I don't think `ID` is the right name for the cursor returned by Redis, but not sure what would be better. https://github.com/go-redis/redis/pull/1780/files#diff-4d7d1923693fc5ce892add2ea2907a744e77ea0b50c1939ccc5067cb48a466a3R1505

### DIFF
--- a/command.go
+++ b/command.go
@@ -1501,14 +1501,11 @@ func (cmd *XPendingExtCmd) readReply(rd *proto.Reader) error {
 
 //------------------------------------------------------------------------------
 
-type XAutoClaim struct {
-	ID       string
-	Messages []XMessage
-}
-
 type XAutoClaimCmd struct {
 	baseCmd
-	val XAutoClaim
+
+	start string
+	val   []XMessage
 }
 
 var _ Cmder = (*XAutoClaimCmd)(nil)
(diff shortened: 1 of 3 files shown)


ericmillin:
@monkey92t The assertion that failed passes locally. Are there any common race conditions you run into?

https://github.com/go-redis/redis/pull/1780/checks?check_run_id=2733660581#step:5:1586

monkey92t:
@ericmillin Yes, there may be an unexpected data race, you can ignore it for now.

monkey92t:
But the current workflow error is not accidental, it needs to be resolved
```go
      Expected
          <string>: 2-0
      to equal
          <string>: 3-0

      /home/runner/work/redis/redis/commands_test.go:4399
```

monkey92t:
@ericmillin  it should be 2-0
```go
1) 1) "stream"
   2) 1) 1) "1-0"
         2) 1) "uno"
            2) "un"
      2) 1) "2-0"
         2) 1) "dos"
            2) "deux"
      3) 1) "3-0"
         2) 1) "tres"
            2) "troix"
127.0.0.1:6379> XAutoClaim stream group consumer 0 - count 2
1) "2-0"
2) 1) 1) "1-0"
      2) 1) "uno"
         2) "un"
   2) 1) "2-0"
      2) 1) "dos"
         2) "deux"
127.0.0.1:6379> 

```

ericmillin:
@monkey92t  Do you have any opinions on this: https://github.com/go-redis/redis/pull/1780/files#diff-4d7d1923693fc5ce892add2ea2907a744e77ea0b50c1939ccc5067cb48a466a3R1556-R1559

Seems to match the Redis semantics, but looks a bit confusing. I'm fairly new to Redis, though, hence my hesitation.

monkey92t:
I made a little modification. What do you think of the new return value method? Any better suggestions?

ericmillin:
@monkey92t That looks great and had crossed my mind. If you're good with it, let's merge it.

Thanks for all your help :)

monkey92t:
OK, thanks

Issue #1789, opened 2021-06-08
Add support for GT and LT in ZADD


chorin1:
## Expected Behavior
Since Redis 6.2 [ZADD](https://redis.io/commands/ZADD) has additional useful options - `LT` (less than) and `GT` (greater than). 
There should be corresponding functions in go-redis -> `ZAddLT` & `ZAddGT`.

## Possible Implementation

Since GT, LT and NX options are mutually exclusive there are two ways to go from here:
1. Add ZAddLT/GT + ZIncrGT/LT + ZAddXXGT/LT (total of 6 functions)
2. Have a single ZAdd function and introduce a Zadd options struct => breaking change.

chorin1:
If there's interest, I can cook up a PR

monkey92t:
We can keep the more commonly used API, such as:
`zadd key score member...`
`zaddNX..`

In addition, we can add an API that supports all parameters, such as `ZAddArgs(ctx, key, *ZAddArgs)`
What do you think?

vmihailenco:
+1 to `ZAddArgs(ctx, key, ZAddArgs)` (not a `*ZAddArgs` since it is a small struct)

monkey92t:
@chorin1 HI~
I am currently upgrading the `sorted set` series of commands. If you have no plans at this time, I will modify the `ZADD` command

chorin1:
@monkey92t go ahead.
I saw that you already transitioned the item to "in-progress" so I didn't start any work on my end.
LMK if you need help.

PR #1792, opened 2021-06-23
Upgrade the <sorted set> series of commands

monkey92t:
New commands and options:

- ZAdd GT and LT options ( #1789 )
- ZRange ByScore,ByLex,Rev and Limit options
- New cmd ZRangeStore
- New cmd ZUnion

  
New API：

- ZAddArgs 
- ZAddArgsIncr 
- ZRangeArgs 
- ZRangeArgsWithScores 
- ZRangeStore 
- ZUnion 
- ZUnionWithScores 

API marked `Deprecated` and removed in V9:

- ZAddNX 
- ZAddXX
- ZAddCh
- ZIncr
- ZAddNXCh
- ZAddXXCh
- ZIncrNX
- ZIncrXX

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -248,12 +248,27 @@ type Cmdable interface {
 
 	BZPopMax(ctx context.Context, timeout time.Duration, keys ...string) *ZWithKeyCmd
 	BZPopMin(ctx context.Context, timeout time.Duration, keys ...string) *ZWithKeyCmd
+
+	// TODO: remove
+	// 		ZAddNX
+	//		ZAddXX
+	//		ZAddCh
+	//		ZIncr
+	//		ZAddNXCh
+	//		ZAddXXCh
+	//		ZIncrNX
+	//		ZIncrXX
+	// 	in v9.
+	// 	use ZAddArgs and ZAddArgsIncr.
+
 	ZAdd(ctx context.Context, key string, members ...*Z) *IntCmd
 	ZAddNX(ctx context.Context, key string, members ...*Z) *IntCmd
 	ZAddXX(ctx context.Context, key string, members ...*Z) *IntCmd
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -2106,96 +2142,181 @@ func (c cmdable) BZPopMin(ctx context.Context, timeout time.Duration, keys ...st
...
-	const n = 3
-	a := make([]interface{}, n+2*len(members))
-	a[0], a[1], a[2] = "zadd", key, "nx"
-	return c.zAdd(ctx, a, n, members...)
+// ZAddArgs WARN: The GT, LT and NX options are mutually exclusive.
+type ZAddArgs struct {
+	NX      bool
+	XX      bool
+	LT      bool
+	GT      bool
```
vmihailenco, line 2150: What do you think about replacing this 4 options with:

```go
    Mode string // either NX, XX, LT, or GT
```

vmihailenco: Nevermind, it is possible to use XX LT together.

commands.go
```diff
@@ -248,12 +248,27 @@ type Cmdable interface {
 
 	BZPopMax(ctx context.Context, timeout time.Duration, keys ...string) *ZWithKeyCmd
 	BZPopMin(ctx context.Context, timeout time.Duration, keys ...string) *ZWithKeyCmd
+
+	// TODO: remove
+	// 		ZAddNX
+	//		ZAddXX
```
vmihailenco, line 254: I would keep these 2 in v9. They looks like a popular option.

vmihailenco (approved):

commands.go
monkey92t: OK, I will make adjustments

commands.go
monkey92t: Then we don't change it

PR #1803, opened 2021-06-28
Connect hook

pfreixes:
Addresses https://github.com/go-redis/redis/discussions/1786

Added a new `PoolHook` interface allowing the user to add two new hooks that would be called before and after the creation of a connection, allowing the user - among other stuff - to measure the connection latency.

User still uses the `AddHook` interface but providing an extended interface which supports the two methods required by the pool.

An example can be found here https://github.com/go-redis/redis/pull/1803/files#diff-ee4b7901a404913de455c4b45c583a3d0bf51e6c3abc06c93c2a1fa6c3eac235R16

# What is missing

Does not support the `clone` behaviour used by functions like `Client.WithTimeout` which for the client hooks it creates a list by using a shallow copy, decoupling hooks from the former and the new cloned client.

Current proposal does not support this, since the pool is shared by all of the cloned clients.

### DIFF
--- a/example_instrumentation_test.go
+++ b/example_instrumentation_test.go
@@ -5,11 +5,20 @@ import (
 	"fmt"
 
 	"github.com/go-redis/redis/v8"
+	"github.com/go-redis/redis/v8/internal/pool"
 )
 
+type contextKey string
+
+var key = contextKey("foo")
+
 type redisHook struct{}
+type redisFullHook struct {
+	redisHook
+}
 
 var _ redis.Hook = redisHook{}
+var _ redis.Hook = redisFullHook{}
 
 func (redisHook) BeforeProcess(ctx context.Context, cmd redis.Cmder) (context.Context, error) {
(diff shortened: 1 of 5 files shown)


vmihailenco:
Have you considered using OpenTelemetry Metrics to measure the data you need? We already have some metrics and probably should not wait for a stable release to add more.

I understand that the more hooks the better, but this PR adds considerable amount of code and complexity.

pfreixes:
We are already measuring op latency and errors by using the current hooks, which I guess is what OpenTelemetry would provide. I would like to make and step forward and also measure create connection latency.

If this PR does not not move forward, we will do that by overriding the dialer.

vmihailenco:
:+1: I just meant that OpenTelemetry should work (at least in theory) with Prometheus/Grafana/whatever. So may be adding it to your stack could help.

pfreixes:
Ill close the PR, if eventually other users are interested we can reborn this.

vmihailenco:
Sounds good - thanks :+1:

PR #1814, opened 2021-07-07
Added RPopCount support to mirror LPopCount

devinsburke:
Added RPopCount support to mirror LPopCount. This satisfies an issue I opened a half hour ago.

PR #1815, opened 2021-07-07
Add the count option to the rpop command(fix #1813)

### LINKED ISSUE #1813: Add commands.go method for RPopCount to mirror LPopCount
Add commands.go method for RPopCount to mirror existing LPopCount. Both command support the count argument, according to https://redis.io/commands/RPOP

monkey92t:
fix #1813


monkey92t:
repeat #1814

Issue #1819, opened 2021-07-13
LIFO connection pool does not give every connection equal chance of being chosen


ktaekwon000:
## Introduction

Currently, the connection pool of go-redis is implemented as a LIFO stack. This has caused some problem within our organisation when using our library, where we use a single redis client to connect to multiple proxy servers for concurrency purposes. (tldr: each connection in the go-redis connection pool is a connection to a single redis proxy, which then acts as an abstraction for a redis cluster behind it. One client connects to multiple redis proxies and ideally the load from the client will be balanced between the proxies). 

From some testing, we found that with the LIFO connection pool, the connections at the end of the array have a disproportionate chance of being selected rather than other connections in the connection pool. This causes an issue with our setup, where we depend on the connection pool for load balancing.

## Expected Behavior

Each connection in the connection pool has similar chances of being used.

## Current Behavior

Some connections in the connection pool are used less often than others.

## Possible Solution

We have tested both of these solutions in our environment and verified that both of these solutions fix the issue at hand.

### Option 1: Randomized, simpler, involves minimal changes

![image](https://user-images.githubusercontent.com/65844703/125382695-b97eee80-e3c8-11eb-954d-f73b7b9ae41b.png)
When a connection is returned to the connection pool, put it in an arbitary slot in the connection pool. The next connection to be returned is randomly chosen from the connection pool too, with this logic.

The primary downside of this logic is that the current tests will not apply to this kind of connection pool any more, as the existing tests depend on the ordering of the connections in the pool being stable.

### Option 2: Convert idleConns into a linked list, convert the connection pool into a FIFO queue

![image](https://user-images.githubusercontent.com/65844703/125383126-8426d080-e3c9-11eb-9071-2a89a57348e7.png)
![image](https://user-images.githubusercontent.com/65844703/125383198-a1f43580-e3c9-11eb-8715-d97f3a66acb0.png)
When retrieving an idle connection from the connection pool, we can retrieve the connection in a FIFO manner, so that the connection pool is implemented in a round robin fashion, where for a connection that has been put into a connection pool to be get again, all existing connections in the pool must be used before that.

The primary downside to this is the complicated nature of the code change, as changing the data structure of idleConns will cause extensive changes in the connection pool. However, as the order of the connections in the connection pool is deterministic, the existing tests can be modified for the new type of connection pool.

## Steps to Reproduce

1. Have a DNS resolver that resolves multiple IP addresses from a single hostname
2. Create a redis client with poolSize set in the options.
[...]

PR #1820, opened 2021-07-13
Allow FIFO pool in redis client

ktaekwon000:
Linked issue: #1819 

Allows the user to choose between fifo and lifo pool via passing in the PoolType variable into redis.Options.

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -11,7 +11,12 @@ import (
 	"github.com/go-redis/redis/v8/internal"
 )
 
+type PoolType string
+
 var (
+	PoolLifo PoolType = ""
+	PoolFifo PoolType = "fifo"
+
 	// ErrClosed performs any operation on the closed client will return this error.
 	ErrClosed = errors.New("redis: client is closed")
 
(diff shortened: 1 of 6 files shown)


options.go
```diff
@@ -76,6 +76,10 @@ type Options struct {
 	// Default is ReadTimeout.
 	WriteTimeout time.Duration
 
+	// Type of Pool
+	// Default is lifo, supports fifo too
+	// Note that fifo has higher overhead compared to lifo
+	PoolType pool.PoolType
```
vmihailenco, line 82: `pool` is a private package. Perhaps `PoolFIFO bool`?

ktaekwon000: Thanks for the headsup, I'll do that .

ktaekwon000: I've changed it, along with the golint issues.

internal/pool/pool.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/go-redis/redis/v8/internal"
 )
 
+type PoolFIFO bool
```
monkey92t, line 14: It seems redundant

monkey92t:
Other clients, we should also add it, for example: cluster, ring, sentinel, universal...

PR #1823, opened 2021-07-15
Backport wrapMultiExec() cmds copy bug fix from v8

wspowell:


wspowell:
This is probably the same bug that was being hit in issue 1386.

monkey92t:
We did not plan to fix v7 before, it will be very difficult, and we have removed `travis CI`

@vmihailenco What do you think?

vmihailenco:
Since out time is limited, we don't support old versions. But we do accept backports from users if tests are passing.

So I think we should merge the first [commit](https://github.com/go-redis/redis/pull/1823/commits/193fcc450425655941982615d08e6e8201386db1), run tests locally (since travis is not working), release a new v7 version, and forget about it (i.e. don't try to fix build / backport other changes). Does that sound good?

monkey92t:
Yes, I think this is a good idea, I have removed the extra.

monkey92t:
I performed the test locally and found no problems, I will merge it and release a new version

vmihailenco:
Thanks :+1:

PR #1825, opened 2021-07-16
Check Failing() before serving random node

AnatolyRugalev:
We encounter errors during Redis failover with `RouteRandomly` set to `true`.

After some debugging we found that `Failing` is never checked when serving random node for the slot.

This should be pretty easy to reproduce:

1. Start cluster using this docker-compose service:
```
  redis-cluster:
    image: grokzen/redis-cluster:6.0.9
    container_name: redis-cluster.go-redis-test
    environment:
      - "IP=0.0.0.0"
    ports:
      - 7000-7005:7000-7005
```
2. Connect to cluster, perform read/write operations
3. `docker-compose exec redis-cluster supervisorctl stop redis-2`

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -595,8 +595,13 @@ func (c *clusterState) slotRandomNode(slot int) (*clusterNode, error) {
 	if len(nodes) == 0 {
 		return c.nodes.Random()
 	}
-	n := rand.Intn(len(nodes))
-	return nodes[n], nil
+	for _, idx := range rand.Perm(len(nodes)) {
+		if nodes[idx].Failing() {
+			continue
+		}
+		return nodes[idx], nil
+	}
+	return c.nodes.Random()
 }
 
 func (c *clusterState) slotNodes(slot int) []*clusterNode {
--- a/cluster.go
+++ b/cluster.go
@@ -598,12 +598,13 @@ func (c *clusterState) slotRandomNode(slot int) (*clusterNode, error) {
 	if len(nodes) == 1 {
 		return nodes[0], nil
 	}
-	for _, idx := range rand.Perm(len(nodes)) {
+	randomNodes := rand.Perm(len(nodes))
+	for _, idx := range randomNodes {
 		if node := nodes[idx]; !node.Failing() {
 			return node, nil
 		}
 	}
-	return nodes[0], nil
+	return nodes[randomNodes[0]], nil
 }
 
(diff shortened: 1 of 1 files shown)


bokan (approved):

cluster.go
```diff
@@ -595,8 +595,13 @@ func (c *clusterState) slotRandomNode(slot int) (*clusterNode, error) {
 	if len(nodes) == 0 {
 		return c.nodes.Random()
 	}
-	n := rand.Intn(len(nodes))
-	return nodes[n], nil
+	for _, idx := range rand.Perm(len(nodes)) {
+		if nodes[idx].Failing() {
+			continue
```
ferhatelmas, line 600: good to revert logic and return early

ferhatelmas (approved):

cluster.go
```diff
@@ -595,8 +595,12 @@ func (c *clusterState) slotRandomNode(slot int) (*clusterNode, error) {
...
 		return c.nodes.Random()
 	}
-	n := rand.Intn(len(nodes))
-	return nodes[n], nil
+	for _, idx := range rand.Perm(len(nodes)) {
+		if !nodes[idx].Failing() {
+			return nodes[idx], nil
+		}
+	}
+	return c.nodes.Random()
```
vmihailenco, line 603: Nice catch - thanks :+1: 

Does it make sense to try random node here? It will result in MOVED error and will re-try the command again on the same failing nodes.

I think we should:
- remove `c.nodes.Random()` and just use first/random failing node as we did previously
- add an optimization for `len(nodes) == 1` since `rand.Perm` is somewhat expensive

```go
if len(nodes) == 1 {
    return nodes[0]
}
```

cluster.go
```diff
@@ -595,8 +595,12 @@ func (c *clusterState) slotRandomNode(slot int) (*clusterNode, error) {
 	if len(nodes) == 0 {
 		return c.nodes.Random()
 	}
-	n := rand.Intn(len(nodes))
-	return nodes[n], nil
+	for _, idx := range rand.Perm(len(nodes)) {
+		if !nodes[idx].Failing() {
```
vmihailenco, line 599: Slightly better:

```go
if node := nodes[idx]; !node.Failing() {
    return node, nil
}
```

monkey92t (approved):

monkey92t:
Thanks!

monkey92t:
There are still some other issues that need to be resolved, we will release a new version later.

PR #1826, opened 2021-07-19
Set up Dependabot for gomod and github-actions dependencies

mmorel-35:


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: weekly
```
vmihailenco, line 6: Weekly is probably rare enough, but we can consider increasing this interval to monthly to lower the noise created by dependabot commits.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
```
vmihailenco, line 10: So this updates the versions of Github actions itself, right? Interesting...

vmihailenco (approved):

.github/dependabot.yml
mmorel-35: Exactly! You can learn more about this [here](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot)

.github/dependabot.yml
mmorel-35: I doubt there will be much noise as the number of dependencies is quite low

monkey92t (approved):

PR #1836, opened 2021-07-27
add cmd: geosearch, geosearchstore

monkey92t:
`GeoRadius` and `GeoRadiusByMember` have been deprecated, we will not change them anymore.

### DIFF
--- a/command.go
+++ b/command.go
@@ -2662,7 +2662,7 @@ type GeoSearchQuery struct {
 	CountAny bool
 }
 
-type GeoSearchWithOptionsQuery struct {
+type GeoSearchLocationQuery struct {
 	GeoSearchQuery
 
 	WithCoord bool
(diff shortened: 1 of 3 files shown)


command.go
```diff
@@ -2637,6 +2637,188 @@ func newGeoLocationParser(q *GeoRadiusQuery) proto.MultiBulkParse {
...
+	baseCmd
+
+	opt GeoSearchWithOptionsQuery
+	val []GeoLocation
+}
+
+var _ Cmder = (*GeoSearchLocationCmd)(nil)
+
+func NewGeoSearchLocationCmd(
+	ctx context.Context, opt GeoSearchWithOptionsQuery, args ...interface{},
```
vmihailenco, line 2741: Is not `GeoSearchWithOptionsQuery` struct too big to be passed as value? Should not it be `*GeoSearchWithOptionsQuery`?

commands.go
```diff
@@ -383,6 +383,9 @@ type Cmdable interface {
 	GeoRadiusStore(ctx context.Context, key string, longitude, latitude float64, query *GeoRadiusQuery) *IntCmd
 	GeoRadiusByMember(ctx context.Context, key, member string, query *GeoRadiusQuery) *GeoLocationCmd
 	GeoRadiusByMemberStore(ctx context.Context, key, member string, query *GeoRadiusQuery) *IntCmd
+	GeoSearch(ctx context.Context, key string, q GeoSearchQuery) *StringSliceCmd
+	GeoSearchWithOptions(ctx context.Context, key string, q GeoSearchWithOptionsQuery) *GeoSearchLocationCmd
```
vmihailenco, line 387: We could rename `GeoSearchWithOptions` to `GeoSearchLocation`. The only problem is that Redis can add a command with such name, but it does not look very likely (they will add another 10 options instead :)

vmihailenco:
Looks good - thanks!

vmihailenco (approved):

Issue #1737, opened 2021-04-23
Retries may be done using broken connections from the pool


julianbrost:
## Expected Behavior

When restarting the Redis server, due to the default of `MaxRetries: 3`, I'd expect go-redis to attempt to reconnect to the server and retry the query.

## Current Behavior

If there are enough old connections in the connection pool, all retry attempts are done on broken connections and an error is returned without even attempting to reconnect.

## Possible Solution

Some options:
1. Perform a ping on the connection before reusing it (adds latency if the connection is fine)
2. Don't count retries on connections from the pool towards the retry limit (might lead to a huge number of retries for a single query)
3. Clear the pool on errors (might be overly pessimistic)
4. Always use a fresh connection for retries

## Steps to Reproduce

The following code reproduces the issue. Instead of restarting the server, it sets a dialer that simply closes each connection after 5 seconds.

```go
package main

import (
	"context"
	"github.com/go-redis/redis/v8"
	"log"
	"net"
	"time"
)

func main() {
	ctx := context.Background()

	// Redis client with a dialer that kills connections after 5 seconds (to simulate a server restart).
	client := redis.NewClient(&redis.Options{
		Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
			log.Print("Dialer called")
			conn, err := net.Dial(network, addr)
			if err == nil {
				go func() {
					time.Sleep(5 * time.Second)
					conn.Close()
				}()
			}
			return conn, err
		},
	})

	// Function to ping the server and log errors if they occur.
	ping := func() {
		_, err := client.Ping(ctx).Result()
		if err != nil {
			log.Print(err)
		}
	}

	// Perform some pings to fill the connection pool.
	for i := 0; i < 10; i++ {
		go ping()
	}

	// Wait for connections to die.
	time.Sleep(10 * time.Second)

	// Perform another ping and log pool stats before and after.
	log.Printf("%#v", client.PoolStats())
	ping()
	log.Printf("%#v", client.PoolStats())
}
```

Example output (note that the dialer is not called for the last ping and the hit count in the pool stats increases by 4, i.e. the initial attempt and all 3 retries were done using stale connections from the pool):
```
2021/04/23 13:40:50 Dialer called
2021/04/23 13:40:50 Dialer called
2021/04/23 13:40:50 Dialer called
2021/04/23 13:40:50 Dialer called
2021/04/23 13:40:50 Dialer called
2021/04/23 13:40:50 Dialer called
[...]

PR #1824, opened 2021-07-16
ConnPool check  fd for bad conns

hidu:
it's work

Benchmark with Ping(No bad conns):

cpu: Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
Benchmark_Ping_before-8   	 4014712	       289.7 ns/op	     164 B/op	       4 allocs/op
Benchmark_Ping_after-8    	 4029973	       286.7 ns/op	     164 B/op	       4 allocs/op

Benchmark_Ping_before-8   	 4039963	       290.5 ns/op	     164 B/op	       4 allocs/op
Benchmark_Ping_after-8    	 4077626	       290.2 ns/op	     164 B/op	       4 allocs/op

Benchmark_Ping_before-8   	 4018032	       285.9 ns/op	     164 B/op	       4 allocs/op
Benchmark_Ping_after-8    	 4074387	       292.3 ns/op	     164 B/op	       4 allocs/op

PR #1837, opened 2021-07-28
Added maxBadConnRetries to limit the number of times to obtain broken…

monkey92t:
This PR is not intended to solve #1737.

It is to limit the number of attempts. If all connections in a connection pool fail, we should not let it detect all connections in the entire connection pool.

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -12,8 +12,7 @@ import (
 )
 
 // maxBadConnRetries is the maximum number of attempts to obtain a valid connection from the connection pool.
-// When this number is exceeded, a new connection is created directly and
-// no longer obtained from the connection pool.
+// When this number is exceeded, a new connection is created directly.
 const maxBadConnRetries = 2
 
 var (
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -11,6 +11,11 @@ import (
 	"github.com/go-redis/redis/v8/internal"
 )
 
+// maxBadConnRetries is the maximum number of attempts to obtain a valid connection from the connection pool.
+// When this number is exceeded, a new connection is created directly and
+// no longer obtained from the connection pool.
+const maxBadConnRetries = 2
+
 var (
 	// ErrClosed performs any operation on the closed client will return this error.
 	ErrClosed = errors.New("redis: client is closed")
@@ -235,7 +240,7 @@ func (p *ConnPool) Get(ctx context.Context) (*Conn, error) {
 		return nil, err
 	}
 
-	for {
+	for i := 0; i < maxBadConnRetries; i++ {
 		p.connsMu.Lock()
 		cn := p.popIdle()
 		p.connsMu.Unlock()
@@ -252,6 +257,10 @@ func (p *ConnPool) Get(ctx context.Context) (*Conn, error) {
 		atomic.AddUint32(&p.stats.Hits, 1)
 		return cn, nil
 	}
+
+	// After the connection pool is empty or after trying maxBadConnRetries times,
+	// we still get a broken connection.
+	// At this time, we directly create a new connection.
 
 	atomic.AddUint32(&p.stats.Misses, 1)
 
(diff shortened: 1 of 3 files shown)


monkey92t:
This is an improvement to #1824.

internal/pool/pool.go
```diff
@@ -235,7 +239,7 @@ func (p *ConnPool) Get(ctx context.Context) (*Conn, error) {
 		return nil, err
 	}
 
-	for {
+	for i := 0; i < maxBadConnRetries; i++ {
```
vmihailenco, line 242: I guess we need some limit here after https://github.com/go-redis/redis/pull/1824 was merged, but this won't help with https://github.com/go-redis/redis/issues/1737.

vmihailenco: And I see that you already realize this :)

vmihailenco:
I think that with this change we may open more than `PoolSize` connections if there are more than `maxBadConnRetries` in the pool. So this does not look right.

I guess we should introduce something like `ErrBadConn` from database/sql and change the code to use it. I don't have more details so it is fine if you leave this for me. And we probably should do this in v9 branch since it is a tricky/risky change. I will try to find some time in August to prepare a PR.

Speaking about v9. I think it is time we start thinking about releasing first beta version so we can start testing it. But before that we should remove/replace deprecated APIs. Could you take care of that?

monkey92t:
I don't quite understand what you mean, there are only two situations where you can create a new connection:

1. The connection pool is empty
2. Get and close maxBadConnRetries=2 broken connections

In both cases, there is no problem creating a new connection.

I agree with `ErrBadConn` plan, but we can't add it in a short time.

Compared to v8.11.0, we have recently added (or fixed) some features, and we need to consider releasing a new version.
But #1824 does not impose some restrictions on `Get() *Conn`, this PR is a temporary risk avoidance solution

monkey92t:
This PR will be replaced by ErrBadConn in V9. I just want users to use the improvements #1811, #1820, #1824, #1825....

vmihailenco:
>In both cases, there is no problem creating a new connection.

Well, you can't create new connections having idle connections in the pool. Otherwise you have > PoolSzie connections in the pool which almost certainly breaks the pool, because `queue chan struct{}` is used for synchronization and it's size is equal to `PoolSize`.

>does not impose some restrictions on Get() *Conn, this PR is a temporary risk avoidance solution

I am not sure it is a big problem since `connCheck` should be fast enough to handle 100 connections. I think it is fine to release v8 as it is now.

monkey92t:
> Well, you can't create new connections having idle connections in the pool. Otherwise you have > PoolSzie connections in the pool which almost certainly breaks the pool, because `queue chan struct{}` is used for synchronization and it's size is equal to `PoolSize`.

I still can’t understand. When I closed maxBadConnRetries=2 connections and created a new connection, why would the connection pool overflow (> PoolSize)?

Suppose now len(idleConns) = 10, maxBadConnRetries=2,
```	
for i := 0; i < maxBadConnRetries; i++ {
		p.connsMu.Lock()
		cn := p.popIdle()
		p.connsMu.Unlock()

		if cn == nil {
			break
		}

		if p.isStaleConn(cn) {
			_ = p.CloseConn(cn)
			continue
		}

		atomic.AddUint32(&p.stats.Hits, 1)
		return cn, nil
	}
```
After execution, len(idleConns) = 8, it seems that there is no problem in creating a new connection.

monkey92t:
> I am not sure it is a big problem since `connCheck` should be fast enough to handle 100 connections. I think it is fine to release v8 as it is now.

OK, I just think it is risky, if you think there is nothing wrong with it, we can close this issue.

vmihailenco:
>After execution, len(idleConns) = 8, it seems that there is no problem in creating a new connection.

There were 8 idle conns at some point in the past, but in the next instant there can be 100 or 0 idle connections because there are other threads that can add/remove connections.

The pool is synchronized around `p.queue chan struct{}`:
1. you must acquire a struct{} from the `p.queue`
2. holding the struct{}, you must use idle conn or open a new conn if there are no idle conns
3. you must put the struct{} to the `p.queue` in the end to let other threads use the pool

Hopefully that helps.

PR #1843, opened 2021-08-02
Update changelog for v9

vmihailenco:
It is not a full list of changes, but it should be a good start


monkey92t (approved):

monkey92t:
Yes, we record important changes in `CHANGELOG.md`, for more trivial, we can record in `Releases` of each version.

PR #1849, opened 2021-08-06
Revert "ConnPool check  fd for bad conns"

monkey92t:
This PR(#1824) will cause a lot of read client -> redis-server io/timeout due to option.ReadTimeout timeout.

Specific process:

Thread A -> GetConn(A) -> SetReadDeadline(default 3 * time.Second) -> ..... -> putConn(A)

After 3 seconds...

Thread B -> GetConn(A) -> connCheck -> Read([]byte) -> io/timeout -> CloseConn

This operation will cause a lot of TIME_WAIT

It should be fixed in v9, and this commit was revoked in v8.

@vmihailenco  @hidu @bingzhuo2008


monkey92t:
@vmihailenco  Do you have a different opinion?

vmihailenco (approved):

vmihailenco:
>It should be fixed in v9, and this commit was revoked in v8.

Agreed, let's consider this for v9.

hidu:
got it

hidu:
before  putConn(A), i think we shuold reset readDeadline  and writeDeadline

I guess adding such a line might solve the problem ？

https://github.com/go-redis/redis/blob/master/internal/pool/conn.go#L66

`func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error) error {
	if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
		return err
	}
  
       defer cn.netConn.SetReadDeadline(time.Time{})  // add this ?
	
       return fn(cn.rd)
}`

 Conn.WithWriter is the same

monkey92t:
> before putConn(A), i think we shuold reset readDeadline and writeDeadline
> 
> I guess adding such a line might solve the problem ？
> 
> https://github.com/go-redis/redis/blob/master/internal/pool/conn.go#L66
> 
> ```go
> func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error) error {
> 	if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
> 		return err
> 	}
> 	defer cn.netConn.SetReadDeadline(time.Time{})  // add this ?
> 	return fn(cn.rd)
> }
> ```
> 
> Conn.WithWriter is the same

Yes, this problem can be solved by setting SetReadDeadline(time) before `putConn` or during `connCheck`.

In the upcoming v9 version, we have redesigned the use of command retry and connection pool.
The reason for revert #1824 is not to be abandoned, but because it is part of v9, but there are some conflicts, we want to add it directly in v9.

> before putConn(A), i think we shuold reset readDeadline and writeDeadline
> 
> I guess adding such a line might solve the problem ？
> 
> https://github.com/go-redis/redis/blob/master/internal/pool/conn.go#L66
> 
> ```go
> func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error) error {
> 	if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
> 		return err
> 	}
> 	defer cn.netConn.SetReadDeadline(time.Time{})  // add this ?
> 	return fn(cn.rd)
> }
> ```
> 
> Conn.WithWriter is the same

Yes, this problem can be solved by setting SetReadDeadline(time) before `putConn` or during `connCheck`.

We did not abandon #1824, but there are too many connections between #1737 and #1824. We will integrate them in v9. 

#1824 needs to work with #1737 to achieve better results.

PR #1852, opened 2021-08-11
Automate release process

vmihailenco:
I've tested the shell scripts on https://github.com/uptrace/bun and they seem to work fine. But I am not sure if release.yml is working properly - we will have to figure it out the hard way :)


monkey92t:
I don’t have much experience in this area, if you know enough, I think we can experiment.

monkey92t (approved):

Issue #1861, opened 2021-08-16
When redis client is closed before all goroutine is completed,p.poolSize and p.idleConnsLen will overflow maybe


mstmdev:
https://github.com/go-redis/redis/blob/bc9d5c8d3af4780baecc5b76e3f37a9c3d18ab42/internal/pool/pool.go#L117-L134

https://github.com/go-redis/redis/blob/bc9d5c8d3af4780baecc5b76e3f37a9c3d18ab42/internal/pool/pool.go#L412

When redis client is closed before all goroutine is completed,p.poolSize and p.idleConnsLen will overflow maybe.

monkey92t:
Yes, but it seems to have a bigger problem, I need to do some experiments.

monkey92t:
fixed by #1863

PR #1863, opened 2021-08-17
after the connection pool is closed, no new connections should be added

monkey92t:
linked: #1861

When the connection pool is closed, we will still add new connections to the connection pool, and this part of the connection cannot be closed.

It may also cause errors such as `Stats`, as described in #1861

I added a count of `runGoroutine`, I am not sure if adding it is a good decision, although it is only for testing.

@vmihailenco Do you have any ideas?

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -89,6 +89,11 @@ type ConnPool struct {
 
 	_closed  uint32 // atomic
 	closedCh chan struct{}
+
+	// runGoroutine is the number of goroutines running in the connection pool.
+	// When the connection pool is closed, it should be zero.
+	// It is only for testing.
+	runGoroutine int32 // atomic
 }
 
 var _ Pooler = (*ConnPool)(nil)
@@ -108,6 +113,7 @@ func NewConnPool(opt *Options) *ConnPool {
 	p.connsMu.Unlock()
 
 	if opt.IdleTimeout > 0 && opt.IdleCheckFrequency > 0 {
+		atomic.AddInt32(&p.runGoroutine, 1)
 		go p.reaper(opt.IdleCheckFrequency)
 	}
 
@@ -121,14 +127,17 @@ func (p *ConnPool) checkMinIdleConns() {
 	for p.poolSize < p.opt.PoolSize && p.idleConnsLen < p.opt.MinIdleConns {
 		p.poolSize++
 		p.idleConnsLen++
+
+		atomic.AddInt32(&p.runGoroutine, 1)
 		go func() {
 			err := p.addIdleConn()
-			if err != nil {
+			if err != nil && err != ErrClosed {
 				p.connsMu.Lock()
 				p.poolSize--
 				p.idleConnsLen--
 				p.connsMu.Unlock()
 			}
+			atomic.AddInt32(&p.runGoroutine, -1)
 		}()
 	}
 }
@@ -140,9 +149,16 @@ func (p *ConnPool) addIdleConn() error {
 	}
 
 	p.connsMu.Lock()
+	defer p.connsMu.Unlock()
+
+	// It is not allowed to add new connections to the closed connection pool.
+	if p.closed() {
+		_ = cn.Close()
+		return ErrClosed
+	}
+
 	p.conns = append(p.conns, cn)
 	p.idleConns = append(p.idleConns, cn)
-	p.connsMu.Unlock()
 	return nil
 }
 
@@ -157,6 +173,14 @@ func (p *ConnPool) newConn(ctx context.Context, pooled bool) (*Conn, error) {
 	}
 
 	p.connsMu.Lock()
+	defer p.connsMu.Unlock()
+
+	// It is not allowed to add new connections to the closed connection pool.
+	if p.closed() {
+		_ = cn.Close()
+		return nil, ErrClosed
+	}
+
 	p.conns = append(p.conns, cn)
 	if pooled {
 		// If pool is full remove the cn on next Put.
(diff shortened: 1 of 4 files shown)


internal/pool/pool.go
```diff
@@ -140,9 +149,16 @@ func (p *ConnPool) addIdleConn() error {
 	}
 
 	p.connsMu.Lock()
+	defer p.connsMu.Unlock()
+
+	// It is not allowed to add new connections to the closed connection pool.
+	if p.closed() {
+		_ = cn.Close()
+		return ErrClosed
+	}
```
vmihailenco, line 158: I believe this is the only change we need. I agree that `runGoroutine` should be removed.

vmihailenco:
This looks good, but let's remove `runGoroutine` :+1:

internal/pool/pool.go
```diff
@@ -121,14 +127,17 @@ func (p *ConnPool) checkMinIdleConns() {
 	for p.poolSize < p.opt.PoolSize && p.idleConnsLen < p.opt.MinIdleConns {
```
vmihailenco, line 127: You could also add `&& !p.closed()` just in case.

vmihailenco: Or even better, check `p.closed()` in `popIdle`?

monkey92t: `checkMinIdleConns` is executed after `connsMu.Lock()`, it protects `poolSize` and `idleConnsLen`, we only need to ensure their correctness in the goroutine (func).

```go
if p.opt.MinIdleConns == 0 || p.closed() {
	return
}
```
We can add a `p.closed()` judgment here, but its function is only to not start the goroutine.

vmihailenco: Perhaps adding something like this to `Pool.Get`:

```
		p.connsMu.Lock()
                if p.Closed() {
                    return nil, ErrClosed
                }
		cn := p.popIdle()
		p.connsMu.Unlock()
```

monkey92t: Although `p.Get()` can still get the connection after `Close()`, it will not cause the connection pool to leak.

monkey92t: We have added detection (`if p.closed()`) in `p.newConn()`, `p.Get()` will not add unmanaged new connections to the closed connection pool.

monkey92t:
I try to do more experiments to see if there are similar problems in the connection pool.

vmihailenco:
Thanks, and let me know when to take another look.

monkey92t:
We will create >poolSize connections.

test example:
```go
ctx := context.Background()
var n int32
connPool := pool.NewConnPool(&pool.Options{
	Dialer: func(ctx context.Context) (net.Conn, error) {
		atomic.AddInt32(&n, 1)
		time.Sleep(10 * time.Millisecond)
		return &net.TCPConn{}, nil
	},
	PoolSize:           10,
	PoolTimeout:        time.Hour,
	IdleTimeout:        time.Minute,
	IdleCheckFrequency: time.Minute,
	MinIdleConns:       10,
})
perform(10, func(_ int) {
	cn, err := connPool.Get(ctx)
	if err != nil {
		t.Fatal(err)
	}
	connPool.Put(ctx, cn)
})
t.Log(atomic.LoadInt32(&n))

// poolSize: 10
// output: 20
```

vmihailenco:
It is expected that `addIdleConns` can create extra connections. Such connections are maked with `pooled = false` and are removed on then next `Put` - see https://github.com/go-redis/redis/blob/e5571ea13ffc16774e02090872024111917cfd59/internal/pool/pool.go#L178-L180

I could not come with a better/cleaner implementation, because the requirement is to not hold the lock when dialing/closing a new connection (which can potentially block for long time).

monkey92t:
> It is expected that `addIdleConns` can create extra connections. Such connections are maked with `pooled = false` and are removed on then next `Put` - see

Yes, I noticed here. but... we should not create a connection with `pooled=true` when the connection pool is full(`p.poolSize` == `opt.PoolSize`).

If we want to fix this, the changes to the connection pool are too large, but it should have the following characteristics:

1. Have the existing efficiency and security of the connection pool.
2. Never create more connections than `opt.PoolSize` (except `pooled=false`).
3. After the connection pool is closed, we should not create a new connection, otherwise it will cause the connection to leak.
4. When creating a new connection, it should not be locked, at least the `p.connsMu` lock should not be used.

vmihailenco:
I am not sure that is possible with the current design and it looks like a different/separate problem to me. It is expected that number of connections in a pool can exceed PoolSize for a brief period, but the size of a pool should shrink down quick enough to not cause any harm.

I would merge `p.closed()` checks in this PR since they improve current situation and then would try to come up with a better pool design in a different PR.

monkey92t:
OK, good

monkey92t:
I reopened the PR, do you have anything else to add? (check p.closed)

internal/pool/pool.go
```diff
@@ -115,15 +115,16 @@ func NewConnPool(opt *Options) *ConnPool {
 }
 
 func (p *ConnPool) checkMinIdleConns() {
-	if p.opt.MinIdleConns == 0 {
+	if p.opt.MinIdleConns == 0 || p.closed() {
```
vmihailenco, line 118: What about moving this from here to `popIdle` where it makes more sense? Something like:

```go
func (p *ConnPool) popIdle() (*Conn, error) {
    if p.closed() {
        return nil, ErrClosed
    }
}
```

vmihailenco (approved):
Have a small suggestions but overall looks good.

monkey92t:
Oh...golangci-lint released a new version... forward

monkey92t:
This error seems to be a false positive

PR #1867, opened 2021-08-24
Add go 1.17 to the build matrix

kishaningithub:


.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.15.x, 1.16.x]
+        go-version: [1.15.x, 1.16.x, 1.17.x]
```
vmihailenco, line 15: Thanks :+1:  Could you also remove 1.15.x?

kishaningithub: Done

kishaningithub:
@vmihailenco Should it be identitical to what we have for go-redis? Also, Do we want golang-cli lint as a separate pipeline or can it be a step within the same pipeline

vmihailenco:
>@vmihailenco Should it be identitical to what we have for go-redis?

I assume this question is about go-redis/rate. Yes, something similar would be fine.

>Also, Do we want golang-cli lint as a separate pipeline or can it be a step within the same pipeline

Separate pipelines can be run in parallel which is nice. Also it is simple enough so why not?

vmihailenco:
Build is passing - merging and thanks.

PR #1869, opened 2021-08-25
Introduce full FIFO pool implementation

jasonjoo2010:
## Key Points

- Using interface instead of specific pointer type
- Introduce FIFO connection pool implementation
- Add necessary tests

## Notes

In this PR the `PoolFIFO bool` is replaced with a more extensible `PoolType`. It just follows our previous implementation and I don't insist on it. But I think a `type` makes more sense than a `boolean`. Certainly, we should also consider the backward compatibility.

## Detail Design

Please refer to Design on the FIFO connection pool (#1868)

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -68,9 +68,10 @@ type ClusterOptions struct {
 	ReadTimeout  time.Duration
 	WriteTimeout time.Duration
 
-	// PoolFIFO uses FIFO mode for each node connection pool GET/PUT (default LIFO).
-	PoolFIFO bool
-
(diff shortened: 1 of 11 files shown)


jasonjoo2010:
> Yep, and that is even more code which is why I guess it rarely happens in practice and people continue to use slices...

Yeah. Or just make it simple, how about open the ability to use a user implemented connection pool when creating a client?
Since we define the interface of the pool but never get it used inside.

vmihailenco:
That is a possibility but requires moving and exporting everything (including *Conn) from internal/pool.

Issue #1874, opened 2021-08-31
Data Race when using PubSub.Unsubscribe() after PubSub.Channel()


YongJieYongJie:
After creating channel using `PubSub.Channel()`, subsequent calls to `PubSub.Unsubscribe()` may result in data race.

## Expected Behavior

There shouldn't be a data race.

## Current Behavior

In `initChannel()`, there is a goroutine that periodically calls `PubSub.Ping()`, which causes data race when there is a separate call to `PubSub.Unsubscribe()` in another goroutine.

## Possible Solution

Perhaps acquire the mutex lock in `Ping()` too:
```go
	c.mu.Lock()
	defer c.mu.Unlock()
```

## Steps to Reproduce

Please see the minimal working example to recreate the race condition: https://github.com/YongJieYongJie/redis/commit/115dcfc24a72a5fca4f3343623eae53391c82079
1. Checkout the branch above
2. Run `go mod init github.com/go-redis/redis && go mod tidy`
3. Run `GORACE="halt_on_error=1" go test -v -race -run TestDataRaceRecreate`
4. The test should result in error like the following:
```
==================
WARNING: DATA RACE
Read at 0x00c000021b28 by goroutine 11:
  bufio.(*Writer).Flush()
      /usr/local/go/src/bufio/bufio.go:608 +0x16b
  github.com/go-redis/redis/internal/proto.(*Writer).Flush()
      /<TRUNCATED>/go-redis/redis/internal/proto/writer.go:158 +0xc7
  github.com/go-redis/redis/internal/pool.(*Conn).WithWriter()
      /<TRUNCATED>/go-redis/redis/internal/pool/conn.go:86 +0x90
  github.com/go-redis/redis.(*PubSub).writeCmd()
      /<TRUNCATED>/go-redis/redis/pubsub.go:87 +0xb4
  github.com/go-redis/redis.(*PubSub).Ping()
      /<TRUNCATED>/go-redis/redis/pubsub.go:261 +0x1ec
  github.com/go-redis/redis.(*PubSub).initChannel.func2()
      /<TRUNCATED>/go-redis/redis/pubsub.go:498 +0x2c6

Previous write at 0x00c000021b28 by goroutine 17:
  bufio.(*Writer).WriteByte()
      /usr/local/go/src/bufio/bufio.go:666 +0x144
  github.com/go-redis/redis/internal/proto.(*Writer).WriteArgs()
[...]

PR #1878, opened 2021-09-03
Fix/pubsub ping mutex

vmihailenco:
Fixes https://github.com/go-redis/redis/issues/1874

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -252,13 +252,16 @@ func (c *PubSub) Ping(ctx context.Context, payload ...string) error {
 	}
 	cmd := NewCmd(ctx, args...)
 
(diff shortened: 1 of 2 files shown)


monkey92t:
I seem to have a misunderstanding, reading resp does not need to lock....let me add related tests

vmihailenco:
>I seem to have a misunderstanding, reading resp does not need to lock....let me add related tests

We want to allow creating/removing subscriptions when reading messages. Otherwise `ReceiveTimeout` would hold the lock almost all the time.

monkey92t:
Yes, I have a deadlock here....😭

YongJieYongJie:
Thanks @monkey92t  @vmihailenco for extremely fast fix!

PR #1882, opened 2021-09-08
chore: cleanup OpenTelemetry example

vmihailenco:


ring.go
```diff
@@ -12,7 +12,8 @@ import (
 	"time"
 
 	"github.com/cespare/xxhash/v2"
-	"github.com/dgryski/go-rendezvous"
+	rendezvous "github.com/dgryski/go-rendezvous" //nolint
```
vmihailenco, line 15: Any ideas why `//nolint` is necessary here?

PR #1883, opened 2021-09-09
Retry pipeline read commands

lenstr:
Hi! First, thanks for such a great library.

In our production environment we noticed that Redis can sometimes return errors such as "LOADING redis is loading the dataset in memory".

The library already handles errors like this in a `shouldRetry` function, but for pipelined commands it never calls.

This PR enables retries for pipelined commands when using `redis.Client`

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -462,14 +462,24 @@ func (c *baseClient) pipelineProcessCmds(
 }
 
 func pipelineReadCmds(rd *proto.Reader, cmds []Cmder) error {
+	var firstErr error
 	for _, cmd := range cmds {
 		err := cmd.readReply(rd)
 		cmd.SetErr(err)
-		if err != nil && !isRedisError(err) {
-			return err
+		if err == nil {
+			continue
 		}
+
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks for bringing this issue and for the PR. The behavior here is rather subtle so we may need to add some comments.

An alternative fix here could be to call `setCmdsErr` in `pipelineReadCmds` and then always checking `cmds[0].Err()` when deciding whether to retry the pipeline. Something like:

```go
func pipelineReadCmds(rd *proto.Reader, cmds []Cmder) error {
	for i, cmd := range cmds {
		err := cmd.readReply(rd)
		cmd.SetErr(err)
		if err != nil && !isRedisError(err) {
			setCmdsErr(cmd[i+1:], err)
			return err
		}
	}
	return nil
}
```

Hard to say which is better...

vmihailenco:
PS To clarify, I think that we need to retry the pipeline only if the first command error is retryable. We don't need to retry if the error is in the middle of a pipeline.

PR #1884, opened 2021-09-10
Allow query parameters for ParseURL

dmke:
### Background

In my Go programs, I usually want to modiffy the database connection options using CLI flags or environment variables. For example, to tune Postgres connections, I can set an environment variable like

```sh
PG="postgres://user@localhost:5432/testdb?sslmode=verify-full&connect_timeout=10" ./myprog
```

This is very handy for containerized applications.

In the Go program, one only needs to pass that string into `ParseConfig`:

```go
config, err := pgx.ParseConfig(os.Getenv("PG"))
conn, err := pgx.Connect(config)
```

### Issue

With the Go Redis library, this is not so simple:

```go
_, err := redis.ParseURL("redis://localhost/0?dial_timeout=10")
//=> redis: no options supported
```

Here, the library requires me to either setup and parse additional flags/env vars, or extract the parameters from the URL before passing it to `ParseURL`.

### Implementation details

This PR addresses this issue, by allowing (a subset of) the `Option` fields to be expressed as query parameters.

I've settled on the following behaviour:

- field names are mapped using snake-case conversion
  - `max_retries=2` → `opt.MaxRetries == 2`
  - `pool_fifo=true` → `opt.PoolFIFO == true`
  - mapping happens manually, I did not use the `reflect` package
- query parameters "network", "addr", "username" and "password" are ignored
  - these are covered by the URL `Scheme`, `Host`, `UserInfo` fields
- only scalar type fields are supported
  - i.e. `bool`, `int`, `time.Duration`
  - `string` is notably absent, because all `Option` string fields are already covered (but adding support is trivially easy)
  - struct fields (like `opt.TLSConfig`) are not supported yet. If they would ever be, someone needs to decide how to represent them as query parameters (cf. `tls_config_insecure_skip_verify=true` vs. `tls_config.insecure_skip_verify=true` vs. `tls_config[insecure_skip_verify]=true`)
  - functions (`opt.OnConnect`) are not supported, and probably will never be
- for duration fields, values must be a valid input for `time.ParseDuration`
  - for convenience, plain integers are interpreted as seconds
  - `read_timeout=60` == `read_timeout=1m` → `opt.ReadTimeout == time.Minute`
- only the last value is interpreted if a parameter is given multiple times
  - `redis://.../0?db=1&db=2` → `opt.DB == 2`
- unknown parameter names will result in an error
  - this is closest to the previous behaviour, which produced an error if any parameter was present

### DIFF
--- a/options.go
+++ b/options.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/url"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -192,9 +193,32 @@ func (opt *Options) clone() *Options {
 // Scheme is required.
 // There are two connection types: by tcp socket and by unix socket.
 // Tcp connection:
-// 		redis://<user>:<password>@<host>:<port>/<db_number>
+//		redis://<user>:<password>@<host>:<port>/<db_number>
 // Unix connection:
 //		unix://<user>:<password>@</path/to/redis.sock>?db=<db_number>
+// Most Option fields can be set using query parameters, with the following restrictions:
+//	- field names are mapped using snake-case conversion: to set MaxRetries, use max_retries
+//	- only scalar type fields are supported (bool, int, time.Duration)
+//	- for time.Duration fields, values must be a valid input for time.ParseDuration();
+//	  additionally a plain integer as value (i.e. without unit) is intepreted as seconds
+//	- to disable a duration field, use value less than or equal to 0; to use the default
+//	  value, leave the value blank or remove the parameter
+//	- only the last value is interpreted if a parameter is given multiple times
+//	- fields "network", "addr", "username" and "password" can only be set using other
+//	  URL attributes (scheme, host, userinfo, resp.), query paremeters using these
+//	  names will be treated as unknown parameters
+//	- unknown parameter names will result in an error
+// Examples:
+//		redis://user:password@localhost:6789/3?dial_timeout=3&db=1&read_timeout=6s&max_retries=2
+//		is equivalent to:
+//		&Options{
+//			Network:     "tcp",
+//			Addr:        "localhost:6789",
+//			DB:          1,               // path "/3" was overridden by "&db=1"
+//			DialTimeout: 3 * time.Second, // no time unit = seconds
+//			ReadTimeout: 6 * time.Second,
+//			MaxRetries:  2,
+//		}
 func ParseURL(redisURL string) (*Options, error) {
 	u, err := url.Parse(redisURL)
 	if err != nil {
@@ -215,10 +239,6 @@ func setupTCPConn(u *url.URL) (*Options, error) {
 	o := &Options{Network: "tcp"}
 
 	o.Username, o.Password = getUserPassword(u)
-
-	if len(u.Query()) > 0 {
-		return nil, errors.New("redis: no options supported")
-	}
 
 	h, p, err := net.SplitHostPort(u.Host)
 	if err != nil {
@@ -250,7 +270,7 @@ func setupTCPConn(u *url.URL) (*Options, error) {
 		o.TLSConfig = &tls.Config{ServerName: h}
 	}
 
-	return o, nil
+	return setupConnParams(u, o)
 }
 
 func setupUnixConn(u *url.URL) (*Options, error) {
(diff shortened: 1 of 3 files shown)


vmihailenco:
This looks good, we could make code a little bit cleaner:

```go
type queryOptions struct {
    q *url.Values
}

// Err returns the first encountered error.
func (q *queryOptions) Err() error {}
func (q *queryOptions) Int64(name string) (int64, ok) {}

q := queryOptions{q: u.Query()}
if db, ok := q.Int64("db"); ok {}
if db, ok := q.Int64("max_retries"); ok {}
if err := q.Err(); err != nil {
    return err
}
```

What do you think?

dmke:
@vmihailenco Done.

options.go
```diff
@@ -262,19 +279,123 @@ func setupUnixConn(u *url.URL) (*Options, error) {
...
+	s := o.string(name)
+	if s == "" {
+		return 0
+	}
+	// try plain number first
+	i, err := strconv.Atoi(s)
+	if err == nil {
+		if i == -1 {
+			// special case: -1 is used to disable timeouts
+			return time.Duration(i)
```
vmihailenco, line 325: I am not sure about this one. We need `-1` in Go because `0` is also a default value. But for DSN I think it should be:
- `` - default timeout
- `?read_timeout=0` - disable timeout
- `?read_timeout=1` - 1 second

What do you think?

options.go
```diff
@@ -262,19 +279,123 @@ func setupUnixConn(u *url.URL) (*Options, error) {
...
+}
+
+func (o *queryOptions) duration(name string) time.Duration {
+	s := o.string(name)
+	if s == "" {
+		return 0
+	}
+	// try plain number first
+	i, err := strconv.Atoi(s)
+	if err == nil {
```
vmihailenco, line 322: It is safer (`i` can't be used later by mistake) and shorter to write: `if i, err := strconv.Atoi(s); err == nil {`

options.go
```diff
@@ -262,19 +279,123 @@ func setupUnixConn(u *url.URL) (*Options, error) {
...
+	o.DialTimeout = q.duration("dial_timeout")
+	o.ReadTimeout = q.duration("read_timeout")
+	o.WriteTimeout = q.duration("write_timeout")
+	o.PoolFIFO = q.bool("pool_fifo")
+	o.PoolSize = q.int("pool_size")
+	o.MinIdleConns = q.int("min_idle_conns")
+	o.MaxConnAge = q.duration("max_conn_age")
+	o.PoolTimeout = q.duration("pool_timeout")
+	o.IdleTimeout = q.duration("idle_timeout")
+	o.IdleCheckFrequency = q.duration("idle_check_frequency")
```
vmihailenco, line 390: Neat!

options.go
dmke: Sure, I can change this, but this would create a difference in configuring the connection:

| direct assignment | URL parameter | behaviour (after `init()`) |
|:-----|:-----|:-----|
| `opt.ReadTimeout = -1` | `?read_timeout=0` | timeout disabled |
| `opt.ReadTimeout = 0` | absent | default timeout |
| `opt.ReadTimeout = 3` | `?read_timeout=3` | 3s timeout |

Personally I don't like that the value depends on the context. I feel this is more consistent:

| direct assignment | URL parameter | behaviour (after `init()`) |
|:-----|:-----|:-----|
| `opt.ReadTimeout = -1` | `?read_timeout=-1` | timeout disabled |
| `opt.ReadTimeout = 0` | `?read_timeout=0` or absent  | default timeout |
| `opt.ReadTimeout = 3` | `?read_timeout=3` | 3s timeout |

Thinking about it though, `?read_timeout=0` implying a default timeout feels wrong too...

dmke: Yeah, that sounds reasonable.

dmke: Should negative timeout values be discouraged at all (i.e. yield an error), or should they treated as "disable timeout"?

vmihailenco: Let's treat as disable timeout to mimic `Options` behavior.

dmke:
I have one question left: In the doc string, I've written

> - query parameters "network", "addr", "username" and "password" are ignored
>   (these are covered by other URL attributes)

However, this code will currently fail:

```go
ParseURL("redis://user@localhost/?username=admin")
//=> redis: unexpected option: username
```

I feel like, this is the correct behaviour, but it's inconsistent with the way other (duplicate) parameters are handled:

```go
opts, _ := ParseURL("redis://user@localhost/0?db=1")
//=> opts.Database == 1
```

We've got basically three options here:

1. Follow doc string (requires some `delete(q.q, "username")` before checking `r.remaining()`).
2. Yield error (needs update to the doc string).
3. Allow updating `opts.Username` via query parameters.

@vmihailenco, do you have a preference?

vmihailenco:
Looks like `?db=123` is an exception needed to support Unix sockets `unix://foo:bar@/tmp/redis.sock?db=1`. Let's not duplicate other options.

>However, this code will currently fail:

I think this code should continue to fail (option #2).

dmke:
Okay, I'll update the doc string later, I'm currently a bit busy with different things.

dmke:
@vmihailenco Done.

vmihailenco:
Thanks - this looks *very* good :+1: I am going to steal this for https://github.com/uptrace/bun/tree/master/driver/pgdriver configuration :)

dmke:
Sure, go ahead 😀

PR #1896, opened 2021-09-18
feat: add acl auth support for sentinels

jsievenpiper:
👋🏾 Hey There!

Hopefully this is a pretty straightforward one. I have a redis installation that I'm trying to leverage that has Redis6 ACL auth against the Sentinels. This exposes a `SentinelUsername` field that propagates down to the underlying client (which already switches between Redis < 6 auth and ACL auth).

While I was there, I added a new test suite for the ACL'd sentinels!

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -40,27 +40,15 @@ const (
 	sentinelPort3      = "9128"
 )
 
-const (
-	aclSentinelUsername = "sentinel-user"
-	aclSentinelPassword = "sentinel-pass"
-	aclSentinelName     = "my_server"
-	aclServerPort       = "10001"
-	aclSentinelPort1    = "10002"
-	aclSentinelPort2    = "10003"
-	aclSentinelPort3    = "10004"
-)
-
 var (
 	sentinelAddrs = []string{":" + sentinelPort1, ":" + sentinelPort2, ":" + sentinelPort3}
-	aclSentinelAddrs = []string {":" + aclSentinelPort1, ":" + aclSentinelPort2, ":" + aclSentinelPort3}
 
 	processes map[string]*redisProcess
 
-	redisMain, aclServer                           *redisProcess
+	redisMain                                      *redisProcess
 	ringShard1, ringShard2, ringShard3             *redisProcess
 	sentinelMaster, sentinelSlave1, sentinelSlave2 *redisProcess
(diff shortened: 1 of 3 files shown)


sentinel_test.go
```diff
@@ -212,3 +212,62 @@ var _ = Describe("NewFailoverClusterClient", func() {
...
+			Password: aclSentinelPassword,
+		})
+
+		addr, err := sentinel.GetMasterAddrByName(ctx, aclSentinelName).Result()
+		Expect(err).NotTo(HaveOccurred())
+
+		server = redis.NewClient(&redis.Options{
+			Addr:       net.JoinHostPort(addr[0], addr[1]),
+			MaxRetries: -1,
+		})
```
vmihailenco, line 245: This looks unused?

main_test.go
```diff
@@ -364,6 +388,28 @@ func startSentinel(port, masterName, masterPort string) (*redisProcess, error) {
...
+		return nil, err
+	}
+
+	for _, cmd := range []*redis.StatusCmd{
+		redis.NewStatusCmd(ctx, "ACL", "SETUSER", aclSentinelUsername, "ON", ">" + aclSentinelPassword, "-@all",
+			"+auth", "+client|getname", "+client|id", "+client|setname", "+command", "+hello", "+ping", "+role",
+			"+sentinel|get-master-addr-by-name", "+sentinel|master", "+sentinel|myid", "+sentinel|replicas",
+			"+sentinel|sentinels"),
+	} {
+		process.Client.Process(ctx, cmd)
```
vmihailenco, line 403: Can we instead run this command on existing sentinel servers? And then run `ACL DELUSER` to disable ACL?

Starting servers is a slow operation so we want to minimize the number of servers we need for testing.

vmihailenco:
Thanks for the PR :+1: If possible, it would be nice to run the new test on existing Sentinel servers. Otherwise, this looks pretty god.

main_test.go
jsievenpiper: Yeah that seems fair, done.

jsievenpiper:
Thanks for taking a look so quickly!

vmihailenco:
Thanks :+1:

jsievenpiper:
You're welcome! Excited to see this in an upcoming release 👍🏾

PR #1897, opened 2021-09-18
feat: add Prometheus instrumentation of pool stats

HatsuneMiku3939:
# feat: add Prometheus instrumentation of pool stats

Hello. Thanks for a wonderful library. `go-redis`  is a major part of our production environments.
This PR adds an extra instrumentation module for exposing pool stats via the Prometheus collector.

I hope this will help people in high traffic production environments.


extra/redisprom/collector.go
```diff
@@ -0,0 +1,81 @@
...
+	misses   *prometheus.Desc
+	timeouts *prometheus.Desc
+
+	totalConns *prometheus.Desc
+	idleConns  *prometheus.Desc
+	staleConns *prometheus.Desc
+}
+
+// NewPoolStatsCollector returns a collector that exports pool metrics about the given redis.UniversalClient.
+func NewPoolStatsCollector(universeClient redis.UniversalClient, name string) prometheus.Collector {
```
vmihailenco, line 21: Probably just `redisprom.NewCollector`? In theory, this collector can gather other stats as well...

vmihailenco:
@HatsuneMiku3939 thanks for the PR. I don't have much experience with Prometheus collector so I can only review the code in general. Maybe someone on @go-redis/team is interested? :)

Overall I do have plans to improve `redisotel` to provide OpenTelemetry Metrics and in theory Prometheus users could use that too. See https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter

But we can support Prometheus directly too. The lack of reviewers and potential maintainers is the only problem...

dim:
May I suggest using https://github.com/bsm/openmetrics/ instead. It's compatible but far more lightweight than the native collector and has zero dependencies

knadh:
@mr-karan would you want to take a look?

mr-karan:
Would be worthwhile to take a look at https://github.com/VictoriaMetrics/metrics which is another lightweight metrics exporter lib. No dependencies as well.

HatsuneMiku3939:
@vmihailenco 
As you say, Prometheus users could use OpenTelemetry Metrics. And, I think the potential maintainer problem is more critical rather than add a new feature in OSS projects.

I will close this PR and publish it as a seprate project. Is this all right?

HatsuneMiku3939:
@dim @mr-karan 
Thanks for your suggestions. I will consider it if I continue to work on this PR.

vmihailenco:
@HatsuneMiku3939 sorry for the delay. Yes, let's publish this a separate module/project so we don't block you without being really helpful. I will add a link to your project to the GitHub readme and probably write an article at redis.uptrace.dev.

And thanks everyone for the input. Looks like there are a lot of good options for metrics...

vmihailenco:
@HatsuneMiku3939 I've created https://github.com/go-redis/redisprom and sent you an invite. Let's continue there.

Thanks again for the PR.

PR #1899, opened 2021-09-20
Add support for BLMove command

alexus1024:


vmihailenco:
Thanks for the contribution, but please fix action checks...

alexus1024:
Hmm. It was merged. Unexpected. Should I still fix these pipeline problems or it is not important anymore?

vmihailenco:
No need, I've fixed that myself.

PR #1906, opened 2021-09-25
feat: add SetVal method for each command

### LINKED ISSUE #1879: Add API to set the result value of command
We already have the function `SetErr` in the interface `Cmder` to set a custom error in command response result. But there are cases where we want to manually set response value to command result. For example when attach hooks to command process we can interrupt the processing of a command and set a custom value to command. 

So please add function `SetVal(interface{})` to the interface `Cmder`.

Stavrospanakakis:
Closes #1879

### DIFF
--- a/command.go
+++ b/command.go
@@ -200,6 +200,10 @@ func (cmd *Cmd) String() string {
 	return cmdString(cmd, cmd.val)
 }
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks

PR #1914, opened 2021-10-02
Fix Redis Cluster issue during roll outs of new nodes with same addr

pfreixes:
This issue was surfaced in a Cloud Provider solution that used for
rolling out new nodes using the same address (hostname) of the nodes
that will be replaced in a Redis Cluster, while the former ones once
depromoted as Slaves would continue in service during some mintues
for redirecting traffic.

The solution basically identifies when the connection could be stale
since a MOVED response will be returned using the same address (hostname)
that is being used by the connection. At that moment we consider the
connection as no longer usable forcing to recycle the connection.

### DIFF
--- a/error.go
+++ b/error.go
@@ -83,6 +83,8 @@ func isBadConn(err error, allowTimeout bool, addr string) bool {
 			// of the connection. Force a DNS resolution when all connections
 			// of the pool are recycled
 			return true
+		} else {
+			return false
 		}
 	}
 
--- a/error.go
+++ b/error.go
@@ -65,7 +65,7 @@ func isRedisError(err error) bool {
 	return ok
 }
 
-func isBadConn(err error, allowTimeout bool) bool {
+func isBadConn(err error, allowTimeout bool, addr string) bool {
 	switch err {
 	case nil:
 		return false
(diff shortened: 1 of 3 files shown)


pfreixes:
Ill investigate why some tests are failing ...

error.go
```diff
@@ -74,9 +74,16 @@ func isBadConn(err error, allowTimeout bool) bool {
...
+		if isReadOnlyError(err) {
+			// Close connections in read only state in case domain addr is used
+			// and domain resolves to a different Redis Server. See #790.
+			return true
+		} else if isMovedSameConnAddr(err, addr) {
+			// Close connections when we are asked to move to the same addr
+			// of the connection. Force a DNS resolution when all connections
+			// of the pool are recycled
+			return true
+		}
```
vmihailenco, line 86: Replace `else if` with `else` and add `return false` as a final clause.

error.go
```diff
@@ -74,9 +74,16 @@ func isBadConn(err error, allowTimeout bool) bool {
...
-		// and domain resolves to a different Redis Server. See #790.
-		return isReadOnlyError(err)
+		if isReadOnlyError(err) {
+			// Close connections in read only state in case domain addr is used
+			// and domain resolves to a different Redis Server. See #790.
+			return true
+		} else if isMovedSameConnAddr(err, addr) {
+			// Close connections when we are asked to move to the same addr
+			// of the connection. Force a DNS resolution when all connections
+			// of the pool are recycled
```
vmihailenco, line 84: Thanks for the comment!

error.go
pfreixes: Moved to a switch, hope that this also fixes the broken tests.

vmihailenco:
Thanks :+1:

pfreixes:
Any chance to have a bugfix release version?

vmihailenco:
Sure, I will push 8.11.4

PR #1918, opened 2021-10-04
fix: add whitespace for avoid unlikely colisions

pfreixes:
@vmihailenco sorry but received some feedback from my colleagues and they suggested, which I agree, that by adding an space before the addr will remove any chance of having a false positive because of overlapping in the hostname.

Tested and works like a charm also.


vmihailenco:
Thanks

PR #1922, opened 2021-10-15
chore: use ctx parameter in cmdInfo

HurSungYun:
cmdInfo has `ctx` as parameter, so I believe it would be more reasonable to use context parameter rather than client context when logging


vmihailenco:
Thanks

PR #1958, opened 2021-12-13
[Perf] simple time period checking dial circuit

longxuecheng:


longxuecheng:
it is just a simple change to avoid large number of routines was created at the time dialing operations have high delay

PR #1960, opened 2021-12-17
Allow struct on HSet

wjdqhry:
Each time, struct had to be converted to map and put it in.
So I thought it would be better to embed it inside the function.


YuriPuck:
I want to use this feature

monkey92t:
Should we use the "json" tag? I feel like it's better handled by the person who needs it.

```go
func structToMap() {
    ....
}

client.Hset(ctx, "key", structToMap(items))
```

wjdqhry:
> Should we use the "json" tag? I feel like it's better handled by the person who needs it.
> 
> ```go
> func structToMap() {
>     ....
> }
> 
> client.Hset(ctx, "key", structToMap(items))
> ```

Thank you for your reply.

I felt that the experience of using the library was reduced because the person who wanted it had to make it from the outside, and I thought it would be ideal to solve it internally.

It doesn't have to be a "json" tag, but I think we need a tag for our hash keys.

monkey92t:
We should think more, like tag="" or tag="-", or for types like:

```go
type Demo struct {
    Pers []int
    Ptr *struct
    Map map[string]*struct
    ...
}

```

wjdqhry:
I modified the tag name. The example has also been modified.

But if you don't think it's necessary, you can close the PR!

michaelfeinbier:
I was wondering why this does not exist yet. 

Thanks for the code - I guess you should maybe use the tag `redis` as it is already used for the other way round if you [scan a HGetAll to a struct](https://redis.uptrace.dev/guide/scanning-hash-fields.html)

wjdqhry:
Sorry i'm just too late..
Thank you for your feedback!

I accepted your feedback and I just changed serialize key "key" to "redis"

monkey92t:
I added support for structure types in the `appendArgs` function, so that `MSet`, `MSetNX` functions can also benefit.

I added the description of using the struct type in the API comments, and added the test function.

We should note that for complex field types of structures, the encoding.BinaryMarshaler interface should be implemented.

Synchronized with v9.

PR #1980, opened 2022-01-05
add OnGetConnect hook

cuitingshi:
OnGetConnect hook enables us to do things related the real connection. For example, in use of this hook, we can pass the remote address through the context, and then do some data report in the AfterProcesss hook in production.

@monkey92t


monkey92t:
I wonder if we really need it?

vmihailenco:
@monkey92t I don't think we need it. We already have `Option.Dialer` which allows customizing dialing. And the interface in this PR is very specialized.

PR #1987, opened 2022-01-07
feat: add RemoteAddr into basecmd

cuitingshi:


vmihailenco:
Thanks, but this is overly specific. There is no reason command should have information about remote addr.

cuitingshi:
> Thanks, but this is overly specific. There is no reason command should have information about remote addr.

You are right. I wonder if we could find a better way to pass the connection remote address to the AfterProcess hooks. Because in our production environment, we really need the remote address in the AfterProcess hooks to report the module call metrics to our monitor system.
@vmihailenco

vmihailenco:
Perhaps we should make this information available in the hooks, but it should not be part of the `Cmd`.

Issue #1990, opened 2022-01-11
cmd.setFirstKeyPos Should be exported for building module cmd


jiangtao244:
https://github.com/go-redis/redis/blob/333fee1a8fd98a2fbff1ab187c1b03246a7eb01f/command.go#L162

jiangtao244:
In my case, I want to use roaring module , but the first key pos of  R.BITOP command must be custom.
https://github.com/aviggiano/redis-roaring/blob/4fa6e4ebebe72cea5d23b4791812e09d4973d35c/src/redis-roaring.c#L913

PR #1991, opened 2022-01-11
feat: export cmder.SetFirstKeyPos to support build module commands

jiangtao244:


jiangtao244:
fix https://github.com/go-redis/redis/issues/1990

vmihailenco:
Thanks. This is an internal API so we can change it at any point.

PR #2012, opened 2022-02-01
otel instrumentation name and version

nelz9999:
Use full path for tracer name, and decorate traces with the library version

For example, as in other packages:
* https://github.com/open-telemetry/opentelemetry-go-contrib/blob/aad3f901baac5d79791fefb43c81d5b16ca22f99/instrumentation/github.com/gorilla/mux/otelmux/mux.go#L39
* https://github.com/open-telemetry/opentelemetry-go-contrib/blob/7e31ebe040306aee2c826972269f938f9f0e7c34/instrumentation/net/http/otelhttp/config.go#L30

### DIFF
--- a/extra/redisotel/redisotel.go
+++ b/extra/redisotel/redisotel.go
@@ -14,8 +14,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks :+1:

PR #2027, opened 2022-02-18
fix invalid type assert in stringArg

### LINKED ISSUE #2026: horrible performance when send req by Do with non-string key

## Current Behavior
when send req by Do() with non-string key in cluster,  the cmder failed to get correct slot id。because  the below func will return "" when the key is not string type.
```
func (cmd *baseCmd) stringArg(pos int) string {
	if pos < 0 || pos >= len(cmd.args) {
		return ""
	}
	s, _ := cmd.args[pos].(string)
	return s
}
``` 
 as a result, the cluster will send req to invalid slot, and get a move resp. the below is the benchmark with client.Do()
```
func BenchmarkClusterDo(b *testing.B) {
	if testing.Short() {
		b.Skip("skipping in short mode")
	}

	ctx := context.Background()
	cluster := newClusterScenario()
	if err := startCluster(ctx, cluster); err != nil {
		b.Fatal(err)
	}
	defer cluster.Close()

	client := cluster.newClusterClient(ctx, redisClusterOptions())
	defer client.Close()
	b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {

			err :=client.Do(ctx, "SET", 10, 10).Err()
			if err != nil {
			 	b.Fatal(err)
			}
		}
	})
}
```
And the benchmark result is horrible as show below. the req will finally success after get an moved from cluster.
```
➜  redis git:(master) ✗ go test  -bench="BenchmarkClusterDo" -run=none 
goos: darwin
goarch: amd64
pkg: github.com/go-redis/redis/v8
cpu: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
BenchmarkClusterDo-8         813           1441423 ns/op             623 B/op         16 allocs/op
PASS
ok      github.com/go-redis/redis/v8    40.381s
```

## Possible Solution
check type assert when the key is non string

## Context (Environment)

any env

lintanghui:
fix #2026


vmihailenco:
Thanks

PR #2033, opened 2022-02-22
Feature/otel attributes

thrawny:
Adds the option of passing extra attributes to the `TracingHook` to make it possible to set for example `net.peer.name` which is helpful to distinguish between different redis instances


vmihailenco:
Looks good - thank you!

PR #2038, opened 2022-02-28
Add customizable retry logic

ocallaco:
Added a field ShouldRetry to all client option structs, to allow
custom behavior on retries. Existing shouldRetry function is renamed
to DefaultShouldRetry, and is the default as the name suggests.

Also exporting pool.ErrPoolTimeout

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -63,6 +63,7 @@ type ClusterOptions struct {
 	MaxRetries      int
 	MinRetryBackoff time.Duration
 	MaxRetryBackoff time.Duration
+	ShouldRetry     ShouldRetryFunc
 
 	DialTimeout  time.Duration
 	ReadTimeout  time.Duration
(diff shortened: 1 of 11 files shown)


vmihailenco:
Thanks for the PR.

First, we need to figure out the right interface for this. for example, there is https://github.com/avast/retry-go and https://github.com/matryer/try.

Probably, `retryTimeout bool` should not be a part of the API and instead handled internally.

vmihailenco:
You may also want to check https://debugged.it/blog/go-patterns-retries/#using-contexts

ocallaco:
I don't think i'll be able to in the near term.  If I find the time, I'll
likely open a new PR

ndyakov:
Understandable, will close this PR then. Thanks for the effort, @ocallaco !

PR #2040, opened 2022-03-02
fix(sentinel): use opt.DB and fallback with 0

### LINKED ISSUE #2025: go-redis/v8 ClusterClient always reads/writes to DB 0

go-redis/redis version: v8.11.4

```       
sentinelOptions = &redis.FailoverOptions{
            MasterName:       options.SentinelMaster,
            SentinelUsername: options.SentinelUser,
            SentinelPassword: options.SentinelPass,
            Username:         options.RedisUser,
            Password:         options.RedisPass,
            SentinelAddrs:    options.SentinelServers,
            RouteRandomly:    true,
            DB:               options.DBNumber,                  // XXX: Setting this to non-zero DB number
            DialTimeout:      30 * time.Second,
}
clusterClient := redis.NewFailoverClusterClient(sentinelOptions)
```

This is how I setup my failover cluster client.
All writes/reads (eg: clusterClient.Set(ctx, key, value)) go into DB 0. I expect these operations to be performed on the selected DB that is passed while creating ClusterClient object.

## Expected Behavior

Operations should go to the selected DB (non-zero in my tests) used during the creation of ClusterClient object

## Current Behavior

With current code all operations go to DB 0 irrespective of the DB value selected during client creation.

## Steps to Reproduce

Create a failover cluster client as above and try to make reads/writes to non-zero DB numbers.

anwam:
fix #2025

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -114,7 +114,7 @@ func (opt *FailoverOptions) sentinelOptions(addr string) *Options {
 		Dialer:    opt.Dialer,
 		OnConnect: opt.OnConnect,
 
-		DB:       opt.DB | 0,
+		DB:       opt.DB,
 		Username: opt.SentinelUsername,
 		Password: opt.SentinelPassword,
 
--- a/sentinel.go
+++ b/sentinel.go
@@ -114,7 +114,7 @@ func (opt *FailoverOptions) sentinelOptions(addr string) *Options {
 		Dialer:    opt.Dialer,
 		OnConnect: opt.OnConnect,
 
-		DB:       0,
+		DB:       opt.DB | 0,
 		Username: opt.SentinelUsername,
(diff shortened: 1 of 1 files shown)


gkodali-zededa:
@anwam Just curious if you verified this fix to see reads/writes going to selected DB (instead of 0)?

sentinel.go
```diff
@@ -114,7 +114,7 @@ func (opt *FailoverOptions) sentinelOptions(addr string) *Options {
 		Dialer:    opt.Dialer,
 		OnConnect: opt.OnConnect,
 
-		DB:       0,
+		DB:       opt.DB,
```
vmihailenco, line 117: These options are for Sentinel client. Do we really want to select a DB in Sentinel Server?

PR #2052, opened 2022-03-19
fix(#1943): XInfoConsumer.Idle should be time.Duration instead of int64

### LINKED ISSUE #1943: XInfoConsumer.Idle should be time.Duration instead of int64
See https://github.com/vmihailenco/taskq/issues/151 for some context

monkey92t:
to v9, fix #1943


monkey92t:
This is an easy fix and it doesn't seem to be a problem.

Issue #2053, opened 2022-03-19
reopen(v9): checkConn for badConn(#1821, #1824), alwaysNewConn(#1737)


monkey92t:
checkConn: #1821, #1824
alwaysNewConn: #1737
link: https://github.com/go-redis/redis/pull/1849#issuecomment-894382680

Works on the v9 branch.

Issue #1821, opened 2021-07-14
conncheck for badconn?


hidu:
Issue tracker is used for reporting bugs and discussing new features. Please use
[stackoverflow](https://stackoverflow.com) for supporting issues.

when  redis server restart, the conns in Pooler can not rw,
they well occur io.EOF  error

we can check the  bad conns like this:
https://github.com/go-sql-driver/mysql/blob/master/conncheck.go

## Expected Behavior

Pooler.Get(context.Context) (*Conn, error)
can remove the bad conn

## Current Behavior

Pooler.Get return a bad conn, it occur io.EOF  error when we rw it

## Possible Solution

check the conn before we use:
https://github.com/go-sql-driver/mysql/blob/master/conncheck.go

vmihailenco:
@hidu looks pretty neat!

monkey92t:
After testing, it will increase the CPU time by about 5-10%. It looks good.

Are you interested in submitting a PR?

hidu:
i'll do it

bingzhuo2008:
With this feature, my connections are closed and re created again and again as syscall.Read always said  "resource temporarily unavailable"

bingzhuo2008:
I just fallbacked to previous version v8.11.0, and my connections stopped re creating. I'm running it in docker debian. Please test this code in docker before next version published, thank you ! @monkey92t @hidu

monkey92t:
> I just fallbacked to previous version v8.11.0, and my connections stopped re creating. I'm running it in docker debian. Please test this code in docker before next version published, thank you ! @monkey92t @hidu

I ran a simple test in debian and it seems to work well (v8.11.1), can you provide the test code and debian version?

bingzhuo2008:
> > I just fallbacked to previous version v8.11.0, and my connections stopped re creating. I'm running it in docker debian. Please test this code in docker before next version published, thank you ! @monkey92t @hidu
> 
> I ran a simple test in debian and it seems to work well (v8.11.1), can you provide the test code and debian version?

I'm using redis cluster which has 20 master nodes and 20 slave nodes, in k8s docker, Debian 4.9.144-3 (2019-02-02) x86_64 GNU/Linux(I guess it's not debian who caused the problemn, maybe it's docker?)

But with v8.11.0 and the same code there is no the problemn.

`
import (
    gredis "github.com/go-redis/redis/v8"
)
 
type RedisCluster struct {
	client *gredis.ClusterClient
}

[...]

PR #2054, opened 2022-03-19
checkConn for badConn and alwaysNewConn(#2053)

naiqianz:
#2053
check bad conn when getConn from pool and add maxBadConnRetries

### DIFF
--- a/internal/pool/conn_check.go
+++ b/internal/pool/conn_check.go
@@ -1,0 +1,49 @@
+// +build linux darwin dragonfly freebsd netbsd openbsd solaris illumos
+
+package pool
+
+import (
+	"errors"
+	"io"
+	"net"
+	"syscall"
+	"time"
+)
+
+var errUnexpectedRead = errors.New("unexpected read from socket")
+
+func connCheck(conn net.Conn) error {
+	//reset deadline
+	_ = conn.SetDeadline(time.Time{})
+
+	sysConn, ok := conn.(syscall.Conn)
+	if !ok {
+		return nil
+	}
+	rawConn, err := sysConn.SyscallConn()
+	if err != nil {
+		return err
+	}
+
+	var sysErr error
+	err = rawConn.Read(func(fd uintptr) bool {
+		var buf [1]byte
+		n, err := syscall.Read(int(fd), buf[:])
+		switch {
+		case n == 0 && err == nil:
+			sysErr = io.EOF
+		case n > 0:
+			sysErr = errUnexpectedRead
+		case err == syscall.EAGAIN || err == syscall.EWOULDBLOCK:
+			sysErr = nil
+		default:
+			sysErr = err
+		}
+		return true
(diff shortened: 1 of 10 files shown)


monkey92t:
It seems that the issue of #1849 was not resolved.

monkey92t:
It seems that we should consider the database/sql solution(https://github.com/go-redis/redis/issues/1737#issuecomment-832434567):

```
func (db *DB) ExecContext(ctx context.Context, query string, args ...interface{}) (Result, error) {
	var res Result
	var err error
	for i := 0; i < maxBadConnRetries; i++ {
		res, err = db.exec(ctx, query, args, cachedOrNewConn)
		if err != driver.ErrBadConn {
			break
		}
	}
	if err == driver.ErrBadConn {
		return db.exec(ctx, query, args, alwaysNewConn)
	}
	return res, err
}
```

monkey92t:
@vmihailenco  Do you have any comments?

monkey92t:
WechatIMG97

internal/pool/conn_check.go
```diff
@@ -0,0 +1,49 @@
...
+	"net"
+	"syscall"
+	"time"
+)
+
+var errUnexpectedRead = errors.New("unexpected read from socket")
+
+func connCheck(conn net.Conn) error {
+	//reset deadline
+	_ = conn.SetDeadline(time.Time{})
```
vmihailenco, line 17: @monkey92t I believe this line resolves https://github.com/go-redis/redis/pull/1849/files. I've been using similar code with go-clickhouse and everything seems to be working fine.

vmihailenco, line 17: @naiqianz could you fix the comment formatting?

```go
// Reset previous timeout.
```

internal/pool/pool.go
```diff
@@ -250,7 +255,7 @@ func (p *ConnPool) Get(ctx context.Context) (*Conn, error) {
 		return nil, err
 	}
 
-	for {
+	for i := 0; i < maxBadConnRetries; i++ {
```
vmihailenco, line 258: Let's split this change in a separate PR and leave only checkConn like it was done in https://github.com/go-redis/redis/issues/1821

vmihailenco:
>It seems that we should consider the database/sql solution

We probably need both this change and database/sql-like retry approach. `checkConn` is about early detection of bad connections and we need it anyway.

internal/pool/conn_check.go
naiqianz: sure, tks for ur suggestion

monkey92t:
We once made a mistake in `SetDeadline` and should add a check for it.

monkey92t:
It looks good, we will add `maxBadConnRetries`, `cachedOrNewConn `, `alwaysNewConn ` later.

monkey92t (approved):

PR #2066, opened 2022-04-04
Add NewXPendingResult for unit testing XPending

evilstreak:
I came across this gap whilst writing a unit test. Hopefully this is the right way to add it!


vmihailenco:
>Would this be feat, as it's changing the code to add something new?

Yes, `feat` looks the right choice here. I will use squah button.

>I don't think it causes a problem with this commit

:+1:

evilstreak:
Thanks!

PR #2079, opened 2022-04-28
nil check before calling methods on Conn.netConn

gidyon:


vmihailenco:
There should not be cases when netConn is nil.

PR #2083, opened 2022-05-01
chore: arrange the arguments 'start' and 'stop' of 'XRevRange(N)' in the way as the order of the redis command.

taoikaihatsu:
In the redis command 'XREVRANGE', 'end' is followed by 'start'.

https://redis.io/commands/xrevrange/
XREVRANGE key end start [COUNT count]

In go-redis 'XRevRange' and 'XRevRangeN', 'start' is followed by 'stop'.

XRevRange(stream, start, stop string)
XRevRangeN(stream, start, stop string, count int64)

There is no problem with the program's behavior, but it might be confusing.
So in 'XRevRange' and 'XRevRangeN', 'stop' is followed by 'start'.


vmihailenco:
It is better to not break existing programs that rely on such order.

PR #2093, opened 2022-05-25
feature: dynamically add and remove shards by the ring client

### LINKED ISSUE #2077: add/remove shards in redis ring
We use redis ring client to shard access to redis for our rate limit infrastructure in Kubernetes https://opensource.zalando.com/skipper/tutorials/ratelimit/#redis-based-cluster-ratelimits.
I would like to add and remove shards on demand while Kubernetes is scaling-out redis instances.
I tried to implement it by closing and recreating the redis ring, but I think it would be better (less locks required) to trigger it via a library call. 

One idea I had was to have a `func() []string` that is called every configurable time.Duration with a time.Ticker to set the Members and propagate these into the library ringShards. Or we could also do the triggering ourselves and the library just provides `ReconfigureShards(shards []string)`.
Do you have a better idea how to make this happen?

I am willing to create a PR if it makes sense for you.

szuecs:
fix #2077 
feature: ring.SetAddrs to add and remove shards by the ring client #2093 
tests: to scale in and out

Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -160,6 +160,7 @@ func (opt *RingOptions) clientOptions() *Options {
 type ringShard struct {
 	Client *Client
 	down   int32
+	addr   string
 }
 
 func newRingShard(opt *RingOptions, name, addr string) *ringShard {
@@ -168,6 +169,7 @@ func newRingShard(opt *RingOptions, name, addr string) *ringShard {
 
 	return &ringShard{
 		Client: opt.NewClient(name, clopt),
+		addr:   addr,
 	}
 }
 
@@ -212,33 +214,68 @@ type ringShards struct {
 	opt *RingOptions
 
 	mu       sync.RWMutex
+	muClose  sync.Mutex
 	hash     ConsistentHash
-	shards   map[string]*ringShard // read only
-	list     []*ringShard          // read only
+	shards   map[string]*ringShard // read only, updated by SetAddrs
+	list     []*ringShard          // read only, updated by SetAddrs
 	numShard int
 	closed   bool
 }
 
 func newRingShards(opt *RingOptions) *ringShards {
-	shards := make(map[string]*ringShard, len(opt.Addrs))
+	c := &ringShards{
+		opt: opt,
+	}
+	c.SetAddrs(opt.Addrs)
+
+	return c
+}
+
+// SetAddrs replaces the shards in use, such that you can increase and
+// decrease number of shards, that you use. It will reuse shards that
+// existed before and close the ones that will not be used anymore.
+func (c *ringShards) SetAddrs(addrs map[string]string) {
+	c.muClose.Lock()
+	defer c.muClose.Unlock()
+	if c.closed {
+		return
+	}
+
+	shards := make(map[string]*ringShard)
+	unusedShards := make(map[string]*ringShard)
+
+	for k, shard := range c.shards {
+		if addr, ok := addrs[k]; ok && shard.addr == addr {
+			shards[k] = shard
+		} else {
+			unusedShards[k] = shard
+		}
+	}
+
+	for k, addr := range addrs {
+		if shard, ok := c.shards[k]; !ok || shard.addr != addr {
+			shards[k] = newRingShard(c.opt, k, addr)
+		}
+	}
+
 	list := make([]*ringShard, 0, len(shards))
-
-	for name, addr := range opt.Addrs {
-		shard := newRingShard(opt, name, addr)
-		shards[name] = shard
-
+	for _, shard := range shards {
 		list = append(list, shard)
 	}
 
-	c := &ringShards{
-		opt: opt,
+	c.mu.Lock()
+	c.shards = shards
+	c.list = list
 
-		shards: shards,
-		list:   list,
+	c.rebalanceLocked()
+	c.mu.Unlock()
+
+	for k, shard := range unusedShards {
+		err := shard.Client.Close()
+		if err != nil {
+			internal.Logger.Printf(context.Background(), "Failed to close ring shard client %s %s: %v", k, shard.addr, err)
+		}
 	}
-	c.rebalance()
-
-	return c
 }
 
 func (c *ringShards) List() []*ringShard {
@@ -355,6 +392,23 @@ func (c *ringShards) rebalance() {
 	c.mu.Unlock()
 }
 
+// rebalanceLocked removes dead shards from the Ring and callers need to hold the locl
+func (c *ringShards) rebalanceLocked() {
+	shards := c.shards
+	liveShards := make([]string, 0, len(shards))
+
+	for name, shard := range shards {
+		if shard.IsUp() {
+			liveShards = append(liveShards, name)
+		}
+	}
+
+	hash := c.opt.NewConsistentHash(liveShards)
+
+	c.hash = hash
+	c.numShard = len(liveShards)
+}
+
 func (c *ringShards) Len() int {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
@@ -363,6 +417,8 @@ func (c *ringShards) Len() int {
 }
 
 func (c *ringShards) Close() error {
+	c.muClose.Lock()
+	defer c.muClose.Unlock()
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -434,6 +490,10 @@ func NewRing(opt *RingOptions) *Ring {
 	go ring.shards.Heartbeat(hbCtx, opt.HeartbeatFrequency)
 
 	return &ring
+}
+
+func (c *Ring) SetAddrs(ctx context.Context, addrs map[string]string) {
+	c.shards.SetAddrs(addrs)
 }
 
 // Do creates a Cmd from the args and processes the cmd.
--- a/ring_test.go
+++ b/ring_test.go
@@ -113,6 +113,79 @@ var _ = Describe("Redis Ring", func() {
 		Expect(ringShard2.Info(ctx, "keyspace").Val()).To(ContainSubstring("keys=100"))
 	})
 
+	Describe("[new] dynamic setting ring shards", func() {
+		It("downscale shard and check reuse shard, upscale shard and check reuse", func() {
+			Expect(ring.Len(), 2)
+
+			wantShard := ring.GetAddr("ringShardOne")
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			ring.SetAddrs(ctx, map[string]string{
+				"ringShardOne": ":" + ringShard1Port,
+			})
+			Expect(ring.Len(), 1)
+			gotShard := ring.GetAddr("ringShardOne")
+			Expect(gotShard).To(Equal(wantShard))
+
+			ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			ring.SetAddrs(ctx, map[string]string{
+				"ringShardOne": ":" + ringShard1Port,
+				"ringShardTwo": ":" + ringShard2Port,
+			})
+			Expect(ring.Len(), 2)
+			gotShard = ring.GetAddr("ringShardOne")
+			Expect(gotShard).To(Equal(wantShard))
+
+		})
+
+		It("uses 3 shards after setting it to 3 shards", func() {
+			Expect(ring.Len(), 2)
+
+			// Start ringShard3.
+			var err error
+			ringShard3, err = startRedis(ringShard3Port)
+			Expect(err).NotTo(HaveOccurred())
+
+			shardName1 := "ringShardOne"
+			shardAddr1 := ":" + ringShard1Port
+			wantShard1 := ring.GetAddr(shardName1)
+			shardName2 := "ringShardTwo"
+			shardAddr2 := ":" + ringShard2Port
+			wantShard2 := ring.GetAddr(shardName2)
+			shardName3 := "ringShardThree"
+			shardAddr3 := ":" + ringShard3Port
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			ring.SetAddrs(ctx, map[string]string{
+				shardName1: shardAddr1,
+				shardName2: shardAddr2,
+				shardName3: shardAddr3,
+			})
+			Expect(ring.Len(), 3)
+			gotShard1 := ring.GetAddr(shardName1)
+			gotShard2 := ring.GetAddr(shardName2)
+			gotShard3 := ring.GetAddr(shardName3)
+			Expect(gotShard1).To(Equal(wantShard1))
+			Expect(gotShard2).To(Equal(wantShard2))
+			Expect(gotShard3).ToNot(BeNil())
+
+			ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			ring.SetAddrs(ctx, map[string]string{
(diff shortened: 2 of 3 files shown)


ring.go
```diff
@@ -220,25 +220,43 @@ type ringShards struct {
...
-		opt: opt,
+func (c *ringShards) SetAddrs(addrs map[string]string) {
+	shards, list := createRingShards(c.opt, addrs)
 
-		shards: shards,
-		list:   list,
+	c.mu.Lock()
+	if !c.closed {
+		c.shards = shards
+		c.list = list
```
AlexanderYastrebov, line 255: I think it should close old shards to prevent connection leak.
Also the common usecase seems to be addition/removal of a shard, e.g. `[a, b, c] -> [a, b, c, d]` - the proposed implementation would throw away perfectly fine shards a, b and c just to recreate them, it would also briefly double number of connections to each of them.
Besides some of existing shards could be down and re-creation would mark them up until next healthcheck.
Maybe the update should attempt to reuse existing shards if their name+addr are in the updated list.

vmihailenco: I agree with this. @szuecs what do you think?

szuecs: Yes of course!

szuecs: I made the change

ring.go
```diff
@@ -220,25 +222,94 @@ type ringShards struct {
...
-		opt: opt,
+func (c *ringShards) SetAddrs(addrs map[string]string) {
+	var currentRingshards map[string]*ringShard
+	toDelete := make(map[string]*ringShard)
+	newShards := make(map[string]*ringShard)
 
-		shards: shards,
-		list:   list,
+	c.mu.RLock()
+	if !c.closed {
```
AlexanderYastrebov, line 257: When closed I think it is better to simply return here

ring.go
```diff
@@ -220,25 +222,94 @@ type ringShards struct {
...
+		currentRingshards = c.shards
+	}
+	c.mu.RUnlock()
+
+	// delete already connected shard addresses
+	for k, currentRingshard := range currentRingshards {
+		if v, ok := addrs[k]; ok {
+
+			if currentRingshard.addr == v {
```
AlexanderYastrebov, line 266: The check could be combined into `ok && currentRingshard.addr == v`.
It is also possible to `delete(currentRingshards, k)` right away https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop

ring.go
```diff
@@ -220,25 +222,94 @@ type ringShards struct {
...
+	}
+
+	// create new connections
+	for name, addr := range addrs {
+		newShards[name] = newRingShard(c.opt, name, addr)
+	}
+	// likely needs to be in critical section, if not we can use the loop before and drop "newShards"
+	for name, shard := range newShards {
+		currentRingshards[name] = shard
+	}
```
AlexanderYastrebov, line 286: This two loops could be combined to eliminate `newShards` altogether.

ring.go
AlexanderYastrebov: Actually I think it is not safe to modify `currentRingshards` as it is just a ref to the `c.shards` (the same is true for `addrs` argument). I think it is better to fill-in the new map+list and then update `c.shards` and `c.list`.

ring.go
szuecs: agree

ring.go
szuecs: I don't want to change the if predicate into one, because there are 2 different else cases, that do the same thing, but maybe there is something we miss.

That's why I don't delete here right away, yes.

ring.go
szuecs: yes, that's on my list, but maybe we need to guard the assignment and then it might make sense to have both separated to not block the map while creating connection pools.

ring.go
szuecs: As far as I understand it should be safe, but and a test with race detector enabled was fine. 
```
ginkgo -race -focus '\[new\] dynamic setting ring shards' .
```
For this to run you need to delete a couple of Examples that panic (also without -race they panic).

ring.go
```diff
@@ -213,32 +215,72 @@ type ringShards struct {
...
-	shards := make(map[string]*ringShard, len(opt.Addrs))
+	shards, list := createRingShards(opt, opt.Addrs)
+
+	c := &ringShards{
+		opt: opt,
+
+		shards: shards,
+		list:   list,
+	}
+	c.rebalance()
```
AlexanderYastrebov, line 233: I think we now can change this to 
```go
c := &ringShards{opt: opt}
c.SetAddrs(opt.Addrs)

return c
```
and remove `createRingShards`.

ring_test.go
```diff
@@ -113,6 +113,45 @@ var _ = Describe("Redis Ring", func() {
 		Expect(ringShard2.Info(ctx, "keyspace").Val()).To(ContainSubstring("keys=100"))
 	})
 
+	Describe("[new] dynamic setting ring shards", func() {
+		It("uses one shard after setting it to one shard", func() {
+			Expect(ring.Len(), 2)
+			ring.SetAddrs(map[string]string{
+				"ringShardOne": ":" + ringShard1Port,
+			})
+			Expect(ring.Len(), 1)
```
AlexanderYastrebov, line 122: Maybe ring should export shards getter for tests and test should verify name+address as well as shard reuse...

ring.go
```diff
@@ -213,32 +215,72 @@ type ringShards struct {
...
+			newRingshards[k] = shard
+		} else {
+			defer shard.Client.Close()
+		}
 	}
-	c.rebalance()
 
-	return c
+	for k, addr := range addrs {
+		if _, ok := c.shards[k]; !ok || c.shards[k].addr != addr {
```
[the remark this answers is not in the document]
szuecs: thanks, done

ring.go
szuecs: makes sense, done.

ring.go
```diff
@@ -213,32 +215,72 @@ type ringShards struct {
 
 	mu       sync.RWMutex
 	hash     ConsistentHash
-	shards   map[string]*ringShard // read only
-	list     []*ringShard          // read only
+	shards   map[string]*ringShard // read only, replaced by SetAddrs
```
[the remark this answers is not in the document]
szuecs: +1

ring_test.go
szuecs: @vmihailenco wdyt?
It's an API change.

AlexanderYastrebov: > Maybe ring should export shards getter for tests
> It's an API change.

This could be done without API change using `export_test.go` trick https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd

ring.go
```diff
@@ -213,32 +215,54 @@ type ringShards struct {
...
+	c.SetAddrs(opt.Addrs)
 
-	for name, addr := range opt.Addrs {
-		shard := newRingShard(opt, name, addr)
-		shards[name] = shard
+	return c
+}
 
-		list = append(list, shard)
+func (c *ringShards) SetAddrs(addrs map[string]string) {
```
AlexanderYastrebov, line 233: I think we may need to protect it from concurrent calls

szuecs: Do you mean concurrent calls to SetAddrs?

ring.go
```diff
@@ -213,32 +215,72 @@ type ringShards struct {
...
+		return
+	}
 
-		shards: shards,
-		list:   list,
+	for k, shard := range c.shards {
+		if addr, ok := addrs[k]; ok && shard.addr == addr {
+			newRingshards[k] = shard
+		} else {
+			defer shard.Client.Close()
```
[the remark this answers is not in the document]
vmihailenco: I am fine with ignore the error, but logging the error is probably the best option.

ring.go
```diff
@@ -213,32 +215,54 @@ type ringShards struct {
...
+		}
+	}
 
-		shards: shards,
-		list:   list,
+	for k, addr := range addrs {
+		if shard, ok := c.shards[k]; !ok || shard.addr != addr {
```
vmihailenco, line 248: Should not you close the shard client when `shard.addr != addr`?

ring.go
```diff
@@ -213,32 +215,54 @@ type ringShards struct {
...
-	c.rebalance()
 
-	return c
+	list := make([]*ringShard, 0, len(newRingshards))
+	for _, shard := range newRingshards {
+		list = append(list, shard)
+	}
+
+	c.mu.Lock()
+	if !c.closed {
```
vmihailenco, line 259: If the client is closed, you probably should close `newRingshards` as well.

vmihailenco:
The code could be cleaner if you restructure it like this:

```go
shards := newRingShards(addrs, c.shards)
c.mu.Lock()
if c.closed {
    // close shards
} else {
    c.shards = shards
    c.list = shards.list()
}
c.mu.Unlock()
```

What do you think?

ring.go
AlexanderYastrebov: Every existing shard with the name in the `addrs` map but different address should be closed by the `defer` in the previous loop.

AlexanderYastrebov: Every existing shard with the name in the `addrs` map but different address should be closed by the `defer` in the previous loop.

szuecs:
@vmihailenco so you are suggesting a type change of `c.shards` to be able to add a `.list()` ?
I think the problem will be that we would need to lock this `.list()` and here we don't need it.
We could also add an intermediate type, but then we would do:
```
shards := newRingShards(addrs, c.shards)
c.mu.Lock()
if c.closed {
    // close shards
} else {
    c.shards = shards.shards() // returns map[string]*ringShard
    c.list = shards.list()
}
c.mu.Unlock()
```

ring.go
szuecs: added godoc

ring.go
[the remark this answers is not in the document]
szuecs: added log

ring.go
```diff
@@ -213,32 +216,66 @@ type ringShards struct {
...
+	}
+
+	for k, shard := range c.shards {
+		if addr, ok := addrs[k]; ok && shard.addr == addr {
+			shards[k] = shard
+		} else {
+			defer func() {
+				err := shard.Client.Close()
+				if err != nil {
+					log.Printf("Failed to close ring shard client %s: %v", shard.addr, err)
```
AlexanderYastrebov, line 248: I think it makes sense to log shard name as well

szuecs: done

ring.go
```diff
@@ -213,32 +216,66 @@ type ringShards struct {
...
+		c.list = list
+	}
+	c.mu.Unlock()
 
-		shards: shards,
-		list:   list,
+	if c.closed {
+		for _, shard := range shards {
+			shard.Client.Close()
+		}
```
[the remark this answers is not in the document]
AlexanderYastrebov: > What I am sure is wrong is the mention of defer as a problem

1. `SetAddrs` defers close of non-reused shard
2. `c.Close` is called concurrently and closes all current shards, including those defered to be closed by `SetAddrs`
3.  `SetAddrs` close non-reused shards once again on exit via defer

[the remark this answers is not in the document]
AlexanderYastrebov: > The proper cleanup should:
> close newly created shards (that `c.Close` did not know about, but do not close reused shards as they should be closed by `c.Close`)

The updated version 
```go
	c.mu.Lock()
	if !c.closed {
		c.shards = shards
		c.list = list
	} else {
		for _, shard := range shards {
			shard.Client.Close()
		}
	}
	c.mu.Unlock()
```
closes all shards in the else branch which is wrong - there might be reused shards already closed by the concurrent `c.Close` call - the else branch should only close shards created in the `SetAddrs` via `newRingShard`.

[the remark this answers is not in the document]
szuecs: Ok after reading multiple times code and your comments I understand your comment :)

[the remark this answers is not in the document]
szuecs: I guess each shard needs to know if it was closed or not, otherwise it seems to be impossible to not guard the whole SetAddrs func. wdyt?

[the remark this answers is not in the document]
szuecs: @AlexanderYastrebov what do you mean by `c.Close`, the func or the member and if func which func Client.Close() or ringShards.Close() ?

I think now it should be fine, but happy to discuss further. :)

vmihailenco:
>@vmihailenco so you are suggesting a type change of c.shards to be able to add a .list() ?

I suggest to slightly refactor `ringShards` type so we can fully replace it in `SetAddrs` holding the lock:

```go
// SetAddrs is safe to be used concurrently.
func (c *Ring) SetAddrs(addrs map[string]string) {
    // Construct new shards without holding the lock. Pass old shards to reuse existing clients.
    shards := newRingShards(addrs, c.shards)

    c.shardsMu.Lock()
    defer c.shardsMu.Unlock()

    if c.closed {
        // Ring is closed. Close new shards.
        shards.Close()
        return
    }

    // Atomically set new shards so they are used by the Ring.
    c.shards = shards
}
```

Does that make sense?

szuecs:
@vmihailenco yeah thanks, makes a lot of sense!

AlexanderYastrebov:
> I suggest to slightly refactor ringShards type so we can fully replace it in SetAddrs holding the lock:

This does not look to be safe for concurrent use: imagine two threads call `SetAddrs`, each creates a new `shards` and then consequently they take a lock and update `c.shards` - the second thread would overwrite `c.shards` and thus leak shards created by the first thread.

ring.go
```diff
@@ -213,32 +215,71 @@ type ringShards struct {
...
+	for _, shard := range shards {
+		list = append(list, shard)
+	}
 
-		shards: shards,
-		list:   list,
+	c.mu.Lock()
+	if c.closed {
+		for _, k := range newShardNames {
+			shards[k].Client.Close()
```
AlexanderYastrebov, line 274: Log an error?

ring.go
```diff
@@ -213,32 +215,71 @@ type ringShards struct {
...
+			defer func() {
+				c.mu.Lock()
+				if !c.closed {
+					err := shard.Client.Close()
+					if err != nil {
+						internal.Logger.Printf(context.Background(), "Failed to close ring shard client %s %s: %v", k, shard.addr, err)
+					}
+				}
+				c.mu.Unlock()
+			}()
```
AlexanderYastrebov, line 254: I think it would be cleaner to track `unusedShards` and close them at the end - this would be consistent with closing `newShardNames`.

ring.go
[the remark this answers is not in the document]
AlexanderYastrebov: I meant `func (c *ringShards) Close() error` as `c.Close` everywhere

AlexanderYastrebov:
As an idea we may have another mutex to guard only `SetAddrs` and `Close` (need to make sure to lock it first in both methods https://stackoverflow.com/questions/38489190/holding-two-mutex-locks-at-the-same-time). 
This way we ensure no two threads execute `SetAddrs/Close` while address change is in progress. `SetAddrs` would only check `c.closed` once and thus will not need to cleanup newly created shards.

vmihailenco:
>As an idea we may have another mutex

Yes, that is what I suggest as well.

szuecs:
@AlexanderYastrebov great idea, I hope this is now good, but also happy to rewrite. :)

szuecs:
@vmihailenco if you give it another try for a review would be great. I had an online session with @AlexanderYastrebov and rewrote some parts to make it nicer. 

In an unrelated follow up PR I would like to enhance the locking in ring.go and for example remove the current rebalance()

Issue #2096, opened 2022-05-28
should not redirect when meet network err in cluster mode


lintanghui:
Issue tracker is used for reporting bugs and discussing new features. Please use
[stackoverflow](https://stackoverflow.com) for supporting issues.

should not redirect req when meet netwrokr err in cluster, in cluster mode ,we default set maxRediect to 3, we should redirect when meet move ,ask or readonly err. but we should not redirect req when meet network err. but as code below:
```
err := c._processPipelineNode(ctx, node, cmds, failedCmds)
				if err == nil {
					return
				}
				if attempt < c.opt.MaxRedirects {
					if err := c.mapCmdsByNode(ctx, failedCmds, cmds); err != nil {
						setCmdsErr(cmds, err)
					}
				} else {
					setCmdsErr(cmds, err)
				}
```
when meet network err, it will retry all cmd, and if we also set maxRetry, it will cause multi retry req.
## Expected Behavior

shoudl not  redirect when meet network err in cluster mode
## Current Behavior
retry all cmd when meet network err in cluster mode

## Possible Solution

check err in cluster mode ,and only redirect in move|ask|readonly err

## Steps to Reproduce

1.
2.
3.
4.

PR #2098, opened 2022-05-29
fix:redirect only when meet move|ask|loading|readonly err

lintanghui:
redirect only when meet move|ask|loading|readonly err, when meet other err, such as network ,should not retry request in redirect loop, network err  request should be retry by maxretries. if we did retry in redirect loop, there may be maxretries*maxredirect request.  in other way, if we set maxretries to 0, but it may actually did retry when meet network err, which is not consistency with what option set.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -834,18 +834,6 @@ func (c *ClusterClient) process(ctx context.Context, cmd Cmder) error {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


lintanghui:
fix #2096

lintanghui:
@vmihailenco PTAL

vmihailenco:
We need more opinions on this and some tests. Otherwise, it is very likely that we break something unintentionally.

PR #2108, opened 2022-06-04
update cluster slots reply checking logic for Redis 7

wangrzneu:
In general a client implementation should just rely on the fact that certain parameters are at fixed positions as specified (referenced from https://redis.io/commands/cluster-slots/)

### DIFF
--- a/command.go
+++ b/command.go
@@ -2700,8 +2700,8 @@ func (cmd *ClusterSlotsCmd) readReply(rd *proto.Reader) error {
 				if err != nil {
 					return nil, err
 				}
-				if n != 2 && n != 3 {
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -2717,7 +2717,7 @@ func (cmd *ClusterSlotsCmd) readReply(rd *proto.Reader) error {
 
 				nodes[j].Addr = net.JoinHostPort(ip, port)
 
-				if n == 3 {
+				if n >= 3 {
```
vmihailenco, line 2720: If we have 4 replies, then we should also read (and probably discard) the 4th reply as well?

vmihailenco:
Thanks for working on this - I will try to add a build for Redis 7.

vmihailenco:
Thanks

Issue #2114, opened 2022-06-07
While using v9 connect to redis < 6.0, the err text is not compatibility in some case


szyhf:
About:

https://github.com/go-redis/redis/blob/4ddd7d18036f93ac0753e1905c7dbd17cde3ca58/redis.go#L231-L237

## Expected Behavior

If I'm using redis version < 6, the return err will match the key words and continue.

## Current Behavior

Return err with text is

```go
ERR unknown command `hello`, with args beginning with: `3`
```

---

```txt
the quotation marks surround "hello" is "`", not "'", I'm not sure if every verison < 6 is "`".
```

## Possible Solution

```go
	// For redis-server <6.0 that does not support the Hello command,
	// we continue to provide services with RESP2.
	if err := conn.Hello(ctx, 3, username, password, "").Err(); err == nil {
		auth = true
	} else if !strings.HasPrefix(err.Error(), "ERR unknown command `hello`") {
		return err
	}
```

Using has prefix instead of string equal.

## Steps to Reproduce

1. Runing a redis server of `redis:5.0-alpine`
2. Create a client and connect to the server

vmihailenco:
It probably should be `strings.HasPrefix(err.Error(), "ERR unknown command")`, right?

Otherwise, it looks good. Could you send a PR?

szyhf:
谢谢您的来信，我一般2天会查看一次邮箱。---------------------------------------Thank you for your letter, I usually check Email every two days.

szyhf:
Thanks for reply.

@vmihailenco I'm not sure if other command should be ignore, if you think it looks good,I'm willing to make a PR :)

vmihailenco:
As I understand, the error is either `ERR unknown command `hello`` or `ERR unknown command "hello"` so `strings.HasPrefix(err.Error(), "ERR unknown command `hello`")` won't work. That is why I suggest `strings.HasPrefix(err.Error(), "ERR unknown command")` - we always execute the same command here so it should be fine.

So please send a PR.

PR #2119, opened 2022-06-10
fix: #2114 for redis-server not support Hello

szyhf:
fix: #2114 Using `strings.HasPrefix` instead of `equal`


vmihailenco:
Thanks

Issue #2117, opened 2022-06-10
Closing pubsub gracefully with cancellation context will log discarding connection


titpetric:
When using context cancellation, the underlying connection isn't closed nicely, triggering a logger error in [pubsub.go:159](https://github.com/go-redis/redis/blob/master/pubsub.go#L159).

```
redis: 2022/06/10 11:26:24 pubsub.go:159: redis: discarding bad PubSub connection: read tcp 127.0.0.1:49920->127.0.0.1:6379: use of closed network connection
```

The `closeTheCn` function is invoked from `reconnect(ctx, err)`, `Close()` - the context cancellation happens before Close() is invoked; pubsub has an `initHealthCheck` which uses a `context.TODO()` value, rather than a cancellation context supplied to Subscribe, Receive, ReceiveMessage.

closeTheCn is checking `!c.closed` to log the error. And then changes c.cn to nil after invoking closeConn;

1. isBadConn would return true, if it received a context.Cancelled/DeadlineExceeded;
2. releaseConn invokes c.reconnect as isBadConn would return true

It's really hard to say what the appropriate fix here would be. Ideally, if a context cancellation occurs, the connections should be silently cleaned up, and the underlying ctx.Err() returned.

titpetric:
```
goroutine 21 [running]:
runtime/debug.Stack()
	/go/src/runtime/debug/stack.go:24 +0x65
runtime/debug.PrintStack()
	/go/src/runtime/debug/stack.go:16 +0x19
github.com/go-redis/redis/v8.(*PubSub).closeTheCn(0xc00054e000, {0x2141680?, 0xc0002c0460})
	/root/tyk/redis/pubsub.go:161 +0xf3
github.com/go-redis/redis/v8.(*PubSub).reconnect(0x2141680?, {0x214c8d8, 0xc00003c540}, {0x2141680?, 0xc0002c0460?})
	/root/tyk/redis/pubsub.go:151 +0x38
github.com/go-redis/redis/v8.(*PubSub).releaseConn(0xc00054e000, {0x214c8d8, 0xc00003c540}, 0xc00010f4a0?, {0x2141680, 0xc0002c0460}, 0x18?)
	/root/tyk/redis/pubsub.go:146 +0x89
github.com/go-redis/redis/v8.(*PubSub).releaseConnWithLock(0xc00054e000, {0x214c8d8?, 0xc00003c540?}, 0x0?, {0x2141680?, 0xc0002c0460?}, 0x0?)
[...]

PR #2120, opened 2022-06-10
Listen for context.Cancelled, try to cancel read if detected, still racy

titpetric:
This is somewhat of an 80% attempt at solving #2117 - as net.Conn doesn't have a cancellation context:

- I've started a background goroutine in Conn.WithReader that tries to listen for that cancellation,
- If cancellation has been caught, the net.Conn read is cancelled with a SetReadDeadline,
- Maintain a context.Canceled value with error wrapping and an explicit early exit cancellation check,

I'm still not entirely happy about this, as some error details get lost on cancellation (reader error returned). The function produces 3 error objects, which could be merged with go-uber/multierr to enable value checks with errors.Is.

### DIFF
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -3,7 +3,9 @@ package pool
 import (
 	"bufio"
 	"context"
+	"fmt"
 	"net"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks, but I think we should be moving away from supporting context timeouts, not adding more complexity. See https://uptrace.dev/blog/posts/go-context-timeout.html

PR #2122, opened 2022-06-14
Improve redis lib garbage collection

evilaffliction:
1. Provide method with a custom reply data reader
2. Provide buffered methods for internal reader


evilaffliction:
accidentally pushed PR to the original git, not to the fork

Issue #2121, opened 2022-06-13
No results but also no error!


SamiAlsubhi:
go-redis v8
go version go1.17.1 darwin/amd64

## Expected Behavior

if a command does not find a key/member than it should return `redis.Nil` in the error part of the result

## Current Behavior

command does not find a match and returns nil value and nil error.

## Steps to Reproduce

When I print val and err i get `[<nil>] <nil>` which does not make sense. Since there are no results, the error should be redis.Nil.

```go
func (r *RedisClient) GetOne(ctx context.Context, item *Identifier) (*SearchResponse, error) {
	pos, err := r.GeoPos(ctx, item.key(), item.Id).Result()

	if err == redis.Nil {
		return &SearchResponse{0, nil}, nil
	} else if err != nil {
		return nil, errors.New("failed to get point")
	}
	fmt.Println(pos, err) // [<nil>] <nil>
	point := pos[0]
	var objects []Object
	objects = append(objects, Object{
		Identifier: item,
		Location: &Location{
			Lat: point.Latitude,
			Lng: point.Longitude,
		},
	})

	return &SearchResponse{1, &objects}, nil

}
```

vmihailenco:
>if a command does not find a key/member than it should return redis.Nil in the error part of the result

go-redis returns redis.Nil only when Redis Server returns `(nil)` response.

SamiAlsubhi:
if you use the same command in `redis-cli` on a key that does not exist:

```
127.0.0.1:6379> GEOPOS DOESNT EXIST
1) (nil)
```
you get `nil` but in this library I am getting no `redis.Nil` and I am getting a list of one nil item in the results `[<nil>]`.

vmihailenco:
Sorry, I've skipped that part. Yes, this is a bug.

SamiAlsubhi:
I am using v8

git-hulk:
@vmihailenco seems GeoPosCommand have to break the SetVal/Val/Result API to fix this issue since need to change from `[]*GeoPos` to `[]interface{}`. If this makes sense, I can submit a PR.

VladimirStepanov:
@vmihailenco @SamiAlsubhi Which result we expect for this?
```bash
127.0.0.1:6379> GEOPOS Sicily Palermo NonExisting Catania
1) 1) "13.36138933897018433"
   2) "38.11555639549629859"
2) (nil)
3) 1) "15.08726745843887329"
   2) "37.50266842333162032"
```

[...]

PR #2135, opened 2022-06-29
feat: add redis.nil field to GeoPos result

VladimirStepanov:
Close https://github.com/go-redis/redis/issues/2121

This will work as:
```go
v, _ := rdb.GeoPos(ctx, "Sicily", "Palermo", "NonExisting").Result()
for _, g := range v {
    if !g.IsRedisNil() {
	fmt.Println(g)
    }
}
```


VladimirStepanov:
@vmihailenco hello, could you review it, plz?

vmihailenco:
Thanks for the PR, but I don't think this is much better than the current behavior. Let's keep it as is.

PR #2137, opened 2022-07-01
Disregard failed Pings in updateLatency() for cluster nodes

keitwb:
It is possible that a ping command might fail very quickly due to an
immediate connection refused message or some other server failure condition.
In this case, the derived latency is not reliable and should not be used
to compare this node to other nodes in the cluster.

This change will only count successful Ping commands in the average latency
and in the case of no successful Pings, set the latency to an arbitrarily
high value to make this node have the least priority when routing by latency.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -204,15 +204,26 @@ func (n *clusterNode) updateLatency() {
 	const numProbe = 10
 	var dur uint64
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
Thanks :+1:

PR #2143, opened 2022-07-05
baseClient#process(): reset retry attempts on successfully written XREAD

### LINKED ISSUE #2131: XREADs with BLOCK 0 on multiple consecutive Redis restarts exceed retries
@lippserd just had the observation that blocking XREADs without timeouts (BLOCK
0) on multiple consecutive Redis restarts and I/O timeouts exceeds Go-Redis
internal retries and eventually leads to errors. @julianbrost
looked at this for clarification, here is his finding:

go-redis only considers a command successful when it returned something,
so a successfully started blocking XREAD consumes a retry attempt each
time the underlying Redis connection is terminated. If this happens
often before any element appears in the stream, this error is
propagated.

## Expected Behavior

Successfully _sent_ XREADs with BLOCK reset the internal retries.

## Current Behavior

Successfully _sent_ XREADs with BLOCK just failed while _reading_ consume internal retries.

## Steps to Reproduce

1.
2.
3.
4.

## Context (Environment)

refs Icinga/icingadb#504

Al2Klimov:
if a retryable error ocurred during waiting for the response.
Otherwise we could run out of retries despite the XREAD
was actually partially successful.

fixes #2131

### DIFF
--- a/command.go
+++ b/command.go
@@ -23,6 +23,7 @@ type Cmder interface {
 	SetFirstKeyPos(int8)
 
 	readTimeout() *time.Duration
+	longRunning() bool
 	readReply(rd *proto.Reader) error
 
 	SetErr(error)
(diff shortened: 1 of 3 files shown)


vmihailenco:
Thanks, but retries but are already a pain point and I don't want to further complicate the retry logic.

PR #2150, opened 2022-07-12
refactor: remove unused context attribute from conn

KnutZuidema:
The existing context attribute of `conn` was never used.


vmihailenco:
Thanks

PR #2153, opened 2022-07-13
fix: Handle panic in ringShards Hash function when Ring got closed

### LINKED ISSUE #2126: go-redis/v8 nil shard in the shards map
## Expected Behavior

no panic due to nil pointer dereference after retrieving a nil shard from the shards map

## Current Behavior

[There is no check if a shard is not nil before returning it](https://github.com/go-redis/redis/blob/a40b4f0f69429e0b11945193cbcc05e565cd9776/ring.go#L305).

It might cause panic due to nil pointer dereference [here](https://github.com/go-redis/redis/blob/a40b4f0f69429e0b11945193cbcc05e565cd9776/ring.go#L677)

## Possible Solution

Check if shard is not nil before returning it.

## Steps to Reproduce

We have a proprietary code which wraps `go-redis/go` and sometimes recreates the Ring structure (and hence repopulating the shards map), all while holding the mutex.
Despite it, sometimes after the ring is recreated, we get panic in the place linked above because the shard retrieved by the name is nil.

## Possible Implementation

```go
func (c *ringShards) GetByName(shardName string) (*ringShard, error) {
	if shardName == "" {
		return c.Random()
	}

	c.mu.RLock()
	shard := c.shards[shardName]
	c.mu.RUnlock()

	if shard == nil {
		return nil, fmt.Errorf("a shard named %q is nil", shardName)
	}

	return shard, nil
}
```

kavu:
Fixes #2126 

Tried to figure out how we can fix the #2126 the panic that raises from `func (c *ringShards) Hash(key string)`. And the easiest way is no add missing wiping of `numShard` on `ringShards`, so `c.hash.Get` (where `hash` is nil) will not be called.

Interesting thing that without moving the `c.mu.RUnlock()` in `func (c *ringShards) Hash(key string)` up and to make it deferred test got completely stuck. I think it is connected to the unhandled panic. In a deferred approach mutex will be released no matter how the function exited. So I changed some of the locks to avoid similar problems in other places.

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -259,10 +259,11 @@ func (c *ringShards) Hash(key string) string {
 	var hash string
 
 	c.mu.RLock()
+	defer c.mu.RUnlock()
+
 	if c.numShard > 0 {
 		hash = c.hash.Get(key)
 	}
-	c.mu.RUnlock()
 
 	return hash
 }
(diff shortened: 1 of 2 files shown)


kavu:
@vmihailenco Hi! Any chance you can check this and #2157 out? Thanks a lot in advance!

vmihailenco:
:+1:

PR #2154, opened 2022-07-14
refactor: remove unused context attributes

KnutZuidema:
For many types the context attribute was not removed after moving to context propagation via function parameters.


vmihailenco:
Thanks

PR #2157, opened 2022-07-14
fix: provide a signal channel to end heartbeat goroutine

kavu:
Hi!

We continue to investigate various problems with `Ring` which arise from our case, briefly described in #2126.

This time it is massive goroutines leaks of Heartbeat goroutines. The reason is that there is no way to exit the ticker `range channel` loop: 

https://github.com/go-redis/redis/blob/e061db8c13fd6b5e007d73370873026e360719de/ring.go#L313-L334

Simple program to demonstrate the problem:

```go
package main

import (
	"context"
	"fmt"
	"log"
	"net/http"
	_ "net/http/pprof"
	"runtime"

)

func main() {
	ctx := context.Background()

	opts := &redis.RingOptions{
		Addrs: map[string]string{
			"shard0": ":6379",
		},
	}

	for i := 0; i < 1000; i++ {
		rdb := redis.NewRing(opts)
		rdb.Ping(ctx)
		rdb.Close()
	}

	// Let's check how many Goroutines we have now
	fmt.Println(runtime.NumGoroutine())

	// Let's start HTTP and check pprof at http://localhost:8080/debug/pprof/goroutine?debug=1
	log.Println(http.ListenAndServe(":8080", nil))
}

```

In this example will see around 1000 goroutines, and `pprof/goroutine` profile will show something like (note the thousand):

```
goroutine profile: total 1003
1000 @ 0x102f93878 0x102f62cc0 0x102f62724 0x103191688 0x102fc2dc4
#	0x103191687	github.com/go-redis/redis/v9.(*ringShards).Heartbeat+0x87	/Users/kavu/go/pkg/mod/github.com/go-redis/redis/v9@v9.0.0-beta.1.0.20220714104342-e061db8c13fd/ring.go:318

1 @ 0x102f93878 0x102f8cf54 0x102fbd094 0x103002edc 0x103003a14 0x103003a05 0x10308637c 0x1030907d8 0x103126058 0x102fc2dc4
#	0x102fbd093	internal/poll.runtime_pollWait+0xa3		/opt/homebrew/Cellar/go/1.18.4/libexec/src/runtime/netpoll.go:302
#	0x103002edb	internal/poll.(*pollDesc).wait+0x2b		/opt/homebrew/Cellar/go/1.18.4/libexec/src/internal/poll/fd_poll_runtime.go:83
``` 

Such an amount of goroutines alone can create a visible pressure on CPU (actually this amount is almost exactly the same as we observed one of our services), but if you will set a pretty low value for a heartbeat frequency it can become even more visible and painful.

I tested in on v8, current v9 beta, and the latest master commit, just to be 100% sure.

So I suggest a relatively straightforward approach — a signal channel for the goroutine to end the loop.

Another additional option to consider is an option to disable heartbeats completely. Sill, it won't be enough, as in cases when it is enabled leak will still be observed. 

I'll gladly discuss any other approach here if you have any concerns here.

Thank you!

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -310,11 +310,10 @@ func (c *ringShards) Random() (*ringShard, error) {
 }
 
 // Heartbeat monitors state of each shard in the ring.
-func (c *ringShards) Heartbeat(frequency time.Duration, closeCh chan struct{}) {
+func (c *ringShards) Heartbeat(ctx context.Context, frequency time.Duration) {
 	ticker := time.NewTicker(frequency)
 	defer ticker.Stop()
 
-	ctx := context.Background()
 	for {
 		select {
 		case <-ticker.C:
@@ -332,7 +331,7 @@ func (c *ringShards) Heartbeat(frequency time.Duration, closeCh chan struct{}) {
 			if rebalance {
 				c.rebalance()
 			}
-		case <-closeCh:
+		case <-ctx.Done():
 			return
 		}
 	}
@@ -392,10 +391,10 @@ func (c *ringShards) Close() error {
 //------------------------------------------------------------------------------
 
 type ring struct {
-	opt                 *RingOptions
-	shards              *ringShards
-	cmdsInfoCache       *cmdsInfoCache //nolint:structcheck
-	hearbeatCloseSignal chan struct{}
+	opt               *RingOptions
+	shards            *ringShards
+	cmdsInfoCache     *cmdsInfoCache //nolint:structcheck
+	heartbeatCancelFn context.CancelFunc
 }
 
 // Ring is a Redis client that uses consistent hashing to distribute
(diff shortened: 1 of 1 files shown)


KnutZuidema:
The idiomatic approach for canceling a goroutine is using a `context.Context` parameter. It provides all of the benefits that the signal channel would, while keeping the interface uniform (a context is used in many other places in this library) and also keeping with Go conventions in general. 
see:  
https://go.dev/blog/context  
https://pkg.go.dev/context

kavu:
@KnutZuidema thank you for pointing it out! My missed opportunity! Reworked a little bit.

MEZk:
@KnutZuidema any comments or u can merge?

KnutZuidema:
@MEZk I am not a member if this repository, so I can not merge, however from my side this looks good 😉 @vmihailenco is the maintainer I believe

vmihailenco:
Thanks for reviews everyone :+1:

PR #2162, opened 2022-07-20
Add Subscribe and PSubscribe to Cmdable

### LINKED ISSUE #845: redis.Cmdable  interface didn't have Subscribe?
Is there some reason for it?
I have found "Subscribe" in Client and ClusterClient.

evilstreak:
We're using the `Cmdable` interface and need to start subscribing to keyspace notifications.

Fixes #845 (though it was closed a while ago)


evilstreak:
OK, this is a bit more involved than I'd realised 😅 I'll follow up with something that works at some point

PR #2163, opened 2022-07-21
feat(extra/redisprometheus): prometheus.Collector implementation for redis clients

KnutZuidema:
I've been using prometheus to aggregate the pool stats provided by the clients and thought the implementation might be helpful for others as well.


vmihailenco:
Thanks

PR #2060, opened 2022-03-25
fix: Use redis over SSH channel(#2057)

slowrookie:
useage:
```golang
     sshConfig := &ssh.ClientConfig{
		User:            "root",
		Auth:            []ssh.AuthMethod{ssh.Password("password")},
		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
		Timeout:         15 * time.Second,
	}

	client, err := ssh.Dial("tcp", "remoteIP:22", sshConfig)
	if err != nil {
		log.Println(fmt.Errorf("%v", err))
	}

	cli := redis.NewClient(&redis.Options{
		Addr: net.JoinHostPort("127.0.0.1", "6379"),
		Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
			return client.Dial(network, addr)
		},
		ReadTimeout:  -1,
		WriteTimeout: -1,
	})
       
       

```

PR #2176, opened 2022-08-03
fix(internal/pool): call SetDeadline even if timeout is zero

### LINKED ISSUE #2175: `pubsub.Receive()` will timeout
#2060 cause `SetReadDeadline` not called, so `PubSub.ReceiveTimeout(ctx, 0)` will not get a read deadline set, thus it will receive the `i/o timeout` error after some time. Related #2139.

## Expected Behavior

`PubSub.ReceiveTimeout(ctx, 0)` will block forever until a message is received, So does `PubSub.Receive()`, `PubSub.ReceiveMessage()` and `PubSub.Channel()`.

## Current Behavior
`PubSub.Receive()` will timeout if read nothing for sometime.
```
received subscribe: hello <nil>
redis: 2022/08/03 22:36:54 pubsub.go:168: redis: discarding bad PubSub connection: read tcp [::1]:53016->[::1]:6379: i/o timeout
received <nil> read tcp [::1]:53016->[::1]:6379: i/o timeout
received subscribe: hello <nil>
redis: 2022/08/03 22:36:57 pubsub.go:168: redis: discarding bad PubSub connection: read tcp [::1]:53035->[::1]:6379: i/o timeout
received <nil> read tcp [::1]:53035->[::1]:6379: i/o timeout
received subscribe: hello <nil>

```

## Possible Solution
Revert #2060

## Steps to Reproduce
```go
package main

import (
    "context"
    "fmt"

)

func main() {
    c := redis.NewClient(&redis.Options{})
    pubsub := c.Subscribe(context.Background(), "hello")
    for {
        m, err := pubsub.Receive(context.Background())
        fmt.Println("received", m, err)
    }
}
```

## Context (Environment)
Redis: 7.0.3
go-redis: master

j178:
This PR revert #2060, which caused #2175.

Fixes #2175.

### DIFF
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -64,8 +64,10 @@ func (cn *Conn) RemoteAddr() net.Addr {
 }
 
 func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error) error {
-	if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
-		return err
+	if timeout >= 0 {
+		if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
+			return err
+		}
 	}
 	return fn(cn.rd)
 }
@@ -73,8 +75,10 @@ func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(r
 func (cn *Conn) WithWriter(
 	ctx context.Context, timeout time.Duration, fn func(wr *proto.Writer) error,
 ) error {
-	if err := cn.netConn.SetWriteDeadline(cn.deadline(ctx, timeout)); err != nil {
-		return err
+	if timeout >= 0 {
+		if err := cn.netConn.SetWriteDeadline(cn.deadline(ctx, timeout)); err != nil {
+			return err
+		}
 	}
 
 	if cn.bw.Buffered() > 0 {
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -64,10 +64,8 @@ func (cn *Conn) RemoteAddr() net.Addr {
 }
 
 func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error) error {
-	if timeout != 0 {
-		if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
-			return err
-		}
+	if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
+		return err
 	}
 	return fn(cn.rd)
 }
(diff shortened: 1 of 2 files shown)


internal/pool/conn.go
```diff
@@ -64,21 +64,17 @@ func (cn *Conn) RemoteAddr() net.Addr {
 }
 
 func (cn *Conn) WithReader(ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error) error {
-	if timeout != 0 {
-		if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
-			return err
-		}
+	if err := cn.netConn.SetReadDeadline(cn.deadline(ctx, timeout)); err != nil {
```
vmihailenco, line 67: I think this should be `timeout >= 0` so it works with PubSub

options.go
```diff
@@ -76,7 +76,7 @@ type Options struct {
 	// Default is 3 seconds.
 	ReadTimeout time.Duration
 	// Timeout for socket writes. If reached, commands will fail
-	// with a timeout instead of blocking.
+	// with a timeout instead of blocking. Use value -1 for no timeout and 0 for default.
```
vmihailenco, line 79: `-1 for no timeout` is meant to disable the timeout, but it looks like you always call `SetReadDeadline` no matter what. Perhaps the comment must be updated?

vmihailenco, line 79: `-1 for no timeout` is meant to disable the timeout, but it looks like you always call `SetReadDeadline` no matter what. Perhaps the comment must be updated?

j178: This sentience is copied from the `ReadTimeout` description above.
I think `no timeout` == `disable the timeout` == `block until read/write succeeds`, do I understand this correctly?

`-1` will be adjusted to `0` here: https://github.com/go-redis/redis/blob/084c0c8914d4d84918a81b867b61d94b6c343ecd/options.go#L152-L157
then in `SetReadDeadline` or `SetWriteDeadline`, the timeout is diabled, and the read/write operation will block until it succeeds.

internal/pool/conn.go
j178: In the `pubsub.ReceiveTimeout(ctx, timeout)` case, we don't say anything about the timeout argument, is it intended to *ignore* negative timeout? 

for example `pubsub.ReceiveTimeout(ctx, -1)` means `do not set a deadline, use the previous set deadline` ?

```go
_, _ = pubsub.ReceiveTimeout(ctx, 3*time.Second) // this call set the deadline to time.Now().Add(3*time.Second)
time.Sleep(3)
_, _ = pubsub.ReceiveTimeout(ctx, -1) // this call will timeout immediately, but it may be intended to block forever
```

options.go
vmihailenco: That is correct, but it looks like we should also support `-2` to disable `Set*Deadline` calls completely.

internal/pool/conn.go
vmihailenco: `ReceiveTimeout(0)` is the same as `Receive()` which means block forever. Negative values should probably disable `Set*Deadline` calls completely.

But I admit it is subtle.

woutslakhorst:
I can confirm this also fixes the disconnect behaviour when using Redis sentinel.

PR #2177, opened 2022-08-03
feat(pubsub): support sharded pub/sub

j178:
Closes #2115 

Sharded pub/sub ref: https://redis.io/docs/reference/cluster-spec/#publishsubscribe

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1528,6 +1528,16 @@ func (c *ClusterClient) PSubscribe(ctx context.Context, channels ...string) *Pub
 	return pubsub
 }
 
+// SSubscribe Subscribes the client to the specified shard channels.
+func (c *ClusterClient) SSubscribe(ctx context.Context, channels ...string) *PubSub {
+	pubsub := c.pubSub()
+	if len(channels) > 0 {
+		_ = pubsub.SSubscribe(ctx, channels...)
+	}
+	return pubsub
+}
+
+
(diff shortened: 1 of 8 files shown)


j178:
Seems like the two failing tests is not related.

vmihailenco:
Thanks

schien-dong:
Tested with "GO111MODULE=off", the changes to commands,go, universal.go and commands.go break the build with the following errors:
`commands.go:355:62: undefined: StringIntMapCmd`
`commands.go:3125:78: undefined: StringIntMapCmd`
`commands.go:3132:9: undefined: NewStringIntMapCmd`
`ring.go:568:18: c.shards undefined (type *Ring has no field or method shards)`

I won't create a Pull Request as I don't know the full story behind this overnight merge. But here are the manual patch to temporarily Revert the damage.

ring.go Line 568, disable Ring::SSubscribe
universal.go Line 193, remove interface SSubscribe
commands.go remove all instances of SPublish, PubSubShardChannels, PubSubShardNumSub

The true fix is to add in the missing line in ring.go, which is likely missed in the merge.

j178:
@schien-dong Thanks for your report. I think some refactor changes(renaming `StringIntMapCmd` to `MapStringIntCmd` etc) to master after this PR is opened cause the break. I'll send a PR to fix this.

The current merge strategy is not safe IMHO, if it is forced to merge the master back, and trigger a new CI run, this break can be avoided.

PR #2194, opened 2022-08-26
WIP Update redis otel with option to not set raw command as an attribute

Cull-Methi:
TBD

### DIFF
--- a/extra/redisotel/redisotel.go
+++ b/extra/redisotel/redisotel.go
@@ -18,8 +18,9 @@ const (
 )
 
 type TracingHook struct {
-	tracer trace.Tracer
-	attrs  []attribute.KeyValue
+	tracer    trace.Tracer
+	attrs     []attribute.KeyValue
+	logRawCmd bool
 }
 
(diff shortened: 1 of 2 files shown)


Cull-Methi:
Updated the PR title to match the commitlint spec (I think) - if someone can show me how to kick it off again that'd be great 🙃

extra/redisotel/redisotel.go
```diff
@@ -136,3 +142,10 @@ func WithAttributes(attrs ...attribute.KeyValue) Option {
 		cfg.attrs = append(cfg.attrs, attrs...)
 	})
 }
+
+// WithIgnoreRawCmd tells the tracing hook not to log raw redis commands.
+func WithIgnoreRawCmd() Option {
```
vmihailenco, line 147: Let's rename this to `WithDBStatement(true/false)` so it can be more easily enabled/disabled - https://uptrace.dev/blog/posts/go-functional-options-named-args.html#bool-options

vmihailenco:
@Cull-Methi aside from the option name this looks pretty good. Could you update the name?

PR #2201, opened 2022-08-31
Use strings.Contains for script error matching

cotiAutonomousCar:


vmihailenco:
Any reason this is better or covers more cases?

PR #2212, opened 2022-09-16
docs: another ScanType example with "Hash" as the KeyType

rustymagnet3000:
Builds on [existing example code](https://github.com/go-redis/redis/blob/e061db8c13fd6b5e007d73370873026e360719de/example_test.go#L253).

Differs from previous example by passing in "Hash" and not throwing away the `keys`.

`keysFromScan, cursor, err = rdb.ScanType(context.TODO(), cursor, "key*", 10, "hash").Result()`


vmihailenco:
Thanks

PR #2232, opened 2022-10-06
fix: improve pipelines retry logic

vmihailenco:
Closes https://github.com/go-redis/redis/pull/1883

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -416,7 +416,6 @@ func (c *baseClient) generalProcessPipeline(
 ) error {
 	err := c._generalProcessPipeline(ctx, cmds, p)
 	if err != nil {
-		setCmdsErr(cmds, err)
 		return err
 	}
 	return cmdsFirstErr(cmds)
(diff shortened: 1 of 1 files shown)


TAW2025:
Yes

redis.go
```diff
@@ -290,27 +290,7 @@ func (c *baseClient) withConn(
...
-		_ = cn.Close()
-		// Wait for the goroutine to finish and send something.
-		<-errc
-
-		err = ctx.Err()
-		return err
-	case err = <-errc:
-		return err
-	}
+	return fn(ctx, cn)
```
adriansmares, line 293: Please note that this is a rather big change compared to `v8`, and it should be documented as part of the changelog. 

Compared to `v8`, it is no longer possible to use `context.WithCancel` due to this change. I've followed the attempts over time to make `net.Conn` work with context cancellations and I know this is a huge headache, but please at least mention next to the `ContextTimeoutEnabled` entry in the changelog that asynchronous cancellation is no longer supported in `v9`.

vmihailenco: Done in https://github.com/go-redis/redis/commit/4cb9d14eb6b7f60f967e7f2c040c5795a847e0a0

PR #2237, opened 2022-10-07
fix: use all provided sections

injeniero:
Fixes the INFO command that was not using all the given sections.


vmihailenco:
Thanks

PR #2250, opened 2022-10-14
feat: port v9 WithDBStatement into v8 

vincentserpoul:
to be able to use tracing without any privacy concern in redis < 7.0, I need to disable the dbstatement.
This is simply a port of the v9 functionality, without changing any default behaviour for current v8 tracing.

I also updated the tracing libs (if that is a problem, I can simply remove the last commit)


vincentscdc:
just to get a quick estimate, no pressure at all, but is there a chance this gets merged and released within the next 2 months, or is the focus elsewhere?

vmihailenco:
Thanks!

PR #2255, opened 2022-10-19
Dont quit scan unmarshal on first failure

cpoole:
Periodically a data type might need to change in a redis struct. I've personally had this occur if something was a string and i'd now like it to be a more specific type (maybe a boolean). The current scan method exits on the first field unmarshal failure. This PR allows scan to continue until the end and returns all field errors concatenated together. 

This allows the caller to check for their specific "migrated" field in the error and ignore it if they're ok with that going back to it's zero value


cpoole:
the go 1.18 build seems to have failed for reasons unrelated to this PR. Im not sure how to rerun the flaky test

ndyakov:
@ofekshenawa this one makes sense to me, but in the rest of the library we often fail on the first error that occurs. Let me know what you think.

PR #2272, opened 2022-11-01
Capture error correctly in withConn

DomParfitt:
# What
Correctly passes the `error` from calling `fn` to `releaseConn` in `withConn`.

# Why
This appears to be a regression introduced in #2232. 

Currently if the connection times out, `fn` will return an error but this will not be passed to `releaseConn` (instead the error from `getConn` will be passed). This will result in `releaseConn` attempting to put the `Conn` back into the connection pool. If the Conn's read buffer is non-empty then the `Conn` will instead be removed but otherwise the `Conn` will be returned to the pool. This can result in the `Conn` being reused and returning incorrect data on read.


redis.go
```diff
@@ -288,11 +288,9 @@ func (c *baseClient) withConn(
 		return err
 	}
 
-	defer func() {
-		c.releaseConn(ctx, cn, err)
-	}()
-
-	return fn(ctx, cn)
+	err = fn(ctx, cn)
```
adriansmares, line 291: I am not sure if this is entirely correct. If `fn` panics the connection will be lost, thus lowering the capacity of the connection pool permanently.

I think a name returned should be used here, and in case of a panic, the connection should always be closed.

PR #2273, opened 2022-11-03
Add withConn tests

DomParfitt:
Adds tests to cover the behaviour witnessed in #2272


vmihailenco:
Thanks! I will move this to internal_test, but otherwise this looks pretty good :+1:

PR #2283, opened 2022-11-15
fix: fixes ring.SetAddrs and rebalance race 

AlexanderYastrebov:
While working on reducing `ring.SetAddrs` lock contention (see https://github.com/go-redis/redis/pull/2190#discussion_r953040289) I have discovered a race condition between `SetAddrs` and `rebalance` which I would like to fix first and separately.

The change consists of two commits:
- a test to reproduce the race
- the fix 

The fix ensures atomic update of `c.hash` and `c.shards`, otherwise `c.hash` may return shard name that is not in the `c.shards` and cause ring operation panic. 

`BenchmarkRingRebalanceLocked` shows rebalance latency if that is a concern:
```
go test . -run=NONE -bench=BenchmarkRingRebalanceLocked -v -count=10 | benchstat /dev/stdin
name                   time/op
RingRebalanceLocked-8  8.50µs ±14%
```

(Note: it essentially reverts https://github.com/go-redis/redis/commit/a46b053aa626a005a30dfb1ac4e096abcce1ef76)

Updates https://github.com/go-redis/redis/issues/2077
FYI @szuecs

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -95,9 +95,10 @@ func GetSlavesAddrByName(ctx context.Context, c *SentinelClient, name string) []
 }
 
 func (c *Ring) ShardByName(name string) *ringShard {
-	return c.sharding.ShardByName(name)
+	shard, _ := c.sharding.GetByName(name)
+	return shard
 }
 
-func (c *ringSharding) ShardByName(name string) *ringShard {
-	return c.shards.m[name]
+func (c *Ring) ShardByKey(key string) (*ringShard, error) {
+	return c.sharding.GetByKey(key)
 }
--- a/export_test.go
+++ b/export_test.go
@@ -102,3 +102,7 @@ func (c *Ring) ShardByName(name string) *ringShard {
 func (c *Ring) ShardByKey(key string) (*ringShard, error) {
 	return c.sharding.GetByKey(key)
 }
+
+func (c *Ring) RebalanceLocked() {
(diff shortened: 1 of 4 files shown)


ring_test.go
```diff
@@ -739,3 +740,89 @@ var _ = Describe("Ring Tx timeout", func() {
...
 	})
 })
+
+type fixedHash string
+
+func (h fixedHash) Get(string) string {
+	return string(h)
+}
+
+func TestRingSetAddrsAndRebalanceRace(t *testing.T) {
```
vmihailenco, line 750: Could you please create and move this test to `internal_test.go` that has `package redis` so we don't need to export more commands in `export_test.go`?

export_test.go
```diff
@@ -95,9 +95,6 @@ func GetSlavesAddrByName(ctx context.Context, c *SentinelClient, name string) []
 }
 
 func (c *Ring) ShardByName(name string) *ringShard {
-	return c.sharding.ShardByName(name)
-}
-
-func (c *ringSharding) ShardByName(name string) *ringShard {
```
AlexanderYastrebov, line 101: This was only used by above `ShardByName`

ring_test.go
AlexanderYastrebov: Moved test and benchmark to internal_test.go

vmihailenco:
Thanks for the fix :+1:

PR #2292, opened 2022-11-21
fix: reduce `SetAddrs` shards lock contention

AlexanderYastrebov:
Introduces a new lock to make `SetAddrs` calls exclusive.
This allows release of shards lock for the duration of potentially long `newRingShards` call.

`TestRingSetAddrsContention` observes number of pings increased from <1000 to ~40_000.

See https://github.com/go-redis/redis/pull/2190#discussion_r953040289

Updates #2077

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -219,6 +219,10 @@ type ringSharding struct {
 	hash      ConsistentHash
 	numShard  int
 	onNewNode []func(rdb *Client)
+
+	// ensures exclusive access to SetAddrs so there is no need
+	// to hold mu for the duration of potentially long shard creation
+	setAddrsMu sync.Mutex
 }
 
 type ringShards struct {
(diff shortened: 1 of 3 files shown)


AlexanderYastrebov:
FYI @szuecs

ring.go
```diff
@@ -213,6 +213,7 @@ func (shard *ringShard) Vote(up bool) bool {
 type ringSharding struct {
 	opt *RingOptions
 
+	addrsMu   sync.Mutex
```
vmihailenco, line 216: Please try to make it clear what this mutex protects, e.g. I would expect something like

```go
addrsMu   sync.Mutex
addrs        type

// another group
```

vmihailenco: After reading the code and the PR comment, I suggest to rename this to `setAddrsMu` and add an explanation similar to the one you have in the PR description.

vmihailenco:
>Introduces a new lock to make SetAddrs calls exclusive.
>This allows release of shards lock for the duration of potentially long newRingShards call.

Neat trick :+1:  but please try to add some comments to the code. Without your explanation it is hard to understand what is happening.

AlexanderYastrebov:
I've addressed comments and added shard cleanup tests.

ring.go
```diff
@@ -245,46 +249,62 @@ func (c *ringSharding) OnNewNode(fn func(rdb *Client)) {
...
+	existing := c.shards
+	c.mu.RUnlock()
+
+	shards, created, unused := c.newRingShards(addrs, existing)
 
-	shards, cleanup := c.newRingShards(addrs, c.shards)
+	c.mu.Lock()
+	if c.closed {
+		cleanup(created)
+		c.mu.Unlock()
```
vmihailenco, line 276: Just wonder if we can call `cleanup(created)` after the mutex is released?

vmihailenco:
Thanks

ring.go
AlexanderYastrebov: I thought about it but decided to keep like that - there would not be much difference as the ring is closed anyway at this point. I can move the unlock for aesthetics though.

PR #2296, opened 2022-11-25
fix(iterator): remove unnecessary mutex

v3s1e:
The mutex was removed in v9 in #2184 as the iterator cannot be used concurrently so I think we should also drop it in v8. Assuming v8 is still supported as it is mentioned in the installation section of the readme file


vmihailenco:
Makes sense, thanks!

PR #2316, opened 2022-12-16
test: `TestRingSetAddrsContention` changed to Benchmark

monkey92t:
`TestRingSetAddrsContention` caused a lot of test errors, it should be changed to Benchmark.

See #2292

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 *.rdb
-testdata/*/
+testdata/*
 .idea/
--- a/bench_test.go
+++ b/bench_test.go
@@ -365,3 +365,78 @@ func BenchmarkClusterSetString(b *testing.B) {
 		}
 	})
 }
+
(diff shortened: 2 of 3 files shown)


AlexanderYastrebov:
I did not implement it as a benchmark originally as go runs benchmark multiple times with increasing N (https://github.com/golang/go/issues/23398) and that was confusing.
Another benefit is that test ensures there is no regression so an alternative could be to lower the fail threshold below 10k to accommodate slugging build workers.

monkey92t:
> I did not implement it as a benchmark originally as go runs benchmark multiple times with increasing N ([golang/go#23398](https://github.com/golang/go/issues/23398)) and that was confusing. Another benefit is that test ensures there is no regression so an alternative could be to lower the fail threshold below 10k to accommodate slugging build workers.

The number of `pings` depends on the performance of the system (network, kernel, redis-server), which leads to multiple build failures, which can confuse other developers.

PR #2325, opened 2022-12-22
readXGroupInfo support entries-read and lag

lianghongle:


lianghongle:
readXGroupInfo support entries-read and lag

PR #2326, opened 2022-12-22
fix: readXGroupInfo support entries-read and lag

lianghongle:
fix: readXGroupInfo support entries-read and lag


lianghongle:
fix: readXGroupInfo support entries-read and lag

PR #2331, opened 2022-12-23
fix(redisotel): correct metrics.DialHook attrs

scott20315:
- pass attrs by value instead of by reference to the otel histogram `Record` function
   - the histograms in the otel sdk perform an in-place sort on the attrs passed into `Record`
   - a race condition can occur when separate goroutines concurrently interact with the various redisotel hooks
   - this PR converts the dialhook to a read operation which is consistent with the behavior of the other hooks
- side benefit of including the error status attr which is another step towards consistency with the other hooks


monkey92t:
I don't know about metrics...
It looks like it's just missing handling `err`.

monkey92t (approved):
seems to be ok.

PR #2334, opened 2022-12-28
test: upgrade redis version(7.0.7)

monkey92t:


Makefile
```diff
@@ -18,7 +18,7 @@ bench: testdeps
 
 testdata/redis:
 	mkdir -p $@
-	wget -qO- https://download.redis.io/releases/redis-7.0.6.tar.gz | tar xvz --strip-components=1 -C $@
+	wget -qO- https://download.redis.io/releases/redis-7.0.7.tar.gz | tar xvz --strip-components=1 -C $@
```
SoulPancake, line 21: Do we need to update the testdata version by this? 
For Redis 7.2 commands support? @monkey92t

monkey92t: Yes~

SoulPancake: So, for the tests to pass can I raise a PR that does this?

monkey92t: I've created a PR but something seems to have changed in 7.2, I'm looking at the redis-server code and need some time to fix it

#2532

SoulPancake: There are already issues for the two commands whose tests would fail with the new versions 
Can I raise PRs for those?

SoulPancake: Referring to #2526 and #2525 @monkey92t

PR #2345, opened 2023-01-07
feat: add zintercard cmd

monkey92t:


vmihailenco (approved):
Look good, thanks!

commands.go
```diff
@@ -2346,6 +2347,22 @@ func (c cmdable) ZInterWithScores(ctx context.Context, store *ZStore) *ZSliceCmd
 	return cmd
 }
 
+func (c cmdable) ZInterCard(ctx context.Context, limit int64, keys ...string) *IntCmd {
+	args := make([]interface{}, 4+len(keys))
```
chayim, line 2351: Even though we're passing a variadic string here for keys, @vmihailenco maybe we should check that the length is at least one? [ZINTERCARD](https://redis.io/commands/zintercard/) places a requirement on a minimum of one key. 
WDYT?

vmihailenco: We could, yes, and we probably should if Redis Server does not validate this by itself or this leads to a crash in go-redis client or something similarly bad.

But generally speaking we don't validate user input since there are lots of commands and the effort would be too big.

chayim: Yeah, makes sense - each client is different ;)

chayim (approved):

PR #2360, opened 2023-01-22
docs: updated "Installation" description in README.md

monkey92t:
After testing, in the case of removing the redis-7 api, v9 can pass the test of redis6-server.

We should allow users using redis6 to install go-redis/v9.

@vmihailenco  Do we have other situations where redis6 cannot be used in v9?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,17 +53,18 @@ module:
 go mod init github.com/my/repo
 ```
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
Looks good, I am using v9 with Redis 6 and it works fine :+1:

PR #2362, opened 2023-01-23
chore: update import path

vmihailenco:


README.md
```diff
@@ -16,17 +16,17 @@
...
 - [Chat](https://discord.gg/rWtp5Aj)
-- [Reference](https://pkg.go.dev/github.com/go-redis/redis/v9)
-- [Examples](https://pkg.go.dev/github.com/go-redis/redis/v9#pkg-examples)
+- [Reference](https://pkg.go.dev/github.com/redis/go-redis/v9)
+- [Examples](https://pkg.go.dev/github.com/redis/go-redis/v9#pkg-examples)
 
 ## Ecosystem
 
-- [Redis Mock](https://github.com/go-redis/redismock)
+- [Redis Mock](https://github.com/redis/go-redismock)
```
monkey92t, line 26: redismock should stay in go-redis, right?

chayim: We certainly think so.. though I'll bet we could help the community here too. I've at least added it to the plate to track (internally).

chayim (approved):

README.md
vmihailenco: Yeah, this should be reverted.

PR #2368, opened 2023-01-25
Adding the release drafter, for release note generation

chayim:


sazzad16 (approved):

vmihailenco:
Does it automatically updates the [changelog](https://github.com/redis/go-redis/blob/master/CHANGELOG.md)? Or it must be done manually?

vmihailenco:
Also having some docs for contributors would help. For example, it looks like if you would name this branch `maintenance-something`, it would get autolabeled with `maintenance` tag and release drafter will do a better job with the changelog generation...

chayim:
I want to inject (separate PR) changelog updating, as well. I view these all as separate PRs, since I like to slice.

After this merges, future releases (from mainlines) will have auto-generated release notes...

PR #2375, opened 2023-01-28
fix(conn): releaseConn should be executed correctly

monkey92t:
In any event...we should release Conn.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1287,9 +1287,13 @@ func (c *ClusterClient) processPipelineNode(
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


chayim (approved):

monkey92t:
When `processPipelineNodeConn`, `processTxPipelineNodeConn`, `fn` panic, we will lose a connection, if multiple panics occur, the connection pool will no longer work.

adriansmares:
Is it correct to release it with the `nil` error in case of a panic ? It would basically return it to the pool no matter the connection state - what if it was in the middle of a `WATCH` ?

PR #2379, opened 2023-01-30
 Adding error handling in SSubscribe for cluster connection s

### LINKED ISSUE #2303: SSubscribe doesn't return errors (including cross slot errors with multiple keys)
SSubscribe doesn't return errors including cross slot errors when multiple keys are specified that don't map to the same hash slot.

## Expected Behavior
Expected multiple return with error 
func (c *ClusterClient) SSubscribe(ctx context.Context, channels ...string) (*PubSub, error)

SoulPancake:
Adding simple error handling in the SSubscribe function in cluster.go

closes #2303

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1700,13 +1700,17 @@ func (c *ClusterClient) PSubscribe(ctx context.Context, channels ...string) *Pub
 }
 
 // SSubscribe Subscribes the client to the specified shard channels.
-func (c *ClusterClient) SSubscribe(ctx context.Context, channels ...string) *PubSub {
+func (c *ClusterClient) SSubscribe(ctx context.Context, channels ...string) (*PubSub, error) {
 	pubsub := c.pubSub()
 	if len(channels) > 0 {
-		_ = pubsub.SSubscribe(ctx, channels...)
+		err := pubsub.SSubscribe(ctx, channels...)
+		if err != nil {
+			return nil, err
+		}
 	}
-	return pubsub
+	return pubsub, nil
 }
+
 
 func (c *ClusterClient) retryBackoff(attempt int) time.Duration {
 	return internal.RetryBackoff(attempt, c.opt.MinRetryBackoff, c.opt.MaxRetryBackoff)
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -554,7 +554,7 @@ var _ = Describe("ClusterClient", func() {
 		})
 
 		It("supports sharded PubSub", func() {
-			pubsub := client.SSubscribe(ctx, "mychannel")
+			pubsub,_ := client.SSubscribe(ctx, "mychannel")
 			defer pubsub.Close()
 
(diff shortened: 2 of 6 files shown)


chayim (approved):

monkey92t (commented):
It breaks compatibility with v9 version, we already have v9.0.1

chayim:
> It breaks compatibility with v9 version, we already have v9.0.1

True - so it would belong in a v9.1. But, we should have error access in general, no?

monkey92t:
> > It breaks compatibility with v9 version, we already have v9.0.1
> 
> True - so it would belong in a v9.1. But, we should have error access in general, no?

No....See https://research.swtch.com/vgo-import

For v9 version, go mod import path: `github.com/redis/go-redis/v9`

old API: `SSubscribe(ctx context.Context, channels ...string) *PubSub`
new API: `SSubscribe(ctx context.Context, channels ...string) (*PubSub, error)`

However, `SSubscribe` is a redis7 command, and redis7 currently only has a small number of users.
We can also ignore it. 😁😁

chayim:
IMHO we shouldn't ignore - and we should generally have consistency. Breaking - sure.  While I agree with:

```
If an old package and a new package have the same import path,
the new package must be backwards compatible with the old package.
```

The reality is I wish all packages did this (and yes, we should strive to). But given the double return construct we all use of x, err it's worth talking about.

> However, SSubscribe is a redis7 command, and redis7 currently only has a small number of users.
We can also ignore it.

Did you mean ignore the breaking part of this change (i.e merge and who cares), or did you mean why bother?

vmihailenco:
`SSubscribe` was modeled after `Subscribe` which offers the same API (with lack of error handling). The idea is that you can ignore the error and PubSub will re-connect on errors without a fuss. So it is a feature, not a bug/overlook.

If you need to handle the error, I'd suggest to try something like `err := rdb.PubSub().SSubscribe(...)`. If that does not work, we should consider supporting it.

Alternatively, we could support additional methods like `SSubscribeError` and `SubscribeError`, but I am not big fan of it.

monkey92t:
> Did you mean ignore the breaking part of this change (i.e merge and who cares), or did you mean why bother?

My thoughts: If we have to change the API, do it as soon as possible to affect fewer users.

I ran the test:
```go
pubsub, err := client.SSubscribe(ctx, "c1", "c2", "c3", "c4")
fmt.Println("SSubscribe error", err)

v, err := pubsub.Receive(ctx)
fmt.Println("SSubscribe receive:", v, "err:", err)
```

got the answer: 
SSubscribe error <nil>
SSubscribe receive: <nil> err: CROSSSLOT Keys in request don't hash to the same slot.

we should close this issue.

monkey92t:
According to redis.doc: "https://redis.io/commands/ssubscribe/"

`ssubscribe` does not respond with any errors.

PR #2408, opened 2023-01-31
Adding support for ZMPOP command

SoulPancake:
Adding support for ZMPOP


monkey92t:
I don't know how the test passes, I run this PR test and I get the error:

```
  [FAILED] Unexpected error:
      <*errors.errorString | 0xc0008c5050>: {
          s: "redis: can't parse reply=\"*1\" reading string",
      }
      redis: can't parse reply="*1" reading string
  occurred
  In [It] at: /Users/monkey/Desktop/redis/commands_test.go:3592
```

SoulPancake:
Closing till I fix and debug the tests , will re-open once it is done

SoulPancake:
@monkey92t @chayim Should I rename the command to ZArrayWithKeyCmd ( matching the standard from 
ZWithKey ) ?

SoulPancake:
@monkey92t Can you please review this if possible

commands.go
```diff
@@ -2472,6 +2473,18 @@ func (c cmdable) ZInterCard(ctx context.Context, limit int64, keys ...string) *I
 	_ = c(ctx, cmd)
 	return cmd
 }
+func (c cmdable) ZMPop(ctx context.Context, order string, count int64, keys ...string) *ZArrayWithKeyCmd {
```
chayim, line 2476: I know we haven't done this previously.. but WDYT about starting to add help strings - and a link to the official command docs at https://redis.io/commands/zmpop. IMHO it helps people understand redis commands.

commands.go
```diff
@@ -2472,6 +2473,18 @@ func (c cmdable) ZInterCard(ctx context.Context, limit int64, keys ...string) *I
...
+	args := make([]interface{}, 2+len(keys), 5+len(keys))
+	args[0] = "zmpop"
+	args[1] = len(keys)
+	for i, key := range keys {
```
chayim, line 2480: good

chayim:
@SoulPancake overall it looks good. My comment was literally only a nit. CI failures don't appear related.

@monkey92t or @vladvildanov you have more experience here. Does this frequently happen with CI in this project?

chayim (approved):

monkey92t:
For the name `ZArrayWithKeyCmd`, in the go language, `slice` is generally used, we can change it to `ZSliceWithKeyCmd`

monkey92t:
For commands with complex parameters, we should add some comments and usage examples.

commands.go
```diff
@@ -2472,6 +2473,18 @@ func (c cmdable) ZInterCard(ctx context.Context, limit int64, keys ...string) *I
...
+		args[2+i] = key
+	}
+	args = append(args, strings.ToLower(order), "count", count)
```
vladvildanov, line 2483: @chayim 

What is the policy about wrong `order` provided? Should we check against correct values (e.g `min`, `max` ) or just rely on redis server error response and assume that user familiar with documentation?

chayim: Clients generally stick to order provided by the redis commands.

commands.go
SoulPancake: Sure will do that

commands.go
SoulPancake: So is there something I should do separately about this part? @chayim @vladvildanov

vladvildanov: @SoulPancake @chayim Maybe it was a little misunderstanding, I'm just wondering should we check value of `string order` argument on the client side or it's okay if server exception will be thrown in case if user provide wrong value? Since, `string order` is not an enum (google says Go doesn't support enums), so we never know which value user will provide

commands.go
SoulPancake: @chayim Can you quote an example for what we should put in the help strings?

monkey92t (approved):

commands.go
monkey92t: Yes, we really should add a comment to help users with this command.

SoulPancake: Added help strings for this commands usage.  Can someone please review the same @monkey92t @chayim

monkey92t (approved):

SoulPancake:
Thanks everyone for their valuable comments and suggestions, Learning a lot @chayim @monkey92t @vladvildanov

PR #2413, opened 2023-02-01
feat(scan): scan time.Time sets the default decoding

monkey92t:
1. scan time.Time, we should set the default decoding format, because we encoded time.Time in RFC3339Nano format in proto.WriteArg.
```go
case time.Time:
		w.numBuf = v.AppendFormat(w.numBuf[:0], time.RFC3339Nano)
		return w.bytes(w.numBuf)
```
We should achieve the consistency of Write(Hset Scan) and Read(HMGet Scan):

```go
type data struct {
	Key1 string    `redis:"key1"`
	Key2 int       `redis:"key2"`
	Time time.Time `redis:"time"`
}

// write set
client.HSet(ctx, "hash", &data{
	Key1: "hello",
	Key2: 100,
	Time: time.Now(),
})

// read get
client.HMGet(ctx, "hash", "key1", "key2", "time").Scan(&data)
```

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

2. `switch case encoding.BinaryMarshaler` is invalid

### DIFF
--- a/internal/hscan/structmap.go
+++ b/internal/hscan/structmap.go
@@ -101,10 +101,10 @@ func (s StructValue) Scan(key string, value string) error {
 
 	if isPtr && v.Type().NumMethod() > 0 && v.CanInterface() {
 		switch scan := v.Interface().(type) {
-		case encoding.TextUnmarshaler:
-			return scan.UnmarshalText(util.StringToBytes(value))
 		case Scanner:
 			return scan.ScanRedis(value)
+		case encoding.TextUnmarshaler:
+			return scan.UnmarshalText(util.StringToBytes(value))
 		}
 	}
 
--- a/internal/hscan/structmap.go
+++ b/internal/hscan/structmap.go
@@ -1,10 +1,13 @@
 package hscan
 
 import (
+	"encoding"
 	"fmt"
 	"reflect"
 	"strings"
 	"sync"
+
+	"github.com/redis/go-redis/v9/internal/util"
 )
 
 // structMap contains the map of struct fields for target structs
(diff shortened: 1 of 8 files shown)


vmihailenco:
>switch case encoding.BinaryMarshaler is invalid

Could you elaborate? I don't see why it should not work...

leoantony72 (dismissed):

monkey92t:
> > switch case encoding.BinaryMarshaler is invalid
> 
> Could you elaborate? I don't see why it should not work...

Oh..
it my test error, my mistake, I recover it.

leoantony72:
hgetall with Scan command doesn't work if there is time.Time field in the struct and it leads to erratic behaviour like not retrieving every field from the hashset

monkey92t:
We can directly assert that the `encoding.TextUnmarshaler` interface, `time.Time` implements the `encoding.TextUnmarshaler` interface, which is more convenient. Consider removing `Scanner` in the future.

leoantony72:
> 

That would remove the use of Scan. Which is more reliant

internal/hscan/structmap.go
```diff
@@ -97,7 +100,10 @@ func (s StructValue) Scan(key string, value string) error {
 	}
 
 	if isPtr && v.Type().NumMethod() > 0 && v.CanInterface() {
-		if scan, ok := v.Interface().(Scanner); ok {
+		switch scan := v.Interface().(type) {
+		case encoding.TextUnmarshaler:
+			return scan.UnmarshalText(util.StringToBytes(value))
+		case Scanner:
```
vmihailenco, line 106: Perhaps swap Scanner and TextUnmarshaler? Scanner looks more specific.

vmihailenco (dismissed):

internal/hscan/structmap.go
leoantony72: ah

chayim (approved):

ryan961:
> We can directly assert that the `encoding.TextUnmarshaler` interface, `time.Time` implements the `encoding.TextUnmarshaler` interface, which is more convenient. Consider removing `Scanner` in the future.

hset struct field can assert the `encoding.TextMarshaler` interface ?
https://github.com/redis/go-redis/issues/2161#issuecomment-1471422636

PR #2416, opened 2023-02-03
chore: add error types that should be retried

yikuaibro:


internal/log.go
```diff
@@ -19,6 +22,20 @@ func (l *logger) Printf(ctx context.Context, format string, v ...interface{}) {
...
+
+func (l *logger) Fatalf(ctx context.Context, format string, v ...interface{}) {
+	_ = l.log.Output(2, fmt.Sprintf(format, v...))
+	os.Exit(1)
+}
+
+func (l *logger) Fatalln(ctx context.Context, v ...interface{}) {
+	_ = l.log.Output(2, fmt.Sprintln(v...))
+	os.Exit(1)
+}
```
vmihailenco, line 37: Changes to the logging are unrelated and should be in a separate PR.

error.go
```diff
@@ -55,6 +55,15 @@ func shouldRetry(err error, retryTimeout bool) bool {
...
 	}
+	if s == "Connection reset by peer" {
+		return true
+	}
+	if strings.HasPrefix(s, "ERR redis temporary failure") {
+		return true
+	}
+	if strings.HasPrefix(s, "FLUSHDB ") {
+		return true
+	}
```
vmihailenco, line 66: Please add a brief comment / reasoning why these errors must be retried.

yikuaibro: ok, I will add a comment / reasoning, thank you.

internal/log.go
yikuaibro: Understand.

PR #2418, opened 2023-02-05
chore: add some error types that should be retried or a bad connection

yikuaibro:
1. Connection reset by peer
Client connection is closed, maybe I should treat it as a bad connection.
2. ERR redis temporary failure
Access timeout, Should retry
3. FLUSHDB
When adding or subtracting data nodes in a cluster, the FLUSHDB command is disabled and returns the error "ERR FLUSHDB is not allowed in migrating mode", which I think is also required as a retry case (although it is relatively rare).


yikuaibro:
Is something wrong,？Maybe it should not be added in the default error handler?

monkey92t:
> Is something wrong,？Maybe it should not be added in the default error handler?

I have explained:

1. `ERR redis temporary failure` is not an error responded by redis-server, at least I did not find it in the redis documentation and source code.
2. please see the code:
```go
        if isRedisError(err) {
		switch {
		case isReadOnlyError(err):
			// Close connections in read only state in case domain addr is used
			// and domain resolves to a different Redis Server. See #790.
			return true
		case isMovedSameConnAddr(err, addr):
			// Close connections when we are asked to move to the same addr
			// of the connection. Force a DNS resolution when all connections
			// of the pool are recycled
			return true
		case isResetError(err):
			// The client connection is closed, usually due to a client buffer exception.
			return true
		default:
			return false
		}
	}

        func isResetError(err error) bool {
	        return err.Error() == "Connection reset by peer"
        }
```
In your PR, `Connection reset by peer` is not a redis error, it is false in `if isRedisError(err)`.

monkey92t:
@yikuaibro 

You need to distinguish what is `redis error` from `tcp error`.

`redis error` is the error message responded by redis-server, for example: `(error) ERR value is not an integer or out of range`, it is a logical error.

And `tcp error` is an error caused by accidents such as the network when we write/read data, and it is a kind of system error.

hope it helps you.

yikuaibro:
> @yikuaibro
> 
> You need to distinguish what is `redis error`, what is `tcp error`.
> 
> `redis error` is the error message responded by redis-server, for example: `(error) ERR value is not an integer or out of range`, it is a logical error.
> 
> And `tcp error` is an error caused by accidents such as the network when we write/read data, and it is a kind of system error.
> 
> hope it helps you.

Thank you very much for your suggestion, I understand, if there is any problem in the next use, I will submit issuse.

PR #2423, opened 2023-02-06
test: update go mod before testing(go mod tidy)

monkey92t:


vmihailenco:
Sometimes updating the dependencies will break the build which can be annoying if all you want is to run the tests...

monkey92t:
> Sometimes updating the dependencies will break the build which can be annoying if all you want is to run the tests...

#2420 #2421

monkey92t:
@vmihailenco We need to update go.sum in order to run the tests.

vmihailenco:
@monkey92t alright, I did not realize that. Could you change `make go_mod_tidy` to not run `go get -u` and make a separate command like `make update_deps` that updates dependencies? Then we can merge this.

monkey92t:
It seems that we only need to run `go mod tidy` when `make test`.

vmihailenco (approved):

Issue #2406, opened 2023-01-31
Add support for PEXPIRETIME command


chayim:
Redis PEXPIRETIME command was added in version 7.0.0 and is currently supported by the library. We should add support as documented [here](https://redis.io/commands/PEXPIRETIME).

monkey92t:
fixed #2426

PR #2426, opened 2023-02-07
Expiretime and PExpireTime

### LINKED ISSUE #2391: Add support for EXPIRETIME command
Redis EXPIRETIME command was added in version 7.0.0 and is currently supported by the library. We should add support as documented [here](https://redis.io/commands/EXPIRETIME).

elena-kolevska:
Fixes #2391  and #2406


monkey92t (approved):
thanks..

PR #2431, opened 2023-02-08
hotfix/v8_conn_pool_check_fd

hardeyhe:
hello, our team hopes to fix this wrong connection problem in v8 and add a tag on this basis
reference mr: https://github.com/redis/go-redis/pull/1824/

### DIFF
--- a/internal/pool/conncheck.go
+++ b/internal/pool/conncheck.go
@@ -1,0 +1,46 @@
+//go:build linux || darwin || dragonfly || freebsd || netbsd || openbsd || solaris || illumos
+// +build linux darwin dragonfly freebsd netbsd openbsd solaris illumos
+
+package pool
+
+import (
+	"errors"
+	"io"
+	"net"
+	"syscall"
+)
+
+var errUnexpectedRead = errors.New("unexpected read from socket")
+
+func connCheck(conn net.Conn) error {
+	sysConn, ok := conn.(syscall.Conn)
+	if !ok {
+		return nil
+	}
+	rawConn, err := sysConn.SyscallConn()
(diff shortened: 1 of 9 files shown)


monkey92t:
Oh, we neglected to port the PR to v8, thanks for the addition.
I did a test at the time, `connCheck` will cause a 20% performance loss. If you have any questions, don't hesitate to ask!

vmihailenco:
Looks good. Don't forget to remove the Draft label once this is ready.

hardeyhe:
> Oh, we neglected to port the PR to v8, thanks for the addition. I did a test at the time, `connCheck` will cause a 20% performance loss. If you have any questions, don't hesitate to ask!

That's really bad, so don't you plan to fix this problem in v8?

hardeyhe:
> > Oh, we neglected to port the PR to v8, thanks for the addition. I did a test at the time, `connCheck` will cause a 20% performance loss. If you have any questions, don't hesitate to ask!
> 
> That's really bad, so don't you plan to fix this problem in v8?

Can I add an option to ConnPool(like `strictConnCheck bool`)? which is disabled by default but if enabled, it will use connCheck in isStaleConn() method. This way, neither the default behavior will be changed nor the performance loss introduced, and the excellent features of v9 can be introduced, leaving the choice to the user. Thanks reply.

monkey92t:
NO, although we lost 20% of performance, `go-redis` still maintains a high performance, which is not much different from `redis-benchmark`

20% performance reduction, in exchange for a stable and effective network connection for us, it is worth it.

monkey92t (approved):
Looks very good 👍🏻

Issue #2276, opened 2022-11-05
Blocking XGroupCreateMkStream does not interrupt on context cancellation


jgirtakovskis:
When XGroupCreateMkStream is called in blocking mode (Block = 0), call does not get interrupted by cancelling context.

## Expected Behavior

Blocking function interrupts when context is cancelled

## Current Behavior

Function continues to block after context cancellation

## Possible Solution

Unsure yet

## Steps to Reproduce

```
package main

import (
	"context"
	"fmt"
	"sync"
	"time"

	"github.com/go-redis/redis/v9"
	"github.com/google/uuid"
)

func main() {
	rdb := redis.NewUniversalClient(&redis.UniversalOptions{
		Addrs:    []string{"localhost:6379"},
		Password: "", // no password set
		DB:       0,  // use default DB
	})

	defer rdb.Close()

	ctx, cancelFn := context.WithCancel(context.Background())

	go func() {
		for idx := 0; idx < 5; idx++ {
			fmt.Printf("Waiting %v...\n", idx)
			time.Sleep(time.Second)
		}
		cancelFn()
		fmt.Printf("Cancelled context and now expect blocking XGroupCreateMkStream to be interrupted...\n")
	}()

	name := "blag"
	streamName := name
	groupName := name + "-blah"

	_, err := rdb.XGroupCreateMkStream(ctx, streamName, groupName, "0").Result()
	fmt.Printf("%v\n", err)

	var wg sync.WaitGroup

	wg.Add(1)
	go func() {
		defer wg.Done()
		objs, err := rdb.XReadGroup(ctx, &redis.XReadGroupArgs{
			Group:    groupName,
			Consumer: uuid.NewString(),
			Streams:  []string{streamName, ">"},
			Count:    100,
			Block:    0,
		}).Result()
		fmt.Printf("%v, %v\n", err, objs)
	}()

	wg.Wait()
	fmt.Printf("Done.\n")
}

```

## Context (Environment)

I have two goroutines concurrently performing XREADGROUP and XADD in blocking mode. XADD is triggered by external events and is not guaranteed to add items to the stream at any particular cadence or pattern. Shutting down reading goroutine is not possible due to the blocking call that does not get interrupted by context concellation.

## Detailed Description

Blocking calls should interrupt when context is cancelled and connection closed.

## Possible Implementation

N/A

berndverst:
Dapr maintainer here. I also need this fixed.

Alger7w:
[...]

PR #2432, opened 2023-02-09
test: Add tests for cancelled context

armsnyder:
I added tests for the bug reported in #2276. They are all failing on `Receive()` as expected, so I disabled them until the issue is fixed.


ndyakov:
@armsnyder doesn't look like the contributor of the fix is responding. We closed the PR you mentioned, so we are gonna close this one for now as well. From our initial research, turns out the context cancelation for such cases is not supported at the moment. Feel free to reopen or actually open one with a proposed fix for the issue. Thank you once again!

PR #2433, opened 2023-02-10
fix: Blocking commands respect canceled context

### LINKED ISSUE #2276: Blocking XGroupCreateMkStream does not interrupt on context cancellation

armsnyder:
## Description

This fixes an issue where blocking commands like `XRead` and `XReadGroup` could not be canceled using the provided context. Similarly with `*PubSub.Receive`.

Fixes #2276
Supersedes #2432 

## Approach

I looked holistically at places in the code where a connection was being used without inspecting the context and applied the same basic pattern. Once a timeout is reached, the recommendation is to close the connection. Note that the change to `*PubSub.ReceiveTimeout` does not explicitly close the connection since that is already handled by the subsequent `*PubSub.releaseConnWithLock`.

Note that with this change, the [`ContextTimeoutEnabled`](https://github.com/redis/go-redis/blob/bdfb4a75ad0de4222af0c16cf4f9cd212822228b/options.go#L91) client option is redundant. Can we deprecate it?

## Testing

The new tests that I added all failed without this change. Note that I did not add new tests for the `processPipeline` and `processTxPipeline` changes, since pipelining blocking commands does not make sense. However, if you can think of a good test case I will be happy to add it.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1291,7 +1291,19 @@ func (c *ClusterClient) processPipelineNode(
 		defer func() {
 			node.Client.releaseConn(ctx, cn, processErr)
 		}()
-		processErr = c.processPipelineNodeConn(ctx, node, cn, cmds, failedCmds)
+
+		errCh := make(chan error, 1)
+
+		go func() {
+			errCh <- c.processPipelineNodeConn(ctx, node, cn, cmds, failedCmds)
+		}()
+
+		select {
+		case processErr = <-errCh:
+		case <-ctx.Done():
+			_ = cn.Close()
+			processErr = ctx.Err()
+		}
 
 		return processErr
 	})
@@ -1472,7 +1484,19 @@ func (c *ClusterClient) processTxPipelineNode(
 		defer func() {
 			node.Client.releaseConn(ctx, cn, processErr)
 		}()
-		processErr = c.processTxPipelineNodeConn(ctx, node, cn, cmds, failedCmds)
+
+		errCh := make(chan error, 1)
+
+		go func() {
+			errCh <- c.processTxPipelineNodeConn(ctx, node, cn, cmds, failedCmds)
+		}()
+
+		select {
+		case processErr = <-errCh:
+		case <-ctx.Done():
+			_ = cn.Close()
+			processErr = ctx.Err()
+		}
 
 		return processErr
 	})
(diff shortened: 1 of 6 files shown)


chayim:
@monkey92t looks reasonable WDYT?

wk8:
Just hit this, could this please get reviewed & merged? Looks reasonable no?

ndyakov:
@armsnyder feel free to sync with master and open a new PR and we will review it. Thank you.

PR #2441, opened 2023-02-13
fix: limit the number of connections created

### LINKED ISSUE #2439: connection pool size
Hi
I am using redis.Client in my project with this configs. 
poolsize: 10
minIdleConn: 5
maxRetries: 5

the used command is LPUSH for a list of 2000 elements. 

	var dialCount uint64
		Addr:       "localhost:6379",
		DB:         0,
		MaxRetries: 5,
			fmt.Printf("%d - dialer called for new connection\n", atomic.AddUint64(&dialCount, 1))
			return net.Dial(network, addr)
		},
		PoolSize:     50,
		MinIdleConns: 10,
	})

	ctx := context.Background()
	var items []int
	for i := 1; i < 2000; i++ {
		items = append(items, i)
	}
	pipe := client.Pipeline()
	defer func(pipe redis.Pipeliner) {
		err := pipe.Close()
		if err != nil {
			fmt.Println("error in closing pipe")
		}
	}(pipe)

	key := "test_key"
	for _, item := range items {
		cmd := pipe.LPush(ctx, key, item)
		if cmd.Err() != nil {
			fmt.Printf("error in list push is: %v\n", cmd.Err())
			return
		}
	}
	_, err := pipe.Expire(ctx, key, time.Duration(72)*time.Hour).Result()
	if err != nil {
		fmt.Printf("error in expiring pipeline is %v\n", err)
		return
	}
	_, err = pipe.Exec(ctx)
	if err != nil {
		fmt.Printf("error in executing pipe is %v\n", err)
		return
	}

	signal := make(chan bool)
	wg := sync.WaitGroup{}

	for i := 0; i < 20000; i++ {
		wg.Add(1)
		go func(ctx context.Context, wg *sync.WaitGroup, n int) {
			defer wg.Done()
			<-signal
			cmd := client.RPop(ctx, key)

			if cmd.Err() != nil {
				if cmd.Err().Error() != "redis: connection pool timeout" && cmd.Err() != redis.Nil {
					fmt.Printf("%d - error: %v\n", n, cmd.Err())
				}
			} else {
				fmt.Printf("%d - value: %v\n", n, cmd.Val())
			}
		}(ctx, &wg, i)
	}
	close(signal)
	wg.Wait()

## Expected Behavior

the dial function should be called for 50 times, which is the size of poolSize.
## Current Behavior

I see more than 50 connections opens. why client opens connections more than poolSize?

monkey92t:
fix #2439

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -112,18 +112,25 @@ func (p *ConnPool) checkMinIdleConns() {
 		return
 	}
 	for p.poolSize < p.cfg.PoolSize && p.idleConnsLen < p.cfg.MinIdleConns {
-		p.poolSize++
-		p.idleConnsLen++
+		select {
+		case p.queue <- struct{}{}:
+			p.poolSize++
+			p.idleConnsLen++
 
-		go func() {
-			err := p.addIdleConn()
-			if err != nil && err != ErrClosed {
-				p.connsMu.Lock()
-				p.poolSize--
-				p.idleConnsLen--
-				p.connsMu.Unlock()
-			}
-		}()
+			go func() {
+				err := p.addIdleConn()
+				if err != nil && err != ErrClosed {
+					p.connsMu.Lock()
+					p.poolSize--
+					p.idleConnsLen--
+					p.connsMu.Unlock()
+				}
+
+				p.freeTurn()
+			}()
+		default:
+			return
+		}
 	}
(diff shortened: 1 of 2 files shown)


chayim (approved):

vmihailenco (approved):
Looks promising! Let's try it.

PR #2440, opened 2023-02-13
Add support for LMPOP

SoulPancake:
Adding command and command test for LMPOP command.

PR #2443, opened 2023-02-14
refactor: change ListElementCmd to KeyValuesCmd.

monkey92t:
change ListElementCmd to KeyValuesCmd. #2440 

Using `KeyValuesCmd` seems more reasonable.


vmihailenco:
I like the new `KeyValuesCmd` name, but I think the old API was cleaner:

```go
// This is better
key, elems, err := client.LMPop(ctx, "left", 3, "list1", "list2").Result()

// This is worse
keyVals, err := client.LMPop(ctx, "left", 3, "list1", "list2").Result()
```

Could we keep the old API or I am missing something?

monkey92t:
> I like the new `KeyValuesCmd` name, but I think the old API was cleaner:
> 
> ```go
> // This is better
> key, elems, err := client.LMPop(ctx, "left", 3, "list1", "list2").Result()
> 
> // This is worse
> keyVals, err := client.LMPop(ctx, "left", 3, "list1", "list2").Result()
> ```
> 
> Could we keep the old API or I am missing something?

OK, we keep the `KeyValuesCmd` name, but the API uses:

```go
key, elems, err := client.LMPop()
```

vmihailenco:
>OK, we keep the KeyValuesCmd name, but the API uses:

Sounds good. I think we already have APIs like this.

vmihailenco (approved):

PR #2444, opened 2023-02-14
Add option to customize span name

RangelReale:
This PR adds an option to customize span name.

```go
	hook := newTracingHook(
		"",
		WithTracerProvider(provider),
		WithSpanName(func(hook TracingHook, defaultName string) string {
			if hook == TracingHookProcess {
				return "redis." + defaultName
			}
			return defaultName
		}),
	)
```

### DIFF
--- a/extra/redisotel/config.go
+++ b/extra/redisotel/config.go
@@ -12,8 +12,9 @@ import (
 type config struct {
 	// Common options.
 
-	dbSystem string
-	attrs    []attribute.KeyValue
+	dbSystem   string
+	attrs      []attribute.KeyValue
+	spanNameFn func(hook TracingHook, defaultName string) string
 
 	// Tracing options.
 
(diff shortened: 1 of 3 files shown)


chayim:
@RangelReale What's your use case?

RangelReale:
> @RangelReale What's your use case?

It's mostly because in Datadog when you are looking at the latest traces, the only thing shown is the span name. So I see "GET <url>" for http, "db.query", "db.tx" for sql, but for redis I see just "get", "put", I wanted to at least show "redis.get" and "redis.put", to at least know what it is for.

ndyakov:
Closing for inactivity, feel free to reopen when synced with master.

RangelReale:
Synced with master.

RangelReale:
Can't see to find a way to reopen.

PR #2448, opened 2023-02-17
fix: group lag can be null

holykol:
*fixes 3+ hours of me trying to understand why things don't work* 

see https://github.com/redis/redis/blob/383d902ce68131cf40d6122ce09e305e672e8555/src/t_stream.c#L1456

change the behavior so that if lag is null, the group lag will be set to 0 instead of returning an error

### DIFF
--- a/command.go
+++ b/command.go
@@ -2057,7 +2057,7 @@ func (cmd *XInfoGroupsCmd) readReply(rd *proto.Reader) error {
 				}
 			case "lag":
 				group.Lag, err = rd.ReadInt()
-				if err != nil {
+				if err != nil && err != Nil {
 					return err
 				}
 			default:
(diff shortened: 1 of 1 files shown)


monkey92t (approved):
Good, thanks!

monkey92t:
This is a new field in redis7, so we don't need to port to v8.

command.go
```diff
@@ -2057,7 +2057,7 @@ func (cmd *XInfoGroupsCmd) readReply(rd *proto.Reader) error {
 				}
 			case "lag":
 				group.Lag, err = rd.ReadInt()
-				if err != nil {
+				if err != nil && err != Nil {
```
vmihailenco, line 2060: Having a comment here would help next time someone changes this code.

PR #2458, opened 2023-02-23
Fix for internal/hscan/structmap.go:89:23: undefined: reflect.Pointer

eepurichandra:


monkey92t (approved):
OK, we accept this PR, but it is recommended to upgrade go1.18 above versions.

Shirtiny:
> 

I really want to upgrade to Go 1.18, but some cloud platforms only support Go 1.7, and it makes me cry.

monkey92t:
> > 
> 
> I really want to upgrade to Go 1.18, but some cloud platforms only support Go 1.7, and it makes me cry.

wt?  1.7?

Shirtiny:
> > > 
> > 
> > 
> > I really want to upgrade to Go 1.18, but some cloud platforms only support Go 1.7, and it makes me cry.
> 
> wt? 1.7?

1.17

monkey92t:
We only support the 2 latest go versions in most cases, which is also the practice of many go libraries.

Issue #2453, opened 2023-02-21
Race condition in AddHook


duongcongtoai:
I suspect current AddHook function will cause race condition with the background connection pool dialer
```
WARNING: DATA RACE
Write at 0x00c000150278 by goroutine 84:
  github.com/redis/go-redis/v9.(*hooksMixin).chain()
      /go/pkg/mod/github.com/redis/go-redis/v9@v9.0.2/redis.go:120 +0x8c
  github.com/redis/go-redis/v9.(*hooksMixin).AddHook()
      /go/pkg/mod/github.com/redis/go-redis/v9@v9.0.2/redis.go:114 +0x527
  github.com/redis/go-redis/extra/redisotel/v9.InstrumentTracing()
      /go/pkg/mod/github.com/redis/go-redis/extra/redisotel/v9@v9.0.2/tracing.go:28 +0x3e6
...
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1446 +0x216
  testing.(*T).Run.func1()
      /usr/local/go/src/testing/testing.go:1493 +0x47

Previous read at 0x00c000150278 by goroutine 85:
  github.com/redis/go-redis/v9.(*hooksMixin).dialHook()
      /go/pkg/mod/github.com/redis/go-redis/v9@v9.0.2/redis.go:169 +0x64
  github.com/redis/go-redis/v9.(*hooksMixin).dialHook-fm()
      <autogenerated>:1 +0x99
  github.com/redis/go-redis/v9.newConnPool.func1()
      /go/pkg/mod/github.com/redis/go-redis/v9@v9.0.2/options.go:489 +0xa4
  github.com/redis/go-redis/v9/internal/pool.(*ConnPool).dialConn()
      /go/pkg/mod/github.com/redis/go-redis/v9@v9.0.2/internal/pool/pool.go:194 +0x186
  github.com/redis/go-redis/v9/internal/pool.(*ConnPool).addIdleConn()
      /go/pkg/mod/github.com/redis/go-redis/v9@v9.0.2/internal/pool/pool.go:134 +0x6f
  github.com/redis/go-redis/v9/internal/pool.(*ConnPool).checkMinIdleConns.func1()
      /go/pkg/mod/github.com/redis/go-redis/v9@v9.0.2/internal/pool/pool.go:122 +0x34
```

## Possible Solution
Allow options struct to receive a hook on init, instead of letting AddHook function modify the hooks after the pool has been created and running

## Steps to Reproduce

[...]

PR #2460, opened 2023-02-26
feat: add lock for hooks.dial

monkey92t:
See #2453 

We added a lock to `DialHook`, but `ProcessHook`, `ProcessPipelineHook` are still not thread-safe.

If the `Options.MinIdleConns` parameter is set, go-redis will create a network connection asynchronously, and there may be a data race (hooks.dial) with the `AddHook()` API, and even a part of the network connection is created before the `AddHook` operation.

We only need to pay for the overhead of the read lock for the DialHook.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -838,7 +838,7 @@ type ClusterClient struct {
 	state         *clusterStateHolder
 	cmdsInfoCache *cmdsInfoCache
 	cmdable
-	hooksMixin
+	*hooksMixin
 }
 
 // NewClusterClient returns a Redis Cluster client as described in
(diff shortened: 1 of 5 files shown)


vmihailenco:
I've tried to review you this, but the existing code looks confusing enough and locking does not make it better :( I will try to get back to this later and make API cleaner and more understandable...

monkey92t:
> I've tried to review you this, but the existing code looks confusing enough and locking does not make it better :( I will try to get back to this later and make API cleaner and more understandable...

Yes...Indeed, we introduced more complex code to solve a problem, and in my view, DialHook and MinIdleConns are conflicting.

zcong1993:
Any updates?

bramsvs:
@vmihailenco any updates on this? :)

duongcongtoai:
ping

ndyakov:
@monkey92t would you like to revisit this after https://github.com/redis/go-redis/pull/3320 ?

ndyakov:
closing as stale

PR #2463, opened 2023-02-27
test: remove testify

monkey92t:


cluster_test.go
```diff
@@ -1454,47 +1451,36 @@ func TestParseClusterURL(t *testing.T) {
...
-				t.Fatalf("expected error: got %+v", actual)
-				return
-			}
-			if tc.err != nil && err != nil {
-				if tc.err.Error() != err.Error() {
-					t.Fatalf("got %q, expected %q", err, tc.err)
-				}
-				return
+			if tc.err != nil {
+				Expect(err).To(Equal(err))
```
armsnyder, line 1459: Minor suggestion, I usually use `MatchError` when matching errors, since it handles error unwrapping.

```suggestion
				Expect(err).To(MatchError(err))
```

monkey92t: ok thanks for your suggestion.

vmihailenco (approved):

PR #2466, opened 2023-02-28
Allowing for running tests on a port other than the fixed 6380

chayim:
In testing various versions of redis, it is sometimes necessary to go back and forth between multiple running redis instances. This change makes it possible to specify the instance, purely by the existing port - which includes a README directive.


chayim:
Interesting - the github cli pushed this here, when it explicitly said it would push to the fork :shrug: 

I'll double check my settings along the way.

monkey92t:
I developed a tool: https://github.com/monkey92t/grte
It's weak, but helps developers quickly test go-redis.

The dependent redis-server has been installed in docker-image,
Quick run test😁: 

exp:
```go
monkey@MonkeyiMac redis % grte go test ./...
[go-redis]  🐳 Prepare docker image ==> goredis/grte:v9
[go-redis]  🐳 Use image cache, ID ==> sha256:f4a6149b55e000ed40b5a68041615e7223b52119178d6ec4190f4714b8b88a49
[go-redis]  🐳 Create docker container...
[go-redis]  🐳 ContainerID: c788618272cefc280a86134d2c5315fdd2d9d48e7bf30b948e3efc6ca2076a2e
[go-redis]  🐳 WorkDir: /Users/monkey/Desktop/redis
[go-redis]  🐳 Command: [go test ./...]
ok  	github.com/redis/go-redis/v9	108.379s
ok  	github.com/redis/go-redis/v9/internal	0.016s
ok  	github.com/redis/go-redis/v9/internal/hashtag	0.024s
ok  	github.com/redis/go-redis/v9/internal/hscan	0.022s
ok  	github.com/redis/go-redis/v9/internal/pool	1.568s
ok  	github.com/redis/go-redis/v9/internal/proto	0.010s
?   	github.com/redis/go-redis/v9/internal/rand	[no test files]
?   	github.com/redis/go-redis/v9/internal/util	[no test files]
[go-redis]  ✅  Success!
monkey@MonkeyiMac redis % 
```

monkey92t:
I don't think this is a good way, go-redis test is more complicated, we need to start redis-server, sentinel, cluster, ring server.

chayim:
Depends on your need. I need to test umpteen redises - and dockers don't solve my problems. When I need to go through tonnes and things, and dockers we rely on my [redisenv](https://github.com/RedisLabsModules/redisenv). It'll probably support direct binary eventually.

chayim:
> I don't think this is a good way, go-redis test is more complicated, we need to start redis-server, sentinel, cluster, ring server.

Depends what you're testing. You can have a variety of inputs. This allows me to move forward with multis - and run the others after the fact. For my use case - very useful.  Again, nothing changes for the user here.

monkey92t:
Go-redis needs to stop and start redis-server during test execution, for example:

```go
err = master.Shutdown(ctx).Err()
....
_, err = startRedis(masterPort)
```

If redis-server is running in other containers, it is difficult for us to start or stop.

vmihailenco (approved):

vmihailenco:
This overrides only a single port out of ~ 10, but no objections to have it :+1:

monkey92t (approved):
OK, we can do that for now, but we should find an easier way to test. 😂

PR #2470, opened 2023-03-02
fix: update go.mod module

yuwenlong:
update go.mod module


yuwenlong:
@monkey92t v9 branch update go.mod package name.

monkey92t:
What problem is this PR trying to solve?

vmihailenco:
I've deleted v9 branch since it was merged into master and is not updated any more.

PR #2475, opened 2023-03-06
Implement `FUNCTION` group of commands

elena-kolevska:
Implemented load, delete and flush. Working on `function list`


elena-kolevska:
Solves https://github.com/redis/go-redis/issues/2400, https://github.com/redis/go-redis/issues/2397 and https://github.com/redis/go-redis/issues/2395

monkey92t (commented):
Good, please continue...

command.go
```diff
@@ -3840,3 +3840,167 @@ func (cmd *ZSliceWithKeyCmd) readReply(rd *proto.Reader) (err error) {
...
+		case "name":
+			function.Name, err = rd.ReadString()
+			if err != nil {
+				return RdsFunction{}, err
+			}
+		case "description":
+			function.Description, err = rd.ReadString()
+			if err != nil {
+				if err == proto.Nil {
+					function.Description = "" // TODO Talk to Chayim
```
elena-kolevska, line 3981: @monkey92t are we ok setting the function description to an empty string here? We're getting back a `redis: nil` but it didn't feel like we would need some kind of a nil object here, since it's just one property of a bigger object.

elena-kolevska:
@vmihailenco , @monkey92t, if the approach I had with `FUNCTION LIST` is ok with you, I'll continue implementing the `WITHCODE` variant of `FUNCTION LIST` and all the rest of the `FUNCTIONS` group.

commands_test.go
```diff
@@ -6086,6 +6086,159 @@ var _ = Describe("Commands", func() {
...
+
+		It("Flushes all libraries asynchronously", func() {
+			FunctionLoad := client.FunctionLoad(ctx, lib1Code)
+			Expect(FunctionLoad.Err()).NotTo(HaveOccurred())
+
+			functionFlush := client.FunctionFlushAsync(ctx)
+			Expect(functionFlush.Err()).NotTo(HaveOccurred())
+
+			args := []interface{}{"function", "flush", "async"}
+			Expect(functionFlush.Args()).To(Equal(args))
```
elena-kolevska, line 6207: If someone has a better idea of testing specifically the `async` flushing, I'd love to hear it. Otherwise, I think that just testing the command string and the end result (no functions on the server) should be good enough.

elena-kolevska:
> `redis.RdsFunction` looks weird 😁

Indeed. I was reluctant to name it just `Function` though, cause even though it's not a reserved word in Go, when you're reading it in code, Redis Functions would probably not be the first thing to come to mind.
If you think that's not a concern though, I'm totally open to changing it.

monkey92t:
> > `redis.RdsFunction` looks weird 😁
> 
> Indeed. I was reluctant to name it just `Function` though, cause even though it's not a reserved word in Go, when you're reading it in code, Redis Functions would probably not be the first thing to come to mind. If you think that's not a concern though, I'm totally open to changing it.

`redis.Function{}` looks fine since go-redis only works with `redis` right? `redis.RdsFunction` looks weird as `pool.PoolName`.

elena-kolevska:
Ok, fixed. Thanks for the insight.

commands.go
```diff
@@ -395,6 +395,13 @@ type Cmdable interface {
 	ScriptKill(ctx context.Context) *StatusCmd
 	ScriptLoad(ctx context.Context, script string) *StringCmd
 
+	FunctionLoad(ctx context.Context, code string) *StringCmd
+	FunctionLoadReplace(ctx context.Context, code string) *StringCmd
+	FunctionDelete(ctx context.Context, libName string) *StringCmd
+	FunctionFlush(ctx context.Context) *StringCmd
+	FunctionFlushAsync(ctx context.Context) *StringCmd
+	FunctionList(ctx context.Context) *FunctionsListCmd
```
vmihailenco, line 403: Let's rename FunctionsListCmd to FunctionListCmd for consistency with Redis `FunctionList`?

command.go
```diff
@@ -3840,3 +3840,167 @@ func (cmd *ZSliceWithKeyCmd) readReply(rd *proto.Reader) (err error) {
...
+
+func (cmd *FunctionsListCmd) readReply(rd *proto.Reader) (err error) {
+	cmd.val, err = readRdsLibrarySlice(rd)
+	return err
+}
+
+func readRdsLibrarySlice(rd *proto.Reader) ([]Library, error) {
+	n, err := rd.ReadArrayLen()
+	if err != nil {
+		return []Library{}, err
```
vmihailenco, line 3898: nitpick: you can replace `[]Library{}` with `nil`

command.go
```diff
@@ -3840,3 +3840,167 @@ func (cmd *ZSliceWithKeyCmd) readReply(rd *proto.Reader) (err error) {
...
+	for i := 0; i < n; i++ {
+		k, err := rd.ReadString()
+		if err != nil {
+			return Library{}, err
+		}
+
+		switch k {
+		case "library_name":
+			library.Name, err = rd.ReadString()
+			if err != nil {
```
vmihailenco, line 3926: looks like you double check the `err` here

command.go
vmihailenco: Can't say for @monkey92t , but it looks fine too me :+1:

vmihailenco:
Looks good so far :+1:

monkey92t:
I think:

```go
func FunctionList(libraryNamePattern string, withCode bool) []Library
```

We should try to maintain consistency with redis commands, and should not increase the cost for users to learn go-redis.

elena-kolevska:
Yeah, I definitely see the point there; I just feel that it wouldn't look very pretty when we call the method without the optional arguments:

```
redis.FunctionList("", false)
```

The following would actually be equivalent, and it looks better, but then we would have to teach people to use it that way; it's not obvious:
```
redis.FunctionList("*", false)
```

At the very minimum, I could make the library name pattern its own struct and pass nil. Still not great, but I think it looks a bit better than an empty string. What do you think?
```
redis.FunctionList(nil, false)

// Or with an argument:
redis.FunctionList(LibraryNamePattern{"mylib"}, false)
```

commands_test.go
SoulPancake: @elena-kolevska To specifically test the asynchronous behavior we can use a goroutine to perform some blocking operation, wait for a few seconds, and then call FunctionFlushAsync. We can then verify that the FunctionFlushAsync method does not block and returns without an error. Is that the right way to think?

vmihailenco:
I think it should be either

```go
func FunctionList(libraryNamePattern string, withCode bool) []Library
```

or Option 5

```go
rdb.FunctionList(FunctionListArgs{
	LibraryName: "mylib",
	WithCode: true,
})				// []Library{}
```

The latter is somewhat more readable.

>I don't like returning a single Library object wrapped in an array when q.LibraryName is not null

We could add a helper function to retrieve a single library:

```go
rdb.FunctionList(...).Result() // []Library
rdb.FunctionList(...).First() // *Library
```

vmihailenco:
Also check:

```go
// SORT list LIMIT 0 2 ASC
vals, err := rdb.Sort(ctx, "list", &redis.Sort{Offset: 0, Count: 2, Order: "ASC"}).Result()

// ZRANGEBYSCORE zset -inf +inf WITHSCORES LIMIT 0 2
vals, err := rdb.ZRangeByScoreWithScores(ctx, "zset", &redis.ZRangeBy{
    Min: "-inf",
    Max: "+inf",
    Offset: 0,
    Count: 2,
}).Result()

// ZINTERSTORE out 2 zset1 zset2 WEIGHTS 2 3 AGGREGATE SUM
vals, err := rdb.ZInterStore(ctx, "out", &redis.ZStore{
    Keys: []string{"zset1", "zset2"},
    Weights: []int64{2, 3}
}).Result()
```

monkey92t:
I've considered "Option 5" too, if you all like it then we'll use it. 😄

elena-kolevska:
Great that we have an agreement :)

I implemented everything as you suggested, with the only difference being that the `.First()` method uses value instead of pointer semantics. I'm relatively new to Go, and I'm totally open to opinions here; I was just following the guideline that if it's a struct type and you're not expecting it to be mutated, you should use value semantics so you're not polluting the heap. As I said, though: new to Go, totally open to learning! ;)

I also implemented a `FirstVal()` method for completeness.

monkey92t:
Library is large enough that we can use ptr. What is the point of adding FirstVal()?

elena-kolevska:
Ok, I'll change it to a pointer then.

Re: FirstVal(), my reasoning was to have the matching "single-library" method for `Val()` as we have for `Result()`. I'm not a fan of the naming, though, I gotta admit.

```
rdb.FunctionList(...).Result() // []Library, err
rdb.FunctionList(...).First() // *Library, err

rdb.FunctionList(...).Val() // []Library
rdb.FunctionList(...).FirstVal() // *Library
```

commands_test.go
elena-kolevska: Yeah, great idea. Do you think we should add that to the test site, or would that fall more into the "testing the server" category?
I'm kind of inclined to think that for the commands that return a simple `OK` response where we don't need to populate a more complex object, checking that we've sent the correct string and we've gotten an OK should be good enough. What do you say?

monkey92t:
@elena-kolevska I tidied up the code.....I hope you won't mind. 😊

elena-kolevska:
Of course not! Quite the contrary, it’s great for my learning. 🙏

monkey92t:
@elena-kolevska  Do you want to add more commands? such: `function kill`, `function status`, or do we add it later?

elena-kolevska:
Since I'll only be able to come back to this next week, I'd suggest we close this PR now, and I'll add what's missing next week. There was also a PR on `FUNCTION KILL` already; we should check if we can use it.

monkey92t (approved):
OK, we don't need to close it, this PR can already do some work, we will add `function kill`, `function status` later.

monkey92t:
@elena-kolevska Waiting for you to finish the rest of the work. 😄

elena-kolevska:
I'm very much looking forward to it too ;)
Thank you all for your help with this PR.

vmihailenco:
@elena-kolevska thanks for your contribution and putting an extra effort into API design! It's clean and simple :+1:

PR #2479, opened 2023-03-09
fixed #2462  v9 continue support dragonfly,  it's Hello command return "NOAUTH Authentication required" error

### LINKED ISSUE #2462: NOAUTH Authentication required when use go-redis connect dragonfly
i want use go-redis/v9  to connect dragonfly (redis_version:df-v0.15.0, a db like redis), but i go a err : NOAUTH Authentication required.
but it ok when i use  go-redis/v9 to connect my local redis, 
and it also ok when i use go-redis/v8 to connect dragonfly.
this is a bug?  why go-redis/v9 not support dragonfly with password?
`
var redisdb *redis.Client  

 redisdb = redis.NewClient(&redis.Options{  
      Addr:     "127.0.0.1:6379", // 指定
      Password: "1234",
      DB:       2, // redis一共16个库，指定其中一个库即可
      PoolSize: 2000,
 })    

t.Log(redisdb.Info(context.Background()))
`

xinzhanguo:
See https://github.com/redis/go-redis/issues/2462

dragonfly dont support hello command, it's return "NOAUTH Authentication required" when use Hello command.

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -286,7 +286,7 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 	if err := conn.Hello(ctx, 3, username, password, "").Err(); err == nil {
 		auth = true
 	} else if !strings.HasPrefix(err.Error(), "ERR unknown command") &&
-	// this check is for compatibility DragonflyDB.
+		// this check is for compatibility DragonflyDB.
 		!strings.HasPrefix(err.Error(), "NOAUTH Authentication") {
 		return err
 	}
--- a/redis.go
+++ b/redis.go
@@ -285,7 +285,8 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 	// we continue to provide services with RESP2.
 	if err := conn.Hello(ctx, 3, username, password, "").Err(); err == nil {
 		auth = true
-	} else if !strings.HasPrefix(err.Error(), "ERR unknown command") {
+	} else if !strings.HasPrefix(err.Error(), "ERR unknown command") &&
+		!strings.HasPrefix(err.Error(), "NOAUTH Authentication") {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


monkey92t:
We should maintain compatibility with redis, not `dragonfly`. 
We shouldn't support multiple systems at the same time, like `protobuf` supporting JSON is a bad thing. 
If `dragonfly` and `redis` fork in the future, it will be a disaster for go-redis.

/cc @vmihailenco

vmihailenco:
If this is all that is needed to support DragonflyDB, then I am fine with keeping it. It is a tiny change that does not cost us anything.

redis.go
```diff
@@ -285,7 +285,8 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 	// we continue to provide services with RESP2.
 	if err := conn.Hello(ctx, 3, username, password, "").Err(); err == nil {
 		auth = true
-	} else if !strings.HasPrefix(err.Error(), "ERR unknown command") {
+	} else if !strings.HasPrefix(err.Error(), "ERR unknown command") &&
+		!strings.HasPrefix(err.Error(), "NOAUTH Authentication") {
```
vmihailenco, line 289: @xinzhanguo could you a comment mentioning that this is check is for DragonflyDB?

xinzhanguo: i checked, it's ok, DragonflyDB code is here https://github.com/dragonflydb/dragonfly/blob/8cf81151166865e2201fe10bf660e140e0e2fc13/src/server/main_service.cc#L612

Of course, the best way is that the third-party system supports `Hello` command or return standard error messages, but the reality is that the already running service will not be updated, even the code is modified, there will be problems when using v9 requests.

monkey92t: Can code comments be added? This line of code is marked for compatibility with dragonfly.

monkey92t:
We accept this PR, but we do not guarantee compatibility with dragonfly.

monkey92t (approved):

vmihailenco (approved):

PR #2483, opened 2023-03-13
feat: add client info command

### LINKED ISSUE #2482: CLIENT INFO missing

Support for CLIENT INFO is missing.

## Expected Behavior

Client interface that returns a number of fields

```
id=12 addr=127.0.0.1:51300 laddr=127.0.0.1:6379 fd=7 name= age=3 idle=1 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 qbuf=26 qbuf-free=20448 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=22298 events=r cmd=client|info user=default redir=-1 resp=2
```

## Current Behavior

No support

## Steps to Reproduce

n/a

## Context (Environment)

v9.0.2

monkey92t:
We can also consider improving `client list` in the future.

fix #2482

### DIFF
--- a/command.go
+++ b/command.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/redis/go-redis/v9/internal"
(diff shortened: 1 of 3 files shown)


command.go
```diff
@@ -3992,3 +3993,269 @@ func (cmd *FunctionListCmd) readFunctions(rd *proto.Reader) ([]Function, error)
...
+	ClientDenyBlocking        ClientFlags = 1 << 41 /* Indicate that the client should not be blocked. currently, turned on inside MULTI, Lua, RM_Call, and AOF client */
+	ClientReplRDBOnly         ClientFlags = 1 << 42 /* This client is a replica that only wants RDB without replication buffer. */
+	ClientNoEvict             ClientFlags = 1 << 43 /* This client is protected against client memory eviction. */
+	ClientAllowOOM            ClientFlags = 1 << 44 /* Client used by RM_Call is allowed to fully execute scripts even when in OOM */
+	ClientNoTouch             ClientFlags = 1 << 45 /* This client will not touch LFU/LRU stats. */
+	ClientPushing             ClientFlags = 1 << 46 /* This client is pushing notifications. */
+)
+
+// ClientInfo is redis-server ClientInfo, not go-redis *Client
+type ClientInfo struct {
```
chayim, line 4057: I'm wary of supporting a specific type like this. It's not unlikely that Redis will (and has) changed these things in the past. How much would you hate a map[string]string or something else that's fluid.

monkey92t: > I'm wary of supporting a specific type like this. It's not unlikely that Redis will (and has) changed these things in the past. How much would you hate a map[string]string or something else that's fluid.

Whether it is performance or API design is not a good option, map is suitable for unknown k-v.

monkey92t: I agree that `ClientInfo` can introduce name ambiguity, especially since we already have `*Client`. But I still suggest that the name should be consistent with redis-server.

vmihailenco: I agree with @chayim that this is going to break a lot, but I doubt that `map[string]string` is a better solution. `map[string]string` is much harder to use and will break silently...

command.go
```diff
@@ -3992,3 +3993,269 @@ func (cmd *FunctionListCmd) readFunctions(rd *proto.Reader) ([]Function, error)
...
+	ID                 int64         // redis version 2.8.12, a unique 64-bit client ID
+	Addr               string        // address/port of the client
+	LAddr              string        // address/port of local address client connected to (bind address)
+	FD                 int64         // file descriptor corresponding to the socket
+	Name               string        // the name set by the client with CLIENT SETNAME
+	Age                time.Duration // total duration of the connection in seconds
+	Idle               time.Duration // idle time of the connection in seconds
+	Flags              ClientFlags   // client flags (see below)
+	DB                 int           // current database ID
```
vmihailenco, line 4066: Let's change `int` to `int64` (or `int32` if you are sure). `int` is either `int32` or `int64` depending on Go arch, but these fields don't depend on that.

vmihailenco, line 4066: Let's change `int` to `int64` (or `int32` if you are sure). `int` is either `int32` or `int64` depending on Go arch, but these fields don't depend on that.

vmihailenco:
I would merge this as is since `struct` is much more convenient to use then `map[string]string` and we can change it in a backwards compatible manner.

I would also consider not adding this API at all since it will break a lot and will require some efforts to maintain.

@chayim what do you think?

vmihailenco (approved):

command.go
monkey92t: I think that: when we are sure that it will not exceed int32, I suggest we use the common int type. When we cannot determine the value size, we should use int64.

Issue #2398, opened 2023-01-31
Add support for FUNCTION KILL command


chayim:
Redis FUNCTION KILL command was added in version 7.0.0 and is currently supported by the library. We should add support as documented [here](https://redis.io/commands/FUNCTION-KILL).

chayim:
@elena-kolevska Did you get started on this as we were discussing? Can you and @dwibedis sync to make sure we're not doubling up work.

elena-kolevska:
Hey @chayim  and @dwibedis . No, I haven't gotten to `KILL` yet. Feel free to start working on it. Thanks!

monkey92t:
`FUNCTION` is a combination of multiple commands, like `SCRIPT`.

dwibedis:
Hi, I have raised a draft PR:- https://github.com/redis/go-redis/pull/2465/files.
Please help in checking if I am in the right direction or not,
as next step, I will have to write tests. right?

elena-kolevska:
> Also, as I was going through other commands which were using `NewStatusCmd`, `NewTimeCmd` etc. Should I make a `FunctionCmd` struct, with function and library name as args and start using it?

Hi @dwibedis. No need to create the new structs; I'm working on them as part of the `FUNCTION LIST` implementation.

PR #2486, opened 2023-03-14
Function stats, function kill, fcall and fcall_ro

elena-kolevska:


elena-kolevska:
Not ready for review yet; I need help with a few things.

commands_test.go
```diff
@@ -6305,6 +6314,116 @@ var _ = Describe("Commands", func() {
...
+
+			<-started
+			time.Sleep(1 * time.Second)
+			r, err = client.FunctionStats(ctx).Result()
+
+			select {
+			case <-done:
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(r.Engines)).To(Equal(1))
+				Expect(r.RunningScript.Name).To(Equal(lib.Functions[0].Name))
```
elena-kolevska, line 6417: @vmihailenco @monkey92t I need some help making this test pass. I can't seem to get any result for `running_script` from `function stats`.

elena-kolevska:
Let's talk about the API design of `FUNCTION STATS` in this issue https://github.com/redis/go-redis/issues/2402#issuecomment-1469087768

commands_test.go
monkey92t: > 

We need a way to block lua.

elena-kolevska: That's the thing, blocking commands are not allowed inside of functions. I was hoping that the 5M iterations of `SET` would be enough. On my machine, they take around 3-4 seconds.

command.go
```diff
@@ -3992,3 +3992,231 @@ func (cmd *FunctionListCmd) readFunctions(rd *proto.Reader) ([]Function, error)
...
+
+	cmd.val = result
+	return nil
+}
+
+func (cmd *FunctionStatsCmd) readRunningScript(rd *proto.Reader) (RunningScript, bool, error) {
+	err := rd.ReadFixedMapLen(3)
+	if err != nil {
+		if err == Nil {
+			return RunningScript{}, false, nil // TODO How to handle nil response?
```
monkey92t, line 4109: remove `TODO`

command.go
```diff
@@ -3992,3 +3992,231 @@ func (cmd *FunctionListCmd) readFunctions(rd *proto.Reader) ([]Function, error)
...
+		}
+
+		for i := 0; i < 2; i++ {
+			key, err := rd.ReadString()
+			switch key {
+			case "libraries_count":
+				engine.LibrariesCount, err = rd.ReadInt()
+			case "functions_count":
+				engine.FunctionsCount, err = rd.ReadInt()
+			}
```
monkey92t, line 4166: add default

command.go
```diff
@@ -3992,3 +3992,231 @@ func (cmd *FunctionListCmd) readFunctions(rd *proto.Reader) ([]Function, error)
...
+	for i := 0; i < n; i++ {
+		rs, _, err := cmd.readRunningScript(rd)
+		if err != nil {
+			return nil, false, err
+		}
+		runningScripts = append(runningScripts, rs)
+		isRunning = true
+	}
+
+	return runningScripts, isRunning, nil
```
monkey92t, line 4221: ha~, `return runnScripts, len(runningScripts) > 0, nil`, isRunning is redundant.

commands.go
```diff
@@ -3349,6 +3352,38 @@ func (c cmdable) FunctionRestore(ctx context.Context, libDump string) *StringCmd
 	return cmd
 }
 
+func (c cmdable) FunctionStats(ctx context.Context) *FunctionStatsCmd {
+	cmd := NewFunctionStatsCmd(ctx, "function", "stats")
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+func (c cmdable) FCall(ctx context.Context, function string, keys []string, args ...interface{}) *Cmd {
```
monkey92t, line 3361: It feels weird here: `client.FCall(ctx, "func", []string{"key1", "key2", "arg1", "arg2"})`

I think:
```go
 FCall(ctx context.Context, function string, keys []string, args []interface{})
```

OR

```go
type FuncCall struct {
        FuncName string
        Keys []string
        Args []interface{}
}
 FCall(ctx context.Context, f FuncCall)
```

monkey92t: or: FCallArgs

commands.go
```diff
@@ -3349,6 +3352,38 @@ func (c cmdable) FunctionRestore(ctx context.Context, libDump string) *StringCmd
...
+
+func fcallArgs(command string, function string, keys []string, args ...interface{}) []interface{} {
+	cmdArgs := make([]interface{}, 3+len(keys), 3+len(keys)+len(args))
+	cmdArgs[0] = command
+	cmdArgs[1] = function
+	cmdArgs[2] = len(keys)
+	for i, key := range keys {
+		cmdArgs[3+i] = key
+	}
+	return appendArgs(cmdArgs, args)
```
monkey92t, line 3384: cmdArgs = append(cmdArgs, args...)

commands.go
```diff
@@ -3349,6 +3352,38 @@ func (c cmdable) FunctionRestore(ctx context.Context, libDump string) *StringCmd
...
+	cmdArgs := fcallArgs("fcall", function, keys, args...)
+	cmd := NewCmd(ctx, cmdArgs...)
+	cmd.SetFirstKeyPos(3)
+	_ = c(ctx, cmd)
+	return cmd
+}
+func (c cmdable) FCallRo(ctx context.Context, function string, keys []string, args ...interface{}) *Cmd {
+	cmdArgs := fcallArgs("fcall_ro", function, keys, args...)
+	cmd := NewCmd(ctx, cmdArgs...)
+	cmd.SetFirstKeyPos(3)
```
monkey92t, line 3371: should:
```go
// exp: fcall_ro func 0 arg1 arg2
if len(keys) > 0 {
        cmd.SetFirstKeyPos(3)
}
```

commands.go
```diff
@@ -3349,6 +3352,38 @@ func (c cmdable) FunctionRestore(ctx context.Context, libDump string) *StringCmd
...
+func (c cmdable) FunctionStats(ctx context.Context) *FunctionStatsCmd {
+	cmd := NewFunctionStatsCmd(ctx, "function", "stats")
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+func (c cmdable) FCall(ctx context.Context, function string, keys []string, args ...interface{}) *Cmd {
+	cmdArgs := fcallArgs("fcall", function, keys, args...)
+	cmd := NewCmd(ctx, cmdArgs...)
+	cmd.SetFirstKeyPos(3)
```
monkey92t, line 3364: should:
```go
// exp: fcall func 0 arg1 arg2
if len(keys) > 0 {
        cmd.SetFirstKeyPos(3)
}
```

monkey92t:
@elena-kolevska I have posted some comments, some of which are errors that need to be corrected, some are suggestions, and some are my thoughts on parameter naming.

Also, you need to resolve the conflict in the command.go file :)

commands.go
monkey92t: Yes, we have discovered an issue with eval/eval_ro. When key is not passed, we incorrectly treat arg parameter as key, which is incorrect. We should fix eval/eval_ro.

monkey92t: I'll fix eval/eval_ro.

commands.go
monkey92t: "I did not reference the eval/ro command, it is just my personal opinion.

elena-kolevska: Of course, I was just saying I can't spot the difference between the first of the two options you suggested and the one I did.

monkey92t: We cannot assume that eval/ro is correct just because it was implemented early. If there are better API implementation options in the future, the implementation of eval/ro command can also be modified.

elena-kolevska: I agree 100%. Indeed, I'm using older implementations as a reference because I'm still learning the library, but I understand there's always space to improve :)

monkey92t: ~ok, then we can stick with the same implementation as eval/ro and consider improvements in the future if there are better options.

commands_test.go
```diff
@@ -6369,6 +6378,118 @@ var _ = Describe("Commands", func() {
...
+			case <-done:
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(r.Engines)).To(Equal(1))
+				rs, isRunning := r.RunningScript()
+				Expect(isRunning).To(BeTrue())
+				Expect(rs.Name).To(Equal(lib.Functions[0].Name))
+				Expect(rs.Duration > 0).To(BeTrue())
+			case <-time.After(10 * time.Second):
+				log.Println("Long-running function call timed out")
+			}
```
monkey92t, line 6487: I didn't understand the test here. Why do we need to wait for the "done" signal? The FCall command will block until LUA execution is complete, but when the Lua-time-limit is reached, Redis can accept the "function kill/stats" command. If FCall takes one minute to execute, this test will time out. Additionally, we should execute "function-kill" after the test ends because if LUA execution takes too long, it will continue to block subsequent tests.

elena-kolevska: Good point about `function kill`. The function we're calling is relatively predictable and should never time out, but we should protect against it anyway! I'd have to change the function itself though, make it read-only, cause functions that have modified the keyspace in some way can't be killed. I'm on it!

Now, on to the more complicated issue :) I did some digging in the Redis source code and found that the`function stats` command lives in the same group as `function kill`, `quit`, `shutdown` commands. They all have the `CMD_ALLOW_BUSY` flag which lets them run while another command is running for a longer time. Otherwise, we would never get any results under the `running_script` field.

My point in using the `done` sync was to wait for fcall to complete before moving on to the next test, but your idea is better, we can just kill the function after we run the checks and that should force the goroutine to return.

Let me do these fixes and we can maybe look at the result again, see how can we make this pass.

elena-kolevska: Btw, I saw that https://github.com/redis/go-redis/pull/2465 was closed in the meantime, so I went ahead and implemented `function kill` in this same PR, cause I needed it.

command.go
```diff
@@ -3993,9 +3993,234 @@ func (cmd *FunctionListCmd) readFunctions(rd *proto.Reader) ([]Function, error)
...
+		default:
+			return RunningScript{}, false, fmt.Errorf("redis: function stats unexpected running_script key %s", key)
+		}
+
+		if err != nil {
+			return RunningScript{}, false, err
+		}
+	}
+
+	return RunningScript{}, true, nil
```
monkey92t, line 4137: should: return runningScript, true, nil

elena-kolevska: Oh, that's an embarrassing error! Leftover from the initial empty implementation...
But unfortunately, our `function stats` problem comes before that, we're not getting anything for `running_script` from the server itself.

command.go
```diff
@@ -3993,9 +3993,234 @@ func (cmd *FunctionListCmd) readFunctions(rd *proto.Reader) ([]Function, error)
...
+			return err
+		}
+
+		switch key {
+		case "running_script":
+			result.rs, result.isRunning, err = cmd.readRunningScript(rd)
+		case "engines":
+			result.Engines, err = cmd.readEngines(rd)
+		case "all_running_scripts": // Redis Enterprise only
+			result.allrs, result.isRunning, err = cmd.readRunningScripts(rd)
```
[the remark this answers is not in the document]
elena-kolevska: Its the latter, Redis Enterprise responds with `engines + running_script + all_running_scripts`. 
I inspected the RESP response "manually", in telnet and the code should work. But I can't confirm that until I get the `function stats` implementation right.

chayim:
#2398

elena-kolevska:
The test for `function stats` still doesn't pass, but I tested the code with an external server, where I triggered an `FCALL` manually, and the proper response is returned.
I repeated the test for Redis Enterprise too.

elena-kolevska:
I got it! The problem was that the goroutine was reusing the same connection. I changed it to use its own connection and now, finally, the test passes :)

monkey92t (approved):
Looks very good!

PR #2492, opened 2023-03-19
Add support for CLIENT NO-EVICT

### LINKED ISSUE #2385: Add support for CLIENT NO-EVICT command
Redis CLIENT NO-EVICT command was added in version 7.0.0 and is currently supported by the library. We should add support as documented [here](https://redis.io/commands/CLIENT-NO-EVICT).

SoulPancake:
fixes #2385

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -125,6 +125,8 @@ type Cmdable interface {
 
 	Command(ctx context.Context) *CommandsInfoCmd
 	ClientGetName(ctx context.Context) *StringCmd
+	ClientNoEvictOn(ctx context.Context) *StatusCmd
+	ClientNoEvictOff(ctx context.Context) *StatusCmd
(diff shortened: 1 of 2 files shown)


SoulPancake:
@monkey92t 
Here's the implementation
Followed the same standard followed by Jedis.

monkey92t:
`client-eviction` is not just a command, it is a state of the client connection, targeting a single network connection rather than the entire client.

See: https://redis.io/docs/reference/clients/

monkey92t:
This requires a complicated implementation. Initially we can set it at initialization. If we want to change during the client life cycle, we need to record its state for each network connection.

SoulPancake:
Understood, I'll be closing this
Will understand the problem statement better to tackle this later.

PR #2500, opened 2023-03-24
Add support for COMMAND GETKEYS & COMMAND GETKEYSANDFLAGS

### LINKED ISSUE #2388: Add support for COMMAND GETKEYSANDFLAGS command
Redis COMMAND GETKEYSANDFLAGS command was added in version 7.0.0 and is currently supported by the library. We should add support as documented [here](https://redis.io/commands/COMMAND-GETKEYSANDFLAGS).

SoulPancake:
Fixes #2388

### DIFF
--- a/command.go
+++ b/command.go
@@ -4220,24 +4220,14 @@ func (cmd *KeyFlagsCmd) readReply(rd *proto.Reader) error {
 		return nil
 	}
 
-	typ, err := rd.PeekReplyType()
-	if err != nil {
-		return err
-	}
-	array := typ == proto.RespArray
-
-	if array {
-		cmd.val = make([]KeyFlags, n)
-	} else {
-		cmd.val = make([]KeyFlags, n/2)
-	}
+	cmd.val = make([]KeyFlags, n)
 
 	for i := 0; i < len(cmd.val); i++ {
-		if array {
-			if err = rd.ReadFixedArrayLen(2); err != nil {
-				return err
-			}
+
+		if err = rd.ReadFixedArrayLen(2); err != nil {
+			return err
 		}
+
 		if cmd.val[i].Key, err = rd.ReadString(); err != nil {
 			return err
 		}
--- a/command.go
+++ b/command.go
@@ -4215,7 +4215,7 @@ func (cmd *KeyFlagsCmd) readReply(rd *proto.Reader) error {
 		return err
 	}
 
-	// If the n is 0, can't continue reading.
+	// If n is 0, can't continue reading.
 	if n == 0 {
 		cmd.val = make([]KeyFlags, 0)
 		return nil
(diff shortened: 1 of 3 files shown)


SoulPancake:
Hey @monkey92t @chayim @vmihailenco 
I have done this implementation and I have written the test cases from the official docs.
Please review it.

command.go
```diff
@@ -4168,3 +4168,91 @@ func (cmd *LCSCmd) readPosition(rd *proto.Reader) (pos LCSPosition, err error) {
...
+	if err != nil {
+		return err
+	}
+	array := typ == proto.RespArray
+
+	if array {
+		cmd.val = make([]KeyFlags, n)
+	} else {
+		cmd.val = make([]KeyFlags, n/2)
+	}
```
monkey92t, line 4233: I don't understand what's going on here. Why do we need to check the type of response in Redis? Does Redis respond with multiple data types?

commands.go
```diff
@@ -129,6 +129,8 @@ type Cmdable interface {
 
 	Command(ctx context.Context) *CommandsInfoCmd
 	CommandList(ctx context.Context, filter *FilterBy) *StringSliceCmd
+	CommandGetKeys(ctx context.Context, command string, args ...string) *StringSliceCmd
+	CommandGetKeysAndFlags(ctx context.Context, command string, args ...string) *KeyFlagsCmd
```
monkey92t, line 133:
```go
...(ctx context.Context, commands ...interface{})
```

commands.go
```diff
@@ -568,6 +570,32 @@ func (c cmdable) CommandList(ctx context.Context, filter *FilterBy) *StringSlice
 	return cmd
 }
 
+func (c cmdable) CommandGetKeys(ctx context.Context, command string, cmdArgs ...string) *StringSliceCmd {
+	args := make([]interface{}, 3+len(cmdArgs))
+	args[0] = "command"
+	args[1] = "getkeys"
+	args[2] = command
+    for i, cmdArg := range cmdArgs {
```
monkey92t, line 578: Didn't your editor(ide) help you correct the code formatting? Perhaps you should consider using the gofmt tool.

monkey92t: Perhaps you should pay attention to the difference between TAB and SPACE, as this is a common mistake.

You can search on google for replace TAB with SPACE (you ide).

commands.go
SoulPancake: Each of these commands would take only one type of command and their corresponding argument, correct?
In that case, can't we use a string for the command and then the args followed by it?

command.go
SoulPancake: As you can see https://github.com/redis/go-redis/blob/6790337e5ddff23427a083cbca1140e039bc693b/command.go#L1556
Similarly I've done this for ZSliceCmd as well as this
I'm not sure in which scenario it happens
Do you suggest we should directly make the array without checking the type of response we are getting because the protocol is fixed?

monkey92t: This is done to be compatible with both the RESP2 and RESP3 protocols. In older versions of Redis, map type responses are represented as array types, so we need to ensure that they continue to work properly in RESP2.

 There is no such situation with command-getkeys.

commands.go
SoulPancake: Should we maintain by calling it commandAndArgs? @monkey92t

monkey92t: We don't need to distinguish between commands and args.

SoulPancake: I was just trying to decide on apt. nomenclature for this, just commands also works.

SoulPancake:
I've done all the changes as requested, Can you please take a look? @monkey92t

monkey92t (approved):

monkey92t:
I still have some doubts about this PR, and I will consider improving it later.

PR #2502, opened 2023-03-25
Add support for acl dryrun command

aksh-02:
Fixes https://github.com/redis/go-redis/issues/2381

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -450,6 +450,8 @@ type Cmdable interface {
 	GeoSearchStore(ctx context.Context, key, store string, q *GeoSearchStoreQuery) *IntCmd
 	GeoDist(ctx context.Context, key string, member1, member2, unit string) *FloatCmd
 	GeoHash(ctx context.Context, key string, members ...string) *StringSliceCmd
+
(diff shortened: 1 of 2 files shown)


aksh-02:
@monkey92t @chayim please review these changes

commands.go
```diff
@@ -450,6 +450,8 @@ type Cmdable interface {
 	GeoSearchStore(ctx context.Context, key, store string, q *GeoSearchStoreQuery) *IntCmd
 	GeoDist(ctx context.Context, key string, member1, member2, unit string) *FloatCmd
 	GeoHash(ctx context.Context, key string, members ...string) *StringSliceCmd
+
+	AclDryRun(ctx context.Context, username, command string, args ...interface{}) *StringCmd
```
monkey92t, line 454: // or args or commands
should: ACLDryRun(ctx, username, command ...interface)

for us, there is no need to distinguish between command and args.  It's ACL, not Acl, just like CNCF and not Cncf.

monkey92t (approved):
tks..

commands.go
aksh-02: comment incorporated

PR #2513, opened 2023-03-31
fix: reader float parser

iliyanm:
### Description

Changed the reader float parser to support **nan** values.
Based on the redis [documentation](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md#:~:text=and%20LF%20characters.-,Double,-The%20general%20form) possible float values should be:

```
",inf\r\n"
",-inf\r\n"
",nan\r\n"
```

### Example
When I execute `FT.INFO *index name*` parameter `records_per_doc_avg` can be `-nan` and the float parser returns:
```go
strconv.ParseFloat: parsing "-nan": invalid syntax
```

### DIFF
--- a/internal/proto/reader.go
+++ b/internal/proto/reader.go
@@ -163,7 +163,11 @@ func (r *Reader) ReadReply() (interface{}, error) {
 	case RespInt:
 		return util.ParseInt(line[1:], 10, 64)
 	case RespFloat:
-		return r.readFloat(line)
+		f, err := r.readFloat(line)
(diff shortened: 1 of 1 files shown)


iliyanm:
@monkey92t could you please review

monkey92t:
I don't understand why use *float64? Maybe math.Nan()?

iliyanm:
@monkey92t you are right. I changed it, please review again.

monkey92t (approved):

Issue #2509, opened 2023-03-30
AUTH error when working with Envoy proxy after upgrading from v8 to v9


vandyvilla:
Issue tracker is used for reporting bugs and discussing new features. Please use
[stackoverflow](https://stackoverflow.com) for supporting issues.

The v8 client is able to talk to a Redis cluster fronted by an Envoy proxy. After upgrading to v9, it shows AUTH errors. 

## Expected Behavior
The client should still be able to talk to the proxy. 

## Current Behavior
it shows AUTH errors. 
`NOAUTH Authentication required.`

## Steps to Reproduce

Code Snippet: 
`rOpts := &redis.Options{
    Addr: proxy-addr,
    Username: proxy-username,
    Password: proxy-password
}`
`client := redis.NewClient(rOpts)`
`cmdStatus := client.Set(ctx, "key", "value", 0)`

monkey92t:
The certification process for versions 8 and 9 is as follows:

in v8, `AuthACL`(username+password) or `Auth`(only password)
```go
if password != "" {
        if username != "" {
                client.AuthACL(ctx, username, password)
        } else {
               client.Auth(ctx, password)
        }
}
```

in v9, the `HELLO` command is used for authentication, for redis servers that do not support the hello command (<Redis 6), authentication will continue to be performed using the v8 method.
```go
err := conn.Hello(3, username, password)
if err != nil {
        // call v8 auth
}
```

monkey92t:
The authentication method of v9 works well with all Redis servers, but I am not sure how your proxy works. It is possible that your proxy supports the hello command but does not perform authentication.

If the hello command is successfully executed in go-redis v9, authentication using AuthACL/Auth will not be performed again. 

In the redis-doc: "AUTH <username> <password>: directly authenticate the connection in addition to switching to the specified protocol version. This makes calling [AUTH](https://redis.io/commands/auth) before HELLO unnecessary when setting up a new connection."

monkey92t:
There is a lack of error types in redis-server, and it only responds with human-readable error messages. When dealing with certain errors, we have to compare error strings to get a result.

PR #2515, opened 2023-03-31
feat: no longer verify HELLO error messages

monkey92t:
When the server responds with the RESP protocol and the result is not a normal execution result of the HELLO command, we consider it to be an indication that the server does not support the HELLO command.

The server may be a redis-server that does not support the HELLO command, or it could be DragonflyDB or a third-party redis-proxy. They all respond with different error string results for unsupported commands, making it difficult to rely on error strings to determine all results.

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"strings"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


monkey92t:
#2509

vmihailenco (approved):

PR #2529, opened 2023-04-03
feat: read the structure to increase the judgment of the omitempty op…

monkey92t:
read the structure to increase the judgment of the omitempty option, the same as json


monkey92t:
@vmihailenco @chayim Please approve~

Issue #2527, opened 2023-04-02
Support ZRANK: Added the optional `WITHSCORE` argument.


chayim:
Documented [here](https://redis.io/commands/ZRANK).

SoulPancake:
I'd like to work on this too @chayim

PR #2531, opened 2023-04-03
Add support for zrankwithscore and zrevrankwithscore commands 

### LINKED ISSUE #2528: Support ZREVRANK: Added the optional `WITHSCORE` argument.
Documented [here](https://redis.io/commands/ZREVRANK).

SoulPancake:
Fixes #2528 #2527

### DIFF
--- a/command.go
+++ b/command.go
@@ -4710,3 +4710,64 @@ func (cmd *ClusterShardsCmd) readReply(rd *proto.Reader) error {
 
 	return nil
 }
+
+// -----------------------------------------
+
+type RankScore struct {
+	Rank  int64
+	Score string
+}
+
+type RankWithScoreCmd struct {
+	baseCmd
+
+	val RankScore
+}
+
+var _ Cmder = (*RankWithScoreCmd)(nil)
+
+func NewRankWithScoreCmd(ctx context.Context, args ...interface{}) *RankWithScoreCmd {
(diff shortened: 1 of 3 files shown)


SoulPancake:
I think all 7.2 cmd tests would fail currently because of the test version, can we upgrade that? 
@monkey92t @chayim @vmihailenco

SoulPancake:
@monkey92t Any idea why this is returning nil?

monkey92t:
> @monkey92t Any idea why this is returning nil?

According to the redis docs, the response may be Nil......

SoulPancake:
@monkey92t The response must be nil when the member or key doesn't exist in the sorted set right

command.go
```diff
@@ -4724,3 +4724,64 @@ func (cmd *ClusterShardsCmd) readReply(rd *proto.Reader) error {
 
 	return nil
 }
+
+// -----------------------------------------
+
+type RankScore struct {
+	Rank  int64
+	Score string
```
monkey92t, line 4732: For Score, why did you use the string type?

monkey92t:
In the RESP3 protocol, score uses the Float:
```shell
ZRANK myzset "three" withscore
*2
:2
,3
```

monkey92t (approved):

monkey92t (approved):

PR #2536, opened 2023-04-04
feat: Add ACL LOG

### LINKED ISSUE #2524: Support ACL LOG: Added entry ID, timestamp created, and timestamp last updated.
Documented [here](https://redis.io/commands/ACL-LOG).

SoulPancake:
Resolves #2524

### DIFF
--- a/command.go
+++ b/command.go
@@ -4710,3 +4710,100 @@ func (cmd *ClusterShardsCmd) readReply(rd *proto.Reader) error {
 
 	return nil
 }
+
+// -------------------------------------------
+
+type ACLLogEntry struct {
+	Count                int64
+	Reason               string
+	Context              string
+	Object               string
+	Username             string
+	AgeSeconds           float64
+	ClientInfo           string
+	EntryID              int64
+	TimestampCreated     int64
+	TimestampLastUpdated int64
+}
+
+type ACLLogCmd struct {
+	baseCmd
+
+	val []ACLLogEntry
+}
+
+var _ Cmder = (*ACLLogCmd)(nil)
+
+func NewACLLogCmd(ctx context.Context, args ...interface{}) *ACLLogCmd {
+	return &ACLLogCmd{
+		baseCmd: baseCmd{
+			ctx:  ctx,
+			args: args,
+		},
+	}
+}
+
+func (cmd *ACLLogCmd) SetVal(val []ACLLogEntry) {
+	cmd.val = val
+}
+
+func (cmd *ACLLogCmd) Val() []ACLLogEntry {
+	return cmd.val
+}
+
+func (cmd *ACLLogCmd) Result() ([]ACLLogEntry, error) {
+	return cmd.Val(), cmd.Err()
+}
+
+func (cmd *ACLLogCmd) String() string {
+	return cmdString(cmd, cmd.val)
+}
+
+func (cmd *ACLLogCmd) readReply(rd *proto.Reader) error {
+	n, err := rd.ReadArrayLen()
+	if err != nil {
+		return err
+	}
+
+	cmd.val = make([]ACLLogEntry, n/2)
+	for i := 0; i < n/2; i++ {
+		entry := &cmd.val[i]
+		for j := 0; j < 10; j += 2 {
+			key, err := rd.ReadString()
+			if err != nil {
+				return err
+			}
+
+			switch key {
+			case "count":
+				entry.Count, err = rd.ReadInt()
+			case "reason":
+				entry.Reason, err = rd.ReadString()
+			case "context":
+				entry.Context, err = rd.ReadString()
+			case "object":
+				entry.Object, err = rd.ReadString()
+			case "username":
+				entry.Username, err = rd.ReadString()
+			case "age-seconds":
+				entry.AgeSeconds, err = rd.ReadFloat()
+			case "client-info":
+				entry.ClientInfo, err = rd.ReadString()
+			case "entry-id":
+				entry.EntryID, err = rd.ReadInt()
+			case "timestamp-created":
+				entry.TimestampCreated, err = rd.ReadInt()
+			case "timestamp-last-updated":
+				entry.TimestampLastUpdated, err = rd.ReadInt()
+			default:
+				return fmt.Errorf("redis: unexpected key %q in ACL LOG reply", key)
+			}
+
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
--- a/commands.go
+++ b/commands.go
@@ -495,6 +495,8 @@ type Cmdable interface {
 	GeoHash(ctx context.Context, key string, members ...string) *StringSliceCmd
 
 	ACLDryRun(ctx context.Context, username string, command ...interface{}) *StringCmd
+	ACLLog(ctx context.Context, count ...int64) *ACLLogCmd
+	ACLLogReset(ctx context.Context) *StatusCmd
 }
 
 type StatefulCmdable interface {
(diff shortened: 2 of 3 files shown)


SoulPancake:
Need some help on this @monkey92t

SoulPancake:
@monkey92t Can you please review this

command.go
```diff
@@ -4784,3 +4784,100 @@ func (cmd *RankWithScoreCmd) readReply(rd *proto.Reader) error {
...
+// -------------------------------------------
+
+type ACLLogEntry struct {
+	Count                int64
+	Reason               string
+	Context              string
+	Object               string
+	Username             string
+	AgeSeconds           float64
+	ClientInfo           string
```
monkey92t, line 4797: wait #2483

monkey92t:
@SoulPancake We need to wait for the merge of #2483. In this PR, the ClientInfo field's type comes from #2483. According to our development guidelines, #2483 requires approval from other maintainers (developers cannot approve their own PR).

monkey92t:
We can first think about the solutions for commands such as #2492, #2553, #2551.

monkey92t:
@SoulPancake #2483 merged....please continue....

SoulPancake:
Can you take a look now @monkey92t ?

monkey92t:
> Can you take a look now @monkey92t ?

.........

I mentioned earlier, in #2483, we defined the ClientInfo struct. ACLLogEntry.ClientInfo should use ClientInfo instead of the string type. We already have the parseClientInfo method to parse it.

output:

```shell
127.0.0.1:6379> acl log 1
1)  1# "count" => (integer) 1
    2# "reason" => "auth"
    3# "context" => "toplevel"
    4# "object" => "AUTH"
    5# "username" => "someuser"
    6# "age-seconds" => (double) 192531.833
    7# "client-info" => "id=2351 addr=127.0.0.1:52326 laddr=127.0.0.1:6379 fd=8 name= age=1447772 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 qbuf=48 qbuf-free=16842 argv-mem=25 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=18753 events=r cmd=auth user=default redir=-1 resp=2 lib-name=mylibname lib-ver=mylibver"
    8# "entry-id" => (integer) 0
    9# "timestamp-created" => (integer) 1682153764713
   10# "timestamp-last-updated" => (integer) 1682153764713
```

monkey92t:
str:     7# "client-info" => "id=2351 addr=127.0.0.1:52326 laddr=127.0.0.1:6379 fd=8 name= age=1447772 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 qbuf=48 qbuf-free=16842 argv-mem=25 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=18753 events=r cmd=auth user=default redir=-1 resp=2 lib-name=mylibname lib-ver=mylibver"


We should use the ClientInfo struct defined in #2482 for parsing, instead of using a string. This is why I asked you to wait for the #2482 merge.

```go

// ClientInfo is redis-server ClientInfo, not go-redis *Client
type ClientInfo struct {
	ID                 int64         // redis version 2.8.12, a unique 64-bit client ID
	Addr               string        // address/port of the client
	LAddr              string        // address/port of local address client connected to (bind address)
	FD                 int64         // file descriptor corresponding to the socket
	Name               string        // the name set by the client with CLIENT SETNAME
	Age                time.Duration // total duration of the connection in seconds
	Idle               time.Duration // idle time of the connection in seconds
	Flags              ClientFlags   // client flags (see below)
	DB                 int           // current database ID
	Sub                int           // number of channel subscriptions
	PSub               int           // number of pattern matching subscriptions
	SSub               int           // redis version 7.0.3, number of shard channel subscriptions
	Multi              int           // number of commands in a MULTI/EXEC context
	QueryBuf           int           // qbuf, query buffer length (0 means no query pending)
	QueryBufFree       int           // qbuf-free, free space of the query buffer (0 means the buffer is full)
	ArgvMem            int           // incomplete arguments for the next command (already extracted from query buffer)
	MultiMem           int           // redis version 7.0, memory is used up by buffered multi commands
	BufferSize         int           // rbs, usable size of buffer
	BufferPeak         int           // rbp, peak used size of buffer in last 5 sec interval
	OutputBufferLength int           // obl, output buffer length
	OutputListLength   int           // oll, output list length (replies are queued in this list when the buffer is full)
	OutputMemory       int           // omem, output buffer memory usage
	TotalMemory        int           // tot-mem, total memory consumed by this client in its various buffers
	Events             string        // file descriptor events (see below)
	LastCmd            string        // cmd, last command played
	User               string        // the authenticated username of the client
	Redir              int64         // client id of current client tracking redirection
	Resp               int           // redis version 7.0, client RESP protocol version
	LibName            string        // redis version 7.2, client library name
	LibVer             string        // redis version 7.2, client library version
}
```

SoulPancake:
@monkey92t Gotchu. 
Thanks! Working on migrating this

monkey92t:
It is recommended to use pointers for large structures.

command.go
```diff
@@ -5057,3 +5057,104 @@ func parseClientInfo(txt string) (info *ClientInfo, err error) {
...
+			case "username":
+				entry.Username, err = rd.ReadString()
+			case "age-seconds":
+				entry.AgeSeconds, err = rd.ReadFloat()
+			case "client-info":
+				txt, err := rd.ReadString()
+				if err != nil {
+					return err
+				}
+				entry.ClientInfo, err = parseClientInfo(strings.TrimSpace(txt)) // nolint:ineffassign,staticcheck
```
monkey92t, line 5142: Here, we need to handle "err". "err" has become a local variable here and is no longer the external "err".

commands.go
```diff
@@ -3946,3 +3948,25 @@ func (c cmdable) ModuleLoadex(ctx context.Context, conf *ModuleLoadexConfig) *St
 	_ = c(ctx, cmd)
 	return cmd
 }
+
+func (c cmdable) ACLLog(ctx context.Context, count ...int64) *ACLLogCmd {
```
monkey92t, line 3952: Why is "...int64" used here? ACL-LOG has only one parameter.  `acl log 10` or `acl log rest`.

command.go
```diff
@@ -5057,3 +5057,104 @@ func parseClientInfo(txt string) (info *ClientInfo, err error) {
...
+	ClientInfo           *ClientInfo
+	EntryID              int64
+	TimestampCreated     int64
+	TimestampLastUpdated int64
+}
+
+type ACLLogCmd struct {
+	baseCmd
+
+	val []ACLLogEntry
```
monkey92t, line 5079: recommend using: []*ACLLogEntry

SoulPancake:
@vmihailenco @chayim @monkey92t 
Please review once
and suggest if I should separate the `ACL LOG ` command with no additional params

chayim:
@elena-kolevska I think this is yours.

elena-kolevska:
I think we should separate the two commands because even though they are the same command in Redis, logically they do very different things and return very different responses.
I suggest that `AclLog` keeps the current implementation, and we create a new one for `AclLogReset`.

SoulPancake:
> I suggest that `AclLog` gets a single int argument
How will be able to cover the scenario where the single int arg is not provided or just "ACL LOG" is run with no additional params? @elena-kolevska

commands_test.go
```diff
@@ -1985,6 +1985,62 @@ var _ = Describe("Commands", func() {
 
 			Expect(args).To(Equal(expectedArgs))
 		})
+
+		It("should ACL LOG", func() {
```
elena-kolevska, line 1989: I think we should cause a few acl logs for the tests, so we can check we're indeed reading in the resp response. We would have caught that the current code returns an error.
Not sure if that's easy to do with the current testing set-up, but @chayim  is working on a more flexible solution at the moment.

commands.go
elena-kolevska: I think this is so we can either add or omit the count parameter, since it's not required. So we could do either of the following:

```
redis.ACLLog(ctx)           // For all log entries
redis.ACLLog(ctx, n)       // For n log entries
redis.ACLLogReset(ctx)  // For resetting log entries
```

monkey92t:
I think:

```go
ACLLog(ctx context.Context, count int64) *ACLLogCmd
ACLLogReset(ctx context.Context) *StatusCmd
```

We can ignore the 'acl log' command

SoulPancake:
> I think:
> 
> ```go
> ACLLog(ctx context.Context, count int64) *ACLLogCmd
> ACLLogReset(ctx context.Context) *StatusCmd
> ```
> 
> We can ignore the `acl log` command

Sure, I can make that change
@elena-kolevska You good with this?

monkey92t:
Here is my point of view:

1. `acl log` is not a commonly used command.
2. `acl log` is an abbreviation of `acl log 10` and does not have any other meaning.
3. `ACLLog(ctx, count ...int64)` can mislead people into thinking that multiple count parameters can be passed.

monkey92t (dismissed):
I think this is good, and I look forward to hearing others' opinions.

elena-kolevska:
Looks good to me too. I would just add a doc comment saying that a reasonable value for `count` is 10, and that's the Redis default value.

monkey92t (approved):
good~

PR #2540, opened 2023-04-05
doc: pass ctx in BLpop example for infinite wait time

pranitbauva1997:
The current example will not compile.


SoulPancake (approved):
Good catch

PR #2541, opened 2023-04-05
actions: build on go1.20.x

pranitbauva1997:
go1.20 was released on 1st Feb 2023, and has good adoption thus better to test against this as well.

If the build cleans, should the go1.18.x be removed or not?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


SoulPancake:
I think there are other roadmap items and migrations to take care of ( https://github.com/redis/go-redis/issues/2472 )
before proceeding on this.
@monkey92t thoughts?

monkey92t:
OK, we can add 1.20, but 1.18 I suggest to keep it for now, many people still use it.

monkey92t (approved):

PR #2551, opened 2023-04-18
feat: Add support for new redis command CLIENT NO-TOUCH

ktsivkov:
Issue: https://github.com/redis/go-redis/issues/2519

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -407,6 +407,7 @@ type Cmdable interface {
 	ClientID(ctx context.Context) *IntCmd
 	ClientUnblock(ctx context.Context, id int64) *IntCmd
 	ClientUnblockWithError(ctx context.Context, id int64) *IntCmd
+	ClientNoTouch(ctx context.Context, value bool) *BoolCmd
 	ConfigGet(ctx context.Context, parameter string) *MapStringStringCmd
 	ConfigResetStat(ctx context.Context) *StatusCmd
 	ConfigSet(ctx context.Context, parameter, value string) *StatusCmd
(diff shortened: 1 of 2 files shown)


ktsivkov:
@monkey92t hey, could you please review this one?

monkey92t:
Based on my understanding, this is a client state, not a regular command. It labels a network connection, and we should figure out how to handle similar commands.

ktsivkov:
Some brainstorming, based on my limited understanding we need to..

- be able to "tag" connections (multiple conn* handle pool and cluster?)
i.e.
```go
type baseClient struct {
	...
	connPool pool.Pooler // This here
```
- after executing (such labaling commands), we need to send all followup commands to this exact connection until they are disabled again
- find a way to handle connection drops (if a labeled connection is lost, we need to be able to handle it)

Needs some more thought but what comes to my mind on the fly is either one of those...
- We could use some kind of a Builder interface
```go
// ClientNoTouch /or any other labeling command/ Creates and tags a new connection and returns it for use
cl := redis.ClientNoTouch()
cl.Get(...)
cl.Close()
```
- We could use sync.map to create a map of the connections and their labels, such as `client no touch`
- We could introduce / or re-use / a feature to group connections that are initiated with such pre-executed commands `redis.CreateGroup(ctx, &GroupConfig{ClientNoTouch:true})`

ndyakov:
Hello @ktsivkov, would you like to resurrect this discussion and start over? If so, let me know. If you don't have the time or don't plan to work on this, let's close the PR.

ndyakov:
Closing for inactivity.

PR #2553, opened 2023-04-19
Add support for CLIENT SETINFO

### LINKED ISSUE #2520: Add support for new redis command CLIENT SETINFO
7.2.0 adds support for the new redis command CLIENT SETINFO. We need to add support, documented [here](https://redis.io/commands/CLIENT-SETINFO).

SoulPancake:
- Fixes #2520


SoulPancake:
@monkey92t Would we consider merging #2483  before continuing on this
As it can be useful to test this command
Or alternatively do you have a suggestion on how we can test this?

monkey92t:
CLIENT-SETINFO is similar to #2492, it represents the state of the connection, not just the command.

What do you think we should do?

SoulPancake:
According to the docs 
We should be able to test these by checking whether those version/fields are there in the client info right?

PR #2569, opened 2023-04-29
chore(deps): Updating otel/metric and fixing breaking changes

ViBiOh:
Related to the recent changes in the opentelemetry-go package https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.15.0, we cannot use directly `"go.opentelemetry.io/otel/attribute".KeyValue` but need to wrap them with the option pattern


vmihailenco (approved):

vmihailenco:
@ViBiOh thanks, but looks like `unit` package is gone but still referenced in some packages...

ViBiOh:
> @ViBiOh thanks, but looks like `unit` package is gone but still referenced in some packages...

Hi @vmihailenco , thanks for your reply. It seems that the issue is related to the uptake-go library. I made a pull-request to update it, but it seems that `otel/trace` use the `atomic.Pointer` type, which is a go1.19 feature, so it won't compile in 1.18.

vmihailenco:
@ViBiOh okay, I've just tagged uptrace-go v1.15. Could you update dependencies again and remove Go 1.18 from the build list?

vmihailenco (approved):

vmihailenco:
@ViBiOh Thanks

PR #2570, opened 2023-04-29
Enable specifying the protocol version on connection setup. Avoid CLIENT SETNAME command when using HELLO

filipecosta90:


ofekshenawa (approved):

SoulPancake (approved):

ash2k:
Hi! I wonder if this will solve the problem I described here https://github.com/redis/go-redis/discussions/2537? Thanks.

monkey92t:
I don't quite understand what problem this approach is meant to solve.

monkey92t:
I suggest adding a hook to address the related issue. If the user sets a custom initialization function, we can use it instead of the default connection initialization method.

redis.go
```diff
@@ -312,7 +314,7 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 			pipe.ReadOnly(ctx)
 		}
 
-		if c.opt.ClientName != "" {
+		if !clientNameSet && c.opt.ClientName != "" {
```
chayim, line 317: IMHO you can send this either way, and no need for the if check:

```
127.0.0.1:6379> hello 3 setname foo
1# "server" => "redis"
2# "version" => "6.2.12"
3# "proto" => (integer) 3
4# "id" => (integer) 4
5# "mode" => "standalone"
6# "role" => "master"
7# "modules" => (empty array)
127.0.0.1:6379> hello 3 setname ""
1# "server" => "redis"
2# "version" => "6.2.12"
3# "proto" => (integer) 3
4# "id" => (integer) 4
5# "mode" => "standalone"
6# "role" => "master"
7# "modules" => (empty array)
127.0.0.1:6379> 
```

WDYT @filipecosta90

PR #2573, opened 2023-05-02
Adding multiple redis versions to the matrix

chayim:


SoulPancake:
@chayim 
Can we also include something for this https://github.com/redis/go-redis/issues/2544
with this current approach?

ndyakov:
We are planing to test agains different redis versions in containers. Closing this PR.

PR #2591, opened 2023-05-12
make error message more reasonable

ethanvc:
Look at the following code snippet:
```golang
func main() {
	rdb := redis.NewClient(&redis.Options{
		Addr:     "3.3.3.3:10000", // make connection io timeout
		PoolSize: 1, // make connection pool timeout
	})
	for i := 0; i < 3; i++ {
		go get(rdb, "0")
	}
	time.Sleep(10000 * time.Hour)
}

func get(rdb *redis.Client, key string) {
	c, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
	defer cancel()
	v := rdb.Get(c, key)
	fmt.Println(v)
}
```
output for the code snippet with this pr:
```
get 0: redis: connection pool timeout
get 0: redis: connection pool timeout
get 0: dial tcp 3.3.3.3:10000: i/o timeout
```
When QPS become very high, rdb.Get may return error with the old logic:
`context deadline exceeded`
which is very confusing and hard to lookup the real problem.

In this pr, i referenced the implementation of context.WithTimeout, to use the smaller duration of configed timeout and timeout in current context. by doing this, the error will be `redis: connection pool timeout`, so users will know why timeout occurred.

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -297,7 +297,6 @@ func (p *ConnPool) waitTurn(ctx context.Context) error {
 	}
 
 	var realPoolTimeout time.Duration
-
 	timer := timers.Get().(*time.Timer)
 	deadline, ok := ctx.Deadline()
 	if ok {
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -300,7 +300,7 @@ func (p *ConnPool) waitTurn(ctx context.Context) error {
 	timer := timers.Get().(*time.Timer)
 	deadline, ok := ctx.Deadline()
 	if ok {
-		realPoolTimeout = deadline.Sub(time.Now())
+		realPoolTimeout = time.Until(deadline)
 		if realPoolTimeout > p.cfg.PoolTimeout {
 			realPoolTimeout = p.cfg.PoolTimeout
 		}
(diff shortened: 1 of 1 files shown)


ethanvc:
@vmihailenco please help to cr this code, this pr will be real helpful to let developer know what the timeout reason if they missed config the PoolSize parameter.

monkey92t:
I'm sorry, I didn't understand specifically what issue you are describing....

ethanvc:
> I'm sorry, I didn't understand specifically what issue you are describing....

shortly speaking, when you set timeout smaller then the PoolTimeout, redis may report `context deadline exceeded` error message. This message is not that useful.

But if redis library report `redis: connection pool timeout` error message, user will know maybe the time to increase the PoolSize.

monkey92t:
> But if redis library report `redis: connection pool timeout` error message, user will know maybe it's the time to increase the PoolSize.

I don't believe modifying the error message would be beneficial; in fact, it could potentially mislead users.

ethanvc:
@monkey92t  maybe you can reference the implementation of context.WitDeadline, the related doc is:
> WithDeadline returns a copy of the parent context with the deadline adjusted to be no later than d. **If the parent's deadline is already earlier than d, WithDeadline(parent, d) is semantically equivalent to parent**. The returned context's Done channel is closed when the deadline expires, when the returned cancel function is called, or when the parent context's Done channel is closed, whichever happens first.

In our case, return `context deadline exceeded `does not give any information about the reason why redis cost time unexpected.

And for my case, return connection pool timeout is very helpful to indicate the need to increase pool size.

Issue #2592, opened 2023-05-12
XInfoStreamFull returns nil in Groups whereas there is a group listed


a-agmon:
I have a stream with a registered consumer group. 
I can see these details when I run redis-cli 
```
127.0.0.1:6379> XINFO STREAM GROUPTESTSTREAM-XFI-00573-2023-05-12-1816 FULL
 1) "length"
 2) (integer) 1
 3) "radix-tree-keys"
 4) (integer) 1
 5) "radix-tree-nodes"
 6) (integer) 2
 7) "last-generated-id"
 8) "1683904609269-0"
 9) "max-deleted-entry-id"
10) "0-0"
11) "entries-added"
12) (integer) 1
13) "recorded-first-entry-id"
14) "1683904609269-0"
15) "entries"
16) 1) 1) "1683904609269-0"
       2) 1) "test"
          2) "test"
17) "groups"
18) 1)  1) "name"
        2) "NEWGROUP1-EXV-00337-2023-05-12-1816"
        3) "last-delivered-id"
        4) "0-0"
        5) "entries-read"
        6) (nil)
        7) "lag"
        8) (integer) 1
        9) "pel-count"
       10) (integer) 0
       11) "pending"
       12) (empty array)
       13) "consumers"
       14) (empty array)
```

However when I call the function `XInfoStreamFull` on the same stream through the API then Groups slice is nil
![image](https://github.com/redis/go-redis/assets/54080741/e28847db-0767-4a0c-9062-e020d71afec1)

How to reproduce
1. Produce message to stream SX
2. Create consumer group GX to stream SX (I specifically used XGroupCreateMkStream with SX and start 0 `client.XGroupCreateMkStream(ctx, streamKey, consumerGroup, "0").Err()`
3. call the info full function `client.XInfoStreamFull(ctx, streamKey, 0).Result()`
Groups collection returns null

SoulPancake:
@a-agmon Version of go-redis and redis?

a-agmon:
Apologies  @SoulPancake , 
the versions used are:
 Redis 7.0.4 (00000000/0) 64 bit
github.com/redis/go-redis/v9 v9.0.4

Also updated in the issue. 
I should also say that by updating the latest versions of both, the problem persists ( there are no groups in the collection which is nil, and also err is Redis.Nil

SoulPancake:
> also err is Redis.Nil

[...]

PR #2595, opened 2023-05-14
fix the reading of the "entries-read" field in XInfoStreamFull

monkey92t:
#2592


command.go
```diff
@@ -2373,7 +2373,7 @@ func readStreamGroups(rd *proto.Reader) ([]XInfoStreamGroup, error) {
 				}
 			case "entries-read":
 				group.EntriesRead, err = rd.ReadInt()
-				if err != nil {
+				if err != nil && err != Nil {
```
SoulPancake, line 2376: @monkey92t Shouldn't this be redis.Nil?

monkey92t: > @monkey92t Shouldn't this be redis.Nil?

??????  Your question is quite strange...

We are modifying the redis package itself, so there is no need to use `redis.Nil`, we can directly use `Nil`.
For example, in the io package, we don't need to write `io.Reader`.

SoulPancake: Thank you for your explanation, @monkey92t. I understand now, we are indeed within the context of the Redis package itself, so it does make sense to use Nil directly. I appreciate your patience and for correcting my misunderstanding. I'm still learning and your clarification really helps. 
I was confused as I had previously seen redis.Nil in the tests and now I realise that they are in a different package.

monkey92t: @SoulPancake  Our test code belongs to the `redis_test` package. If you look at `commands_test.go`, you should be able to see it. Therefore, in the test files, we need to write `redis.Nil`.

SoulPancake (approved):

vmihailenco (approved):

PR #2604, opened 2023-05-22
Add RunLoad to script

simao:
Add new `RunLoad` method to Script. It tries to run the script using `EvalSha`, if that fails, it loads the script using SCRIPTLOAD and then tries again.

This is what the redis manual recommends libraries to do:

From https://redis.io/docs/manual/programmability/eval-intro/

> In this case, the application should first load it with SCRIPT LOAD
> and then call EVALSHA once more to run the cached script by its SHA1
> sum. Most of Redis' clients already provide utility APIs for doing
> that automatically. Please consult your client's documentation
> regarding the specific details

Please let me know if you think this should be merged, I will write some tests.

### DIFF
--- a/script.go
+++ b/script.go
@@ -63,6 +63,21 @@ func (s *Script) EvalShaRO(ctx context.Context, c Scripter, keys []string, args 
 	return c.EvalShaRO(ctx, s.hash, keys, args...)
 }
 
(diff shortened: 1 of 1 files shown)


ofekshenawa:
Hey @simao, Before we merge, could you write write unit tests?

PR #2607, opened 2023-05-23
chore(deps): Update otel/metric to stable API

ash2k:
Pick up https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.16.0.

The example also needs to be updated, but uptrace-go needs to be updated first.

Closes https://github.com/redis/go-redis/pull/2578.


vmihailenco:
Thanks! uptrace-go is blocked by https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3874

pellared:
Are you not using uptrace-go only in [OTel example](https://github.com/redis/go-redis/blob/master/example/otel/client.go)?

Personally, I would suggest moving the example here: https://github.com/uptrace/uptrace-go/tree/master/example and refactor the existing example to use only the official "vanilla" OTel Go components. 

PS. I am not a maintainer here. I just noticed this PR via https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3874 and I hope that this idea may help you.

vmihailenco:
@ash2k uptrace-go is updated - could you update this PR?

>Personally, I would suggest moving the example here

We had lots of examples in uptrace-go, but it was hard to update them and it resulted in a lot of noise in the repo for uptrace-go users.

Besides, the example here is in a separate module so it does not affect the redis client.

ash2k:
@vmihailenco PTAL

vmihailenco (approved):

ash2k:
@vmihailenco Any plans to release a new version soon? :)

vmihailenco:
@ash2k https://github.com/redis/go-redis/pull/2608

Issue #2563, opened 2023-04-26
Structured Loggers


davidsonff:
Would it be possible to enhance SetLogger to include some structured loggers such as https://github.com/uber-go/zap? Currently, it fails the interface requirements.

SoulPancake:
What do you think about possibly using the experimental package slog as it is supported by the official go dev team since last year?

davidsonff:
Sure! I think that zap and logrus are super popular, though?

PR #2609, opened 2023-05-25
feat: Adding Structured logger support

SoulPancake:
Resolves - #2563 
This is a brief idea/starter regarding the support of structured Loggers ( Zap in this example )
We can discuss the API or whether we want to do this at all in our case


monkey92t:
We should add some suitable logging interfaces instead of adapting specific logging systems, as there are many such logging systems. We can provide an implementation of the ZAP API that supports the go-redis logging interface.

SoulPancake:
@monkey92t 
My implementation includes an updated Logging interface with multiple methods that allow for structured logging at different severity levels which I find to be a standar. I have also created a ZapLogger on the root level that implements this new interface using the ZAP logging library.

In order to maintain backward compatibility, I have kept the Printf method in the Logging interface.
What are your thoughts about this ?

SoulPancake:
> We can provide an implementation of the ZAP API that supports the go-redis logging interface.
Did you mean that we need to provide it as an example or as an API that can be used?

monkey92t:
I noticed that you've designed a Logging interface, but I don't think it's suitable for go-redis. At present, it seems that go-redis doesn't require that many levels of log recording.

Also, concerning the ZAP API, the associated dependencies should not be in go-redis's go.mod, as it increases the dependency on external libraries. Most users do not need it, yet they pay the cost of downloading the ZAP library.

monkey92t:
When designing interfaces, we only design those that are currently needed or might be needed in the future. Adding those that are clearly unnecessary doesn't make sense.

Please note, we are not adapting to the ZAP interface, but rather ZAP is adapting to go-redis. As a mature logging module, ZAP should have multi-level log handling, but go-redis doesn't need something that complex.

SoulPancake:
Makes sense, In that case do you suggest we provide just an example of how it is to be used with Zap?
And continue to use the same interface?  @monkey92t

vmihailenco:
@SoulPancake I wonder if you've seen log/slog package that is coming in Go 1.21. I have not checked it yet, but I suspect we should use that as a starting point.

Also github.com/go-logr/logr is what most libs are using now.

SoulPancake:
Ofcourse @vmihailenco That was my [initial suggestion](https://github.com/redis/go-redis/issues/2563#issuecomment-1537182429) 
We could use that but it is still relatively experimental in nature.
Let me take a look into github.com/go-logr/logr for the time being.

ndyakov:
@SoulPancake closing this one as stale.

PR #2623, opened 2023-06-09
Set correct cluster slot for scan commands, similarly to Java's Jedis client

pete-woods:
- At present, the scan command is dispatched to a random slot.
- It seems better if we dispatched the scanX family of commands to a `match`-seeded slot.
- See: https://github.com/redis/jedis/blob/869dc0bb6625b85c8bf15bf1361bde485a304338/src/main/java/redis/clients/jedis/ShardedCommandObjects.java#L101


pete-woods:
Hi @vmihailenco. Is there any chance you could have a look over this, and let me know if this is a change you'd be interested in? Thanks!

emilywoods:
Hi @vmihailenco. Just following up on this pr. Is it a change that you would consider?

pete-woods:
@chayim Hi! This is a change we've been carrying in a fork for about 6 months now.

Do you have any time to have a look over and see if this is something this project would be willing to accept?

The idea is basically lifted from the Redis clients for other languages.

axilis-marko:
Would it be an option for scan to implicitly run over all shards if no hash-tag is included?

pete-woods:
The Java clients actually error if you don't provide a hash tag.

I kinda think that behaviour is better, as it avoids the surprising result of your scan not really working at all, but it would be a breaking change for this library.

ndyakov:
Hello @pete-woods , feel free to sync with master and we can check this.

pete-woods:
Okay - did a quick rebase

pete-woods:
@ndyakov please let me know what you think, now the pr is up to date

ndyakov:
@pete-woods Looks good, will double check with Jedis and see if we can include this in the next release, thank you.

ndyakov:
After couple of discussions I don't see how this can break the current behavior, it can only improve it for when there is a hashtag in the pattern. Thank you for the contribution!

pete-woods:
Great stuff. I'd also consider erroring for cluster mode, when there's no tag set like the Java clients do, but that's a separate decision.

PR #2625, opened 2023-06-10
feat: OnDisconnect hook to track server disconnections

Ja7ad:
example:

```go
package main

import (
	"context"
	"log"
	"time"
)

func main() {
	cancelCh := make(chan struct{})

		Addr:         "localhost:6379",
		Username:     "foo",
		Password:     "bar",
		DB:           0,
		OnDisconnect: redis.DefaultOnDisconnect(5*time.Second, cancelCh),
	})

	if err := client.Ping(context.Background()); err != nil {
		log.Fatalln(err)
	}

}

```


vmihailenco:
This looks like some kind of connection monitor, not a disconnect hook.

Not sure if we need one. Instead you can provide a custom dialer and override `net.Conn.Close` method.

Issue #2068, opened 2022-04-06
support for redis stack


LiHeng:
Does go-redis plan to support redis stack?

shra1dhar:
Hi mate, good day
Any word on this. Or are we open to accepting PRs on this?

chayim:
Hi @shra1dhar there's currently a Bloom PR that's near ready. Coming soon is JSON + Search. TImeseries is something that @ofekshenawa was going to start on shortly. Perhaps you two would like to collaborate?

Fyzbyt:
any news on this?

yanikitat:
@ofekshenawa Hi! Do you already have a PR for Redis Stack support? Can you share this PR here?

parkerroan:
Would love to track the progress if possible. Any update on the search support? 

Looks like redisearch-go is not actively maintained so this would be a great addition: 
https://github.com/RediSearch/redisearch-go/issues/187

ndyakov:
Newer go-Redis versions are supporting Redis CE

PR #2649, opened 2023-07-18
Bloom support

elena-kolevska:
part of  #2068


probabilistic.go
```diff
@@ -0,0 +1,1142 @@
...
+
+type BFInfoArgs int
+
+const (
+	BFCAPACITY BFInfoArgs = iota
+	BFSIZE
+	BFFILTERS
+	BFITEMS
+	BFEXPANSION
+)
```
vmihailenco, line 105: What about replacing these constants with methods (and removing constants):

```
BFInfoCapacity()
BFInfoSize()
...
BFInfoExpansion()
```

It is the same amount of work / API surface, but methods are more readable / easier to use.

probabilistic.go
```diff
@@ -0,0 +1,1142 @@
+package redis
+
+import (
+	"context"
+	"fmt"
+	"github.com/redis/go-redis/v9/internal/proto"
+)
+
+type ProbabilisticCmdable interface {
```
vmihailenco, line 9: Let's unexport this unless there is a reason to keep this public?

probabilistic.go
```diff
@@ -0,0 +1,1142 @@
...
+	BFAdd(ctx context.Context, key, element interface{}) *BoolCmd
+	BFCard(ctx context.Context, key string) *IntCmd
+	BFExists(ctx context.Context, key, element interface{}) *BoolCmd
```
vmihailenco, line 12: `key string`?

probabilistic.go
```diff
@@ -0,0 +1,1142 @@
...
+
+	TOPKAdd(ctx context.Context, key string, elements ...interface{}) *StringSliceCmd
+	TOPKCount(ctx context.Context, key string, elements ...interface{}) *IntSliceCmd
+	TOPKIncrBy(ctx context.Context, key string, elements ...interface{}) *StringSliceCmd
+	TOPKInfo(ctx context.Context, key string) *TOPKInfoCmd
+	TOPKList(ctx context.Context, key string) *StringSliceCmd
+	TOPKListWithCount(ctx context.Context, key string) *MapStringIntCmd
+	TOPKQuery(ctx context.Context, key string, elements ...interface{}) *BoolSliceCmd
+	TOPKReserve(ctx context.Context, key string, k int) *StatusCmd
+	TOPKReserveWithOptions(ctx context.Context, key string, k int, width, depth int64, decay float64) *StatusCmd
```
vmihailenco, line 52: `TopKAdd`, `TopKCount` etc?

vmihailenco:
You probably want to use `int64/uint64` instead of `int/uint` everywhere, because `int` is either `int32` or `int64` depending on the architecture and I don't think that behavior is useful in the client.

probabilistic.go
elena-kolevska: Ah, missed this when I converted "element" to interface (it used to be a string). Thanks for catching it.

probabilistic.go
elena-kolevska: Yeah, that actually makes sense since "Top" is not an abbreviation. Good point. Will change it.

probabilistic.go
elena-kolevska: The more I look at this, the more I think Nic did a nice thing with the constants because the IDE makes the needed arguments very easily discoverable.
If we keep a `BFInfoArgs` type with `Capacity()`, `Size()` etc methods, we would have to instantiate an empty object in code, when calling `client.BFInfoArg`. Not terrible, but also not the prettiest. And since the parameter then would be a string, the users might not know we provide the helper methods above and might pass in a hard-coded string.
Does this sound about right @vmihailenco , or I missed your point?

vmihailenco: Instead of this:

```go
rdb.BFInfoArg(ctx, key, redis.BFCAPACITY)
```

User will write this:

```go
rdb.BFInfoCapactiy(ctx, key)
```

The implementation will look like this:

```go
func (c) BFInfoCapacity(ctx context.Context, key string) {
    return c.bfInfo(ctx, key, "capacity")
}
```
Internally, we can keep the constants and `BFInfoArg` method, but unexport them so users can't (mis)use them.

elena-kolevska: Ahh ok! So I had misunderstood you before :) 
Let me work in it...

elena-kolevska: Done!

elena-kolevska:
The tests are not passing because the code needs Redis Stack 7.2 RC3 or above. @chayim is working on a way to automate tests with a Redis Stack option.

probabilistic.go
```diff
@@ -0,0 +1,1134 @@
...
+	BFInfoFilters(ctx context.Context, key string) *BFInfoCmd
+	BFInfoItems(ctx context.Context, key string) *BFInfoCmd
+	BFInfoExpansion(ctx context.Context, key string) *BFInfoCmd
+	BFInsert(ctx context.Context, key string, options *BFInsertOptions, elements ...interface{}) *BoolSliceCmd
+	BFMAdd(ctx context.Context, key string, elements ...interface{}) *BoolSliceCmd
+	BFMExists(ctx context.Context, key string, elements ...interface{}) *BoolSliceCmd
+	BFReserve(ctx context.Context, key string, errorRate float64, capacity int64) *StatusCmd
+	BFReserveExpansion(ctx context.Context, key string, errorRate float64, capacity, expansion int64) *StatusCmd
+	BFReserveNonScaling(ctx context.Context, key string, errorRate float64, capacity int64) *StatusCmd
+	BFReserveArgs(ctx context.Context, key string, options *BFReserveOptions) *StatusCmd
```
vmihailenco, line 25: I know that at this point the `Args` suffix looks like some kind of convention, but it never meant to be so and we don't really enforce it. So feel free to come with a better name :)

`BFReserveExt` or `BFReserveArgsOptions` or `BFReserveWithOptions` might be it, but I don't feel strongly about it.

vmihailenco (approved):

probabilistic.go
```diff
@@ -0,0 +1,1134 @@
+package redis
+
+import (
+	"context"
+	"fmt"
+	"github.com/redis/go-redis/v9/internal/proto"
+)
+
+type probabilisticCmdable interface {
```
mcauto, line 9: Why not public interface?

elena-kolevska:
I'm closing this, since the commits went in another pr that's already merged. Thanks @ofekshenawa  for working on it.

Issue #2561, opened 2023-04-25
"dialHook"  doesn't work in some case


hesining:
Issue tracker is used for reporting bugs and discussing new features. Please use
[stackoverflow](https://stackoverflow.com) for supporting issues.

## Expected Behavior

"dialHook" method can work 

the test code can print  "aaaaa" and "bbbbb"

## Current Behavior

it doesn't work when use connection pool and call "WithTimeout" before "AddHook"

the test code can only print "aaaaa"

## Possible Solution
 maybe we can call the "newConnPool" method again after clone the client

```go
func NewClient(opt *Options) *Client {
	//...
	c.connPool = newConnPool(opt, c.dialHook)
	return &c
}

func (c *Client) WithTimeout(timeout time.Duration) *Client {
	clone := *c
	clone.baseClient = c.baseClient.withTimeout(timeout)
	clone.init()
	return &clone
}
```

## Steps to Reproduce

1. create test file and run it
```go
package redis_test

import (
	"context"
	"fmt"
	"github.com/redis/go-redis/v9"
	"net"
	"testing"
	"time"
)

func Test_Redis(t *testing.T) {
	c := redis.NewClient(&redis.Options{
		Addr:       "xxx",
		Password:   "111",
		PoolSize:   222,
		DB:         1,
		MaxRetries: -1,
	}).WithTimeout(time.Second * 5)

	clusterHook := &hook{
		dialHook: func(hook redis.DialHook) redis.DialHook {
			fmt.Println("aaaaaaa")
			return func(ctx context.Context, network, addr string) (n net.Conn, e error) {
				fmt.Println("bbbbb")
				n, e = hook(ctx, network, addr)
				return
			}
		},
	}
	c.AddHook(clusterHook)

	c.WithTimeout(time.Second * 5)

	c.Keys(context.Background(), "xx")
}

type hook struct {
	dialHook            func(hook redis.DialHook) redis.DialHook
	processHook         func(hook redis.ProcessHook) redis.ProcessHook
	processPipelineHook func(hook redis.ProcessPipelineHook) redis.ProcessPipelineHook
}

func (h *hook) DialHook(hook redis.DialHook) redis.DialHook {
	if h.dialHook != nil {
		return h.dialHook(hook)
	}
	return hook
}

func (h *hook) ProcessHook(hook redis.ProcessHook) redis.ProcessHook {
	if h.processHook != nil {
		return h.processHook(hook)
	}
	return hook
}

[...]

PR #2655, opened 2023-07-20
withTimeout create NewConnPool

taytzehao:
Created a new Connection pool for the cloned client. The copy performed by the `clone := *c` operation is just a shallow copy. This causes the ConnPool in the clone to be referring to the ConnPool in the old Client. When a new hook is added to the client, the hook is added to the newly cloned hook rather than the old one. 

Added unit test case.

#2561


ndyakov (approved):
@taytzehao initially i was thinking this makes sense, but after more investigation I am wondering if this was not by design. Needs deeper investigation.

ndyakov:
Well @taytzehao I do think this is by design to reuse the same pool, don't you think? Closing this PR, but feel free to reopen.

PR #2668, opened 2023-08-09
Add SCANDUMP and LOADCHUNK for bf and cf

ofekshenawa:


probabilistic.go
```diff
@@ -165,6 +169,79 @@ func (c cmdable) BFExists(ctx context.Context, key string, element interface{})
 	return cmd
 }
 
+func (c cmdable) BFLoadChunk(ctx context.Context, key string, iterator int64, data interface{}) *StatusCmd {
+	args := []interface{}{"bf.loadchunk", key, iterator, data}
```
chayim, line 173: Given history (and nothing else) commands really should be all in CAPS everywhere. Yes, we're not doing that currently, but no time like the present.

probabilistic.go
```diff
@@ -165,6 +169,79 @@ func (c cmdable) BFExists(ctx context.Context, key string, element interface{})
...
+	cmd := NewStatusCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+func (c cmdable) BFScanDump(ctx context.Context, key string, iterator int64) *ScanDumpCmd {
```
chayim, line 179: WDYT let's start doing what we do in other clients - add some boilerplate to the function explaining what it is and does - link to the redis.io docs?

chayim (changes requested):
Great first PR. Nice and small - the comments are nits, code-wise makes sense and looks in line. I see this as starting things off to be like some of the others.

chayim (approved):

PR #2669, opened 2023-08-09
Updating client license to clear, MIT

chayim:
This pull request exists to unify all of our clients until the MIT license. In the case of some repositories this is more permissive, than the prior BSD-3-Clause.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 MIT License
 
-Copyright (c) 2022-2023, Redis, inc.
+Copyright (c) 2013-2023 The github.com/redis/go-redis Authors.
+All rights reserved.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,25 +1,21 @@
-Copyright (c) 2013 The github.com/redis/go-redis Authors.
-All rights reserved.
+MIT License
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are
-met:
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,25 +1,21 @@
-Copyright (c) 2013 The github.com/redis/go-redis Authors.
-All rights reserved.
+MIT License
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are
-met:
+Copyright (c) 2022-2023, Redis, inc.
```
vmihailenco, line 3: Really?

vmihailenco:
You can't "just" change the license while also transferring the copyright.

LICENSE
chayim: @uglide ? I was just lining things up... Someone help me - I am opinion free here - this is just a copy/paste from redis-py

uglide: @chayim we shouldn't change the copyright. The aim is to align all clients under Redis org to the same MIT license. 

@vmihailenco you are the copyright owner for most of the code. Is it something that you can consider? Both licenses are permissive, true-OSS.

uglide (changes requested):
Please revert changes to copyright

chayim:
> Please revert changes to copyright

Got it - thanks. Live + learn. Question @vmihailenco @uglide one of the things we discussed was _"Copyright, the maintainers of this library"_ is that reasonable? Am I overthinking all of this?

LICENSE
vmihailenco: As far as I am concerned the MIT license is fine, but the copyright should stay the same.

"The Authors" means the people who contributed the code (which includes Redis, inc.). Maintainers change with time and don't have anything to do with it.

PR #2680, opened 2023-08-22
Copy client hooks to child connections.

Crystalix007:
Enables parent hooks on child connections to allow mocking to hook into the commands run on those connections.

Also fixes a flapping test on Redis Gears, caused by an assumption in the order of returned functions.


ndyakov:
Hey @Crystalix007 can you check the implementation that is part of https://github.com/redis/go-redis/pull/3320 . I think it covers your usecase.

PR #2684, opened 2023-08-29
Update command.go

liuqian1990:
https://redis.io/commands/slowlog-get/

The unit of data obtained by slowlog get itself is already us, no further conversion is required


ofekshenawa:
Hey,
Thank you for your contribution and for bringing this to our attention. However, according to the Redis command reference [here](https://redis.io/commands/slowlog-get/), the `slowlog get` command returns the execution time in microseconds, not nanoseconds. 
The documentation specifically states: "The amount of time needed for its execution, in microseconds.

WDYT?

PR #2687, opened 2023-09-01
Add MigrateWithOptions to handle multiple keys and copy/replace

aidandj:
I wasn't sure the best way to handle adding this. Ideally in a breaking change, but I saw `RestoreReplace` existed, so this was kind of along those lines.

Happy to rename/re-arrange the interface, just wanted to get something started.


ndyakov:
Hello @artificial-aidan and thank you for this PR. You were on the right track. Let's rename it to `MigrateWithArgs` (see https://github.com/redis/go-redis/blob/7b9bd6c36e3696e619a9fc5880d5fff002ed32a0/search_commands.go#L15 for example) and create `options` struct for the additional options - similar to `FTAggregateOptions` 

And please resolve conflicts.

aidandj:
Not stale

PR #2688, opened 2023-09-03
Add Redis Timeseries support

ofekshenawa:
This pull request adds support for [RedisTimeSeries](https://github.com/redistimeseries/redistimeseries). As part of this move forward, we discovered a better naming convention for probabalistic commands - one in line with the standard library.

Proposal: this becomes part of 9.2.0, due to the breaking change.   

I realize that this may be contentious - but given the relative newness of Probabilistic Filters, I'd rather not wait for v10.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1398,12 +1398,12 @@ func (cmd *MapStringSliceInterfaceCmd) readReply(rd *proto.Reader) (err error) {
 			return err
 		}
 		cmd.val[k] = make([]interface{}, nn)
-		for i := 0; i < nn; i++ {
+		for j := 0; j < nn; j++ {
 			value, err := rd.ReadReply()
 			if err != nil {
 				return err
 			}
-			cmd.val[k][i] = value
+			cmd.val[k][j] = value
 		}
 	}
 
--- a/command.go
+++ b/command.go
@@ -1351,6 +1351,65 @@ func (cmd *MapStringIntCmd) readReply(rd *proto.Reader) error {
 	return nil
 }
 
+// ------------------------------------------------------------------------------
+type MapStringSliceInterfaceCmd struct {
+	baseCmd
+	val map[string][]interface{}
+}
+
+func NewMapStringSliceInterfaceCmd(ctx context.Context, args ...interface{}) *MapStringSliceInterfaceCmd {
+	return &MapStringSliceInterfaceCmd{
+		baseCmd: baseCmd{
+			ctx:  ctx,
+			args: args,
+		},
+	}
+}
+
+func (cmd *MapStringSliceInterfaceCmd) String() string {
+	return cmdString(cmd, cmd.val)
+}
+
+func (cmd *MapStringSliceInterfaceCmd) SetVal(val map[string][]interface{}) {
+	cmd.val = val
+}
+
+func (cmd *MapStringSliceInterfaceCmd) Result() (map[string][]interface{}, error) {
+	return cmd.val, cmd.err
+}
+
+func (cmd *MapStringSliceInterfaceCmd) Val() map[string][]interface{} {
+	return cmd.val
+}
+
+func (cmd *MapStringSliceInterfaceCmd) readReply(rd *proto.Reader) (err error) {
+	n, err := rd.ReadMapLen()
+	if err != nil {
+		return err
+	}
+	cmd.val = make(map[string][]interface{}, n)
+	for i := 0; i < n; i++ {
+		k, err := rd.ReadString()
+		if err != nil {
+			return err
+		}
+		nn, err := rd.ReadArrayLen()
+		if err != nil {
+			return err
+		}
+		cmd.val[k] = make([]interface{}, nn)
+		for i := 0; i < nn; i++ {
+			value, err := rd.ReadReply()
+			if err != nil {
+				return err
+			}
+			cmd.val[k][i] = value
+		}
+	}
+
+	return nil
+}
+
 //------------------------------------------------------------------------------
 
 type StringStructMapCmd struct {
--- a/probabilistic.go
+++ b/probabilistic.go
@@ -24,7 +24,7 @@ type probabilisticCmdable interface {
 	BFReserve(ctx context.Context, key string, errorRate float64, capacity int64) *StatusCmd
 	BFReserveExpansion(ctx context.Context, key string, errorRate float64, capacity, expansion int64) *StatusCmd
 	BFReserveNonScaling(ctx context.Context, key string, errorRate float64, capacity int64) *StatusCmd
-	BFReserveWithArgs(ctx context.Context, key string, options *BFReserveOptions) *StatusCmd
+	BFReserveArgs(ctx context.Context, key string, options *BFReserveOptions) *StatusCmd
 	BFScanDump(ctx context.Context, key string, iterator int64) *ScanDumpCmd
 	BFLoadChunk(ctx context.Context, key string, iterator int64, data interface{}) *StatusCmd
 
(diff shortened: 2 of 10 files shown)


command.go
```diff
@@ -1351,6 +1351,65 @@ func (cmd *MapStringIntCmd) readReply(rd *proto.Reader) error {
...
+		k, err := rd.ReadString()
+		if err != nil {
+			return err
+		}
+		nn, err := rd.ReadArrayLen()
+		if err != nil {
+			return err
+		}
+		cmd.val[k] = make([]interface{}, nn)
+		for i := 0; i < nn; i++ {
```
chayim, line 1401: Used i within the sub loop. IMHO another var?

command.go
```diff
@@ -1351,6 +1351,65 @@ func (cmd *MapStringIntCmd) readReply(rd *proto.Reader) error {
...
+	for i := 0; i < n; i++ {
+		k, err := rd.ReadString()
+		if err != nil {
+			return err
+		}
+		nn, err := rd.ReadArrayLen()
+		if err != nil {
+			return err
+		}
+		cmd.val[k] = make([]interface{}, nn)
```
chayim, line 1400: Definitely cheap then calling append over and over again - nice.

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+	cmd := NewIntCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// TSAdd - Adds one or more observations to a t-digest sketch.
+// This function also allows for specifying additional options such as:
+// Retention, ChunkSize, Encoding, DuplicatePolicy and Labels.
+// For more information - https://redis.io/commands/ts.add/
+func (c cmdable) TSAddArgs(ctx context.Context, key string, timestamp interface{}, value float64, options *TSOptions) *IntCmd {
```
chayim, line 151: In general, why didn't you go with integer for timestamp? It's a unix time millis - though it can also be a '*'.

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+		if options.ChunkSize != 0 {
+			args = append(args, "CHUNK_SIZE", options.ChunkSize)
+
+		}
+		if options.Encoding != "" {
+			args = append(args, "ENCODING", options.Encoding)
+		}
+
+		if options.DuplicatePolicy != "" {
+			args = append(args, "DUPLICATE_POLICY", options.DuplicatePolicy)
```
chayim, line 166: I think this should be [ON_DUPLICATE](https://redis.io/commands/ts.add/).

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+		}
+		if options.Labels != nil {
+			args = append(args, "LABELS")
+			for label, value := range options.Labels {
+				args = append(args, label, value)
+
```
chayim, line 172: extra space

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+		}
+	}
+	cmd := NewIntCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// TSCreate - Creates a new time-series key.
+// For more information - https://redis.io/commands/ts.create/
+func (c cmdable) TSCreate(ctx context.Context, key string) *StatusCmd {
```
chayim, line 183: Understood - standalone, and a with args

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+	cmd := NewStatusCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// TSAlterArgs - Alters an existing time-series key with additional options.
+// This function allows for specifying additional options such as:
+// Retention, ChunkSize and DuplicatePolicy.
+// For more information - https://redis.io/commands/ts.alter/
+func (c cmdable) TSAlterArgs(ctx context.Context, key string, options *TSAlterOptions) *StatusCmd {
```
chayim, line 228: I might call this TSAlter directly. Yes, it takes args - but there's no way to trigger an argumentless alter.

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+	args := []interface{}{"TS.ALTER", key}
+	if options != nil {
+		if options.Retention != 0 {
```
chayim, line 231: Given how often we use these same blocks - can we cut it down to a helper function returning the arg, and then call repeatedly WDYT?

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+		}
+	}
+	cmd := NewStatusCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// TSCreateRule - Creates a compaction rule from sourceKey to destKey.
+// For more information - https://redis.io/commands/ts.createrule/
+func (c cmdable) TSCreateRule(ctx context.Context, sourceKey string, destKey string, aggregator string, bucketDuration int) *StatusCmd {
```
chayim, line 256: does it make more sense, in the library to be a supported enum instead - using the [iota enum pattern](https://golangbyexample.com/iota-in-golang/)? You'd have to ensure those are available, but it's IMHO better than a string with docs. WDYT?

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+	cmd := NewStatusCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// TSCreateRuleArgs - Creates a compaction rule from sourceKey to destKey with additional option.
+// This function allows for specifying additional option such as:
+// alignTimestamp.
+// For more information - https://redis.io/commands/ts.createrule/
+func (c cmdable) TSCreateRuleArgs(ctx context.Context, sourceKey string, destKey string, aggregator string, bucketDuration int, options *TSCreateRuleOptions) *StatusCmd {
```
chayim, line 267: Same as above

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+	cmd := NewIntCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// TSIncrByArgs - Increments the value of a time-series key by the specified addend with additional options.
+// This function allows for specifying additional options such as:
+// Timestamp, Retention, ChunkSize, Uncompressed and Labels.
+// For more information - https://redis.io/commands/ts.incrby/
+func (c cmdable) TSIncrByArgs(ctx context.Context, key string, addend float64, options *TSIncrDecrOptions) *IntCmd {
```
chayim, line 292: Maybe instead, to match the standard library this/these should be renamed:
TSIncrByWithArgs

I realize we should probably rename Bloom in probabalistic as a result. Yes, breaking, WDYT.

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+		}
+	}
+	cmd := NewIntCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// TSDecrBy - Decrements the value of a time-series key by the specified subtrahend.
+// For more information - https://redis.io/commands/ts.decrby/
+func (c cmdable) TSDecrBy(ctx context.Context, Key string, subtrahend float64) *IntCmd {
```
chayim, line 323: I know - the docs call this variable subtrahend (and addend in TS.INCRBY), IMHO this isn't clear.  Maybe we should just use 'timestamp' as a variable name.

redis_timeseries.go
```diff
@@ -0,0 +1,877 @@
...
+
+func (cmd *TSTimestampValueCmd) Result() (TSTimestampValue, error) {
+	return cmd.val, cmd.err
+}
+
+func (cmd *TSTimestampValueCmd) Val() TSTimestampValue {
+	return cmd.val
+}
+
+func (cmd *TSTimestampValueCmd) readReply(rd *proto.Reader) (err error) {
```
chayim, line 440: confused by the multiple.

redis_timeseries.go
ofekshenawa: Yes, that is the exact reason. I didn’t use an integer for the timestamp because it can also be a character like *.

probabilistic.go
```diff
@@ -24,7 +24,7 @@ type probabilisticCmdable interface {
 	BFReserve(ctx context.Context, key string, errorRate float64, capacity int64) *StatusCmd
 	BFReserveExpansion(ctx context.Context, key string, errorRate float64, capacity, expansion int64) *StatusCmd
 	BFReserveNonScaling(ctx context.Context, key string, errorRate float64, capacity int64) *StatusCmd
-	BFReserveArgs(ctx context.Context, key string, options *BFReserveOptions) *StatusCmd
+	BFReserveWithArgs(ctx context.Context, key string, options *BFReserveOptions) *StatusCmd
```
chayim, line 27: IMHO let's pull all bloom things into a separate PR please.

redis_timeseries.go
ofekshenawa: Thank you for your suggestion. I tried to create a general helper function, but I found that the use of the general function is slower by 0.00248% when I tested the execution time of using the general function and not using it. However, there is a problem to implement a general function when the expected arguments are maps and slices because of type consistency.

chayim (dismissed):

chayim (approved):

PR #2695, opened 2023-09-09
change interfaces probabilistic and redis gears from private to public

peczenyj:
Today, if we look into `UniversalClient` interface documentation,

https://pkg.go.dev/github.com/redis/go-redis/v9#UniversalClient

it includes the `Cmdable` interface 

https://pkg.go.dev/github.com/redis/go-redis/v9#Cmdable

however we have no mention or link to the methods defined in gears or probabilistic interfaces because such interfaces are private.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -505,8 +505,8 @@ type Cmdable interface {
 
 	ModuleLoadex(ctx context.Context, conf *ModuleLoadexConfig) *StringCmd
 
-	gearsCmdable
-	probabilisticCmdable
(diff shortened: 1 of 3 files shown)


peczenyj:
No idea why go 1.20.x tests are failing

chayim:
Cross linking #2701 for visibility, since they're tied together.

chayim (dismissed):

ofekshenawa (approved):

PR #2701, opened 2023-09-10
Changing the suffix for probablistic commands acceptings arguments to WithArgs from Args

ofekshenawa:


chayim:
Given #2695 I think it may actually be fair to version this as 9.1.1 rather than 9.2.0. I'd argue that with the interface not being public (and for 3 weeks) I can recount what I said. 

Yes, the functions themselves are renamed. 

@peczenyj since you've provided the first/only feedback about support here.. WDYT?

chayim:
@ofekshenawa note this is approved. The only question is around versions. If we release today and don't have input, I'll lean into 9.2.0. I'll poke @uglide

chayim (approved):

peczenyj:
@chayim LGTM

my point on the other PR was to make the interface probabilistic public to be easier find such methods on the UniversalClient internet via godoc and package doc in web.

This PR sounds good to me

PR #2708, opened 2023-09-14
Identify client on connect

### LINKED ISSUE #2707: Update CLIENT-SETINFO to support suffixes
CLIENT SETINFO should be extended to support community members inserting their own suffixes, for Redis usage.

1. Clients should by default set-lib during connect with their name and version.
2. Clients should allow users to not send this
3. We should allow a user do send a suffix to the LIB-NAME portion - this allows us to learn about how things are used.
4. Our clients continue to send their version upstream for LIB-VER

Strings should have the client name, and an optional suffix. They should be formatted ClientName(<suffix>). For example:

1. If you're NRedisStack - LIB-NAME: NRedisStack(StackExchange.Redis-v2.6.8;.NET-6.2.18) LIB-VERSION: 0.9.1
2. If you're a project using NRedisStack (say SemtanticKernel): LIB-NAME (NRedisStack(SemanticKernel-v0.1.7;.NET-6.2.18) LIB-VERSION: 0.9.1

Considerations:
1. Users should not send brackets, or braces, and we should error accordingly.
2. We should replace spaces a user sends with hyphens

ofekshenawa:
closes #2707

### DIFF
--- a/bench_decode_test.go
+++ b/bench_decode_test.go
@@ -47,6 +47,7 @@ func NewClusterClientStub(resp []byte) *ClientStub {
 			return stub.stubConn(initHello), nil
 		},
 		DisableIndentity: true,
+
 		ClusterSlots: func(_ context.Context) ([]ClusterSlot, error) {
 			return []ClusterSlot{
 				{
--- a/bench_decode_test.go
+++ b/bench_decode_test.go
@@ -30,6 +30,7 @@ func NewClientStub(resp []byte) *ClientStub {
 		Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
 			return stub.stubConn(initHello), nil
 		},
+		DisableClientSetInfo: true,
 	})
 	return stub
 }
@@ -45,6 +46,7 @@ func NewClusterClientStub(resp []byte) *ClientStub {
 		Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
 			return stub.stubConn(initHello), nil
 		},
+		DisableClientSetInfo: true,
 		ClusterSlots: func(_ context.Context) ([]ClusterSlot, error) {
 			return []ClusterSlot{
 				{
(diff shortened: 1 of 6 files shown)


cluster.go
```diff
@@ -277,15 +278,15 @@ func (opt *ClusterOptions) clientOptions() *Options {
...
-
-		TLSConfig: opt.TLSConfig,
+		PoolFIFO:             opt.PoolFIFO,
+		PoolSize:             opt.PoolSize,
+		PoolTimeout:          opt.PoolTimeout,
+		MinIdleConns:         opt.MinIdleConns,
+		MaxIdleConns:         opt.MaxIdleConns,
+		ConnMaxIdleTime:      opt.ConnMaxIdleTime,
+		ConnMaxLifetime:      opt.ConnMaxLifetime,
+		DisableClientSetInfo: opt.DisableClientSetInfo,
```
chayim: ```DisableIndentity```

chayim (approved):

redis.go
```diff
@@ -299,7 +299,14 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 		// difficult to rely on error strings to determine all results.
 		return err
 	}
-
+	if !c.opt.DisableIndentity {
```
ljun20160606, line 302: Please move the code after line 318. When protocol is 2, the application will get NOAUTH error since AUTH has not requested.

ljun20160606, line 302: Please move the code after line 318. When protocol is 2, the application will get NOAUTH error since AUTH has not requested.

bench_decode_test.go
```diff
@@ -30,6 +30,7 @@ func NewClientStub(resp []byte) *ClientStub {
 		Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
 			return stub.stubConn(initHello), nil
 		},
+		DisableIndentity: true,
```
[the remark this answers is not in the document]
ofekshenawa: No, we don't want to send client setinfo on benchmark tests 😅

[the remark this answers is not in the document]
pdf: I suspect the question about whether this is a type might be because `Indentity` is a very uncommon word (I'm not sure it's even a generally recognised word), whereas `Identity` would be much more common.

PR #2725, opened 2023-09-21
Add stream interface back to `Cmdable`

gabrielgio:
Without it we can't call stream functions from `Cmdable`.

That broke some our code since we rely on `UniversalClient` to abstract from `ClusterClient` and  `Client`. Adding back seems to have solve the isse since those two client do implement `StreamCmdable`.


minghsu0107:
@gabrielgio  We should also add `StreamCmdable` back to `redis.Pipeliner`

ofekshenawa (approved):
LGTM

PR #2728, opened 2023-09-21
refactor(gears): remove redundant nil check

Juneezee:
From the Go specification:

> "1. For a nil slice, the number of iterations is 0." https://go.dev/ref/spec#For_range

Both `Keys` and `Arguments` are `[]string`. Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/R57hixwdkhX


SoulPancake (approved):
LGTM

ofekshenawa (approved):

PR #2746, opened 2023-10-08
use go workspace mode to improve the work on multiple modules

NeoCN:
1. use go workspace mode to improve the work on multiple modules
2. unify the go version in all the go.mod files to `1.19`


vmihailenco:
`go workspace` does not work with `go replace` meaning tha the release script will be broken.

NeoCN:
hi @vmihailenco , thanks for the reply. Just wonder if the `go workspace` is the right way to go, if it is, then I will try to fix the release script, if not, then the PR will be closed.

tzq0301:
But [google](https://groups.google.com/g/golang-codereviews/c/eHzO5T1Gbdc) & [gitignore](https://github.com/github/gitignore/blob/main/Go.gitignore) both add the `go.work` into the file `.gitignore` ... So I think it's not the right way.

vmihailenco:
@NeoCN last time I tried it is not possible to fix the release script. It is a known issue and somewhere in the Go workspace FAQ they say that `go workspace` is not a release tool.

PR #2748, opened 2023-10-09
Return raw output [resp]

ofekshenawa:


ofekshenawa:
Hey @monkey92t ,
I am currently working on a feature that allows the raw output from a Redis server to be returned directly. This would provide the option to receive the response straight from the server. I would greatly appreciate it if you could take a look and assist me with this task.

internal/pool/conn.go
```diff
@@ -64,14 +64,17 @@ func (cn *Conn) RemoteAddr() net.Addr {
 }
 
 func (cn *Conn) WithReader(
-	ctx context.Context, timeout time.Duration, fn func(rd *proto.Reader) error,
+	ctx context.Context, timeout time.Duration, fn []func(rd *proto.Reader) error,
```
chayim, line 67: I think this would make our library, immediately much slower. Adding a loop here - means parsing speeds increase.

command.go
```diff
@@ -25,6 +25,7 @@ type Cmder interface {
 
 	readTimeout() *time.Duration
 	readReply(rd *proto.Reader) error
+	readRawReply(rd *proto.Reader) error
```
chayim, line 28: Do you think it might be possible, during readReply, prior to the parsing to store the RAW resp - based on user input of redis.Options? My thinking (maybe crazy) is that:

1. If we have readReply store the raw prior to parsing, it means it's optionally parsed - so it's now faster.
2. If we store by default we dramatically increase memory footprint (I assume x2) per command executed - so we want to avoid that.
3. If it's an optional redis argument of some sort, we make it user possible, without impact.

WDYT?

monkey92t:
I'm concerned it might cause a bigger burden; doing this could lead to more memory allocations, and it doesn't seem cost-effective.

rueian:
Hi @ofekshenawa,

I think this feature is cool and like to know more about its use cases. Do you also aim to provide a result-streaming functionality? I saw people build a new client for this purpose https://github.com/ammario/redjet.

ndyakov:
@ofekshenawa can we chat about the usecases of this one. Do we plan to proceed with it or can we close it?

Issue #2745, opened 2023-10-06
[Feature request] HSet struct with pointer


numberinho:
### Discussed in https://github.com/redis/go-redis/discussions/2744

<div type='discussions-op-text'>

<sup>Originally posted by **numberinho** October  6, 2023</sup>
I'm trying to HSet() structs with pointers.

Example:

```
cl := redis.NewClient(&redis.Options{
	Addr:     "localhost:6386",
	Password: "", // no password set
	DB:       0,  // use default DB
	PoolSize: 100,
})

type pointer struct {
	PtrInt *int    `redis:"ptrInt"`
	PtrStr *string `redis:"ptrStr"`
}

i := 1
s := "s"
p := pointer{&i, &s}

err := cl.HSet(context.Background(), "key", p).Err()
fmt.Println(err)

// redis: can't marshal *int (implement encoding.BinaryMarshaler)
```

How would you solve this? In my current application I have to store like 30 different structs, all having pointers to "simple data types" like *string and *int. The marshaling of them would be a super cool feature!

Thank you!</div>

tzq0301:
I create PR #2753 to fix it.

tzq0301:
@vmihailenco @ofekshenawa

PR #2753, opened 2023-10-11
feat: support write the types of pointer of simple data types (#2745)

tzq0301:


ofekshenawa:
Hey,
Do you think you might be able to write a unit test that validates this? WDYT?

tzq0301:
> Hey, Do you think you might be able to write a unit test that validates this? WDYT?

Hello. I have added a unit test for `WriteArg` to test the behaviour of the simple data types and the pointer type of them. 

The result of the unit test shows that `WriteArg` the value of the pointer of the simple data types have the same behaviour as the non-pointer types after the modification of `internal/proto/writer.go` in this PR.

ofekshenawa (approved):

Issue #2756, opened 2023-10-14
missing fields in RingOptions and ClusterOptions


chenjie199234:
version: v9.2.1
1.missing ContextTimeoutEnabled and MaxActiveConns in RingOptions
2.missing MaxActiveConns in ClusterOptions

These fields exist in Options.
Are these fields useless in RingOptions and ClusterOptions?

chenjie199234:
is there anyone can pay some attention on this

PR #2757, opened 2023-10-14
fix missing fields in different Options

chenjie199234:
#2756


chayim:
@ofekshenawa does this make sense in the next release?

ofekshenawa (approved):
LGTM

PR #2770, opened 2023-10-22
recognise read only error returned from Lua script

zhengjilei:


ofekshenawa:
Hey,
Do you think you might be able to write a unit test that validates this? WDYT?

error.go
```diff
@@ -137,7 +137,14 @@ func isLoadingError(err error) bool {
...
 func isReadOnlyError(err error) bool {
-	return strings.HasPrefix(err.Error(), "READONLY ")
+	redisError := err.Error()
+	if strings.HasPrefix(redisError, "READONLY ") {
+		return true
+	}
+
+	// For a Lua script that includes a write command, the error string
+	// contains "-READONLY" rather than beginning with "READONLY "
+	return strings.Contains(redisError, "-READONLY")
```
ndyakov, line 147: Can we be more strict here? I assume one can easily use a command with working on a "something-readonly" key and trigger an error that will return true for containing "-readonly"

ndyakov (changes requested):
Please check and address the comments that were left from @ofekshenawa and me.

ndyakov:
Closing this for inactivity.

PR #2781, opened 2023-11-01
newConn race fix, fix check pool exhausted, replacing the MaxActiveConns parameter with PoolSizeStrict

nvorobev:
This library implements non-standard pool behavior. Where the pool size is essentially not an honest pool size and can increase without limitation, which under heavy loads will lead to the server being flooded with connections.

PoolSizeStrict - enabling classic pool mode, when it is guaranteed that no more connections will open to the server than specified in PoolSize.

An example of classic pool behavior:

https://github.com/gomodule/redigo/blob/master/redis/pool.go#L147C1-L147C1

https://github.com/gomodule/redigo/blob/master/redis/pool.go#L247

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -164,6 +164,18 @@ func (p *ConnPool) NewConn(ctx context.Context) (*Conn, error) {
 }
 
 func (p *ConnPool) newConn(ctx context.Context, pooled bool) (*Conn, error) {
+	var poolExhausted bool
+
+	p.connsMu.Lock()
+	if p.cfg.MaxActiveConns > 0 {
+		poolExhausted = p.poolSize >= p.cfg.MaxActiveConns
+	}
+	p.connsMu.Unlock()
+
+	if poolExhausted {
+		return nil, ErrPoolExhausted
+	}
+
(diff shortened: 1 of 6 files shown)


ofekshenawa:
@nvorobev Could you please add unit tests for this issue?

ndyakov:
@nvorobev thank you for the contribution, please add tests so we can review the PR and proceed.

internal/pool/pool.go
```diff
@@ -180,6 +179,12 @@ func (p *ConnPool) newConn(ctx context.Context, pooled bool) (*Conn, error) {
 	p.connsMu.Lock()
 	defer p.connsMu.Unlock()
 
+	// It is not allowed to add new connections to the closed connection pool.
+	if p.closed() {
+		_ = cn.Close()
+		return nil, ErrClosed
+	}
+
```
ndyakov, line 187: it should check if pool is closed before potential dialconn call.

ndyakov (changes requested):

ndyakov:
It has been couple of months, closing this as stale PR. @nvorobev feel free to open a new one based on the latest state of master.

PR #2794, opened 2023-11-10
Add FailoverClusterClient support for Universal client

a-menshchikov:
It would be nice to add suport of FailoverClusterClient into UniversalClient constructor.


a-menshchikov:
@ndyakov done.

PR #2805, opened 2023-11-19
docs: Change go mod init command example

gowizzard:
Doesn't it make sense to use the command from the official Go tutorial for the description? Here is the link to it: https://go.dev/doc/tutorial/create-module


ndyakov:
@gowizzard thank you for the contribution. Since this `README` is mainly consumed on github and most users will probably look at it and start using the library within their project (which is probably hosted on github), I do prefer the `github.com/my/repo` example that is currently used.

gowizzard:
@ndyakov All right, no problem. That's just how it's described in the official Go documentation, so I thought it would make sense.

PR #2818, opened 2023-11-30
Change Z member type to string

### LINKED ISSUE #2817: zset member returned as an interface type instead of a string and is not type-safe
## Expected Behavior

According to the [redis doc](https://redis.io/docs/data-types/sorted-sets/), zset members should always be strings. We want zset members to be returned as strings so that we don't have to do type assertion explicitly.

## Current Behavior

zset member is returned as an interface: https://github.com/redis/go-redis/blob/21bd40a47e56e61c0598ea1bdf8e02e67d1aa651/sortedset_commands.go#L730

## Possible Solution

Define the Z struct's member field as a string type:
https://github.com/redis/go-redis/blob/21bd40a47e56e61c0598ea1bdf8e02e67d1aa651/sortedset_commands.go#L728-L731

ofekshenawa:
Closes #2817

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -3708,28 +3708,28 @@ var _ = Describe("Commands", func() {
 		It("should ZAdd bytes", func() {
 			added, err := client.ZAdd(ctx, "zset", redis.Z{
 				Score:  1,
-				Member: []byte("one"),
+				Member: "one",
 			}).Result()
 			Expect(err).NotTo(HaveOccurred())
 			Expect(added).To(Equal(int64(1)))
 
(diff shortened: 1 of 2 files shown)


dvora-h (approved):

jarno-rootz:
You don't do semantic versioning in go-redis? This is a breaking change.

AlexanderYastrebov:
This change breaks our code https://github.com/zalando/skipper/pull/2806
```
# github.com/zalando/skipper/net
net/redisclient.go:404:47: cannot use val (variable of type int64) as string value in struct literal
```
as we used int64

PR #2820, opened 2023-12-03
Add BITFIELD_RO Command

### LINKED ISSUE #2796: Bitfield_RO support
Bitfield read only is very useful for targeting only read only nodes, since BitField is considered a write command by default.
https://redis.io/commands/bitfield_ro/

ofekshenawa:
Closes #2796

### DIFF
--- a/bitmap_commands.go
+++ b/bitmap_commands.go
@@ -1,6 +1,8 @@
 package redis
 
-import "context"
+import (
+	"context"
+)
 
 type BitMapCmdable interface {
 	GetBit(ctx context.Context, key string, offset int64) *IntCmd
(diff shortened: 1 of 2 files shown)


bitmap_commands.go
```diff
@@ -127,3 +129,21 @@ func (c cmdable) BitField(ctx context.Context, key string, values ...interface{}
...
+
+// BitFieldRO - Read-only variant of the BITFIELD command.
+// It is like the original BITFIELD but only accepts GET subcommand and can safely be used in read-only replicas.
+// - BitFieldRO(ctx, key, "<Encoding0>", "<Offset0>", "<Encoding1>","<Offset1>")
+func (c cmdable) BitFieldRO(ctx context.Context, key string, values ...interface{}) *IntSliceCmd {
+	args := make([]interface{}, 2, 2+len(values))
+	args[0] = "BITFIELD_RO"
+	args[1] = key
+	if len(values)%2 != 0 {
+		panic("BitFieldRO: invalid number of arguments, must be even")
```
chayim, line 141: This doesn't feel like the rest of the library. Maybe we should do nothing here? WDYT?

ofekshenawa: I just followed https://github.com/redis/go-redis/blob/3c56d748ef83ac14cd487e4f3f7b91214c6f3ab6/bitmap_commands.go#L100C30-L100C30

chayim: Gah.. I'd wager the whole thing can change; not in this PR (fair). I don't love all these panics.

chayim (approved):

bitmap_commands.go
ofekshenawa: Yeah, me too. I'll open an issue for this

PR #2826, opened 2023-12-09
add server address and port span attributes to redis otel trace instrumentation

esara:
address https://github.com/redis/go-redis/issues/2825


extra/redisotel/tracing.go
```diff
@@ -41,6 +68,19 @@ func InstrumentTracing(rdb redis.UniversalClient, opts ...TracingOption) error {
...
+				opts = append(opts, WithAttributes(
+					semconv.ServerAddress(host),
+				))
+				// Parse the port string to an integer
+				port, err := strconv.Atoi(portString)
+				if err == nil {
+					opts = append(opts, WithAttributes(
+						semconv.ServerPort(port),
+					))
+				}
```
vmihailenco, line 82: Could you please add a helper function so we don't repeat this block 3 times?

esara: done :-) thanks!

extra/redisotel/config.go
```diff
@@ -63,17 +62,11 @@ func newConfig(opts ...baseOption) *config {
...
+	conf.attrs = append(conf.attrs, semconv.DBSystemRedis)
 
 	return conf
 }
 
-func WithDBSystem(dbSystem string) Option {
-	return option(func(conf *config) {
-		conf.dbSystem = dbSystem
-	})
-}
```
vmihailenco, line 75: @esara could you please restore this? It is a public API and it is still useful for those who use this client with other systems that provide Redis compatible API.

vmihailenco:
@esara LGTM, but I think we should keep `WithDBSystem`...

extra/redisotel/config.go
esara: of course, sorry for not realizing, restored

extra/redisotel/config.go
```diff
@@ -63,7 +62,7 @@ func newConfig(opts ...baseOption) *config {
 		opt.apply(conf)
 	}
 
-	conf.attrs = append(conf.attrs, semconv.DBSystemKey.String(conf.dbSystem))
+	conf.attrs = append(conf.attrs, semconv.DBSystemRedis)
```
vmihailenco, line 65: @esara looks like this should be reverted as well.

vmihailenco (approved):

extra/redisotel/config.go
esara: no problem, I was trying for semconv, where possible, but it is not crucial to this pr - reverted

vmihailenco (approved):

PR #2830, opened 2023-12-11
Support Monitor Command

### LINKED ISSUE #2041: Support the 'MONITOR' command
The README has stated something like `Redis 3 commands except QUIT, MONITOR, and SYNC.` for 8 years or so. I am unclear if this deliberate, or if there might be an appetite to support the `MONITOR` command.

My company has a requirement for this. We may be able to contribute support, but of course guidance around implementation would be desirable. Can it be confirmed if `MONITOR` support fits the scope of this library?

In terms of implementation, I believe it would be something like `SUBSCRIBE`, but I would need to drill in further to understand the nuance. Thanks!

ofekshenawa:
Closes #2041

### DIFF
--- a/command.go
+++ b/command.go
@@ -5382,73 +5382,60 @@ func (cmd *InfoCmd) Item(section, key string) string {
 	}
 }
 
+type MonitorStatus int
+
+const (
+	MonitorStatusIdle MonitorStatus = iota
+	MonitorStatusStart
+	MonitorStatusStop
+)
+
 type MonitorCmd struct {
 	baseCmd
-	ch chan<- string
+	ch     chan string
+	status MonitorStatus
 }
 
-func NewMonitorCmd(ctx context.Context, ch chan<- string) *MonitorCmd {
+func NewMonitorCmd(ctx context.Context, ch chan string) *MonitorCmd {
 	return &MonitorCmd{
 		baseCmd: baseCmd{
-			ctx: ctx,
-			args: []interface{}{
-				"monitor",
-			},
+			ctx:  ctx,
+			args: []interface{}{"monitor"},
 		},
-		ch: ch,
+		ch:     ch,
+		status: MonitorStatusIdle,
 	}
 }
-
-// func (cmd *InfoCmd) SetVal(val map[string]map[string]string) {
-// 	cmd.val = val
-// }
-
-// func (cmd *InfoCmd) Val() map[string]map[string]string {
-// 	return cmd.val
-// }
-
-// func (cmd *InfoCmd) Result() (map[string]map[string]string, error) {
-// 	return cmd.Val(), cmd.Err()
-// }
 
 func (cmd *MonitorCmd) String() string {
 	return cmdString(cmd, nil)
 }
 
 func (cmd *MonitorCmd) readReply(rd *proto.Reader) error {
-
-	go func() {
-		for {
+	go cmd.readMonitor(rd)
+	return nil
+}
+func (cmd *MonitorCmd) readMonitor(rd *proto.Reader) error {
+	for cmd.status == MonitorStatusStart {
+		if pk, _ := rd.Peek(1); len(pk) != 0 {
 			line, err := rd.ReadString()
 			if err != nil {
-				return
+				return err
 			}
 			cmd.ch <- line
-			fmt.Println(line)
 		}
-	}()
-	return nil
 
+	}
+	if cmd.status == MonitorStatusStop {
+		close(cmd.ch)
+	}
+	return nil
+}
+
+func (cmd *MonitorCmd) Start() {
+	cmd.status = MonitorStatusStart
 }
 
 func (cmd *MonitorCmd) Stop() {
-	close(cmd.ch)
+	cmd.status = MonitorStatusStop
 }
-
-// func (cmd *MonitorCmd) WaitAndRecieve() *string {
-// 	if cmd == nil {
-// 		c.cmd = NewCmd(ctx)
-// 	}
-
-// 	err = cn.WithReader(context.Background(), timeout, func(rd *proto.Reader) error {
-// 		return c.cmd.readReply(rd)
-// 	})
-
-// 	c.releaseConnWithLock(ctx, cn, err, timeout > 0)
-
-// 	if err != nil {
-// 		return nil, err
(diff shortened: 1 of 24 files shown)


commands.go
```diff
@@ -700,3 +699,17 @@ func (c cmdable) ModuleLoadex(ctx context.Context, conf *ModuleLoadexConfig) *St
...
 }
+
+// Monitor - represents a Redis MONITOR command, which allows you to capture
+// and process all commands processed by a Redis server. Note that:
+// -- using MONITOR blocks the connection to the server for itself,
+// so it is recommended to createa separate connection dedicated to monitoring purposes.
+// -- The provided channel for monitor output should be of type string.
+// -- The MonitorCmd runs concurrently in the background. Use the Start method
+// to initiate monitoring and the Stop method to terminate it.
+// Redis MONITOR command: https://redis.io/commands/monitor
```
chayim, line 710:
```
/* Monitor - represents a Redis MONITOR command, allowing the user to capture 
and process all commands sent to a Redis server. This mimics the behavior of 
MONITOR in the redis-cli.

Notes:
- Using MONITOR blocks the connection to the server for itself. It needs a dedicated connection
- The user should create a channel of type string 
- This runs concurrently in the background. Trigger via the Start and Stop functions
See further: Redis MONITOR command: https://redis.io/commands/monitor
```

monitor_test.go
```diff
@@ -0,0 +1,47 @@
...
+		client.Set(ctx, "bap", 8, 0)
+		client.Get(ctx, "bap")
+		lst := []string{}
+		for i := 0; i < 5; i++ {
+			s := <-ress
+			lst = append(lst, s)
+		}
+		mn.Stop()
+		Expect(lst[0]).To(ContainSubstring("OK"))
+		Expect(lst[1]).To(ContainSubstring("\"set\" \"foo\" \"bar\""))
```
chayim, line 43: Backticking the contains, saves all the escaping ([example](https://goplay.tools/snippet/ASxOsQcKunE))

monitor_test.go
```diff
@@ -0,0 +1,47 @@
...
+	AfterEach(func() {
+		Expect(client.Close()).NotTo(HaveOccurred())
+	})
+
+	It("should monitor", Label("monitor"), func() {
+		ress := make(chan string)
+		client1 := redis.NewClient(&redis.Options{Addr: ":6379"})
+		mn := client1.Monitor(ctx, ress)
+		mn.Start()
+		time.Sleep(100 * time.Millisecond)
```
chayim, line 31: IMHO comment the why. Let's be kind to our future selves

monitor_test.go
```diff
@@ -0,0 +1,47 @@
...
+		Expect(client.FlushDB(ctx).Err()).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		Expect(client.Close()).NotTo(HaveOccurred())
+	})
+
+	It("should monitor", Label("monitor"), func() {
+		ress := make(chan string)
+		client1 := redis.NewClient(&redis.Options{Addr: ":6379"})
```
chayim, line 28: IMHO small debt here, to remove (all) the hardcoded 6379 and add to a variable in main_test?

command.go
```diff
@@ -5381,3 +5381,61 @@ func (cmd *InfoCmd) Item(section, key string) string {
 		return cmd.val[section][key]
 	}
 }
+
+type MonitorStatus int
+
+const (
+	MonitorStatusIdle MonitorStatus = iota
```
chayim, line 5388: Excellent use of iota!

command.go
```diff
@@ -5381,3 +5381,61 @@ func (cmd *InfoCmd) Item(section, key string) string {
...
+	MonitorStatusStart
+	MonitorStatusStop
```
chayim, line 5390: Should these elements be public or private?

command.go
```diff
@@ -5381,3 +5381,61 @@ func (cmd *InfoCmd) Item(section, key string) string {
...
+		status: MonitorStatusIdle,
+	}
+}
+
+func (cmd *MonitorCmd) String() string {
+	return cmdString(cmd, nil)
+}
+
+func (cmd *MonitorCmd) readReply(rd *proto.Reader) error {
+	go cmd.readMonitor(rd)
```
chayim, line 5415: Can we do something with the following pattern maybe?

ctx, cancel := context.WithCancel(context.Background())
	go func(ctx context.Context) {
		for {
			select {
			case <-ctx.Done():
				fmt.Println("Goroutine has stopped!")
				return
			default:
				fmt.Println("This is the test goroutine")
				time.Sleep(time.Second * 3)
			}
		}
	}(ctx)

Then maybe we can have stop trigger a cancel() to properly shut down the goroutine? Alternatively - perhaps we can set a variable on the MonitorCmd struct and then check it. Stop would set the variable, the running routine would check and end accordingly -stop returning as usual

command.go
```diff
@@ -5381,3 +5381,61 @@ func (cmd *InfoCmd) Item(section, key string) string {
...
+const (
+	MonitorStatusIdle MonitorStatus = iota
+	MonitorStatusStart
+	MonitorStatusStop
+)
+
+type MonitorCmd struct {
+	baseCmd
+	ch     chan string
+	status MonitorStatus
```
chayim, line 5396: IMHO status shouldn't be in the public interface since someone can technically set it when creating, if by hand. WDYT, what else can we do?

monitor_test.go
```diff
@@ -0,0 +1,48 @@
...
+	AfterEach(func() {
+		Expect(client.Close()).NotTo(HaveOccurred())
+	})
+
+	It("should monitor", Label("monitor"), func() {
+		ress := make(chan string)
+		client1 := redis.NewClient(&redis.Options{Addr: rediStackAddr})
+		mn := client1.Monitor(ctx, ress)
+		mn.Start()
+		// Wait for monitor to start listening.
```
chayim, line 31: `// Wait for the Redis server to be in monitoring mode`

command.go
ofekshenawa: We've made the status lowercase to mark it as private, and the high-level 'Monitor' command accepts only channel. Do you think it is good enough?

chayim (approved):

PR #2831, opened 2023-12-12
return raw value instead of funcation calling in Result()

lowang-bh:
Reduce function calling in Result() function.


ofekshenawa (approved):

ofekshenawa:
Hi @lowang-bh, Can you please update your branch with master?

PR #2840, opened 2023-12-18
test: add ut for util_test

rfyiamcool:
### summary

1. add ut for util functions.
2. add benchmark for `toLower`.

I'm curious why the native tolower ( `strings.ToLower()` ) method is not used. Through stress testing, I learned that the speed of the native tolower is not good. 😁  The ToLower implemented by go-redis is three times faster than strings.ToLower.

#### toLower benchmark

use strings.ToLower()

```
goos: darwin
goarch: amd64
pkg: github.com/redis/go-redis/v9/internal
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkToLowerStd
BenchmarkToLowerStd-12    	10940630	        99.88 ns/op	      24 B/op	       1 allocs/op
PASS
ok  	github.com/redis/go-redis/v9/internal	1.823s
```

use utils.ToLower

```
goos: darwin
goarch: amd64
pkg: github.com/redis/go-redis/v9/internal
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkToLowerInternal
BenchmarkToLowerInternal-12    	35665322	        33.31 ns/op	      24 B/op	       1 allocs/op
PASS
ok  	github.com/redis/go-redis/v9/internal	2.377s
```


ofekshenawa (approved):
LGTM!

PR #2861, opened 2024-01-06
Revert #2818 due to it be a breaking change

ofekshenawa:
Revert #2818 due to a breaking change


sjpotter:
commit message should make clear why this change is occuring.

sjpotter (approved):

sjpotter:
Approved, but I'd rename the PR to "Revert #XYZ due to it be a breaking change".  As that's the point of the PR, inteface vs string is somewhat irrelevant to that point.

PR #2863, opened 2024-01-09
Remove unnecessary Node project definition file

limotley:
It seems this file was added in error, as 1. this is Go-based project and 2. It points to an index.js file that does not exist. Removing as cleanup; please comment otherwise if this file is in fact necessary


vmihailenco:
This was needed to generate the changelog using some JS package. Looks like it is not used anymore.

limotley:
Looks like the build is failing due to the issue addressed in https://github.com/redis/go-redis/pull/2864

chayim (approved):

PR #2869, opened 2024-01-18
feat: Allow for pubsub connections to optionally be taken from pool

vtermanis:
- Suggestion for https://github.com/redis/go-redis/issues/2867
- Only applicable to non-cluster client
   - **Question**: Should it apply to all client types?


vtermanis:
Only the RE Tests suite fails - it's not able to find an (internal?) docker image. I believe there's nothing I can do about that.

ndyakov:
@vtermanis thank you for your contribution. What will happen if we exhaust the pool with pub-sub connections, since they are not reusable with resp2?

vtermanis:
> What will happen if we exhaust the pool with pub-sub connections, since they are not reusable with resp2?

@ndyakov - this should be no different to exhausting the pool with non-pubsub connections, I believe.

1. Normal commands
    1. Connection taken from pool
    2. Connection used for some commands
    3. Connection returned to pool **or** connection closed (`connPool.Remove`, if an error occurred)
2. Pub-sub
    1. Connection taken from pool
    2. Connection used for pubsub
    3. Connection closed (`connPool.Remove`, regardless of success failure)

For both 1.iii & 2.iii, if a connection is closed instead of being returned for re-use, a new connection can be created again, within the total pool limit.
(Caveat: I have not looked at whether the core pool code has changed since my proposal.)

ndyakov:
@vtermanis would you mind checking the pool related login in https://github.com/redis/go-redis/pull/3447 , I did decide to have a separate pool for pubsubs. Before going this route I did try different approaches, including using the connection in the pool for pubsub, but at the end it is way cleaner to have a separate pool for pubsub. We may want to work on it and improve (by having a separate limit maybe? and things like that), Let me know what you think, i will close this one for now. Once we have merged https://github.com/redis/go-redis/pull/3447 I will be available to help, review, assist you if you would like to improve the pooling.

Issue #2850, opened 2023-12-25
Improve connection speed by running fewer commands


chayim:
Today, connecting to a Redis instance, involves running multiple commands, and parsing the response repeatedly, this includes usage such as [setinfo](https://redis.io/commands/client-setinfo), which itself is run twice. The example below illustrates, that it's possible to pass all connection configuration information within a pipeline. In turn, the HELLO call returns the version of Redis, which could perhaps be used later for other things.

This is marked as _good-first-issue_ not because it's necessarily easy to implement, inasmuch as it's a good entrypoint, to learning how connections to redis occur.

```
127.0.0.1:6379> multi
OK
127.0.0.1:6379(TX)> auth bob smurf
QUEUED
127.0.0.1:6379(TX)> select 5
QUEUED
127.0.0.1:6379[5](TX)> hello 3
QUEUED
127.0.0.1:6379[5](TX)> client setinfo lib-name predis
QUEUED
127.0.0.1:6379[5](TX)>  client setinfo lib-ver 55.5.0
QUEUED
127.0.0.1:6379[5](TX)> client tracking on optin
QUEUED
127.0.0.1:6379[5](TX)> CLIENT caching yes
QUEUED
127.0.0.1:6379[5](TX)> exec
1) OK
2) OK
3) 1# "server" => "redis"
   2# "version" => "7.2.3"
   3# "proto" => (integer) 3
   4# "id" => (integer) 26
   5# "mode" => "standalone"
   6# "role" => "master"
   7# "modules" => (empty array)
4) OK
5) OK
6) OK
7) OK
127.0.0.1:6379[5]> 
```

prabhjotsingh18:
Hey @chayim, 
total newbie here, I would love to take this up, but I would require assistance. 
If you can help me with stating the point and blueprint of how this can be implemented, that would be great.

lzakharov:
Hello! If I understood correctly, the task is to optimize Redis connection initialization method [*baseClient.initConn](https://github.com/redis/go-redis/blob/master/redis.go#L280). I'm not sure that it's possible to move `hello` command inside the pipeline, but it can be easily done for two `ClientSetInfo` calls.

prabhjotsingh18:
Sure @chayim,  @ofekshenawa if you can guide me where to begin and what concepts I need to familiarise myself with.

pglushko:
@prabhjotsingh18  are you still working on this?

prabhjotsingh18:
hey @pglushko , 
[...]

PR #2880, opened 2024-02-04
Speed up connections by sending SetInfo via a pipeline

ofekshenawa:


chayim:
part of #2850

chayim (approved):

git-hulk:
@ofekshenawa @chayim This will break users using Redis version < 7.2 which doesn't support the `CLIENT SETINFO` command. The previous version works well since the command error was ignored, bringing them into the pipeline will throw an error.

BTW, I think it'd be better to set `DisableIndentity` to `true` by default, for the sake of 7.2 is too new to most users. I will be happy to send a PR to resolve this if it sounds good to you!

svenwltr:
Hello! This was released in a minor update release, but it would break all our apps, since we are using KeyDB and as far as I see it does not support `CLIENT SETINFO`.

Looking at the references to this PR, we are not the only ones having issues.

Are there plans for fixing this or workarounds available?

homme:
Follow #2911 if you've found your way here because this change is breaking your apps.

PR #2885, opened 2024-02-05
fix ConnPool race in newConn

OlegStotsky:
Currently there is a race in newConn. This means that any code that uses this client will fail tests with -race flag enabled. This PR fixes this.

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -168,9 +168,12 @@ func (p *ConnPool) newConn(ctx context.Context, pooled bool) (*Conn, error) {
 		return nil, ErrClosed
 	}
 
+	p.connsMu.Lock()
 	if p.cfg.MaxActiveConns > 0 && p.poolSize >= p.cfg.MaxActiveConns {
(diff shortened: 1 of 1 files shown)


OlegStotsky:
@ofekshenawa please take a look

Aden-Q:
@OlegStotsky You should mention my issue [here](https://github.com/redis/go-redis/issues/2876) (if you saw it). And I don't think it's the correct way to fix the issue. Even though it can protect the variable to be mutated while being read. There still can be interleaving execution flows, causing some errors on the semantic level (for example, read the value in one goroutine, afterwards it is updated by another goroutine). I think we need to we protect the full scope of the critical section, instead of doing: lock-unlock, then lock-unlock again.

OlegStotsky:
@Aden-Q I've added another check after we acquire the mutex again. Should be correct now.

OlegStotsky:
@vmihailenco could you take a look please

ofekshenawa (approved):

PR #2908, opened 2024-02-15
Fix monitor on go 1.19

ofekshenawa:
This PR fixes a data race issue in the monitor command. 
Additionally, We are skipping the example tests because they contain dependencies not supported in Go v1.19


monitor_test.go
```diff
@@ -46,3 +49,52 @@ var _ = Describe("Monitor command", Label("monitor"), func() {
 		Expect(lst[3]).To(ContainSubstring(`"set" "bap" "8"`))
 	})
 })
+
+func TestMonitorCommand(t *testing.T) {
```
gerzse, line 53: I noticed that this new test is written with the `testing` framework. I guess this is intentional, right? Just some lines above there is a similar test written with Ginkgo.

gerzse (dismissed):

monitor_test.go
ofekshenawa: Correct, this is just to make our debugging easier in the future

gerzse (approved):

PR #2915, opened 2024-02-19
bug: Fix SETINFO ensuring it is set-and-forget

### LINKED ISSUE #2910: redis eval error: ERR unknown subcommand 'setinfo'. Try CLIENT HELP.
Running into this error since updating from 9.4.0 to 9.5.0:
```
redis eval error: ERR unknown subcommand 'setinfo'. Try CLIENT HELP.
```

Not sure if this is an issue with how our worker queue package is interfacing with redis, but since the recent release made changes to `setinfo` I figured I would raise the issue just in case. I will try and update this issue with pertinent info as I investigate further

Running on go 1.22.0, the worker queue package is called `asynq`. Thanks!

### LINKED ISSUE #2911: 9.5.0 breaks app with request for setinfo on older redis

## Expected Behavior

golang client connects to redis.

If the server does not support `CLIENT SETINFO`, the golang client shouldn't try to use it, or should at least not error on connection.

## Current Behavior

Upon the first command:
`ERR unknown subcommand 'setinfo'. Try CLIENT HELP.`

## Possible Solution

`opt.DisableIndentity = true`

## Steps to Reproduce

1. Make a hello world app using v9.5.0 redis client, and connect to a recent but not totally current redis such as 7.1.0
2. Run it

https://goplay.tools/snippet/rbwI-UrfSkd

Won't run on goplay.tools because there are too many packages to download. Copy the code into a local scratch folder and run it.

## Context (Environment)

Apps don't start up, I had to revert to 9.4.0

## Detailed Description

This is caused by 9.5.0 change that moves LibraryInfo logic into a pipeline. The prior code did not check for errors. Now it fails the startup pipeline.

### LINKED ISSUE #2914: ClientSetInfo adds runtime.Version() to libName
ClientSetInfo adds the output of runtime.Version() to the libName. The runtime.Version() is not guaranteed to not contain characters not allowed by redis. For example in our case it is: `go1.22.0 X:boringcrypto`, and spaces are not allowed in libName.

The only way to prevent this is to set DisableIndentity (presumably a typo from DisableIdentity). However, if you are also using Sentinel then that option is not copied over in `sentinelOptions` so you'll always send the illegal libName.

```
sentinel.go:552: sentinel: GetMasterAddrByName master="default" failed: ERR LIB-NAME cannot contain spaces, newlines or special characters.
libName = go-redis(,go1.22.0 X:boringcrypto)
```

## Expected Behavior

It shouldn't matter what the go version string is when connecting to a redis server.

DisableIndentity should prevent it from being sent in all cases

## Current Behavior

It matters what the go version string is, potentially breaking when that string is illegal in libName

DisableIndentity does not always prevent libName from being sent (e.g. Sentinel).

## Possible Solution

Remove the Go version from the libName, or strip illegal characters from it. Also copy disableIndentity in all relevant places.

## Steps to Reproduce

Build a redis client program using a Go version that has a version string with a space in it

ofekshenawa:
Updating SETINFO so that when it is sent, it is set-and-forget. This enabled SETINFO to run only where appropriate (i.e redis 7.2+). Once ready, this release should:

closes #2910 
closes #2911 
closes #2914

### DIFF
--- a/options.go
+++ b/options.go
@@ -143,7 +143,7 @@ type Options struct {
 	readOnly bool
 
 	// Disable set-lib on connect. Default is false.
-	DisableIndentity bool
+	DisableIdentity bool
 
 	// Add suffix to client name. Default is empty.
 	IdentitySuffix string
--- a/options.go
+++ b/options.go
@@ -143,7 +143,7 @@ type Options struct {
 	readOnly bool
 
 	// Disable set-lib on connect. Default is false.
-	DisableIdentity bool
+	DisableIndentity bool
 
 	// Add suffix to client name. Default is empty.
 	IdentitySuffix string
--- a/redis.go
+++ b/redis.go
@@ -346,8 +346,10 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 		if c.opt.IdentitySuffix != "" {
 			libName = c.opt.IdentitySuffix
 		}
-		conn.ClientSetInfo(ctx, WithLibraryName(libName))
-		conn.ClientSetInfo(ctx, WithLibraryVersion(libVer))
+		p := conn.Pipeline()
+		p.ClientSetInfo(ctx, WithLibraryName(libName))
+		p.ClientSetInfo(ctx, WithLibraryVersion(libVer))
+		p.Exec(ctx)
 	}
 
 	if c.opt.OnConnect != nil {
--- a/redis.go
+++ b/redis.go
@@ -313,16 +313,6 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 		return err
 	}
 
-	if !c.opt.DisableIdentity {
-		libName := ""
-		libVer := Version()
-		if c.opt.IdentitySuffix != "" {
-			libName = c.opt.IdentitySuffix
-		}
-		conn.ClientSetInfo(ctx, WithLibraryName(libName))
-		conn.ClientSetInfo(ctx, WithLibraryVersion(libVer))
-	}
-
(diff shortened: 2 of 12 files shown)


git-hulk:
@ofekshenawa What about disabling identity by default as mentioned in https://github.com/redis/go-redis/pull/2880#issuecomment-1951063051? guess most of the users are running Redis < 7.2 and enabling this may affect the performance of the connect stage.

redis.go
```diff
@@ -313,6 +313,16 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 		return err
 	}
 
+	if !c.opt.DisableIndentity {
```
dvora-h, line 316: I think it should be `DisableIdentity`(typo)

And I didn't understand what is the reason to run it not in the pipeline?

ofekshenawa: https://github.com/redis/go-redis/issues/2911
Client setinfo needs to be "fire and forget" and when executing in pipline we check for errors in pipline.

options.go
```diff
@@ -143,7 +143,7 @@ type Options struct {
 	readOnly bool
 
 	// Disable set-lib on connect. Default is false.
-	DisableIndentity bool
+	DisableIdentity bool
```
erikdubbelboer, line 146: Renaming this will be a breaking change for everyone setting it now.

ofekshenawa: Thank you!

dvora-h (approved):

redis.go
chayim: While it should be DisableIdentity, that would now be a breaking change for existing users. We'll have to get to that in V10.

chayim (approved):

dvora-h (approved):

PR #2923, opened 2024-02-20
Fix DisableIdentity and IdentitySuffix option forwarding"

### LINKED ISSUE #2911: 9.5.0 breaks app with request for setinfo on older redis

reenjii:
Fix #2911

DisableIdentity and IdentitySuffix options are not correctly forwarded by sentinelOptions and clusterOptions.

Because 9.5.0 now uses redis 7.2 setinfo options by default, this breakd older redis versions unless DisableIdentity is used.

It still breakd FailoverClients since DisableIdentity is not forwarded to base client.

### DIFF
--- a/bench_decode_test.go
+++ b/bench_decode_test.go
@@ -30,7 +30,7 @@ func NewClientStub(resp []byte) *ClientStub {
 		Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
 			return stub.stubConn(initHello), nil
 		},
-		DisableIndentity: true,
+		DisableIdentity: true,
 	})
 	return stub
 }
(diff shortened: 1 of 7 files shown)


vtolstov:
why this still open ?

ndyakov:
renaming planned for v10, sentinel (failover) should be fixed in master.

PR #2926, opened 2024-02-23
add: hscan data struct

7134g:
I think the hash stored using `encoding.BinaryMarshaler`, 
then parse time should also use `encoding.BinaryUnmarshaler` parsing
Now write using `encoding.BinaryMarshaler`
`Scan` reads using `encoding.TextUnmarshaler`
Look strange

so add case:

`encoding.BinaryUnmarshaler`
`encoding.TextMarshaler`

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -2,6 +2,7 @@ package redis_test
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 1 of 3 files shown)


ndyakov:
Hello @7134g , would you like to review the failing lint and do the required change so we can proceed with this?

ndyakov:
@7134g i will close this for inactivity, feel free to reopen.

PR #2927, opened 2024-02-23
Add CredentialsProvider field to UniversalOptions

dcherubini:
In https://github.com/redis/go-redis/pull/2097 go-redis started to support passing a CredentialsProvider callback to get the username and password when initializing a connection, instead of having to provide them when creating the client. This is very useful in scenarios where you have these credentials stored in a separate service (e.g., AWS Secret Manager or Azure Key Vault) and you want to rotate those credentials periodically.

However, this option was not added in all clients and was not being passed down from the universal client to the concrete clients either. This PR aims to normalize that.

(Note that the CredentialsProviders field was already added to ClusterOptions in https://github.com/redis/go-redis/pull/2791, which is why they are not being touched in this change)

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -70,10 +70,11 @@ type RingOptions struct {
 	Dialer    func(ctx context.Context, network, addr string) (net.Conn, error)
 	OnConnect func(ctx context.Context, cn *Conn) error
 
-	Protocol int
-	Username string
-	Password string
-	DB       int
+	Protocol            int
+	Username            string
+	Password            string
+	CredentialsProvider func() (username string, password string)
+	DB                  int
 
 	MaxRetries      int
 	MinRetryBackoff time.Duration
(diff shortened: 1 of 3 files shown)


ndyakov:
Hello @dcherubini, We are introducing a new `StreamingCredentialsProvider` in #3320  and currently there is the `CredentialsProviderContext` (it's not a great name, I know). Would you mind updating the options in such a way to address those two as well? I was planing to do it myself, but wanted to keep you contribution and work with you on this PR instead of opening a new one.

ndyakov:
@dcherubini the StreamingCredentialsProvider is added. Would you like to update this PR to add all credentials providers:
https://github.com/redis/go-redis/blob/86d418f94077de29be613b9a2c4a324799f96d81/options.go#L68-L84

ndyakov:
@htemelski-redis , @ofekshenawa I did push the needed changes here, let me know if you would like to me to address something else so we can merge this one.

ring.go
```diff
@@ -154,10 +172,11 @@ func (opt *RingOptions) clientOptions() *Options {
...
 		OnConnect:  opt.OnConnect,
 
-		Protocol: opt.Protocol,
-		Username: opt.Username,
-		Password: opt.Password,
-		DB:       opt.DB,
+		Protocol:            opt.Protocol,
+		Username:            opt.Username,
+		Password:            opt.Password,
+		CredentialsProvider: opt.CredentialsProvider,
```
htemelski-oss: Should we add `CredentialsProviderContext` and `StreamingCredentialsProvider` here?

ndyakov: They should be added, let me verify the diff.

htemelski-redis (approved):

ring.go
ndyakov: Updated.

PR #2934, opened 2024-03-05
fix the maxActiveConns sematic

lawlielt:
we can get conn use the `Get` or `NewConn` method, they are all will be add to the `p.conns`.

the sematic of the `MaxActiveConns` opt is the maximum number of connections in the pool


lawlielt (commented):
fix the maxActiveConns sematic

ndyakov:
Hello @lawlielt , would you please add unit tests that cover the change that you did

ndyakov:
@lawlielt I am not 100% sure if we would like to calculate the len since we have the poolSize. Can you check the changes in https://github.com/redis/go-redis/pull/3447 related to the poolSize and let me know there if you would like to change it. For now I'm gonna close this one.

PR #2940, opened 2024-03-13
feat(ring): specify custom health check func via HeartbeatFn option

strobil:
The current implementation of the Redis ring does not support specifying custom shard health check rules.
We use a Redis shard to distribute the load across multiple Redis slave instances.
However, if any shard encounters replication issues, the Redis ring continues to utilize this shard because it responds to ping requests successfully.
Thus, the Redis ring does not remove the malfunctioning shard, even though it is inconsistent.

This pull request proposes adding a **HeartbeatFn** option to the **RingOptions**, which can be configured during ring construction. This addition enables the definition of custom health check logic, such as checks based on the current replication status of the shard.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -18,10 +18,12 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@v9
+      # First step: Handle regular issues (excluding needs-information)
+      - name: Mark regular issues as stale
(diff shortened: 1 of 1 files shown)


ndyakov:
Thank you, @strobil! I do think having a custom healthcheck in the ring can benefit the users. Would you mind resolving conflicts so we can merge master and proceed with review of this PR?

ndyakov:
@strobil i do see the test continue to fail, would you mind taking a look

PR #2946, opened 2024-03-22
feat: support round robin replica selection

ukriish:
In one of our production clusters, where we use `SlaveForKey` function to selectively run `EVALSHA`, we noticed that the replica selection strategy was random and resulted in an uneven traffic pattern. During high throughput this started showing up with noticeable resource utilisation on some of the nodes compared to other.  Ideally this wouldn't have made too big an issue but since we are running lua scripts and `evalsha`, it was noticeable. 

I checked to see if there was a way for clients to inject a routing strategy via ClusterOptions but it isn't supported. Ideally, I think that might make more sense, especially since the client already supports `SlaveForKey`, `MasterForKey`, `ForEachMaster` and `ForEachSlave` methods. Immediately for our use case I decide to just add a new route strategy but it might make sense to support custom route functions. 

Please let me know if this makes sense to be merged back here.


ndyakov:
@ukriish this is an interesting idea. Would you like to work on improving the implementation a bit? I don't understand why are it will loop 10 times to try to select a slave node. There is some benefit in trying couple of nodes or len(nodes) for exampe, but why `10` ?

ukriish:
@ndyakov Yea, will be happy to improve on it. I agree, a couple of nodes or number of replica makes more sense, the only reason I chose 10 was to replicate the logic in `slotSlaveNode` func.

ndyakov:
@ukriish would you like to continue the work on this ?

PR #2968, opened 2024-04-12
Attempt to Close Sentinel when the NewFailoverClusterClient is Closed.

DevineLiu:
We should have a means to reclaim the sentinels' connection resources established by NewFailoverClusterClient.

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -62,6 +62,8 @@ type ClusterOptions struct {
 
 	OnConnect func(ctx context.Context, cn *Conn) error
 
+	OnClose func() error
+
 	Protocol            int
 	Username            string
 	Password            string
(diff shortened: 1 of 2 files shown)


monkey92t:
I didn't understand this PR. What does it mean?

DevineLiu:
When using `NewFailoverClusterClient` for initialization, at [line 788](https://github.com/redis/go-redis/blob/f7f34feddf26644661276e90ebd87e748d70a050/sentinel.go#L788), the code initializes  connections to the Sentinels, and at [line 792](https://github.com/redis/go-redis/blob/f7f34feddf26644661276e90ebd87e748d70a050/sentinel.go#L792), the client subscribes  to  the Sentinel. However, the long-lived connection between the client and the Sentinel doesn't get closed when `Close` is called on the `ClusterClient` returned by `NewFailoverClient`.

This issue occurs when the `Close` method is called on `ClusterClient`, and the long-lived connection to the Sentinel is not closed along with it.

ndyakov:
@DevineLiu I think this is handled in the latest version, please review and if needed open separate PR.

PR #2980, opened 2024-04-23
Remove skipping span creation by checking parent spans

### LINKED ISSUE #2953: Remove skipping span creation by checking parent spans
## Current Behavior

The `redisotel` skips the span creation by only checking the `IsRecording` of the parent span. This breaks the span creation defined by [opentelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/e87154c02b40e4b7bb6bf776a12e352372b21134/specification/trace/sdk.md#sdk-span-creation).

> Create a span depending on the decision returned by ShouldSample: see description of [ShouldSample's](https://github.com/open-telemetry/opentelemetry-specification/blob/e87154c02b40e4b7bb6bf776a12e352372b21134/specification/trace/sdk.md#shouldsample) return value below for how to set IsRecording and Sampled on the Span, and the [table above](https://github.com/open-telemetry/opentelemetry-specification/blob/e87154c02b40e4b7bb6bf776a12e352372b21134/specification/trace/sdk.md#recording-sampled-reaction-table) on whether to pass the Span to SpanProcessors.

Considering a user configures an always-on sampler for debugging and expects every operation taken by Redis that has instrumentation. With the current implementation, this would drop spans even if the user wants to see it.

https://github.com/open-telemetry/opentelemetry-go/blob/ba5d1268082fcc4ccb8fa967eb5133fa23e4f2ba/sdk/trace/sampling.go#L128-L130

## Possible Solution

Remove codes like this before calling `tracer.Start`

https://github.com/redis/go-redis/blob/f3fe61148b2b8fe0a669dc23620690407f5f92af/extra/redisotel/tracing.go#L94-L96

## More context

My concern only comes from OTel's perspective; let me know if it is a special case for go-redis. I am also happy to raise a PR to solve this issue.

XSAM:
Resolves #2953

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## Unreleased
+
+### Changed
+
+* `go-redis` won't skip span creation if the parent spans is not recording. ([#2980](https://github.com/redis/go-redis/issues/2980))
+  Users can use the OpenTelemetry sampler to control the sampling behavior.
+  For instance, you can use the `ParentBased(NeverSample())` sampler from `go.opentelemetry.io/otel/sdk/trace` to keep
+  a similar behavior (drop orphan spans) of `go-redis` as before.
+
 ## [9.0.5](https://github.com/redis/go-redis/compare/v9.0.4...v9.0.5) (2023-05-29)
 
 
--- a/extra/redisotel/tracing.go
+++ b/extra/redisotel/tracing.go
@@ -91,10 +91,6 @@ func newTracingHook(connString string, opts ...TracingOption) *tracingHook {
 
 func (th *tracingHook) DialHook(hook redis.DialHook) redis.DialHook {
 	return func(ctx context.Context, network, addr string) (net.Conn, error) {
-		if !trace.SpanFromContext(ctx).IsRecording() {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## Unreleased
+
+### Changed
+
+* `go-redis` won't skip span creation if the parent spans is not recording. ([#2980](https://github.com/redis/go-redis/issues/2980))
+  Users can use the OpenTelemetry sampler to control the sampling behavior.
+  For instance, you can use the `ParentBased(NeverSample())` sampler from `go.opentelemetry.io/otel/sdk/trace` to keep
+  a similar behavior (drop orphan spans) of `go-redis` as before.
+
```
XSAM, line 9: I am not sure it is the way the `redis` community makes the changelog, but I think it is good to have such a notice for users.

XSAM:
The CI failure seems unrelated to this PR.

CHANGELOG.md
vmihailenco: It was auto-generated, but is not updated any more.

vmihailenco (approved):

PR #2981, opened 2024-04-24
Handle IPv6 in isMovedError

daviddzxy:
If the address in the error is an IPv6 address it adds squared brackets around the host part of the address.

### DIFF
--- a/error.go
+++ b/error.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"strings"
 
+	"github.com/redis/go-redis/v9/internal"
 	"github.com/redis/go-redis/v9/internal/pool"
 	"github.com/redis/go-redis/v9/internal/proto"
 )
@@ -129,7 +130,9 @@ func isMovedError(err error) (moved bool, ask bool, addr string) {
 	if ind == -1 {
 		return false, false, ""
 	}
+
 	addr = s[ind+1:]
+	addr = internal.GetAddr(addr)
 	return
 }
 
--- a/internal/util.go
+++ b/internal/util.go
@@ -67,20 +67,17 @@ func ReplaceSpaces(s string) string {
 }
 
 func GetAddr(addr string) string {
-	ind := strings.LastIndex(addr, ":")
+	ind := strings.LastIndexByte(addr, ':')
 	if ind == -1 {
 		return ""
 	}
 
-	port := addr[ind+1:]
-	host := addr[:ind]
-	if string(host[0]) == "[" && string(host[len(host)-1]) == "]" {
-		host = host[1 : len(host)-1]
+	if strings.IndexByte(addr, '.') != -1 {
+		return addr
 	}
 
-	if net.ParseIP(host) == nil {
-		return ""
+	if addr[0] == '[' {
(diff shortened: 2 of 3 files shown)


monkey92t:
should: 
```go
func getAddr(addr string) string {
	ind := strings.LastIndex(addr, ":")
	if ind == -1 {
		return ""
	}

	if string(addr[ind-1]) == "]" {
		return ""
	}

	return net.JoinHostPort(addr[:ind], addr[ind+1:])
}

func TestAddr(t *testing.T) {
	if v := getAddr("127.0.0.1:1234"); v != "127.0.0.1:1234" {
		t.Fatalf("getAddr failed: %v", v)
	}
	if v := getAddr("[::1]:1234"); v != "[::1]:1234" {
		t.Fatalf("getAddr failed: %v", v)
	}
	if v := getAddr("::1:1234"); v != "[::1]:1234" {
		t.Fatalf("getAddr failed: %v", v)
	}
	if v := getAddr("fd01:abcd::7d03:6379"); v != "[fd01:abcd::7d03]:1234" {
		t.Fatalf("getAddr failed: %v", v)
	}
	if v := getAddr("[fd01:abcd::7d03]:6379"); v != "[fd01:abcd::7d03]:6379" {
		t.Fatalf("getAddr failed: %v", v)
	}
	if v := getAddr("127.0.0.1"); v != "" {
		t.Fatalf("getAddr failed: %v", v)
	}
	if v := getAddr("a:b:c"); v != "" {
		t.Fatalf("getAddr failed: %v", v)
	}
	if v := getAddr("127"); v != "" {
		t.Fatalf("getAddr failed: %v", v)
	}
}
```

daviddzxy:
I moved the GetAddr function into internal/util.go and the unit tests were moved to internal/util_test.go. What do you think?

monkey92t (changes requested):
It seems that the functionality is implemented correctly, but there are too many unnecessary queries and checks. IPv4 is used in most cases, so we should prioritize IPv4 checks. Additionally, we should use the more efficient method IndexByte() instead of Index().

monkey92t:
In the current scenario, we should trust the validity of the data returned by the redis-server. It is known that the redis-server may return three types of data: 127.0.0.1:6379 (IPv4), ::1:6379 (incorrect IPv6), [::1]:6379 (correct IPv6).

monkey92t:
```go
func getAddr(addr string) string {
	idx := strings.LastIndexByte(addr, ':')
	if idx == -1 {
		return ""
	}

	// ipv4
	if strings.IndexByte(addr, '.') != -1 {
		return addr
	}

	// ipv6
	if addr[0] == '[' {
		return addr
	}
	return net.JoinHostPort(addr[:idx], addr[idx+1:])
}
```

monkey92t:
Both LastIndex() and Index() are complex functions that search for one string within another string, rather than searching for a single character. This is a completely different approach. In the current situation, we only need to search for a single character, not a string.

monkey92t (approved):

callmeadi:
When are we releasing a tagged version for this?

PR #2995, opened 2024-05-13
ReadFromAny (master+replica) without cluster mode (Master-Replica setup)

singh-bhawani:
Right now, with go-redis, we can't read from both the Master and its Replica without being in cluster mode. Also, when we use "ReplicaOnly", it randomly chooses Replica to read. This causes a problem because caching is mainly used for read heavy usecase, which means the Master only used for write throughput and it's resource are underutilized.

To solve this, I added "ReadFromAny" to "FailoverOptions". This means that read-only commands will come from both Replicas as well as Master. This way, we can make better use of the caching resource.


ndyakov:
@singh-bhawani thank you for the contribution! Would you please add tests to cover the code you are introducing?

ndyakov (changes requested):
Please add tests.

ndyakov:
Closing this for inactivity. It is a good idea and it would be even better to use some sort of round-robin when we have `ReadFromAny` set. If anyone would like to contribute this, feel free to open a PR.

PR #2998, opened 2024-05-24
fix: fix #2681

### LINKED ISSUE #2681: Options.CredentialsProvider should support Context and returning an error
## Expected Behavior

The interface for `Options.CredentialsProvider` should be `func(ctx context.Context) (username string, password string, err error)`

`baseClient.initConn` should return any error from `Options.CredentialsProvider`

## Current Behavior

`Options.CredentialsProvider` doesn't support context and doesn't pass errors down the call chain.

## Context (Environment)

I want to use Options.CredentialsProvider to implement dynamic password fetching from cloud providers. e.g. https://github.com/redis/go-redis/discussions/2343

## Detailed Description

Looking at the original commit https://github.com/redis/go-redis/pull/2097, the intent of CredentialsProvider seems to be to memory scrub plaintext passwords, but now people want to use it to dynamically fetch passwords from cloud providers. As the library generally supports Context, i'll skip the explanation for why its idiomatic and appropriate to pass context into code that is intended to make network requests. This would be a breaking change and require a version increment.

## Possible Implementation

`CredentialsProvider func(ctx context.Context) (username string, password string, err error)`

```
func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
	if cn.Inited {
		return nil
	}
	cn.Inited = true
	username, password := c.opt.Username, c.opt.Password
	if c.opt.CredentialsProvider != nil {
		var credentialsProviderErr error
		username, password, credentialsProviderErr = c.opt.CredentialsProvider(ctx)
		if credentialsProviderErr != nil {
			return credentialsProviderErr
		}
	}
..
}
```

monkey92t:


redis.go
```diff
@@ -283,8 +283,13 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 	}
 	cn.Inited = true
 
+	var err error
 	username, password := c.opt.Username, c.opt.Password
-	if c.opt.CredentialsProvider != nil {
+	if c.opt.CredentialsProviderContext != nil {
+		if username, password, err = c.opt.CredentialsProviderContext(ctx); err != nil {
```
vmihailenco, line 289: You could move `var err error` inside this block so you would not need to make the changes below.

vmihailenco (approved):

redis.go
monkey92t: Yes, it can be modified in another PR to avoid re-review.

PR #3007, opened 2024-05-31
support parsing client-cache invalidation messages

mhelmich:
This PR adds support for parsing redis client-side caching invalidation messages.
Today, these messages are not parsed properly and an error is returned instead.
With this PR, messages are parsed into a `Message` and the array of invalidated cache key is passed to the caller inside `PayloadSlice` of the message.
This PR does not add a cache to the client or enables the client-side caching feature. It merely plumbs invalidation messages through to the caller. A caching mechanism can (and has to) be built on top of this change.

Details about redis' client-side caching feature can be found here: https://redis.io/docs/latest/develop/use/client-side-caching/


mhelmich:
Hi @chayim and @ofekshenawa 👋 
I'm roping you into this PR because you were also part of the relevant discussion (https://github.com/redis/go-redis/discussions/2370). I'm hoping to get some traction on this PR.

pubsub_test.go
```diff
@@ -567,4 +569,99 @@ var _ = Describe("PubSub", func() {
...
+		invalidateClient := redis.NewClient(&invalidateOpts)
+		invalidations := invalidateClient.Subscribe(ctx, "__redis__:invalidate")
+
+		go func() {
+			defer func() {
+				invalidations.Close()
+				invalidateClient.Close()
+
+				mu.Lock()
+				invalidateClientID = -1
```
atharva29, line 633: Why don't you use `atomic.Int32` to update this variable
such that !
1. Mutex will get removed
2.on line [617] (https://github.com/redis/go-redis/pull/3007#diff-9068507d4bce3bd07f1c5247d3beb979edda47aa6ed0262c82d192b0b22c4ddaR617) you are already locking the mutex, so there is a chance of double locking mutex

iangcarroll:
This would be a useful addition to the library IMO. `CLIENT TRACKING` seems underused and would be great to have support for here.

PR #3014, opened 2024-06-03
Use the default 15-seconds TCP-Keepalive

tahajahangir:
This reverts commit 4bf8b5a128b70cad518e5deeeeeec86315428837

TCP-Keepalives are enable by default in go since go 1.12 [in this commit](https://github.com/golang/go/commit/5bd7e9c54f946eec95d32762e7e9e1222504bfc1) few months after #825 were merged.

Now the default value of 15-seconds is better and safer and we can remove this.


ndyakov:
@tahajahangir the default `KeepAlive` interval is too short and will increase cpu usage. I would prefer to wait to update to `go1.23+` to include https://github.com/golang/go/issues/62254 and have a finer grain configuration. Maybe a longer initial interval and then shorter intervals will work good with `go-redis`.

PR #3016, opened 2024-06-04
refactor: pass custom log to pubsub

bailantaotao:
If a user subscribes to many channels, it becomes difficult to keep track of which channel's data has been missed.

https://github.com/redis/go-redis/discussions/3015


ndyakov:
@bailantaotao please sync with master

ndyakov:
@bailantaotao changing the loging interface here is a breaking change we cannot have at the moment. We will consider your suggestion for the next iteration for logging, but I will close this PR. Thank you.

PR #3060, opened 2024-07-17
Add ConnChecker to customize conn health check

pragkent:
**Allow user to customize conn health check.**

One of the scenario we met is that we'd like to drain some proxy node before upgrading. If we'd like to take down some node, instructions could be sent to clients to drop connections to these nodes.

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -68,6 +68,7 @@ type Options struct {
 	MaxActiveConns  int
 	ConnMaxIdleTime time.Duration
 	ConnMaxLifetime time.Duration
+	ConnChecker     func(net.Conn) error
 }
 
 type lastDialErrorWrap struct {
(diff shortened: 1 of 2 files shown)


monkey92t:
You should use the Option.Dialer property instead, compared to ConnCheck, as it can create the network connection you want, as long as it conforms to the net.Conn interface.

ConnCheck checks *tls.Conn and syscall.Conn interfaces and uses RawConn to check connection health. You can implement the syscall.Conn interface in custom network connections to achieve this purpose.

pragkent:
> You should use the `Option.Dialer` property instead, compared to `ConnCheck`, as it can create the network connection you want, as long as it conforms to the `net.conn` interface.
> 
> `ConnCheck` checks `*tls.Conn` and `syscall.Conn` interfaces and uses `RawConn` to check connection health. You can implement the `syscall.Conn` interface in custom network connections to achieve this purpose.

That should work, I'll try. Thanks.

ndyakov:
Closing for lack of response.

PR #3085, opened 2024-08-07
Do not assume that all non-IP hosts are loopbacks

suever:
The `ClusterClient` is configured such that if the cluster nodes that are discovered have a loopback IP provided as their address (e.g. 127.0.0.1), that the address of the origin that was used to connect to the cluster should be used instead (if it is not also a loopback). The logic for this is captured [here](https://github.com/redis/go-redis/blob/00d98485f8357d016139ece133bd3272e7ff2526/osscluster.go#L637-L639).

The latest change to this code was in [this PR](https://github.com/redis/go-redis/commit/c1c2753ae87f98290db54fa6de8c07d3fcc5d02c) which was a fix for some errors in the behavior of #589. However, in fixing #589, the logic behind determining if the origin's address was a loopback address was incorrectly changed. 

In #589, if the origin host was not an IP address, then [it was not considered to be a loopback address](https://github.com/redis/go-redis/pull/589/files#diff-556b0079cdcdaaa712eaa26594117c9d72b568bbda914495d8f46ee847f98c40R977-R980). Since hostname resolution doesn't happen, there is no way as it is currently written to determine if it is a loopback address so I think returning that it is not a loopback address is a valid option. 

In the fix implemented in #839, if a hostname is used to connect to the cluster rather than an IP address, it is *assumed to be a loopback address* based on [this change](https://github.com/redis/go-redis/commit/c1c2753ae87f98290db54fa6de8c07d3fcc5d02c#diff-556b0079cdcdaaa712eaa26594117c9d72b568bbda914495d8f46ee847f98c40R504) and this therefore prevents the substitution of the origin for a loopback address of a node as should happen.

This PR changes the behavior back to that of #589, where a hostname is not considered to be a loopback address and this allows a cluster client to substitute that hostname for a loopback address of a discovered cluster node that announces its own IP as a loopback address.

Another possible solution would be to *always* change loopback addresses of nodes to the origin address regardless of whether the origin is a loopback or not, but I opted for the minimal touch here to preserve the logic.

This was discovered when using a redis cluster in a docker-compose file where one service referenced the redis-cluster service by name (e.g. `redis-cluster:7000`), but the redis cluster was constructed using `127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002`.


suever:
Failing tests are unrelated to this change and are due to the changes in 00d98485f8357d016139ece133bd3272e7ff2526 which moved `redis-stack-server` to the `latest` (`master` branch build is currently failing as well)

osscluster.go
```diff
@@ -720,7 +720,7 @@ func replaceLoopbackHost(nodeAddr, originHost string) string {
 func isLoopback(host string) bool {
 	ip := net.ParseIP(host)
 	if ip == nil {
-		return true
+		return false
```
ndyakov, line 723: Shoudn't `localhost` return `true` here?

ndyakov: Actually, based on the network setup, many hostnames can be considered loopbacks... we should definitely handle `localhost`. `host.docker.internal` or even anything `*.docker.internal` can be considered loopback

ndyakov (approved):
@ofekshenawa thank you for continuing the work on this!

suever:
Sorry for the slow response to this, just seeing the notification of the clarification now @ofekshenawa ! Thanks for working to get this merged 🙏

PR #3098, opened 2024-08-31
Support Resp 3 Redis Search Unstable Mode

ofekshenawa:


redis.go
```diff
@@ -412,6 +412,20 @@ func (c *baseClient) process(ctx context.Context, cmd Cmder) error {
...
 }
 
+func (c *baseClient) isProblematicMethodsOfSearchResp3(ctx context.Context, cmd Cmder) bool {
+	if c.opt.Protocol != 3 {
+		return false
+	}
+
+	switch cmd.(type) {
+	case *AggregateCmd, *FTInfoCmd, *FTSpellCheckCmd, *FTSearchCmd, *FTSynDumpCmd:
+		internal.Logger.Printf(ctx, "Some RESP3 results for Redis Query Engine responses may change. Refer to the readme for guidance")
```
vladvildanov, line 422: The logic should be different here. So the algorithm is:

1. If `protocol == 2` - we allow to use Search module without limitations.
2. If `protocol == 3 && UnstableResp3SearchModule !== True` - we throw this exception "Some RESP3 results for Redis Query Engine responses may change. Refer to the readme for guidance"
3. If `protocol == 3 && UnstableResp3SearchModule == True` - we allows to use Search module with the limitation that command replies will be raw.

redis.go
```diff
@@ -427,8 +441,12 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder, attempt int) (bool
 			atomic.StoreUint32(&retryTimeout, 1)
 			return err
 		}
-
-		if err := cn.WithReader(c.context(ctx), c.cmdTimeout(cmd), cmd.readReply); err != nil {
+		readReplyFunc := cmd.readReply
+		// Apply unstable RESP3 search module.
+		if c.opt.UnstableResp3SearchModule && c.isProblematicMethodsOfSearchResp3(ctx, cmd) {
```
vladvildanov, line 446: I would suggest to do `c.opt.UnstableResp3SearchModule` check inside if `c.isProblematicMethodsOfSearchResp3`. And I think it will lead to method rename. Let's keep this unstable functionality as much as possible in the single place to be able to remove this easier in a future.

The method name can be something like `assertStableCommand()`. ANd you may call it only in case if `c.opt.protocol !== 2`

vladvildanov (changes requested):

redis.go
```diff
@@ -427,8 +440,12 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder, attempt int) (bool
 			atomic.StoreUint32(&retryTimeout, 1)
 			return err
 		}
-
-		if err := cn.WithReader(c.context(ctx), c.cmdTimeout(cmd), cmd.readReply); err != nil {
+		readReplyFunc := cmd.readReply
+		// Apply unstable RESP3 search module.
+		if c.opt.Protocol != 2 && c.assertStableCommand(cmd) {
```
vladvildanov, line 445: In this case method should be renamed to `assertUnstableCommand()` to match semantically what we're trying to achieve. Cause if it's unstable we want it to return a raw output, in case if flag was set.

vladvildanov (approved):

vladvildanov (approved):

PR #3101, opened 2024-09-02
WriteArg support uintptr

### LINKED ISSUE #3100: Set don't support unitptr
```
var id uintptr = 123
err := rdb.Set(ctx, "test1", id, 0)
if err != nil {
    t.Fatal(err)
}
```

oldme-git:
fixed #3100


internal/proto/writer.go
```diff
@@ -109,6 +109,10 @@ func (w *Writer) WriteArg(v interface{}) error {
 		return w.uint(v)
 	case *uint64:
 		return w.uint(*v)
+	case uintptr:
+		return w.uint(uint64(v))
+	case *uintptr:
+		return w.uint(uint64(*v))
```
ndyakov, line 115: @oldme-git what will happen if the pointer is `nil` ?

PR #3105, opened 2024-09-04
fix: let Set to support custom type

### LINKED ISSUE #3104: Set and Set* don't support custom type
```
func TestI(t *testing.T) {
  type myInt int
  var i myInt = 123

  err := rdb.Set(ctx, "key", i, 0).Err()
  if err != nil {
    t.Fatal(err)
  }
}

result: redis: can't marshal * (implement encoding.BinaryMarshaler)
```

oldme-git:
**Dear review:**
Thanks for your reviewing

Set and other Set function can't use custom type. e.g.
```
func TestI(t *testing.T) {
  type myInt int
  var i myInt = 123

  err := rdb.Set(ctx, "key", i, 0).Err()
  if err != nil {
    t.Fatal(err)
  }
}

result: redis: can't marshal * (implement encoding.BinaryMarshaler)
```
It should add rereflect assertion after type assertion. I did it for performance and other types of considerations. for examole: `time.Time` or `encoding.BinaryMarshaler`

Have a good day!

fixed: #3104


ndyakov:
Hello @oldme-git, Would you mind adding a test case per type added in the reflect select ?

PR #3127, opened 2024-09-16
docs: update default timeout value from 3 seconds to 5 seconds

kooskoos20:
Update incorrect default timeout value  from 3 seconds to 5 seconds


ndyakov:
@kooskoos20 in https://github.com/redis/go-redis/blob/d0f921357d9751a3e11221a7abd072546ac67654/options.go#L187 i can se the default is 3 seconds, not 5. Where did you get that information that the default is `5` ?

PR #3130, opened 2024-09-20
Enable saving of command keys

ofekshenawa:
Adds support for extracting and persisting the Redis keys used by each command—particularly multi‐key commands.

**Key changes:**
- New Keys() method on the Cmder interface, allowing callers to retrieve all keys a command operates on.
- Extended baseCmd to track a keys []string slice and implement Keys() []string.
- Helper for interleaved args: introduced getInterleavedArgumentsWithOffset(offset int) []string to extract key names from commands with key/value pairs (e.g. MSET, MSETNX).
- Command implementations in string_commands.go updated to populate cmd.keys for:
- - Single‐key commands (e.g. GET, SET, INCRBY, APPEND)
- - Multi‐key commands (MGET, MSET, MSETNX)

### DIFF
--- a/command.go
+++ b/command.go
@@ -210,9 +210,9 @@ func (cmd *baseCmd) readRawReply(rd *proto.Reader) (err error) {
 	return err
 }
 
-// getInterleavedArguments returns arguments at even indices starting from index 0.
+// getInterleavedArguments returns arguments at even indices starting from index 1.
 func (cmd *baseCmd) getInterleavedArguments() []string {
-	return cmd.getInterleavedArgumentsWithOffset(0)
+	return cmd.getInterleavedArgumentsWithOffset(1)
 }
 
 // getInterleavedArgumentsWithOffset returns arguments at even indices starting from the specified offset.
--- a/command.go
+++ b/command.go
@@ -122,30 +122,6 @@ func cmdString(cmd Cmder, val interface{}) string {
 	return util.BytesToString(b)
 }
 
-func extractKeys(cmd Cmder) []string {
-	firstKeyPos := cmdFirstKeyPos(cmd)
-	if firstKeyPos == -1 {
-		return nil
-	}
-
-	args := cmd.Args()
-	keys := []string{}
-
-	switch cmd.Name() {
-	case "mget":
-		for i := int(firstKeyPos); i < len(args); i++ {
-			keys = append(keys, cmd.stringArg(i))
-		}
-
-	default:
-		if int(firstKeyPos) < len(args) {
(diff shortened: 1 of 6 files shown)


command.go
```diff
@@ -119,6 +119,35 @@ func cmdString(cmd Cmder, val interface{}) string {
...
+	firstKeyPos := cmdFirstKeyPos(cmd)
+	if firstKeyPos == -1 {
+		return nil
+	}
+
+	args := cmd.Args()
+	keys := []string{}
+
+	switch cmd.Name() {
+	case "mget":
```
vladvildanov, line 137: This approach isn't scalable, each time we need a new type of extracting we have to add a command here. `Cmder` is an interface that should be abstract as possible and should not contain any implementation specific (mget is an implementation specific in this case).

Cmder should return a keys, but shouldn't be aware about the keys position of specific command (this is implementation details), instead the specific command that implements this interface should define keys position for itself.

vladvildanov (changes requested):

command.go
```diff
@@ -202,6 +210,22 @@ func (cmd *baseCmd) readRawReply(rd *proto.Reader) (err error) {
 	return err
 }
 
+// getInterleavedArguments returns arguments at even indices starting from index 1.
+func (cmd *baseCmd) getInterleavedArguments() []string {
```
vladvildanov, line 214: Nice! Make total sense

string_commands.go
```diff
@@ -32,37 +32,43 @@ type StringCmdable interface {
 
 func (c cmdable) Append(ctx context.Context, key, value string) *IntCmd {
 	cmd := NewIntCmd(ctx, "append", key, value)
+	cmd.keys = append(cmd.keys, key)
```
vladvildanov, line 35: Why do you need to append key here? You pass key as an argument in constructor `NewIntCmd(ctx, "append", key, value)`, so it's expected that by calling `cmd.Keys()` you'll get back your keys.

vladvildanov (changes requested):

ndyakov:
@vladvildanov , @ofekshenawa should we close this one? Can someone explain what was the need and the plan for this change?

ofekshenawa:
This PR enables saving the keys used by commands—especially multi-key commands—for use in CSC.

ndyakov:
Hey @ofekshenawa are we planning to continue the work on this one? Should we revisit it after the request-response policy? I do think the command info that we have at hand may help.

ofekshenawa:
Implemented in CSC feature

PR #3132, opened 2024-09-25
Fix field name spellings

### LINKED ISSUE #3150: Typo in FT.CREATE command (SEPARATOR vs SEPERATOR)
The keyword for separator is `SEPARATOR` rather than `SEPERATOR`.

## Expected Behavior

https://redis.io/docs/latest/commands/ft.create/

> SEPARATOR {sep} for TAG attributes, indicates how the text contained in the attribute is to be split into individual tags. The default is ,. The value must be a single character.

## Current Behavior

https://github.com/redis/go-redis/blob/ad131f49b0210b3252648b880921b9f6d2fe452e/search_commands.go#L1006-L1008

## Possible Solution

Fix the typo from `SEPERATOR ` to `SEPERATOR`.

andy-stark-redis:
I don't know if it's as easy as this, but I found a few struct field names that were misspelled, so I've attempted to correct them.

Closes #3150

### DIFF
--- a/search_commands.go
+++ b/search_commands.go
@@ -16,7 +16,7 @@ type SearchCmdable interface {
 	FTAliasAdd(ctx context.Context, index string, alias string) *StatusCmd
 	FTAliasDel(ctx context.Context, alias string) *StatusCmd
 	FTAliasUpdate(ctx context.Context, index string, alias string) *StatusCmd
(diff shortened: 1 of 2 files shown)


vladvildanov:
Thank you @andy-stark-redis. This functionality wasn't yet released, so it's not a breaking change.

vladvildanov (approved):

PR #3134, opened 2024-09-26
Updated package version

vladvildanov:
Updated package version to represent current version

### DIFF
--- a/version.go
+++ b/version.go
@@ -2,5 +2,5 @@ package redis
 
 // Version is the current release version.
 func Version() string {
-	return "9.7.0b1"
+	return "9.7.0-beta.1"
 }
--- a/version.go
+++ b/version.go
@@ -2,5 +2,5 @@ package redis
 
 // Version is the current release version.
 func Version() string {
-	return "9.6.1"
+	return "9.7.0b1"
 }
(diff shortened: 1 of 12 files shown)


version.go
```diff
@@ -2,5 +2,5 @@ package redis
 
 // Version is the current release version.
 func Version() string {
-	return "9.6.1"
+	return "9.7.0-beta.1"
```
ofekshenawa, line 5: Typo? 
Also, we have already released a beta version in this format: 9.7.0-b1, so maybe we want to stick to this format

ofekshenawa (changes requested):
Please note that the version needs to be updated in additional places. Make sure to check all relevant files and documentation where the version is referenced to ensure consistency.
f.e https://github.com/redis/go-redis/pull/2917/files

version.go
vladvildanov: I guess we have to stick to the conventions in go community

https://go.dev/doc/modules/version-numbers

vladvildanov: Previously the format was broken, f.e 9.6.0b1

vladvildanov:
@ofekshenawa But if we update it in submodules they will point to a version that doesn't exists yet right? My IDE is complaining when I'm doing so

Issue #3137, opened 2024-09-30
Pool TLS connections aren't long-lived anymore


rentziass:
👋 We've been investigating an increase in CPU load both in our apps and our Redis clusters for a while now, and after noticing that a big portion of our CPU time was being spent on TLS handshakes we started looking at our connections and noticed exorbitant amounts of [stale connections](https://github.com/redis/go-redis/blob/d9eeed131a32b4aeaf1e1915b6717f8cd2af9124/internal/pool/pool.go?plain=1#L427). Importantly, these are all **TLS connections** from connections pools. We tracked it down to us upgrading to go-redis 9.5.4 and when we downgraded to 9.5.2 we went back to the usual occasional stale connection:

![Image](https://github.com/user-attachments/assets/c98c35b5-4346-43a3-be93-c60a79603889)

## Expected Behavior

Under normal conditions have only a handful of connections being reported as stale.

## Current Behavior

Connections are barely being re-used.

## Possible Solution

It looks like the issue started with [this commit](https://github.com/redis/go-redis/commit/0858ed24e6d08c1940bcb86ccbf8913c761d9ec0), specifically [reading directly](https://github.com/redis/go-redis/blob/0858ed24e6d08c1940bcb86ccbf8913c761d9ec0/internal/pool/conn_check.go#L37) from the `(*tls.Conn).NetConn()`; from [`NetConn()` docs](https://pkg.go.dev/crypto/tls#Conn.NetConn):

> NetConn returns the underlying connection that is wrapped by c. Note that writing to or reading from this connection directly will corrupt the TLS session.

Unfortunately, until [Go supports non-blocking readable connection testing](https://github.com/golang/go/issues/15735) I'm not sure there's a good reliable way to check the health of TLS connections here. Until then reverting this direct read on the underlying connection is probably the way to go?

MySQL driver has also had a [similar issue](https://github.com/go-sql-driver/mysql/issues/1392).

## Steps to Reproduce

1. Create a ClusterClient, connecting to a cluster using TLS (ours is an Azure Enterprise cluster, not that it should make a difference). Configuring the client to have MinIdleConns adds a multiplier to this and makes much more apparent.
2. Observe the number of stale connections grow linearly over time as more and more operations are done
3. Downgrade to `< 9.5.4` (9.5.2 is the first prior available version right now) and observe almost no stale connection (with the cluster in healthy conditions of course

## Detailed Description

[...]

PR #3138, opened 2024-10-01
Remove direct read from TLS underlying conn

rentziass:
Addresses #3137.

This essentially reverts https://github.com/redis/go-redis/commit/0858ed24e6d08c1940bcb86ccbf8913c761d9ec0, to avoid corrupting TLS sessions by directly reading from the underlying `net.Conn`. This is blocking upgrades to `> 9.5.2` if using TLS.

`conn_check.go` is quite different on the [v9.6 branch](https://github.com/redis/go-redis/blob/v9.6/internal/pool/conn_check.go) ([it changed here](https://github.com/redis/go-redis/pull/3066/files)), should I open another PR to fix that as well?

### DIFF
--- a/internal/pool/conn_check.go
+++ b/internal/pool/conn_check.go
@@ -3,7 +3,6 @@
 package pool
 
 import (
-	"crypto/tls"
 	"errors"
 	"io"
 	"net"
@@ -17,10 +16,6 @@ func connCheck(conn net.Conn) error {
 	// Reset previous timeout.
 	_ = conn.SetDeadline(time.Time{})
 
-	// Check if tls.Conn.
-	if c, ok := conn.(*tls.Conn); ok {
-		conn = c.NetConn()
-	}
 	sysConn, ok := conn.(syscall.Conn)
 	if !ok {
 		return nil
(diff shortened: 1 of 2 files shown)


rentziass:
@ofekshenawa @vladvildanov sorry for the direct ping, would anyone of you be able to have a look at this please?

vladvildanov:
@rentziass Hey! Thanks for reaching us out! Just for the context, currently it reads from TLS underlying socket, so it means that it reads unencrypted data?

rentziass:
@vladvildanov 👋 it's all in the [issue](https://github.com/redis/go-redis/issues/3137) but the tldr is that reading from `(*tls.Conn).NetConn()` directly corrupts TLS sessions ([from NetConn() docs](https://pkg.go.dev/crypto/tls#Conn.NetConn)) and we found ourselves creating millions of connections versus the usual few thousands because of that. To be honest I'm surprised no other cluster user noticed this, it puts some noticeable pressure on the Redis cluster CPU. We thought it was an organic increase in load but it was just the constant re-creation of connections.

vladvildanov (approved):

Issue #3139, opened 2024-10-01
Can't get UUID with Scan()


lordspinach:
## Expected Behavior

Can use UUID type for storing structs in hash tables

## Current Behavior

Got `invalid UUID length: 16` error from (uuid *UUID) UnmarshalText

## Possible Solution

Add `encoding.BinaryUnmarshaler` to switch case at `/internal/hscan/structmap.go:107`

## Steps to Reproduce

1. Create structure with UUID field
2. Set it to hash table with HSet()
3. Get it back with HGetAll()
4. Try to write given map[string]string to structure with Scan()

## Detailed Description

I am facing a problem related to hash tables. I want to store objects in it, so I use HSet. Then `encoding.BinaryMarshaler` is used to marshal the UUID, but when I try to get the object back using Scan(), `encoding.TextUnmarshaler` is used instead of `encoding.BinaryUnmarshaler` to unmarshal it. 
This is a incorrect use of these interfaces. As it is written in the description of these interfaces 

> UnmarshalBinary must be able to decode the form generated by MarshalBinary 

And

> UnmarshalText must be able to decode the form generated by MarshalText

They are not interchangeable and Binary should be used with Binary and Text with Text

zeze1004:
Can I contribute to this by adding support for `encoding.BinaryUnmarshaler` in the Scan function?

ndyakov:
@zeze1004 kindly pinging you for https://github.com/redis/go-redis/pull/3144.

ostheperson:
From my investigation the issue is that go-redis writes uuid.UUID using MarshalBinary(), but then reads it back with UnmarshalText(). This causes a mismatch, as UnmarshalText() expects a 36-character string but receives a 16-byte binary.

While go-redis documentation suggests implementing encoding.BinaryMarshaler for custom types, for string-like data like uuid.UUID in hash fields, the order of preference during writing matters.

Proposal: Modify the internal/proto/writer.go  WriteArg() to prioritize using encoding.TextMarshaler if a type implements it, before checking for encoding.BinaryMarshaler. This ensures uuid.UUID is consistently stored as a 36-character string.

ostheperson:
From investigation it seems uuid.UUID is written using MarshalBinary(), but it's read back with UnmarshalText() causeing a mismatch, as UnmarshalText() expects a 36-character string but receives a 16-byte binary.

While documentation suggests implementing encoding.BinaryMarshaler for custom types, the order of what's preferred when writing matters.

[...]

PR #3144, opened 2024-10-03
feat: add encoding.BinaryUnmarshaler in Scan

zeze1004:
issue: #3139 

Fixed issue where UUID fields could not be unmarshaled correctly.


zeze1004:
Hi there, I was wondering if you could take a moment to review this PR when you have the time. Thanks! cc. @ofekshenawa, @monkey92t

ndyakov:
This one looks good to me. @zeze1004 do you think we can add full test with write and read from/to redis instance?

ndyakov (changes requested):
@zeze1004 please add integration test to verify the data is correctly stored and retrieved from/to redis.

zeze1004:
> @zeze1004 please add integration test to verify the data is correctly stored and retrieved from/to redis.

Sorry, I just checked, I'll add the test and mention it

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/bsm/gomega v1.27.10
 	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f
+	github.com/google/uuid v1.6.0
```
justinhwang, line 10: Should probably avoid adding a dependency just for tests, that would force people who import `go-redis` to import this as well

PR #3146, opened 2024-10-07
Support json resp 2

ofekshenawa:


json.go
```diff
@@ -60,7 +60,7 @@ type JSONArrTrimArgs struct {
 type JSONCmd struct {
 	baseCmd
 	val      string
-	expanded []interface{}
+	expanded interface{}
```
vladvildanov, line 63: Isn't it a breaking change?

json_test.go
```diff
@@ -17,644 +19,800 @@ var _ = Describe("JSON Commands", Label("json"), func() {
...
-
-			cmd2 := client.JSONArrAppend(ctx, "append2", "$..a", 10)
-			Expect(cmd2.Err()).NotTo(HaveOccurred())
-			Expect(cmd2.Val()).To(Equal([]int64{2, 3}))
+	setupRedisClient := func(protocolVersion int) *redis.Client {
+		return redis.NewClient(&redis.Options{
+			Addr:          "localhost:6379",
+			DB:            0,
+			Protocol:      protocolVersion,
+			UnstableResp3: true,
```
vladvildanov, line 27: Are we using unstable flag because of test cases with JSON + Search or it's needed for JSON standalone?

ofekshenawa: Yes, only for JSON + SEARCH test

vladvildanov (approved):

PR #3148, opened 2024-10-10
Allow for tracking state in Limiter

AndreasBergmeier6176:
Extend Allow and ReportResult functions to handle a Context. Allow can override the passed in Context.
The returned Context is then further passed down to ReportResult. Using this Context it is then possible to store values/track state between Allow and ReportResult calls.

Without this tracking HalfOpen/Generation state is hard to implement efficiently for Circuit Breakers.


ndyakov:
@AndreasBergmeier6176  Returning context from `getConn` doesn't look good to me. Let's see if we can find another way to have a meaningful context in the limiter.

PR #3169, opened 2024-10-23
fix(redisotel): upgrade the semconv version and optimize some attribute information 

flc1125:
https://opentelemetry.io/docs/specs/semconv/database/redis/


flc1125:
It will affect the upgrade of the Go version and is temporarily closed.

PR #3174, opened 2024-10-29
bug: fix Query sentinel changed to concurrent query to avoid context … 

kwenzh:
Switch from sequential sentinel query to concurrent query to avoid context exhaustion


ndyakov:
@kwenzh you got yourself in an interesting race condition :D Feel free to resolve it and we can discuss this approach.

PR #3176, opened 2024-10-29
Support Probabilistic commands with RESP 2 protocol

ofekshenawa:
Support the following modules with RESP 2 protocol:
* Bloom
* Cuckoo
* CMS
* TopK
* T-Digest

### DIFF
--- a/probabilistic.go
+++ b/probabilistic.go
@@ -321,31 +321,33 @@ func (cmd *BFInfoCmd) Result() (BFInfo, error) {
 func (cmd *BFInfoCmd) readReply(rd *proto.Reader) (err error) {
 	result := BFInfo{}
 
+	// Create a mapping from key names to pointers of struct fields
+	respMapping := map[string]*int64{
+		"Capacity":                 &result.Capacity,
+		"CAPACITY":                 &result.Capacity,
+		"Size":                     &result.Size,
+		"SIZE":                     &result.Size,
+		"Number of filters":        &result.Filters,
+		"FILTERS":                  &result.Filters,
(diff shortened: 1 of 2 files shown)


probabilistic.go
```diff
@@ -319,37 +319,67 @@ func (cmd *BFInfoCmd) Result() (BFInfo, error) {
 }
 
 func (cmd *BFInfoCmd) readReply(rd *proto.Reader) (err error) {
-	n, err := rd.ReadMapLen()
+	result := BFInfo{}
+
+	readAndAssignValue := func(key string) error {
```
vladvildanov, line 324: I propose to simplify it a bit and reduce code repetions and redability by using a references to fields instead of `switch` operator. Here's an example how it might be possible.
```
func main() {
	result := redis.BFInfo{}
	key := "Number of filters"

	// Keep references to the specified field.
	respMapping := map[string]*int64{
		"Capacity":          &result.Capacity,
		"CAPACITY":       &result.Capacity,
		"Size":              &result.Size,
		"SIZE":              &result.Size,
		"Number of filters": &result.Filters,
		"Filters":           &result.Filters,
	}

	ref := respMapping[key]
	*ref = 128 // Here *ref, err = rd.ReadInt();...

	fmt.Println(result.Filters)
}
```

vladvildanov (changes requested):

probabilistic.go
vladvildanov: In general, I would say it's better to have a separate type like `BfInfoMapper`, but for now we may keep it as a part of read method

vladvildanov (approved):

PR #3182, opened 2024-11-04
added `Do` method for raw query by single conn from `pool.Conn()`

Oleglacto:


Oleglacto:
Hello! I want to call `Do` method when I got Conn from redis.Pool. This PR can help me do this

Oleglacto:
@vmihailenco can u make review please?

ndyakov:
@Oleglacto can you please describe your usecase and add tests?

Oleglacto:
@ndyakov It is necessary to execute a raw query on a connection from the pool. Currently, this capability is not available, and workarounds such as `Get`, `Set`, etc., have to be used.

ndyakov:
@Oleglacto on top of my head i cannot think why this should not be possible. Any ideas? How will this behave if you are working with a connection that is in a pub/sub state?

do_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	AfterEach(func() {
+		client.Close()
+	})
+
+	It("should pong with Do cmd", func() {
+		result := client.Conn().Do(ctx, "PING")
+		Expect(result.Result()).To(Equal("PONG"))
+		Expect(result.Err()).NotTo(HaveOccurred())
+	})
```
ndyakov, line 26: test is failing

ndyakov (changes requested):

Oleglacto:
@ndyakov Thanks for waiting. Type Conn from [redis.go](https://github.com/redis/go-redis/blob/master/redis.go#L820) doesn't  support pub/sub. I added  this method for cmdble interface and it will works fine. 

I will trying to fix my test.

ndyakov:
@Oleglacto any specific reason why this is not in `commands.go` ? I don't see the need to add additional file for it.

Oleglacto:
I created a new file because I didn't know where to add the changes.

I'll move it now

ndyakov (approved):

PR #3190, opened 2024-11-15
fix: Fixed issue with context cancelled error leading to connection spikes on Primary instances

EXPEbdodla:
Issue: After upgrading from 9.5.1 to 9.7.0, we noticed the spikes in connections to Master/ Primary nodes and also reads are happening from Master nodes. Also noticed increase in pool_conn_total_current metrics. 

Environment: AWS Cloud, Elasticache Redis with Cluster Mode

Cause: After debugging, we noticed that nodes are marked as failed when the `context.Cancelled` error is raised https://github.com/redis/go-redis/blob/930d904205691ff06104fcc3ac108177077def35/osscluster.go#L1324.  This is introduced in v9.5.2. 

We tested this by deploying the change from my Fork and noticed improvements. Elasticache Cluster Current Connections Screenshot:
image

FYI: My first PR to go-redis. Happy to fix if any concerns.

### DIFF
--- a/error.go
+++ b/error.go
@@ -38,6 +38,15 @@ type Error interface {
 
 var _ Error = proto.RedisError("")
 
+func isContextError(err error) bool {
+	switch err {
+	case context.Canceled, context.DeadlineExceeded:
+		return true
+	default:
+		return false
+	}
+}
+
 func shouldRetry(err error, retryTimeout bool) bool {
 	switch err {
 	case io.EOF, io.ErrUnexpectedEOF:
--- a/osscluster.go
+++ b/osscluster.go
@@ -1350,7 +1350,9 @@ func (c *ClusterClient) processPipelineNode(
 	_ = node.Client.withProcessPipelineHook(ctx, cmds, func(ctx context.Context, cmds []Cmder) error {
 		cn, err := node.Client.getConn(ctx)
 		if err != nil {
-			node.MarkAsFailing()
+			if !isContextError(err) {
+				node.MarkAsFailing()
+			}
 			_ = c.mapCmdsByNode(ctx, failedCmds, cmds)
 			setCmdsErr(cmds, err)
 			return err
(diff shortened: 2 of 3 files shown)


EXPEbdodla:
can I get a review on this PR please? Thank you.

ndyakov:
@EXPEbdodla looking into it

ndyakov:
Looks related to https://github.com/redis/go-redis/issues/3282

error.go
```diff
@@ -38,6 +38,15 @@ type Error interface {
 
 var _ Error = proto.RedisError("")
 
+func isContextError(err error) bool {
+	switch err {
+	case context.Canceled, context.DeadlineExceeded:
```
fishy, line 43: I'm not familiar with the code to know how the errors get passed here, but this approach would only match directly `context.Canceled` and `context.DeadlineExceeded`, it won't match error wrapping those errors.

In general (https://pkg.go.dev/errors#section-documentation) we should use `errors.Is(err, context.Canceled)` over `err == context.Canceled`, unless there's a reason we only want to match direct exact errors.

(not sure whether this is the reason this PR does not fix the issue we saw in https://github.com/redis/go-redis/issues/3282#issuecomment-2672158309)

ndyakov: Generally errors are not wrapped in the `go-redis` package and this error should occur within the package. I do agree that with the new version there are tons of things that can be improved, error handling included. Currently we are focused on improving the stability, testing infrastructure , bug fixes and redis 8 support. Once those are addressed we will move to other improvements that are nice to have... I will continue looking into your issue and continue the discussion in its thread.

EXPEbdodla:
> @EXPEbdodla let's add a test for this change and we are good to go with the pr.

Thanks @ndyakov . I'll look into adding tests for this.

EXPEbdodla:
> > @EXPEbdodla let's add a test for this change and we are good to go with the pr.
> 
> Thanks @ndyakov . I'll look into adding tests for this.

@ndyakov Added tests to this PR.

ndyakov (approved):

PR #3191, opened 2024-11-21
Test against Redis CE

ofekshenawa:


.github/workflows/build.yml
```diff
@@ -42,4 +42,47 @@ jobs:
...
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+      
+      # Set up Docker Compose environment
+      - name: Set up Docker Compose environment
+        run: |
+          docker compose --profile all up -d --build
```
ggivo, line 75: I think we don't need `--build` since we use pre-build images

.github/workflows/build.yml
```diff
@@ -42,4 +42,47 @@ jobs:
...
+      
+      - name: Run tests
+        env:
+          USE_PREBUILT_REDIS: "true"
```
ggivo, line 79: nit: I will suggest something like  "USE_CONTAINERIZED_REDIS"

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+services:
+
+  redis:
```
ggivo, line 5: I will suggest using a bit more descriptive name 'redis' -> 'redis-standalone'
In case we add additional services later  e.g. redis-cluster, redis-master-slave...

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    image: redislabs/client-libs-test:8.0-M02
+    container_name: redis-standalone
+    environment:
+      - REDIS_CLUSTER=no
+      - PORT=6379
+      - TLS_PORT=6666
+    command: --loadmodule /usr/local/lib/redis/modules/redisbloom.so --loadmodule /usr/local/lib/redis/modules/redisearch.so --loadmodule  /usr/local/lib/redis/modules/redistimeseries.so --loadmodule /usr/local/lib/redis/modules/rejson.so 
+    ports:
+      - 6379:6379
+      - 6380:6379
```
ggivo, line 15: Is this intentional? Exposing the same Redis instance on two different ports?

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+      - 6666:6666 # TLS port
+    volumes:
+      - "./dockers/standalone:/redis/work"
```
ggivo, line 18: having the service name as part of the volume can help during troubleshooting
'./dockers/standalone' -> `./dockers/redis-standalone/work`

main_test.go
```diff
@@ -82,8 +83,9 @@ var _ = BeforeSuite(func() {
 	}
 	var err error
 	RECluster, _ = strconv.ParseBool(os.Getenv("RE_CLUSTER"))
-
-	if !RECluster {
+	USE_PREBUILT_REDIS, _ = strconv.ParseBool(os.Getenv("USE_PREBUILT_REDIS"))
+	// panic("RECluster: " + strconv.FormatBool(RECluster) + " USE_PREBUILT_REDIS: " + strconv.FormatBool(USE_PREBUILT_REDIS))
```
ggivo, line 87: nit :  Remove commented code if no longer needed

ggivo (commented):
Hey, looks good.
Left some minor questions/notes.

ggivo (commented):
LGTM

PR #3197, opened 2024-11-25
Update redisotel to use semconv and add operation

urdarinda:


ndyakov:
@urdarinda let's not update the go version in this PR. We will are looking into improving the test infra and releasing support for redis 8, once this is done we can upgrade the version and verify things won't break : )

urdarinda:
Sure, makes sense @ndyakov . I'll try to see if I can do this with the existing version otherwise will close and reopen at a later time

PR #3201, opened 2024-12-03
SortByWithCount FTSearchOptions fix

Cgol9:
In search_commands.go, if the `SortByWithCount` option is true, the query argument that is added is misspelled. it is currently `WITHCOUT`, but should be `WITHCOUNT`.

`if options.SortByWithCount {
				queryArgs = append(queryArgs, "WITHCOUNT")
			}`
			
This change fixes the spelling.


ofekshenawa (approved):

Cgol9:
@ofekshenawa can a new tag be created for this change?

PR #3213, opened 2024-12-19
Enable dialect 2 on default

ofekshenawa:


ndyakov:
@ofekshenawa since this is a breaking change, let's add note in the README (bolded) that the default dialect for search is 2

ndyakov (approved):

PR #3219, opened 2025-01-05
Fix race condition in clusterNodes.Addrs()

shawnwgit:
Resolve a Race Condition in clusterNodes.Addrs() Method. 
The clusterNodes.Addrs() method previously returned a reference to a string slice, which introduced the potential for concurrent read operations by callers while the slice was being modified, potentially leading to undefined behavior. This race condition could occur during GC or other operations affecting the slice.

Fix: Updated the clusterNodes.Addrs() method to create and return a copy of the original addrs slice. By ensuring that callers receive an independent copy, we eliminate the possibility of concurrent access issues.

Fixes https://github.com/redis/go-redis/issues/3218


ndyakov (approved):
Great catch, thank you!

Issue #3220, opened 2025-01-08
When using Resp 2 FTAggregateWithArgs does not return an error when using the "as" option


michaeltch:
I have provided an example program that doesn't panic when the command run on the database should return an error.

## Expected Behavior

Panic with error message `(error) Unknown argument `AS` at position 4 for <main>`

## Current Behavior

The resulting pointer is nil and error is empty causing no panic.
`result: 0x0`

## Steps to Reproduce

```
package main

import (
	"context"
	"strings"

	"github.com/redis/go-redis/v9"
)

func createIndexAndLoadData(client *redis.Client, ctx context.Context) {
	if _, err := client.FTCreate(ctx,
		"temp",
		&redis.FTCreateOptions{Prefix: []interface{}{"user:"}},
		&redis.FieldSchema{FieldName: "__key", FieldType: redis.SearchFieldTypeText},
	).Result(); err != nil && !strings.Contains(err.Error(), "Index already exists") {
		panic(err)
	}

	if _, err := client.HSet(ctx, "user:1", []string{"__key", "v1"}).Result(); err != nil {
		panic(err)
	}
}

func runAggregateQuery(client *redis.Client, ctx context.Context) {
	filterOpts := &redis.FTAggregateOptions{
		Load: []redis.FTAggregateLoad{
			{
				Field: "@__key",
				As:    "key",
			},
		},
	}
	result, err := client.FTAggregateWithArgs(ctx, "temp", "*", filterOpts).Result()
	if err != nil {
		panic(err)
	}
	println("result:", result)
}

func main() {
	ctx := context.Background()
	client := redis.NewClient(&redis.Options{
		Addr:     "localhost:6379",
		Protocol: 2,
	})

	createIndexAndLoadData(client, ctx)

	runAggregateQuery(client, ctx)
}
```

Removing the `As` field fixes the issue and a proper response is returned.

## Context (Environment)

Attempting to recreate a separate issue. The database is OSS 6.2.7.

ndyakov:
The merge is fixed and we are preparing 9.7.1 release with the fixes.

PR #3224, opened 2025-01-09
fix(search&aggregate):fix error overwrite and typo  #3220

bitsark:
* LOAD has NO AS param(https://redis.io/docs/latest/commands/ft.aggregate/)

* fix typo: WITHCOUT -> WITHCOUNT


bitsark:
Fixes #3220

ndyakov:
@bitsark the RE build is falling, a test related to ft.aggregate states it expects an error, but got nil. Seems related, can you look at it?

bitsark:
> @bitsark the RE build is falling, a test related to ft.aggregate states it expects an error, but got nil. Seems related, can you look at it?

sure, i will check the problem

ndyakov:
@bitsark I do see `LOAD {field} AS {name}` examples in https://github.com/RediSearch/RediSearch/blob/e155b7db40eeff5f29b47c6308810b2dda906c7c/docs/docs/indexing/_index.md?plain=1#L608 , and with a local test I can see the AS parameter working. Not sure if it is correctly implemented in the library, this is something we can add a test for ?

bitsark:
> @bitsark I do see `LOAD {field} AS {name}` examples in https://github.com/RediSearch/RediSearch/blob/e155b7db40eeff5f29b47c6308810b2dda906c7c/docs/docs/indexing/_index.md?plain=1#L608 , and with a local test I can see the AS parameter working. Not sure if it is correctly implemented in the library, this is something we can add a test for ?

@ndyakov i will recheck this problem

bitsark:
> @bitsark I do see `LOAD {field} AS {name}` examples in https://github.com/RediSearch/RediSearch/blob/e155b7db40eeff5f29b47c6308810b2dda906c7c/docs/docs/indexing/_index.md?plain=1#L608 , and with a local test I can see the AS parameter working. Not sure if it is correctly implemented in the library, this is something we can add a test for ?

@ndyakov the bug fixed in the pr is like:
`FT.AGGREGATE itemIdx '*' LOAD 4 as aliasLoadName name $.price AS originalPrice `
is not same as the example.

ndyakov:
@bitsark i can see the **As** field removed in the diff. I do think the calculation of the arguments look odd and is probably broken, but the **As** should be left since it is possible to have an alias for a loaded field, don't you think? Can you provide an example go code to reproduce the bug you are referring to?

bitsark:
> @bitsark i can see the **As** field removed in the diff. I do think the calculation of the arguments look odd and is probably broken, but the **As** should be left since it is possible to have an alias for a loaded field, don't you think? Can you provide an example go code to reproduce the bug you are referring to?

@ndyakov , I reproduce the bug now. you are right.  I misunderstood the bug. I will  commit a new fix.

ndyakov (approved):

ndyakov (dismissed):
@bitsark let's fix the tests. I would suggest trying to update the RE version instead of having such condition in the tests.

ndyakov:
@bitsark merged an update to fix the linter.

ndyakov (approved):

Issue #3238, opened 2025-01-17
Add Go 1.22.x and 1.23.x and drop Go 1.19.x


ozennou:
- Update Go versions: Add support for Go 1.22.x and 1.23.x in the build.yml workflow, while removing support for the outdated and unsupported Go 1.19.x version. This ensures compatibility with newer Go versions and keeps the build process up-to-date.

- Cache dependency files: Implement caching for Go module dependencies (e.g., go.sum, go.mod, etc.) to significantly speed up pipeline execution. Caching will reduce redundant downloads of dependencies across builds, improving overall efficiency.

=> i will create a PR.

ndyakov:
Hello @ozennou,
Feel free to review https://github.com/redis/go-redis/pull/3247 regarding the newer go versions in the CI workflow.

ndyakov:
Closing since https://github.com/redis/go-redis/pull/3247 was merged

ndyakov:
Oops, I closed this too early. @ozennou if you would like to look into the caching part feel free to do so.

PR #3239, opened 2025-01-17
fix: Fixed issue adding Go 1.22.x and 1.23.x and drop Go 1.19.x

ozennou:
Update Go versions: Add support for Go 1.22.x and 1.23.x in the build.yml workflow, while removing support for the outdated and unsupported Go 1.19.x version. This ensures compatibility with newer Go versions and keeps the build process up-to-date.
#3238

### DIFF
--- a/.github/workflows/test-redis-enterprise.yml
+++ b/.github/workflows/test-redis-enterprise.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.21.x]
+        go-version: ["1.21.x", "1.22.x", "1.23.x"]
         re-build: ["7.4.2-54"]
 
     steps:
--- a/.github/workflows/test-redis-enterprise.yml
+++ b/.github/workflows/test-redis-enterprise.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ["1.21.x", "1.22.x", "1.23.x"]
+        go-version: ["1.21.x"]
         re-build: ["7.4.2-54"]
 
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,9 +44,6 @@ jobs:
         with:
           files: coverage.txt
           token: ${{ secrets.CODECOV_TOKEN }}
-
-      - name: check the *.sum files location
-        run: ls -la **/*.sum
   
   test-redis-ce:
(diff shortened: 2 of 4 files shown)


ndyakov:
Based on https://go.dev/doc/devel/release `1.20.x` and `1.21.x` are not supported as well. If it is decided to drop support for 1.19, does it make sense for the https://github.com/redis/go-redis/blob/master/go.mod#L3 to be updated as well? It still states 1.18, but if tests are not executed on 1.18 and 1.19, maybe stating 1.20 as min versions should be done.

ozennou:
Thank you for pointing that out! If support for 1.18 and 1.19 is being dropped, it does make sense to update the go.mod file to reflect 1.20 as the minimum version. I can include this change in the PR or handle it separately—let me know your preference!

urdarinda:
qq, should this drop support for 1.20 and 1.21 as well?

ozennou:
> qq, should this drop support for 1.20 and 1.21 as well?

True, I think I should add only the 1.22 and 1.23 without dropping 1.19

urdarinda:
> > qq, should this drop support for 1.20 and 1.21 as well?
> 
> True, I think I should add only the 1.22 and 1.23 without dropping 1.19

What I meant was, we should probably drop support for 1.20 and 1.21 as well. maybe

ozennou:
Got it, thanks!

urdarinda:
@ndyakov any plans to merge this one?

.github/workflows/test-redis-enterprise.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.21.x]
+        go-version: ["1.21.x"]
```
SoulPancake, line 18: Are quotes required here? @ozennou

ozennou: No, just a best practice

ndyakov:
I will get back to you in the following week. Currently I am playing around
with the test setup. I ran into some failing tests if we decide to change
the go version in the go.mod . For this reason I will postpone the said
change for now.

ndyakov:
@ozennou i worked on some improvements for the tests. Can you check https://github.com/redis/go-redis/pull/3247 and see if it addresses your concerns related to the go versions in the CI matrix?

ndyakov:
Closing in favor of https://github.com/redis/go-redis/pull/3247 that was merged.

PR #3244, opened 2025-01-22
fix(command): add missing `io-thread` key in `client info`

ndyakov:
Redis 8.0 introduces new key `io-thread` in the response
for `client info`. The key needs to be parsed. If an unknown key
is observed, the client will return an error.

This PR also introduces the 8.0-M3 build in the docker compose and a small improvement to the readability.

### DIFF
--- a/command.go
+++ b/command.go
@@ -5114,6 +5114,7 @@ type ClientInfo struct {
 	OutputListLength   int           // oll, output list length (replies are queued in this list when the buffer is full)
 	OutputMemory       int           // omem, output buffer memory usage
 	TotalMemory        int           // tot-mem, total memory consumed by this client in its various buffers
+	IoThread           int           // io-thread id
 	Events             string        // file descriptor events (see below)
 	LastCmd            string        // cmd, last command played
 	User               string        // the authenticated username of the client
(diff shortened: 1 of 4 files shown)


command.go
```diff
@@ -5114,6 +5114,7 @@ type ClientInfo struct {
 	OutputListLength   int           // oll, output list length (replies are queued in this list when the buffer is full)
 	OutputMemory       int           // omem, output buffer memory usage
 	TotalMemory        int           // tot-mem, total memory consumed by this client in its various buffers
+	IoThread           int           // io-thread id
```
vladvildanov, line 5117: Could you please add a test case to verify a new field in response schema?

docker-compose.yml
```diff
@@ -3,7 +3,7 @@
 services:
 
   redis-stanalone:
-    image: redislabs/client-libs-test:8.0-M02
+    image: redislabs/client-libs-test:8.0-M03
```
vladvildanov, line 6: Given the fact that 8.0-M03 brings more things to implement (ACL + CONFIG), I think it makes sense to make it as a separate PR, WDYT?

vladvildanov (changes requested):

command.go
ndyakov: The integration test in https://github.com/redis/go-redis/blob/9f1522a91e666ee40f807945e49396b09e4f87c2/example_test.go#L178 will actually test client info parser. Exactly having that test failing pointed me to the missing key.

docker-compose.yml
ndyakov: I agree.

command.go
vladvildanov: Ah, okay! Counts as testing as well

vladvildanov (approved):

PR #3250, opened 2025-01-28
create entraid credentials provider

ofekshenawa:
- Create Credential Provider for EntraID
- Create Token-Manager

### DIFF
--- a/entra_id/credentials_provider.go
+++ b/entra_id/credentials_provider.go
@@ -1,0 +1,57 @@
+package entra_id
+
+import (
+	"log"
+	"time"
+)
+
+// EntraIdIdentityProvider defines the interface for an identity provider
+type EntraIdIdentityProvider interface {
+	RequestToken(forceRefresh bool) (string, time.Duration, error)
+}
+
+// EntraIdCredentialsProvider manages credentials and token lifecycle
+type EntraIdCredentialsProvider struct {
+	tokenManager *TokenManager
+	isStreaming  bool
+}
+
+// NewEntraIdCredentialsProvider initializes a new credentials provider
+func NewEntraIdCredentialsProvider(idp EntraIdIdentityProvider, refreshInterval time.Duration, telemetryEnabled bool) *EntraIdCredentialsProvider {
+	refreshFunc := func() (string, time.Duration, error) {
+		return idp.RequestToken(false)
+	}
+
+	tokenManager := NewTokenManager(refreshFunc, refreshInterval, telemetryEnabled)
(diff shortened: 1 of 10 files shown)


entra_id/token_manager.go
```diff
@@ -0,0 +1,129 @@
+package entra_id
```
ndyakov, line 1: In the official [go styleguide](https://google.github.io/styleguide/go/decisions.html#package-names) it is suggested to not use `_` in package names. Let's rename the package to `entraid` or even if we end up with a different name once we move the repo, we can follow the styleguide.

entra_id/credentials_provider.go
```diff
@@ -0,0 +1,57 @@
+package entra_id
+
+import (
+	"log"
+	"time"
+)
+
+// EntraIdIdentityProvider defines the interface for an identity provider
+type EntraIdIdentityProvider interface {
```
ndyakov, line 9: If we think about how this interface will look when called outside of this package it will be `entra_id.EntraIdIdentityProvidor` .  As you can see there is a repetition - `EntraId` which we drop. Here is a brief section related to repetition in the [go styleguide](https://google.github.io/styleguide/go/decisions#repetition)

entra_id/credentials_provider.go
```diff
@@ -0,0 +1,57 @@
...
+	RequestToken(forceRefresh bool) (string, time.Duration, error)
+}
+
+// EntraIdCredentialsProvider manages credentials and token lifecycle
+type EntraIdCredentialsProvider struct {
```
ndyakov, line 14: This type can also be renamed to `CredentialsProvider`

entra_id/go.mod
```diff
@@ -0,0 +1,13 @@
+module entra_id
+
+go 1.18.0
+
+toolchain go1.23.1
```
ndyakov, line 5: we are not required to specify a `toolchain`. If we specify that this module needs `go 1.18` any available toolchain that is with a higher version will be used. Let's remove this line

vladvildanov:
@ofekshenawa @ndyakov is right. All EntraID specific code should be outside of go-redis and should be optional for end user. If user needs EntraID authentication he should add an additional dependency to it's project. However, all the common interfaces (e.g TokenManager, IdentityProvider, CredentialProvider) should be a part of go-redis, so it will be possilble to plug any CredentialProvider from outside. Check "Packaging" section in the design document

ndyakov:
closing, will reimplement

PR #3255, opened 2025-01-29
feat: support Elasticache cluster mode by setting IsClusterMode in config

reeshijoshi:
- This is required because Elasticache supports setting up cluster mode with configuration endpoint. So the check len(Addrs) > 1 is not enough for determining cluster mode.
- If IsClusterMode is true, the client will be created as a cluster client.

### DIFF
--- a/universal.go
+++ b/universal.go
@@ -69,6 +69,9 @@ type UniversalOptions struct {
 	DisableIndentity bool
 	IdentitySuffix   string
 	UnstableResp3    bool
+
+	// IsClusterMode can be used when only one Addrs is provided (e.g. Elasticache supports setting up cluster mode with configuration endpoint).
+	IsClusterMode bool
 }
 
 // Cluster returns cluster options created from the universal options.
@@ -244,7 +247,7 @@ var (
 func NewUniversalClient(opts *UniversalOptions) UniversalClient {
 	if opts.MasterName != "" {
 		return NewFailoverClient(opts.Failover())
-	} else if len(opts.Addrs) > 1 {
+	} else if len(opts.Addrs) > 1 || opts.IsClusterMode {
 		return NewClusterClient(opts.Cluster())
 	}
 	return NewClient(opts.Simple())
--- a/universal_test.go
+++ b/universal_test.go
@@ -60,4 +60,21 @@ var _ = Describe("UniversalClient", func() {
 		a := func() { client.FTInfo(ctx, "all").Result() }
 		Expect(a).ToNot(Panic())
 	})
+
+	It("should connect to clusters if IsClusterMode is set even if only a single address is provided", Label("NonRedisEnterprise"), func() {
+		client = redis.NewUniversalClient(&redis.UniversalOptions{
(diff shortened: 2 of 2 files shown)


ndyakov:
@reeshijoshi thank you for the contribution. Would you be able to add tests for this new option?

reeshijoshi:
> @reeshijoshi thank you for the contribution. Would you be able to add tests for this new option?

Added a test that asserts client type

universal.go
```diff
@@ -69,6 +69,10 @@ type UniversalOptions struct {
 	DisableIndentity bool
 	IdentitySuffix   string
 	UnstableResp3    bool
+
+	// This is required because Elasticache supports setting up cluster mode with configuration endpoint. So the check len(Addrs) > 1 is not enough for determining cluster mode.
```
ndyakov, line 73:
```suggestion
	// IsClusterMode can be used when only one Addrs is provided (e.g. Elasticache supports setting up cluster mode with configuration endpoint).
```

ndyakov (commented):
@reeshijoshi Overall looks good to me, would you mind changing the comment as suggested or anything starting with the `IsClusterMode` as suggested in https://tip.golang.org/doc/comment

universal.go
reeshijoshi: Done! Thanks @ndyakov

ndyakov (approved):
Looks good to me, thanks @reeshijoshi

universal_test.go
```diff
@@ -38,4 +38,13 @@ var _ = Describe("UniversalClient", func() {
 		})
 		Expect(client.Ping(ctx).Err()).NotTo(HaveOccurred())
 	})
+
+	It("should connect to clusters if IsClusterMode is set even if only a single address is provided", Label("NonRedisEnterprise"), func() {
```
vladvildanov, line 42: @ndyakov This test is an integration test, right? So we basically test if handshake had happened and it's possible to connect to cluster via single endpoint?

ndyakov: @reeshijoshi , what Vlad said is valid. Let's check if the `cluster slots` response includes all slots.

reeshijoshi: @ndyakov I think this should only be a test for whether setting `IsClusterMode: true` actually returns an instance of ClusterClient. I don't think verifying whether all the connectivity is setup correctly should be in the scope of this PR?

ndyakov: What this will test is if the cluster client is working properly with a single endpoint. It can be in a separate place, but let’s add such test before including the possibility of initiating cluster client with single endpoint.

ndyakov:
@reeshijoshi would you please add additional test so we can merge this?

universal_test.go
reeshijoshi: @ndyakov apologies I've been away on a vacation. I've pushed a a new test checking length of cluster slots. Let me know if this solves it!

reeshijoshi:
> @reeshijoshi would you please add additional test so we can merge this?

done

ndyakov (approved):

PR #3258, opened 2025-01-31
feat(tests): validate that ConfigSet and ConfigGet work with Modules

ndyakov:
- [x] Add deprecation warnings
- [x] Test `config set` and `config get` for module keys
- [x] Test `config set search-default-dialect` and then `ft.config get default_dialect`
- [x] Found issue with Resp3 parsing of `ft.config get`, resolved it. Since `ft.config` will be dropped, it was not justified to improve the solution any further. 
- [x] added additional tests, including negative (read only and wrong values) 
- [x] Fixed bug with FT.CONFIG GET command returns error in RESP3

### DIFF
--- a/search_commands.go
+++ b/search_commands.go
@@ -831,20 +831,32 @@ func (c cmdable) FTAlter(ctx context.Context, index string, skipInitialScan bool
 	return cmd
 }
 
-// FTConfigGet - Retrieves the value of a RediSearch configuration parameter.
+// Deprecated: FTConfigGet is deprecated in Redis 8.
+// All configuration will be done with the CONFIG GET command.
+// For more information check [Client.ConfigGet] and [CONFIG GET Documentation]
+//
+// Retrieves the value of a RediSearch configuration parameter.
 // The 'option' parameter specifies the configuration parameter to retrieve.
-// For more information, please refer to the Redis documentation:
-// [FT.CONFIG GET]: (https://redis.io/commands/ft.config-get/)
+// For more information, please refer to the Redis [FT.CONFIG GET] documentation.
+//
+// [CONFIG GET Documentation]: https://redis.io/commands/config-get/
+// [FT.CONFIG GET]: https://redis.io/commands/ft.config-get/
 func (c cmdable) FTConfigGet(ctx context.Context, option string) *MapMapStringInterfaceCmd {
 	cmd := NewMapMapStringInterfaceCmd(ctx, "FT.CONFIG", "GET", option)
 	_ = c(ctx, cmd)
 	return cmd
 }
 
-// FTConfigSet - Sets the value of a RediSearch configuration parameter.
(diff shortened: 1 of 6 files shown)


search_commands.go
```diff
@@ -831,20 +831,32 @@ func (c cmdable) FTAlter(ctx context.Context, index string, skipInitialScan bool
...
+//
+// Sets the value of a RediSearch configuration parameter.
 // The 'option' parameter specifies the configuration parameter to set, and the 'value' parameter specifies the new value.
-// For more information, please refer to the Redis documentation:
-// [FT.CONFIG SET]: (https://redis.io/commands/ft.config-set/)
+// For more information, please refer to the Redis [FT.CONFIG SET] documentation.
+//
+// [CONFIG SET Documentation]: https://redis.io/commands/config-set/
+// [FT.CONFIG SET]: https://redis.io/commands/ft.config-set/
 func (c cmdable) FTConfigSet(ctx context.Context, option string, value interface{}) *StatusCmd {
```
vladvildanov, line 860: @ndyakov Is there's a way to deprecate a command in Go?

ndyakov: @vladvildanov the part of the godoc which states `Deprecated:` is how IDEs, other static tools and documentation will know to mark this method as deprecated. There is no other way that I am familiar of. ref: https://go.dev/wiki/Deprecated

vladvildanov: @ndyakov Okay, could we do at least that for FT.CONFIG GET and SET commands, specifying that they're deprecated since 8.0

search_commands.go
```diff
@@ -831,20 +831,32 @@ func (c cmdable) FTAlter(ctx context.Context, index string, skipInitialScan bool
 	return cmd
 }
 
-// FTConfigGet - Retrieves the value of a RediSearch configuration parameter.
+// Deprecated: FTConfigGet is deprecated in Redis 8.
```
ndyakov, line 834: @vladvildanov deprecation comment is here

search_commands.go
```diff
@@ -831,20 +831,32 @@ func (c cmdable) FTAlter(ctx context.Context, index string, skipInitialScan bool
...
+// [CONFIG GET Documentation]: https://redis.io/commands/config-get/
+// [FT.CONFIG GET]: https://redis.io/commands/ft.config-get/
 func (c cmdable) FTConfigGet(ctx context.Context, option string) *MapMapStringInterfaceCmd {
 	cmd := NewMapMapStringInterfaceCmd(ctx, "FT.CONFIG", "GET", option)
 	_ = c(ctx, cmd)
 	return cmd
 }
 
-// FTConfigSet - Sets the value of a RediSearch configuration parameter.
+// Deprecated: FTConfigSet is deprecated in Redis 8.
```
ndyakov, line 850: @vladvildanov since those were missed, I will move them to the last line of the godoc

vladvildanov (approved):

PR #3271, opened 2025-02-08
fix: nil pointer dereferencing in writeArg

aliforever:
- Added example for hset using struct
- Added nil pointer checks to `WriteArg (internal/proto/writer.go:62)` 
- Added test cases

Issue: https://github.com/redis/go-redis/issues/3208


ndyakov:
Hello @aliforever, Looking at the example test, don't you think we should add a check with `omitempty` to verify the `nil` pointers will be retrieved as such? Other than that, lgtm. Will review once we have that test updated.

aliforever:
> Hello @aliforever, Looking at the example test, don't you think we should add a check with `omitempty` to verify the `nil` pointers will be retrieved as such? Other than that, lgtm. Maybe remove the dump package from the test so the linter passes? For some reason it doesn't like it. Will review once we have that test updated.

Hi @ndyakov, I think linter was due to missing go.mod file in commit. 

Regarding omitempty, I added it to a bool struct, is that what you meant?

ndyakov:
@aliforever yes, thank you. Will merge once CI completes.

ndyakov (approved):

Issue #3208, opened 2024-12-12
omitempty tag does not work for empty time.Time struct field when using HSet


gh73962:
Issue tracker is used for reporting bugs and discussing new features. Please use
[stackoverflow](https://stackoverflow.com) for supporting issues.

## Expected Behavior
```go
type MyHash struct { 
    Key1 string `redis:"key1"`;
    Key2 time.Time `redis:"key2,omitempty"` 
}
```
should be
```
hset myhash key1 value1
```
## Current Behavior
```
hset myhash key1 value1 key2 0001-01-01T00:00:00Z
```
## Possible Solution

https://github.com/redis/go-redis/blob/master/commands.go#L142

```go
func isEmptyValue(v reflect.Value) bool {
	switch v.Kind() {
	case reflect.Array, reflect.Map, reflect.Slice, reflect.String:
		return v.Len() == 0
	case reflect.Bool:
		return !v.Bool()
	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
		return v.Int() == 0
	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
		return v.Uint() == 0
	case reflect.Float32, reflect.Float64:
		return v.Float() == 0
	case reflect.Interface, reflect.Pointer:
		return v.IsNil()
++++++++++++++++++++++++++++++++++++++++++++++
        case reflect.Struct:
                return v.IsZero()
++++++++++++++++++++++++++++++++++++++++++++++
	}
	return false
}
```

## Steps to Reproduce

```go
type MyHash struct { 
    Key1 string `redis:"key1"`;
    Key2 time.Time `redis:"key2,omitempty"` 
}

data := MyHash{
     Key1 : "value1",
     Key2 : time.Time{},
}

if err := rdb.HSet(ctx, "key", model1).Err(); err != nil {
	panic(err)
}
```

## Detailed Description

omitempty tag does not work for empty time.Time struct field when using HSet

## Possible Implementation

If you agree with the proposal I made in the title "Possible Solution," I can proceed to complete this PR.

aliforever:
Also if you use `*time.Time` without `omitempty` tag, it would panic with nil pointer dereference. Is that the intended behavior?

SoulPancake:
This is the expected behaviour right now K guess but with go 1.24
We have omitzero which solves this directly

[...]

PR #3273, opened 2025-02-10
feat: func isEmptyValue support time.Time

gh73962:
This PR is based on the solution discussed in the  #3208 issue.


ndyakov:
@gh73962 Thank you for the contribution. The test that you have added would have passed with or without the addition you did. Can we check what will be read so we verify the data. Maybe a raw read would work, since if we unmarshal the response to the object it would be the same as before, won't it?

gh73962:
> @gh73962 Thank you for the contribution. The test that you have added would have passed with or without the addition you did. Can we check what will be read so we verify the data. Maybe a raw read would work, since if we unmarshal the response to the object it would be the same as before, won't it?

Of course, I don't have the capability to perform this kind of test on my side. Changes like this should have comprehensive tests to ensure code quality.

ndyakov:
@gh73962 would you like me to help you add those tests? The CI will run a redis server, you should be able to execute code agains it in the tests (you can check the existing tests for Set and Get). Let me know if I can help you somehow.

gh73962:
> @gh73962 would you like me to help you add those tests? The CI will run a redis server, you should be able to execute code agains it in the tests (you can check the existing tests for Set and Get). Let me know if I can help you somehow.

Of course

ndyakov:
@gh73962 any concrete questions on how to write tests? will you be able to do this on your own?

gh73962:
In my first commit of the `commands_test.go` file, the added test covers a scenario that our team frequently encounters.I need you guide, What additional test cases should I include?

ndyakov:
@gh73962 lets add tests for the following scenarious:
- Add a test with custom struct in the hash that wants to hold a zero value (for example `type RequestsSetting struct { Limit int }` and the limit set to zero. Check if the struct is parsed correctly when the data is requested from redis.
- Add a test which validates that the data that was received from redis is what was written to redis. For example add a test where you set the `time.Time` value and validate that the same is parsed later when reading the data.

gh73962:
> @gh73962 lets add tests for the following scenarios:
> 
>     * Add a test with custom struct in the hash that wants to hold a zero value (for example `type RequestsSetting struct { Limit int }` and the limit set to zero. Check if the struct is parsed correctly when the data is requested from redis.
> 
>     * Add a test which validates that the data that was received from redis is what was written to redis. For example add a test where you set the `time.Time` value and validate that the same is parsed later when reading the data.

add to `commands_test.go` `It("should HSet", func()` ?

ndyakov:
@gh73962 yes.

ndyakov:
@gh73962 can we try with a child struct? Would you like me to write the tests or you would do it?

ndyakov (changes requested):
I don't find the testing sufficient. Please check the discussion so far.

gh73962:
@ndyakov I was sick a few days ago, and I'll start improving these tasks tomorrow.

gh73962:
> @gh73962 can we try with a child struct? Would you like me to write the tests or you would do it?

I can test it out and check the impact on the child struct. Should I write the tests, or will you?

gh73962:
@ndyakov pls review and hset.Scan not support struct
图片

commands.go
```diff
@@ -153,6 +153,8 @@ func isEmptyValue(v reflect.Value) bool {
 		return v.Float() == 0
 	case reflect.Interface, reflect.Pointer:
 		return v.IsNil()
+	case reflect.Struct:
```
ndyakov, line 156: If we consider `struct` as unsupported in https://github.com/redis/go-redis/blob/0df0f3c88785aa16c996089b68b731a4e4a2fdc0/internal/hscan/hscan.go#L21 , should we filter better for `time.Time` here ? @gh73962

gh73962: code will be 
```go
    case reflect.Struct:
        if v.Type() == reflect.TypeOf(time.Time{}) {
            return v.IsZero()
        }
        // Only supports the struct time.Time, 
        // subsequent iterations will follow the func Scan support decoder.

```
What do u think about this change?

gh73962:
> @ndyakov pls review and hset.Scan not support struct 图片

Additional note: hGetAll.Scan does not support parsing into nested structs, not that Scan doesn't support structs in general. The current issue is that hset supports implementing the BinaryMarshaler interface, while Scan fails to support nested BinaryUnmarshaler, creating an asymmetry in functionality. I believe redisv9's hset and scan are very developer-friendly, which is why my team migrated from v6 to v9. I hope go-redis can align the capabilities of hset and scan in the future. Thank you very much.
@ndyakov

gh73962:
@ndyakov hi，pls check this PR

gh73962:
@ndyakov pls review

ndyakov (approved):
Added one additional case, overall looks good.

PR #3274, opened 2025-02-10
chore: extract benchmark tests

ndyakov:


docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 
 services:
   redis:
-    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:7.4.1}
+    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:rs-7.4.0-v2}
```
kiryazovi-redis, line 5: will you do a matrix for that in the future?
7.4, 7.2, 8.0?

dockers/sentinel.conf
```diff
@@ -1,5 +1,5 @@
 sentinel resolve-hostnames yes
-sentinel monitor go-redis-test redis 6379 2
+sentinel monitor go-redis-test 127.0.0.1 9121 2
```
kiryazovi-redis, line 2: I imagined more configurations, but I guess the go-client tests require pretty straigth forward redis servers.

docker-compose.yml
ndyakov: @kiryazovi-redis there is a matrix, this is the default value here.

dockers/sentinel.conf
ndyakov: @kiryazovi-redis some configuration is in the test suites themselves. They are setting up a topology that they will check agains later, that is why I left most of it in the suite BeforeAll

kiryazovi-redis (approved):
I went through your automation once again and it looks ok!
I didn't go in-depth with the go-code itself, I hope thats ok with, sorry if I miss something important out.

.github/workflows/build.yml
```diff
@@ -27,15 +33,38 @@ jobs:
...
+        env:
+          REDIS_VERSION: ${{ matrix.redis-version }}
+          CLIENT_LIBS_TEST_IMAGE: "redislabs/client-libs-test:${{ matrix.redis-version }}"
+        run: |
+          set -e
+          redis_version_np=$(echo "$REDIS_VERSION" | grep -oP '^\d+.\d+')
+          
+          # Mapping of redis version to redis testing containers
+          declare -A redis_version_mapping=(
+            ["8.0-M03"]="8.0-M04-pre"
```
vladvildanov, line 46: Are you running module tests against 8.0 as well?

vladvildanov (approved):

PR #3280, opened 2025-02-18
move regexp.MustCompile close to call

alingse:
I'm validating an idea for my linter. I checked the top 2500 projects on github and found this.

All project result can be found here -> https://github.com/alingse/sundrylint/issues/7

I think move the `regexp.MustCompile` out has two advantages

1. Performance benefits
2. Fail early with faulty regexp string

### DIFF
--- a/command.go
+++ b/command.go
@@ -5484,10 +5484,6 @@ func (cmd *InfoCmd) String() string {
 	return cmdString(cmd, cmd.val)
 }
 
-var (
-	moduleRe = regexp.MustCompile(`module:name=(.+?),(.+)$`)
-)
-
 func (cmd *InfoCmd) readReply(rd *proto.Reader) error {
 	val, err := rd.ReadString()
 	if err != nil {
@@ -5506,6 +5502,7 @@ func (cmd *InfoCmd) readReply(rd *proto.Reader) error {
 			cmd.val[section] = make(map[string]string)
 		} else if line != "" {
 			if section == "Modules" {
+				moduleRe := regexp.MustCompile(`module:name=(.+?),(.+)$`)
 				kv := moduleRe.FindStringSubmatch(line)
 				if len(kv) == 3 {
 					cmd.val[section][kv[1]] = kv[2]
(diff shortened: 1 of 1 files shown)


ndyakov:
Hello @alingse thank you for the contribution. You are right that this may increase performance (if `info modules` is used often) and also will fail early if the regex is incorrect. At the same time time it introduces a global variable that can be changed from anywhere in the package. Since this method is not expected to be called that often I do not see the benefit outweighing the possible issues this may introduce. What I can suggest is to compile the regex only when used (i.e. if the section is `modules`) which will speed things a little.

command.go
```diff
@@ -5484,6 +5484,10 @@ func (cmd *InfoCmd) String() string {
 	return cmdString(cmd, cmd.val)
 }
 
+var (
+	moduleRe = regexp.MustCompile(`module:name=(.+?),(.+)$`)
+)
+
```
ndyakov, line 5490: Hello @alingse thank you for the contribution. You are right that this may increase performance (if info modules is used often) and also will fail early if the regex is incorrect. At the same time time it introduces a global variable that can be changed from anywhere in the package. Since this method is not expected to be called that often I do not see the benefit outweighing the possible issues this may introduce. What I can suggest is to compile the regex only when used (i.e. if the section is modules) which will speed things a little.

ndyakov (changes requested):

command.go
alingse: 😂  Yes, I also question the necessity of such changes. You can review the all results in the issue 7 of all top 2500 projects—there are too many instances of this usage, and most of them probably don't require the proposed optimization 🤔, How do you think

ndyakov: I cannot speak for the 2500 projects you have tagged, but for this one - I do not think it is needed. I would assume is a case by case optimization that can benefit high-traffic flows. If you would like to improve this repository, take a brief look at the function. The `MustCompile` can be moved within the if and it will improve the performance in theory without exposing a global variable. Feel free to add that change or close this PR.

alingse: 🤔，ok，I will update it later, move the `MustCompile` within the if branch

ndyakov (approved):

command.go
haunt98: Hi @alingse, I think you move `moduleRe = regexp.MustCompile` inside loop will make func more expensive?

PR #3285, opened 2025-02-19
Add test codes for search_commands.go

tomfrombayesians:
Hello!

Since this is my first PR, I added a very small test. Our team is developing Tom, an open-source AI agent that automates code maintenance tasks like writing unit tests. This code was created through collaboration between Tom and me.

Right now, our project is still private, but if you’re interested, you can sign up for the Beta Release Waitlist through [this link](https://tomai.framer.website/) to get notified when it becomes available.

By the way, I ran into an issue when trying to execute the tests as described in the CONTRIBUTE guide. Even after running a Redis server locally and setting the required port, I couldn’t get the full test suite to run properly. To work around this, I created a TestSuite function inside search_commands_test.go to run the Ginkgo test in isolation.

Please let me know if anything needs to be adjusted! Thanks :)


ofekshenawa:
Hey @tomfrombayesians , Thanks for your contribution!
Just a heads up, we’re about to update our testing infrastructure soon, which means we won’t need to run different servers or settings anymore. In addition, we will update the [README](https://github.com/redis/go-redis/pull/3286/files) to be more accurate.
Thanks again:)

ndyakov:
Hello @tomfrombayesians, there is a file `search_test.go` in which you can add your tests. We did merge couple of fixes related to the search that will be released soon (in 9.7.1). As for the testing infrastructure, you can look at https://github.com/redis/go-redis/pull/3274 , but the work on it is still in progress.

search_commands_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"reflect"
+
+	. "github.com/bsm/ginkgo/v2"
+	. "github.com/bsm/gomega"
+	"github.com/redis/go-redis/v9"
+)
+
+var _ = Describe("FTAggregateQuery", func() {
```
ndyakov, line 11: should either be moved in `search_test.go` or move all the `search_test.go` logic here. Feel free to decide which file would you like to use.

ndyakov (changes requested):

search_commands_test.go
tomfrombayesians: I move all to `search_test.go` and also add some label.

search_test.go
```diff
@@ -805,6 +806,72 @@ var _ = Describe("RediSearch commands Resp 2", Label("search"), func() {
...
+		}
+		args := redis.FTAggregateQuery("q", options)
+		Expect(args).To(ContainElement("TIMEOUT"))
+		found := false
+		for _, a := range args {
+			if reflect.DeepEqual(a, 500) {
+				found = true
+				break
+			}
+		}
```
ndyakov, line 871: I assume this can be rewriten as: 
```suggestion
		for i, a := range args {
			if fmt.Sprintf("%s", a) == "TIMEOUT" {
				Expect(fmt.Sprintf("%s", args[i+1])).To(Equal("500"))
				found = true
				break
			}
		}
```

And this way will check the order of arguments as well. Right now any arg can be 500 and this test will pass.

ndyakov: Oh, and added bonus, we can drop reflect as direct dependancy.

ndyakov (dismissed):
Overall looks good to me. Left a suggestion in one of the tests.

Thank you for the contribution, @tomfrombayesians !

search_test.go
```diff
@@ -805,6 +805,73 @@ var _ = Describe("RediSearch commands Resp 2", Label("search"), func() {
...
+	It("should include TIMEOUT when Timeout > 0", Label("search", "ftaggregate"), func() {
+		options := &redis.FTAggregateOptions{
+			Timeout: 500,
+		}
+		args := redis.FTAggregateQuery("q", options)
+		Expect(args).To(ContainElement("TIMEOUT"))
+		found := false
+		for i, a := range args {
+			if fmt.Sprintf("%s", a) == "TIMEOUT" {
+				Expect(fmt.Sprintf("%s", args[i+1])).To(Equal("500"))
```
ndyakov, line 867:
```suggestion
				Expect(fmt.Sprintf("%d", args[i+1])).To(Equal("500"))
```

PR #3298, opened 2025-03-11
fix: connection pool timeout, increase retries

monkey92t:


ndyakov:
@monkey92t this looks good, can we add a test for it?

ndyakov (approved):
lgtm

PR #3299, opened 2025-03-13
Add vector types INT8 and UINT8 test

ofekshenawa:


ndyakov:
@ofekshenawa i see the tests are failing. As far as I know, there is a `redislabs/client-libs-test:8.0-M05-pre` image that may help with the Redis 8 CE.

PR #3307, opened 2025-03-19
fix zank withscores

briliant1991:
to fix zrank withscores command


ndyakov:
Hello @briliant1991, thanks for contributing. Would you please add a test for that change ?

ndyakov:
Hello @briliant1991 , looking at the redis documentation about ZRANK , the argument is actually WITHSCORE not WITHSCORES - https://redis.io/docs/latest/commands/zrank/

PR #3323, opened 2025-03-25
chore: drop ft.profile that was never enabled

ndyakov:
Unfinished code related to `FTProfile` was added a while back. It was never finished and now we have breaking changes with the ft.profile implementation in the server. Dropping the currently commented out implementation in favor of potentially introducing a new one in the future.


ndyakov:
@htemelski as discussed, let's drop this and introduce a simpler implementation.

elena-kolevska (approved):

PR #3325, opened 2025-03-26
fix: redis: nil should not be retried

ndyakov:
`redis:nil` should not be retried


ndyakov:
`redis: nil` will fall in the default return clause which will return `false` anyway.

PR #3327, opened 2025-03-29
Use correct slot for COUNTKEYSINSLOT command

khasanovbi:
Added selection of the correct slot for the CLUSTER COUNTKEYSINSLOT command based on the argument.
This allows to make a request for the number of keys in a slot not for all shards, but for the necessary one.

### DIFF
--- a/internal_test.go
+++ b/internal_test.go
@@ -352,3 +352,27 @@ var _ = Describe("withConn", func() {
 		Expect(client.connPool.Len()).To(Equal(1))
 	})
 })
+
+var _ = Describe("ClusterClient", func() {
+	var client *ClusterClient
+
+	BeforeEach(func() {
+		client = &ClusterClient{}
+	})
+
+	Describe("cmdSlot", func() {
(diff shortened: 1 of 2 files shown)


ndyakov (changes requested):
@khasanovbi thank you for the contribution! Are there other commands that need to be executed on the `args[2]` slot? If so, let's prepare a slice and check if the `arg[1]` is in this list. Also, let's add tests for the change.

khasanovbi:
Hi @ndyakov I'm checked other commands in docs and don't think that there are other commands that should use slot from command arg.

About the tests, I found ClusterGetKeysInSlot, ClusterCountKeysInSlot tests in osscluster_test.go only. 
Can you tell me if this will not be enough and which cases you suggest to test?

khasanovbi:
Fixed

ndyakov:
@khasanovbi thank you. Can you please add a brief description for the PR. I think the rest looks good.

ndyakov (approved):

khasanovbi:
Done

PR #3335, opened 2025-04-06
3021_add_subscriptionMap_log_subscription

ronak4195:
Add subscriptions map to channel stuct. 

This will resolve datatype mismatch in logger of pubsub.go file


ndyakov (changes requested):
Hello @ronak4195 , would you please add tests. I also failed to see where the subscriptions map will be populated?

ndyakov:
@ronak4195 kind reminder for adding tests here.

PR #3337, opened 2025-04-07
Add  integration tests for Redis 8 behavior changes in Redis Search

ofekshenawa:
This commit adds a suite of integration tests to validate the expected behavior changes and bug fixes introduced for Redisearch in Redis Community Edition 8. These tests cover several areas including query syntax, reducer default values, scoring defaults, alias validations, timeout behavior, and stop word handling. 

**IMPORTANT:** This PR adds tests only to verify that the changes occur correctly. It does not implement any modifications to Redisearch itself.


search_commands.go
```diff
@@ -1954,7 +1954,7 @@ func (c cmdable) FTSearchWithArgs(ctx context.Context, index string, query strin
 				args = append(args, "WITHCOUNT")
 			}
 		}
-		if options.LimitOffset >= 0 && options.Limit > 0 {
```
ndyakov, line 1957: If both the `LimitOffset` and `Limit` are left to the default `0` values, we will end up with `LIMIT 0 0` which is described as a special usecase of the limit, to return count of the results instead of the results themselves (search for `LIMIT 0 0` in the docs: https://redis.io/docs/latest/commands/ft.search/ )

ndyakov (changes requested):
I don't think we should allow `LIMIT 0 0` from the ft.search... We can introduce additional option to note that the execution should only count the results (and add `LIMIT 0 0`) but enabling it for all `ft.search` executions will break the default behaviour without setting the `LimitOffset` and `Limit` options since `0` is the default value for those integer. WDYT @ofekshenawa ?

ndyakov:
blocked by https://github.com/redis/go-redis/pull/3338

search_commands.go
```diff
@@ -293,8 +295,9 @@ type FTSearchSortBy struct {
 // More information about the options can be found
 // in the documentation for FT.SEARCH https://redis.io/docs/latest/commands/ft.search/
 type FTSearchOptions struct {
-	NoContent    bool
-	Verbatim     bool
+	NoContent bool
+	Verbatim  bool
+	// NoStopWords is deprecated since redis 8.0
```
ndyakov, line 300: Maybe with this formatting the IDEs will pick it up as deprecated. Let's try.
```suggestion
	// Deprecated: NoStopWords is deprecated since Redis 8.0
```

search_commands.go
```diff
@@ -261,7 +262,8 @@ type FTAggregateOptions struct {
 	WithCursor        bool
 	WithCursorOptions *FTAggregateWithCursor
 	Params            map[string]interface{}
-	DialectVersion    int
+	// Dialect 1,3 and 4 are deprecated since redis 8.0
```
ndyakov, line 265: I assume we still would like to keep the DialectVersion at least till Redis 10, then we can drop it. Let's deprecate this around Redis 9 (add a task somewhere to remind ourselves).

search_commands.go
```diff
@@ -322,8 +325,9 @@ type FTSearchOptions struct {
 	Limit           int
 	// CountOnly sets LIMIT 0 0 to get the count - number of documents in the result set without actually returning the result set.
 	// When using this option, the Limit and LimitOffset options are ignored.
-	CountOnly      bool
-	Params         map[string]interface{}
+	CountOnly bool
+	Params    map[string]interface{}
+	// Dialect 1,3 and 4 are deprecated since redis 8.0
 	DialectVersion int
```
ndyakov, line 331: Same as above comment

search_commands.go
```diff
@@ -440,7 +444,8 @@ type IndexDefinition struct {
 type FTSpellCheckOptions struct {
 	Distance int
 	Terms    *FTSpellCheckTerms
-	Dialect  int
+	// Dialect 1,3 and 4 are deprecated since redis 8.0
+	Dialect int
```
ndyakov, line 448: Same as comment above

ndyakov (approved):
Thank you for addressing the issues and adding the needed tests. I think it is OK to be merged as it is, let's briefly discuss how we would like to handle the deprecations.

search_commands.go
```diff
@@ -293,8 +295,9 @@ type FTSearchSortBy struct {
 // More information about the options can be found
 // in the documentation for FT.SEARCH https://redis.io/docs/latest/commands/ft.search/
 type FTSearchOptions struct {
-	NoContent    bool
-	Verbatim     bool
+	NoContent bool
+	Verbatim  bool
+	// Deprecated: NoStopWords is deprecated since Redis 8.0
 	NoStopWords  bool
```
ndyakov, line 301: Looking at the changelog for redis 8, I did not see deprecation for `NOSTOPWORDS`. @ofekshenawa can you point me to where you found that, maybe I am missing it.

ndyakov (approved):

PR #3348, opened 2025-04-17
Use iter.Seq to iterate over ScanIterator

khasanovbi:
Support for native iterators has been added in go 1.23.
New Vals method added that return keys iterator directly instead of keys through Next, Val pair.
Minimum go version changed to 1.23.

---

> [!NOTE]
> **Medium Risk**
> Public iterator API change with deprecation and new pagination path; behavior must stay aligned with Next/Val across SCAN, HSCAN, and match/empty-page cases.
> 
> **Overview**
> Adds **Go 1.23 `iter.Seq` support** on `ScanIterator` via a new **`Vals(ctx)`** method so callers can use `for val := range iter.Vals(ctx)` instead of the manual **`Next` / `Val`** loop. **`Next` and `Val` are deprecated** but unchanged for existing code.
> 
> **`Vals`** walks the current scan page, fetches further pages when the Redis cursor is non-zero (same cursor-arg handling as **`scan`/`qscan` vs other commands**), and stops on command errors or when the range loop breaks early.
> 
> Examples (`example_test.go`, `del-keys-without-ttl`) now demonstrate the range style; **`iterator_test.go`** mirrors prior scan/hscan/match scenarios for both **`Next`** and **`Vals`** (often via `slices.Collect`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9703b824a842f6a773a7e298fbb3b8ecd9d33901. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


khasanovbi:
https://github.com/redis/go-redis/issues/3347

khasanovbi:
Hmm, interesting why json tests failed...

iterator.go
```diff
@@ -16,6 +17,8 @@ func (it *ScanIterator) Err() error {
 }
 
 // Next advances the cursor and returns true if more values can be read.
+//
+// Deprecated: support for native iterators has been added in go 1.23. Use Vals instead.
 func (it *ScanIterator) Next(ctx context.Context) bool {
```
ndyakov, line 22: The `Vals` itself will be using `Next`. In general I would vote agains marking those as deprecated since they will continue to be used inside the library and if a developer would prefer to use them instead of the `Vals` I don't think we should discourage that by marking `Next` and `Val` as deprecated. 
Let me know what you think.

ndyakov (changes requested):
@khasanovbi thank you for the suggestion, we were looking into updating the go versions but observed that some tests were failing. We didn't have the time to look more into it back then. Since you decided to try, let's directly update to 1.24 and figure out the tests. 
I also have some reservation about marking the `Next` and `Val` as deprecated. You can see them in my comment. I am curious about what you think.

iterator.go
khasanovbi: I think iterator via Vals is easier and more convenient than the combination Next, Val.

When you use Val, you need to know that Next must be called before it. 
When you use Vals, you just iterate through the values.

Also the code of Vals could be much simpler without calls of Next, Val, without pos var handling, without extra cmd.Err check at each Val call.

```go
// Vals returns iterator over key/field at the current cursor position.
func (it *ScanIterator) Vals(ctx context.Context) iter.Seq[string] {
	return func(yield func(string) bool) {
		if it.cmd.Err() != nil {
			return
		}

		for {
			for _, val := range it.cmd.page {
				if !yield(val) {
					return
				}
			}

			// Return if there is no more data to fetch.
			if it.cmd.cursor == 0 {
				return
			}

			// Fetch next page.
			var cursorIndex int
			switch it.cmd.args[0] {
			case "scan", "qscan":
				cursorIndex = 1
			default:
				cursorIndex = 2
			}

			it.cmd.args[cursorIndex] = it.cmd.cursor

			err := it.cmd.process(ctx, it.cmd)
			if err != nil {
				return
			}
		}
	}
}
```

> they will continue to be used inside the library

In PR Next, Val is used only at this place,  I left them in for backward compatibility only.

ndyakov: I don't disagree that an Iterator is more ergonomic. What I wanted to make sure is that we do not deprecate the current way of iterating over the values before version 10 of `go-redis`. If you feel strongly about deprecating, we can postpone this PR for when `go-redis/v10` will be released.

ndyakov (approved):
Thank you @khasanovbi, looks like we can include this in v9!

khasanovbi:
I’d be happy if this were included in v9)

PR #3349, opened 2025-04-17
fix: better error handling when fetching the master node from the sentinels

ndyakov:
- Make sure all sentinel errors are reported if the `masterAddr` was not resolved.
- Resolving the `masterAddr` concurrently may result in read from `errCh` before read from `done` => first check if the `masterAddr` was resolved, and then proceed to report errors.

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -605,11 +605,11 @@ func (c *sentinelFailover) MasterAddr(ctx context.Context) (string, error) {
 	if masterAddr != "" {
 		return masterAddr, nil
 	}
-	errs := make([]error, len(c.sentinelAddrs))
+	errs := make([]error, 0, len(c.sentinelAddrs))
 	for err := range errCh {
 		errs = append(errs, err)
 	}
-	return "", fmt.Errorf("redis: all sentinels specified in configuration are unreachable: %w", errs)
+	return "", fmt.Errorf("redis: all sentinels specified in configuration are unreachable: %w", errors.Join(errs...))
 }
 
 func (c *sentinelFailover) replicaAddrs(ctx context.Context, useDisconnected bool) ([]string, error) {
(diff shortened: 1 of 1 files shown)


kwenzh:
current cancel() is also can stop goruntine.

Do we need to use select-case done in the goroutine to notify other goroutines to terminate the sentinel query early when the master address has already been queried?

the different is do cancel() more times;

kwenzh:
no others question

kwenzh:
how about this : 

https://github.com/redis/go-redis/pull/3350/files

ndyakov:
> current cancel() is also can stop goruntine.
> 
> Do we need to use select-case done in the goroutine to notify other goroutines to terminate the sentinel query early when the master address has already been queried?
> 
> the different is do cancel() more times;

The cancel for the context can be called only if the master is resolved and will close the other requests in this case, the error for closed context will be reported if the master was not resolve. Can you clarify what is the problem you are observing in this PR?

kwenzh:
> The cancel for the context can be called only if the master is resolved and will close the other requests in this case, the error for closed context will be reported if the master was not resolve. Can you clarify what is the problem you are observing in this PR?

Well, I misunderstood

htemelski (dismissed):
Overall, it looks good.
There's one minor nitpick.

Maybe the logic should be inverted with the happy path return being at the end of the function

ndyakov:
> Maybe the logic should be inverted with the happy path return being at the end of the function

In most cases - yes, but not here. If we were able to get the `masterAddr` from any of the sentinels, even if there are any errors for the rest (and there probably will be since if the context was canceled, there will be deadline errors anyway) the resolved `masterAddrs` is what we would like to be returned from this function.

htemelski:
> > Maybe the logic should be inverted with the happy path return being at the end of the function
> 
> In most cases - yes, but not here. If we were able to get the `masterAddr` from any of the sentinels, even if there are any errors for the rest (and there probably will be since if the context was canceled, there will be deadline errors anyway) the resolved `masterAddrs` is what we would like to be returned from this function.

You are right, I didn't take into account the `once.Do` block

htemelski (approved):

PR #3353, opened 2025-04-19
fix: Fix panic caused when arg is nil

frankxjkuang:
script := `return redis.call('get', KEYS[1])`
keys := []string{"key"}
// When the args parameter is set to nil, it will cause a program panic
// As with the modified appendArg function, when entering the default branch and arg is nil, this code will cause pain
// painc: reflect: call of re reflect.Value.Type on zero Value
// if v.Type().Kind() == reflect.Ptr 
vals, err := rdb.Eval(ctx, script, keys, nil).Result()
if err != nil {
    panic(err)
}
fmt.Println("vals", vals)

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -81,6 +81,8 @@ func appendArg(dst []interface{}, arg interface{}) []interface{} {
 		return dst
 	case time.Time, time.Duration, encoding.BinaryMarshaler, net.IP:
 		return append(dst, arg)
+	case nil:
+		return dst
 	default:
 		// scan struct field
 		v := reflect.ValueOf(arg)


ndyakov:
Nice catch, thank you @frankxjkuang. Wound you mind adding a test case for it?

frankxjkuang:
> 好收获，谢谢@frankxjkuang。您介意为其添加一个测试用例吗？

Hello, should I add test cases for the appendArg function or add a parameter nil for Describe "Eval" and Describe "EvalRO"

ndyakov:
> > 好收获，谢谢@frankxjkuang。您介意为其添加一个测试用例吗？
> 
> Hello, should I add test cases for the appendArg function or add a parameter nil for Describe "Eval" and Describe "EvalRO"

Hello, ideally for both:
- Tests for `Eval` and `EvalRO` will reproduce the previous bug.
- Tests for `appendArg` will test the concrete fix.

frankxjkuang:
> > > 好收获，谢谢@frankxjkuang。您介意为其添加一个测试用例吗？
> > 
> > 
> > Hello, should I add test cases for the appendArg function or add a parameter nil for Describe "Eval" and Describe "EvalRO"
> 
> Hello, ideally for both:
> 
> * Tests for `Eval` and `EvalRO` will reproduce the previous bug.
> * Tests for `appendArg` will test the concrete fix.

Helllo, I have tested the case of adding the args parameter nil for Eval and EvalRO. In addition, since the appendArg function has not been exported, it is not possible to add test cases for them in command_test.go. I do not want to add a test file that would damage the integrity of the existing file structure. I believe this has already verified the case of the args parameter nil.

ndyakov (approved):

PR #3354, opened 2025-04-20
migrate golangci-lint config to v2 format

ofekshenawa:
- bump config schema to version “2”
- re‑add 5m timeout under run
- enable config verification in GitHub Action (verify: true)
- Address several golangci-lint/staticcheck warnings:
- - Use a tagged switch instead of an if/else if chain on readType
- - Drop redundant embedded-field selectors (baseCmd.* and baseClient.*) in favor of direct calls.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v6.5.2
+        uses: golangci/golangci-lint-action@v7.0.0
         with:
           verify: true 
 
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -23,5 +23,5 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6.5.2
         with:
-          verify: false # disable verifying the configuration since golangci is currently introducing breaking changes in the configuration 
+          verify: true 
 
--- a/redis.go
+++ b/redis.go
@@ -677,16 +677,16 @@ func NewClient(opt *Options) *Client {
 func (c *Client) init() {
 	c.cmdable = c.Process
 	c.initHooks(hooks{
-		dial:       c.baseClient.dial,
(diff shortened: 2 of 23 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -23,5 +23,5 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6.5.2
```
elena-kolevska, line 24: We can update the version as well

ndyakov: @ofekshenawa I agree with @elena-kolevska. We can simply update the version in this PR. My initial idea was to update the config in #3328 , but since this is already done here, let's just update the version.

redis.go
```diff
@@ -677,16 +677,16 @@ func NewClient(opt *Options) *Client {
...
 	c.cmdable = c.Process
 	c.initHooks(hooks{
-		dial:       c.baseClient.dial,
-		process:    c.baseClient.process,
-		pipeline:   c.baseClient.processPipeline,
-		txPipeline: c.baseClient.processTxPipeline,
+		dial:       c.dial,
+		process:    c.process,
+		pipeline:   c.processPipeline,
+		txPipeline: c.processTxPipeline,
```
ndyakov, line 683: I think having the embedded field here (`baseClient`) makes it easier to follow the logic and it is more clear what is happening in the client. Let's alter the linters in such a way, to allow us to have the embedded fields in the selectors. I did try different settings for the `staticcheck` linter and found the following to my liking:
```
    staticcheck:
      checks:
        - all
        # Incorrect or missing package comment.
        # https://staticcheck.dev/docs/checks/#ST1000
        - -ST1000
        # Omit embedded fields from selector expression.
        # https://staticcheck.dev/docs/checks/#QF1008
        - -QF1008
```

You can try by adding those in a section `settings:` in the `linters:` section of the `.golangci.yml`. Let me know what you think.

ndyakov (changes requested):
Let's work on setting up the `staticcheck` in such a way that it won't require us to sacrifice on readability just to lint the code.

ndyakov (approved):
Thank you, I think this is a good configuration to start from.

PR #3363, opened 2025-04-29
feat(options): panic when options are nil

### LINKED ISSUE #3358: NewClient panics when given nil Options - add nil parameter check
## Expected Behavior

`_ = redis.NewClient(nil)`

When `opt` parameter is `nil`, `NewClient` should handle it gracefully (e.g., use default configuration or return an error) instead of causing a panic.

## Current Behavior

If `opt` is `nil`, calling `opt.init()` triggers a **nil pointer dereference panic**, leading to program crash.

## Possible Solution

Add a nil check for `opt` parameter and initialize it with default values when it's `nil`:

```go
func NewClient(opt *Options) *Client {
	if opt == nil {
		opt = &Options{} // Use default configuration or actively panic
	}
	// ...
}
```

## Steps to Reproduce

1. Create a `nil` `*Options` instance: `var opt *redis.Options`
2. Call `redis.NewClient(opt)`  
3. The program panics with error: `panic: runtime error: invalid memory address or nil pointer dereference`

## Context (Environment)

Affects developers using the Go-Redis client library. This issue can lead to unexpected application crashes when `NewClient` is called with an uninitialized `Options` pointer.

## Detailed Description

The `NewClient` method does not validate the `opt` parameter for `nil` before calling `opt.init()`. When `opt` is `nil`, this results in a critical runtime panic. This violates defensive programming principles and exposes the library to misuse by clients that might pass invalid configurations.

## Possible Implementation

1. Add a nil check at the beginning of `NewClient`:
```
func NewClient(opt *Options) *Client {
	if opt == nil {
		opt = &Options{} // Initialize with default values
	}
	opt.init() 
	// ... rest of the implementation
}
```
2. Update documentation to clarify that `opt` can be `nil` (with defaults) or explicitly initialized.

ndyakov:
Client creation should panic when options are nil.

Resolves #3358

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -924,6 +924,9 @@ type ClusterClient struct {
 // NewClusterClient returns a Redis Cluster client as described in
 // http://redis.io/topics/cluster-spec.
 func NewClusterClient(opt *ClusterOptions) *ClusterClient {
+	if opt == nil {
+		panic("redis: NewClusterClient nil options")
+	}
 	opt.init()
 
 	c := &ClusterClient{
--- a/redis.go
+++ b/redis.go
@@ -661,6 +661,9 @@ type Client struct {
 
 // NewClient returns a client to the Redis Server specified by Options.
 func NewClient(opt *Options) *Client {
+	if opt == nil {
+		panic("redis: NewClient nil options")
+	}
 	opt.init()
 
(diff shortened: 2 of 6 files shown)


ofekshenawa (approved):
LGTM

PR #3369, opened 2025-05-07
xinfo-groups: support nil lag in XINFO GROUPS

### LINKED ISSUE #3185: XINFO GROUPS can't report that Lag is invalid
XINFO GROUPS returns XInfoGroup.Lag field as 0 when redis reports nil for the Lag. 
The redis nil error seems to be ignored on purpose here. 
https://github.com/redis/go-redis/blob/master/command.go#L2151

This makes differentiating that actually lag is zero or it can not be computed.

## Expected Behavior

I would expect to see a differentiable value from 0 when it is invalid.

## Current Behavior

It returns 0 when redis reports nil for the Lag. 

## Possible Solution

We can return -1 instead of 0 when redis reports nil for the Lag.  

## Steps to Reproduce
Here you can find a description when redis reports nil for the lag. 
```
One or more entries between the group's last-delivered-id and the stream's last-generated-id were deleted
```
https://redis.io/docs/latest/commands/xinfo-groups/

A code example to reproduce: 

```golang
        r, err := redis.ParseURL("redis://:@127.0.0.1:6379")
	panicOnError(err)
	client := redis.NewClient(r)

	ctx := context.Background()
	client.FlushDB(ctx)

	client.XAdd(ctx, &redis.XAddArgs{Stream: "s", ID: "0-1", Values: []string{"foo", "1"}})
	client.XAdd(ctx, &redis.XAddArgs{Stream: "s", ID: "0-2", Values: []string{"foo", "2"}})
	client.XAdd(ctx, &redis.XAddArgs{Stream: "s", ID: "0-3", Values: []string{"foo", "3"}})

	err = client.XGroupCreate(ctx, "s", "g", "0").Err()
	panicOnError(err)
	err = client.XReadGroup(ctx, &redis.XReadGroupArgs{Group: "g", Consumer: "c", Streams: []string{"s", ">"}, Count: 1, Block: -1, NoAck: false}).Err()
	panicOnError(err)

	client.XDel(ctx, "s", "0-2")

	result, err := client.XInfoGroups(context.Background(), "s").Result()
	panicOnError(err)
	for _, group := range result {
		fmt.Printf("%+v\n", group)
	}
```
The output it prints: 
`{Name:g Consumers:1 Pending:1 LastDeliveredID:0-1 EntriesRead:1 Lag:0}` 

What I expected to see:
`{Name:g Consumers:1 Pending:1 LastDeliveredID:0-1 EntriesRead:1 Lag:-1}` 

## Possible Implementation

The code [here](https://github.com/redis/go-redis/blob/master/command.go#L2151) can be changed to 

```
group.Lag, err = rd.ReadInt()

// lag: the number of entries in the stream that are still waiting to be delivered
// to the group's consumers, or a NULL(Nil) when that number can't be determined.
if err != nil {
   err == Nil { 
       group.Lag = -1       
   } else {
       return err
   }   
}			
```

ofekshenawa:
closes #3185

### DIFF
--- a/command.go
+++ b/command.go
@@ -2187,6 +2187,7 @@ func (cmd *XInfoGroupsCmd) readReply(rd *proto.Reader) error {
 
 				// lag: the number of entries in the stream that are still waiting to be delivered
 				// to the group's consumers, or a NULL(Nil) when that number can't be determined.
+				// In that case, we return -1.
 				if err != nil && err != Nil {
 					return err
 				} else if err == Nil {
--- a/command.go
+++ b/command.go
@@ -2189,6 +2189,8 @@ func (cmd *XInfoGroupsCmd) readReply(rd *proto.Reader) error {
 				// to the group's consumers, or a NULL(Nil) when that number can't be determined.
 				if err != nil && err != Nil {
 					return err
+				} else if err == Nil {
+					group.Lag = -1
 				}
 			default:
 				return fmt.Errorf("redis: unexpected key %q in XINFO GROUPS reply", key)
--- a/command.go
+++ b/command.go
@@ -2104,7 +2104,9 @@ type XInfoGroup struct {
 	Pending         int64
 	LastDeliveredID string
 	EntriesRead     int64
-	Lag             int64
+	// Lag represents the number of pending messages in the stream not yet
+	// delivered to this consumer group. Returns -1 when the lag cannot be determined.
+	Lag int64
 }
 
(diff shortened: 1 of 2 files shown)


ndyakov (dismissed):
Thank you, Can we please update the documentation for the `Lag` and note that it will return `-1` when the `Lag` is `nil`.

ndyakov:
Let's just add a comment around command.go#2101 where the response type is defined and we are good to merge.

PR #3370, opened 2025-05-07
Fix: Prevent routing reads to Redis slave nodes in loading state

### LINKED ISSUE #3222: Read nil from a new-joined slave node in Redis cluster.
We enable `ReadOnly=true` for the Redis cluster client. When a new slave node joins the cluster, the client can quickly detect the new slave node by updating the cluster states. Once a new node is detected, requests are immediately forwarded to that node. If the new slave node is still in the master-slave replication process and is waiting for the master node to send the RDB file, the result of the client's read operation will be nil.

## Expected Behavior

A new slave node should not be added to the cluster state before completing master-slave replication.

## Current Behavior

When a slave node joins the cluster, it is immediately added to the cluster state to receive client requests.

## Possible Solution

To check if a node is in replication state when loading cluster state. Send the command `info replication` to the slave node and add it into cluster state only when `master_link_status` is `up`.

## Steps to Reproduce

1. Setup a Redis cluster with at least one slave node for each shard.
2. Fill a bulk of random data into the cluster. This is for extending the replication duration intentionally.
3. Run a Redis cluster client (ReadOnly=true  is enabled) to keep reading existing keys from the cluster.
4. Add a slave node into the cluster.
5. Check if the client reads nil values.

ofekshenawa:
## Problem

When a new slave node is added to a Redis cluster, it enters a loading state while it synchronizes data from its master. If the client has `ReadOnly=true` enabled, it may route read operations to this loading slave node, which can result in nil values or errors being returned to the application.
closes #3222 

## Solution

This PR adds logic to detect when a slave node is in the loading state (by checking for LOADING errors) and temporarily excludes such nodes from the read distribution until they're ready to serve requests.

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -445,6 +445,14 @@ func (n *clusterNode) SetLastLatencyMeasurement(t time.Time) {
 	}
 }
 
+func (n *clusterNode) Loading() bool {
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	err := n.Client.Ping(ctx).Err()
+	return err != nil && isLoadingError(err)
+}
+
 //------------------------------------------------------------------------------
 
 type clusterNodes struct {
@@ -754,7 +762,8 @@ func (c *clusterState) slotSlaveNode(slot int) (*clusterNode, error) {
 	case 1:
 		return nodes[0], nil
 	case 2:
-		if slave := nodes[1]; !slave.Failing() {
+		slave := nodes[1]
+		if !slave.Failing() && !slave.Loading() {
 			return slave, nil
 		}
 		return nodes[0], nil
@@ -763,7 +772,7 @@ func (c *clusterState) slotSlaveNode(slot int) (*clusterNode, error) {
 		for i := 0; i < 10; i++ {
 			n := rand.Intn(len(nodes)-1) + 1
 			slave = nodes[n]
-			if !slave.Failing() {
+			if !slave.Failing() && !slave.Loading() {
 				return slave, nil
 			}
 		}


ndyakov:
@ofekshenawa I am not sure (was not able to find information) if `PING` will return `-LOADING` errors. Let's think of a way to test this somehow? We will need to write tests for it anyway. Sure, we can mock the response and we will get the correct execution, but not sure on top of my head how to actually trigger a redis node to load the rdb for longer periods of time, so we can test. Any ideas?

ndyakov:
Update: `PING` at the moment should return `LOADING`, but it may change. Keep that in mind. Let's think of a way to test this.

ndyakov (approved):
This will work, but we should find a good way to test it. Let's add a note about it.

ste93cry:
@ndyakov Just like the comment above, in the project I'm working on we're pushing thousands of commands through a pipeline, and we've noticed that latency has more than doubled. This change really needs to be rolled back and handled differently, as sending a `PING` with every command just isn’t practical.

ndyakov:
@LINKIWI , @ste93cry thank you for reporting this. The loaded state should / can be cached, you are both correct. @ste93cry for the time being, feel free to use a release before this patch, I will work on adding a cache and will release a new patch soon. 

Let's assume we add a `Loaded` state and perform the `PING` operation only when it is `0`. Once sucessfull we will set the state to `1` and not perform `PING`. One thing that I am not certain is - when should the `Loaded` state be returned to `0`. One good candidate will be when a node is marked as `Failing`, but I assume there may be the case when we don't catch this. I am open to suggestions. For now I will implement the basic caching mechanism to reduce the latency overhead and will look deeper into it when we have more time.

LINKIWI:
Thanks for following up. Regarding when the cached loaded state should return to 0: `Failing` currently handles this problem with a 15s open circuit timeout. I think the loading state can be tracked similarly.

In most cases/for most users, I suspect slaves should only be in the `LOADING` state for less than 30 seconds. So it could make sense to use a more aggressive penalty timeout, maybe on the order of ~5 seconds. The side effect here is that a newly issued command after 5s _could_ once again hit `LOADING`, at which point the slave would once again be marked ineligible for the next 5s.

A separate follow up here could be to transparently fail over command execution to another available slave after observing `LOADING` (and recording this internal state). This would both hide the user-facing errors (when > 1 replicas are available), not compromise steady-state latency, and mostly-avoid redundantly querying slaves during the loading phase.

ndyakov:
@LINKIWI I think introducing a timeout will require a more testing to calculate a good default value of the timeout. Here is a naive approach without a timeout, let's continue the discussion there. I would prefer to have the naive approach merged and released and then, once we are the time we can find a better approach https://github.com/redis/go-redis/pull/3410

PR #3375, opened 2025-05-09
feat: support vectorset

fukua95:
vectorset commands: https://redis.io/docs/latest/develop/data-types/vector-sets/

### DIFF
--- a/command.go
+++ b/command.go
@@ -5620,3 +5620,58 @@ func (cmd *MonitorCmd) Stop() {
 	defer cmd.mu.Unlock()
 	cmd.status = monitorStatusStop
 }
+
+type MapStringFloatCmd struct {
+	baseCmd
+
+	val map[string]float64
+}
+
+var _ Cmder = (*MapStringFloatCmd)(nil)
+
+func NewMapStringFloatCmd(ctx context.Context, args ...interface{}) *MapStringFloatCmd {
+	return &MapStringFloatCmd{
+		baseCmd: baseCmd{
+			ctx:  ctx,
+			args: args,
+		},
+	}
+}
+
+func (cmd *MapStringFloatCmd) SetVal(val map[string]float64) {
+	cmd.val = val
+}
+
+func (cmd *MapStringFloatCmd) Val() map[string]float64 {
+	return cmd.val
+}
+
+func (cmd *MapStringFloatCmd) Result() (map[string]float64, error) {
+	return cmd.val, cmd.err
+}
+
+func (cmd *MapStringFloatCmd) String() string {
+	return cmdString(cmd, cmd.val)
+}
+
+func (cmd *MapStringFloatCmd) readReply(rd *proto.Reader) error {
+	n, err := rd.ReadMapLen()
+	if err != nil {
+		return err
+	}
+
+	cmd.val = make(map[string]float64, n)
+	for i := 0; i < n; i++ {
+		key, err := rd.ReadString()
+		if err != nil {
+			return err
+		}
+
+		nn, err := rd.ReadFloat()
+		if err != nil {
+			return err
+		}
+		cmd.val[key] = nn
+	}
+	return nil
+}
--- a/command.go
+++ b/command.go
@@ -5621,21 +5621,16 @@ func (cmd *MonitorCmd) Stop() {
 	cmd.status = monitorStatusStop
 }
 
-type VectorInfo struct {
-	Name  string
-	Score float64
-}
-
-type VectorInfoSliceCmd struct {
+type VectorScoreSliceCmd struct {
 	baseCmd
 
-	val []VectorInfo
+	val []VectorScore
 }
 
-var _ Cmder = (*VectorInfoSliceCmd)(nil)
+var _ Cmder = (*VectorScoreSliceCmd)(nil)
 
-func NewVectorInfoSliceCmd(ctx context.Context, args ...any) *VectorInfoSliceCmd {
-	return &VectorInfoSliceCmd{
+func NewVectorInfoSliceCmd(ctx context.Context, args ...any) *VectorScoreSliceCmd {
+	return &VectorScoreSliceCmd{
 		baseCmd: baseCmd{
 			ctx:  ctx,
 			args: args,
@@ -5643,29 +5638,29 @@ func NewVectorInfoSliceCmd(ctx context.Context, args ...any) *VectorInfoSliceCmd
 	}
 }
 
-func (cmd *VectorInfoSliceCmd) SetVal(val []VectorInfo) {
+func (cmd *VectorScoreSliceCmd) SetVal(val []VectorScore) {
 	cmd.val = val
 }
 
-func (cmd *VectorInfoSliceCmd) Val() []VectorInfo {
+func (cmd *VectorScoreSliceCmd) Val() []VectorScore {
 	return cmd.val
 }
 
-func (cmd *VectorInfoSliceCmd) Result() ([]VectorInfo, error) {
+func (cmd *VectorScoreSliceCmd) Result() ([]VectorScore, error) {
 	return cmd.val, cmd.err
 }
 
-func (cmd *VectorInfoSliceCmd) String() string {
+func (cmd *VectorScoreSliceCmd) String() string {
 	return cmdString(cmd, cmd.val)
 }
 
-func (cmd *VectorInfoSliceCmd) readReply(rd *proto.Reader) error {
+func (cmd *VectorScoreSliceCmd) readReply(rd *proto.Reader) error {
 	n, err := rd.ReadMapLen()
 	if err != nil {
 		return err
 	}
 
-	cmd.val = make([]VectorInfo, n)
+	cmd.val = make([]VectorScore, n)
 	for i := 0; i < n; i++ {
 		name, err := rd.ReadString()
 		if err != nil {
--- a/command.go
+++ b/command.go
@@ -5629,7 +5629,7 @@ type MapStringFloatCmd struct {
 
 var _ Cmder = (*MapStringFloatCmd)(nil)
 
-func NewMapStringFloatCmd(ctx context.Context, args ...interface{}) *MapStringFloatCmd {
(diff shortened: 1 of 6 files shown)


ndyakov:
Thank you @fukua95, will review as soon as possible.

vectorset_commands.go
```diff
@@ -0,0 +1,284 @@
+package redis
+
+import (
+	"context"
+	"strconv"
+)
+
+type VectorSetCmdable interface {
+	VAdd(ctx context.Context, key, element string, val Vector) *BoolCmd
+	VAddArgs(ctx context.Context, key, element string, val Vector, addArgs VAddArgs) *BoolCmd
```
ndyakov, line 10: If you take a look at https://github.com/redis/go-redis/blob/7b9bd6c36e3696e619a9fc5880d5fff002ed32a0/search_commands.go#L1865-L1864 for example, the functions tend to be named `*WithArgs` and the `Args` themselves are pointers. I think it makes sense to be consistent and do this here as well.

command.go
```diff
@@ -5620,3 +5620,58 @@ func (cmd *MonitorCmd) Stop() {
...
+var _ Cmder = (*MapStringFloatCmd)(nil)
+
+func NewMapStringFloatCmd(ctx context.Context, args ...interface{}) *MapStringFloatCmd {
+	return &MapStringFloatCmd{
+		baseCmd: baseCmd{
+			ctx:  ctx,
+			args: args,
+		},
+	}
+}
```
ndyakov, line 5639: since this is a new api we are introducing. would you like to use `any` instead of `interface{}`? It won't be consistent with the rest of the api, but it is not a breaking change. I don't have a strong opinion on the subject, but saw you were using `any` in other pieces of the code in this PR.

ndyakov (commented):
left two comments, this is just a partial review.

vectorset_commands.go
fukua95: I used to be really confused about that `Args` vs `WithArgs` and `Args` param vs `*Args` param too.

zsort commands use `Args` and `Args` param: https://github.com/redis/go-redis/blob/master/sortedset_commands.go#L18C2-L18C66.

timeseries/search commands use `WithArgs` and `*Args` param.

I updated vector set to use `WithArgs` and `*Args` param, but zsort remains unchanged (breaking changed). WDYT.

command.go
fukua95: I will rewrite `any` to `interface{}` to be consistent with other APIs.

ndyakov:
@fukua95 we will have a discussion next week, overall I think the PR looks good, if the team agrees on the return types, I will merge this, thank you.

fukua95:
> @fukua95 we will have a discussion next week, overall I think the PR looks good, if the team agrees on the return types, I will merge this, thank you.

My bad. I will update it.

fukua95:
> @fukua95 we will have a discussion next week, overall I think the PR looks good, if the team agrees on the return types, I will merge this, thank you.

Thank you for your review. 

One more thing: The `VSim` command in Redis code has the option `Epsilon`, by the doc in Redis.io don't have. I'll modify the code after you decide whether to keep this option.

ndyakov:
@fukua95 we are still not set on the return types, but one thing I will suggest is to use something similar to https://github.com/redis/go-redis/blob/27581fcb437fa7387d1dcfb6a43c5ccf1f4d2273/sortedset_commands.go#L728 for the responses with scores. It will make this more consistent with the rest of the library and in general I don't see the benefit of having a map here - I don't see a lot of usecases that will benefit from the constant time to access an element. WDYT?

fukua95:
> @fukua95 we are still not set on the return types, but one thing I will suggest is to use something similar to
> 
> https://github.com/redis/go-redis/blob/27581fcb437fa7387d1dcfb6a43c5ccf1f4d2273/sortedset_commands.go#L728
> 
> for the responses with scores. It will make this more consistent with the rest of the library and in general I don't see the benefit of having a map here - I don't see a lot of usecases that will benefit from the constant time to access an element. WDYT?

Using a struct is more user-friendly. I will update it soon.

ndyakov:
Thank you @fukua95 , let me know if you need any help. You can also find me on [Discord](https://discord.gg/W4txy5AeKM)

cxljs:
> Thank you @fukua95 , let me know if you need any help. You can also find me on [Discord](https://discord.gg/W4txy5AeKM)

Apologies for the delay @ndyakov  - these days mark a holiday in my country, and I'm currently traveling. 
I'll update the code tonight.

vectorset_commands.go
```diff
@@ -0,0 +1,293 @@
...
+	"strconv"
+)
+
+type VectorSetCmdable interface {
+	VAdd(ctx context.Context, key, element string, val Vector) *BoolCmd
+	VAddWithArgs(ctx context.Context, key, element string, val Vector, addArgs *VAddArgs) *BoolCmd
+	VCard(ctx context.Context, key string) *IntCmd
+	VDim(ctx context.Context, key string) *IntCmd
+	VEmb(ctx context.Context, key, element string, raw bool) *SliceCmd
+	VGetAttr(ctx context.Context, key, element string) *StringCmd
```
ndyakov, line 16: Hey @fukua95 we had a discussion with the team and since we would like to make the API extensible in the future, what do you think about introducing a `VectorAttribute` interface (similar to the `Vector` interface). Right now in your implementation it can have a single `VectorAttributeRawString` struct that will implement it and it will work as simple as - accept the string and pass it to the `Set` command. In the future we can introduce `VerctorAttributeMap` or even better - a combined `VectorAttributeMarshaller` that will accept any type that can be marshaled to a JSON string. Another approach would be to use an empty interface and use similar logic to what is done here: https://github.com/redis/go-redis/blob/20769750085646dd195657bf20edea6f4f6760b9/json.go#L535-L545 for json. The string response on the other hand makes should work since there is https://github.com/redis/go-redis/blob/b67455e0998b8767a47c1efd5828144f6709f85c/command.go#L928

vectorset_commands.go
```diff
@@ -0,0 +1,293 @@
...
+
+type VSimArgs struct {
+	Count    int64
+	EF       int64
+	Filter   string
+	FilterEF int64
+	Truth    bool
+	NoThread bool
+	// The `VSim` command in Redis has the option, by the doc in Redis.io don't have.
+	// Epsilon float64
```
ndyakov, line 291: let's keep this commented out for now. It will be easier to enable it if we decide, than to introduce a breaking change in the api.

vectorset_commands.go
```diff
@@ -0,0 +1,293 @@
+package redis
+
+import (
+	"context"
+	"strconv"
+)
+
+type VectorSetCmdable interface {
```
ndyakov: Since this is not a mature API and overall the vector sets are "experimental" even in Redis, let's mark it as `experimental` in `go-redis` as well. 

Just add a comment, here and even better - a oneliner in the any `V*` function. I can do this as well once this is merged.

vectorset_commands.go
```diff
@@ -0,0 +1,293 @@
...
+	VDim(ctx context.Context, key string) *IntCmd
+	VEmb(ctx context.Context, key, element string, raw bool) *SliceCmd
+	VGetAttr(ctx context.Context, key, element string) *StringCmd
+	VInfo(ctx context.Context, key string) *MapStringInterfaceCmd
+	VLinks(ctx context.Context, key, element string) *StringSliceCmd
+	VLinksWithScores(ctx context.Context, key, element string) *VectorInfoSliceCmd
+	VRandMember(ctx context.Context, key string) *StringCmd
+	VRandMemberCount(ctx context.Context, key string, count int) *StringSliceCmd
+	VRem(ctx context.Context, key, element string) *BoolCmd
+	VSetAttr(ctx context.Context, key, element, attr string) *BoolCmd
```
ndyakov: Setting an empty attributes will remove them. I think having a function `VRemAttr(context.Context, key, elem) *BoolCmd` makes sense to be added for convenience. WDYT @fukua95 ?

ndyakov (commented):
@fukua95 I added some suggestions, let me know if you would like to address them or if you like I can work on those. Thank you for the amazing work so far!

cxljs:
@ndyakov Thank you for the detailed review! I really appreciate your suggestions and I'll work on the updates tonight or tomorrow.

vectorset_commands.go
cxljs: OK, I use `VectorAttributeMarshaller` as the type of `attr` param in `VSetAttr`.

vectorset_commands.go
cxljs: Added.

vectorset_commands.go
cxljs: Yes, `VRemAttr` is convenient.

ndyakov:
@fukua95 just a heads up, for the method that should clear the attributes - let's name it `VClearAttributes` to be able to distinguish it from other redis commands.

cxljs:
Hello, How about naming the method `VClearAttr`? This way it can be consistent with `VSetAttr`.

ndyakov:
@fukua95 I was suggesting something similar to what you are suggesting, but the potential issue is that a user may think there is a redis command `VClearAttr` and that's why we voted and decided to have it completely different from the rest of the redis commands and name it `VClearAttributes`.

cxljs:
Got it, I've made the changes.

ndyakov:
@fukua95 thank you, will review this asap and get back to you.

ndyakov:
@fukua95 I hope you don't mind me pushing the last commit. After trying both implementations for `VSetAttr` I decided agains my initial suggestion and didn't want to ask you to redo it, so I did it.

cxljs:
Not at all. Thank you.

htemelski-redis (approved):

ndyakov:
@cxljs bringing this to your attention https://github.com/redis/redis/pull/14223

cxljs:
Thanks @ndyakov , I'll update it in a bit.

PR #3380, opened 2025-05-16
Export connection pool errors for public consumption

LINKIWI:
The `internal/pool` package has some structured errors:

* `ErrClosed`
* `ErrPoolExhausted`
* `ErrPoolTimeout`

These errors may be propagated to the callsites of top-level commands on the connection. Exposing these errors in the public package layer makes it much easier for external code to detect and classify these errors, e.g. as:

```go
err := ...
if errors.Is(err, redis.ErrPoolTimeout) {
    ...
}
```

This change proposes re-exporting the remaining `ErrPool*` errors, adding to `ErrClosed` which is already re-exported.

See also: #3381

### DIFF
--- a/error.go
+++ b/error.go
@@ -15,6 +15,13 @@ import (
 // ErrClosed performs any operation on the closed client will return this error.
 var ErrClosed = pool.ErrClosed
 
+// ErrPoolExhausted is returned from a pool connection method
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
LGTM, @LINKIWI thanks for the contribution

PR #3383, opened 2025-05-17
chore: optimize function `ReplaceSpaces`

fukua95:
When dealing with small strings, using the `strings.ReplaceAll()` function delivers better performance.


ndyakov:
Thank you @fukua95, looking at the benchmarks, this makes sense!

ndyakov (approved):

PR #3386, opened 2025-05-20
rewrite interface{} to any

cxljs:
Hello @ndyakov , Do you agree with replacing `interface{}` with `any` in non-constraint scenarios? If approved, we can migrate all `interface{}` occurrences.
(similar to the commit: https://github.com/golang/go/commit/2580d0e08d5e9f979b943758d3c49877fb2324cb#diff-eca48ce939b837cddc0194885eff25247d2ef2c481b63a667ee14fafcc0f7c7c).

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,7 @@ For example:
 		"idx:bicycle",
 		"@pickup_zone:[CONTAINS $bike]",
 		&redis.FTSearchOptions{
-			Params: map[string]interface{}{
+			Params: map[string]any{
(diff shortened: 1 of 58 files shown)


fukua95:
> Hello @fukua95, I was considering this and since it will be an API change (although they are the exact same thing), would like to do it just before releasing `v10`, to not introduce such API (documentation, more or less) change in `v9`. WDYT?

100% agree! This is exactly the right approach.

PR #3387, opened 2025-05-23
chore: set the default value for the `options.protocol` in the `init()` of `options`

cxljs:
When the user does not set the `protocol` option, `initConn()` sets `protocol=3`. We can move this behavior of setting the default `protocol` value to `init()` of `options`, so that all default option settings are unified in `init()`.

### DIFF
--- a/options_test.go
+++ b/options_test.go
@@ -222,3 +222,26 @@ func TestReadTimeoutOptions(t *testing.T) {
 		}
 	}
 }
+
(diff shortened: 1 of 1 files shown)


fukua95:
> @fukua95 would you please update the test case to cover this as well? Also, a short description in the PR body will help us later when we are preparing release notes.

My bad. I will update it.

ndyakov (approved):
lgtm

PR #3388, opened 2025-05-23
feat(ring): add GetShardClients and GetShardClientForKey methods to Ring for shard access

### LINKED ISSUE #3341: Make shards in redis.Ring publicly accessible
We use `redis.Ring` to shard load across servers. We shard mainly PubSub related commands,  `PSUBSCRIBE`, `SUBSCRIBE`, `PUNSUBSCRIBE` and `UNSUBSCRIBE`.

Because these commands put the redis connection in a special "pubsub" mode, after which you cannot use it for anything else than pubsub, we try to re-use these connections as much as possible.

What we'd like to do is set up a pubsub connection for every shard in the ring at application start, and then re-use that pubsub connection when we need to subscribe or unsubscribe on a topic.

However for that to be possible, we need two things:

 * access to the list of shards, so we can immediately create a pubsub connection for each shard in the ring
 * access to `ringSharding.GetByKey` so we can determine which shard pubsub connection we can re-use for a given topic

However, on `redis.Ring` the `sharding` attribute, which contains an instance of `*ringSharding`, is not publicly accessible, so we cannot access the shards list or the GetByKey function.

We are working around this for now, but it's rather ugly, because we need to make assumptions in our own code on how topics are distributed over the ring.

Would it be possible to make `redis.Ring.sharding` public, or at least provide some public methods on `redis.Ring` to return the list of shards access the `GetByKey` function?

I would be open to create a merge request if there is a chance of it being accepted.

DengY11:
This PR adds two new methods to the `Ring` client that provide access to individual shard clients, addressing the need to create dedicated connections (e.g., PubSub) for each shard.

When using Redis Ring with PubSub connections, users need to create separate PubSub connections for each shard because PubSub connections enter a special mode and cannot be used for other Redis operations. Currently, there's no way to access the underlying shard clients directly, making it impossible to pre-create and reuse PubSub connections [related issue](https://github.com/redis/go-redis/issues/3341)

This PR introduces two new public methods to the `Ring` struct, users can pre-create and reuse PubSub connections per shard

1. `GetShards() []*Client`
- Returns a list of all currently active shard clients
- Only includes shards that are marked as "up" 
- Allows users to iterate over all available shards

2. `GetShardByKey(key string) (*Client, error)`
- Returns the shard client that would handle a specific key/channel
- Uses the same consistent hashing logic as the Ring's internal routing
- Enables users to determine which shard a particular key would be routed to

Checklist:
- [x] Code follows project style guidelines
- [x] Tests pass 
- [x] New functionality is documented  
- [x] Backward compatibility maintained
- [x] Uses existing patterns and conventions
- [x] Test coverage includes edge cases
- [x] No security vulnerabilities introduced

closes #3341

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -848,7 +848,9 @@ func (c *Ring) Close() error {
 	return c.sharding.Close()
 }
 
-func (c *Ring) GetShards() []*Client {
+// GetShardClients returns a list of all shard clients in the ring.
+// This can be used to create dedicated connections (e.g., PubSub) for each shard.
+func (c *Ring) GetShardClients() []*Client {
 	shards := c.sharding.List()
 	clients := make([]*Client, 0, len(shards))
 	for _, shard := range shards {
@@ -859,7 +861,9 @@ func (c *Ring) GetShards() []*Client {
 	return clients
 }
 
-func (c *Ring) GetShardByKey(key string) (*Client, error) {
+// GetShardClientForKey returns the shard client that would handle the given key.
+// This can be used to determine which shard a particular key/channel would be routed to.
+func (c *Ring) GetShardClientForKey(key string) (*Client, error) {
 	shard, err := c.sharding.GetByKey(key)
 	if err != nil {
 		return nil, err
--- a/ring.go
+++ b/ring.go
@@ -847,3 +847,22 @@ func (c *Ring) Close() error {
 
 	return c.sharding.Close()
 }
+
+func (c *Ring) GetShards() []*Client {
+	shards := c.sharding.List()
+	clients := make([]*Client, 0, len(shards))
+	for _, shard := range shards {
+		if shard.IsUp() {
+			clients = append(clients, shard.Client)
+		}
+	}
+	return clients
+}
+
+func (c *Ring) GetShardByKey(key string) (*Client, error) {
+	shard, err := c.sharding.GetByKey(key)
+	if err != nil {
+		return nil, err
+	}
+	return shard.Client, nil
+}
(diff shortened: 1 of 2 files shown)


DengY11:
hi @ndyakov 
Thanks for looking at this!
yep, the tests are in ring_test.go (Ginkgo/Gomega) under the Describe("Ring GetShards and GetShardByKey", ...) block. Sorry if that wasn't clear!
Quickly on why ring for us: we use it for a busy Pub/Sub system. Ring's simpler client-side sharding works well with our current standalone Redis setup. We don't need the full Redis OSS Cluster complexity for this specific Pub/Sub part.
These new methods are super helpful because they let us:
Grab all shard clients to set up dedicated Pub/Sub connections upfront.
Figure out which shard's Pub/Sub connection to use for each channel, using Ring's own logic.
This is key for our Pub/Sub performance, and it's tricky to do cleanly without these helpers.

ndyakov:
@DengY11 now I can see the tests, for some reason the diff was not complete. Anyway, would you mind renaming the new functions to `GetShardClients` and `GetShardClientForKey` ? As for the usecase, we are collecting information regarding `Ring` since we are discussing extracting it from the client since Redis Cluster covers mostly the same usecases. Thank you for providing that information.

jaapz:
@DengY11 thanks for the implementation!

> Our use case, particularly for streamlining Pub/Sub sharding with standalone instances where Ring's simplicity is a benefit, hopefully adds a useful perspective to those discussions.

This is exactly how we are using it too, and it works very well. We can easily shard pub/sub across multiple redis instances without having to set up a redis cluster. Scaling to more redis instances is as simple as starting another redis process, adding it to the ring and off you go. Afaik Redis Cluster is also an enterprise feature? Although that may have changed after the recent relicensing kerfuffle.

PR #3393, opened 2025-05-28
Add the missing NewFloatSliceResult for testing

AndBobsYourUncle:
When I was developing for another project, I ran into a situation in which I needed to test `ZMScore` in my unit tests.

However, I found out that there existed no helper method that returned a mocked `FloatSliceCmd` with a populated result.

After digging into the code, it looked to me like there existed `NewStringSliceResult`, `NewBoolSliceResult`,  but no `NewFloatSliceResult` method.

This PR simply adds that method so that it can be used for testing in other projects.

### DIFF
--- a/result.go
+++ b/result.go
@@ -82,6 +82,14 @@ func NewBoolSliceResult(val []bool, err error) *BoolSliceCmd {
 	return &cmd
 }
 
+// NewFloatSliceResult returns a FloatSliceCmd initialised with val and err for testing.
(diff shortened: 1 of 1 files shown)


ndyakov (approved):
Hello @AndBobsYourUncle , this looks good, thank you for the contribution. I know it may seem useless, but would you mind adding tests for all `New*Result` functions in this file? I can do it in a separate PR for the existing ones, and you can add just for the `NewFloatSliceResult` if you don't have the time.

ndyakov:
I will merge this and add tests separately, thank you.

PR #3396, opened 2025-06-07
fix: check if the shard exists to avoid returning nil

### LINKED ISSUE #2711: panic in ring

## Steps to Reproduce

1.
2.
3.
4.

fukua95:
close #2711

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -405,7 +405,12 @@ func (c *ringSharding) GetByName(shardName string) (*ringShard, error) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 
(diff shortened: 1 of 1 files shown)


ndyakov (approved):

ndyakov:
@fukua95 thank you for fixing this in the Ring client. We are not trying to suggest users to use Cluster instead of Ring and are not actively improving it at the moment. If you would like you can take a look at the rest of the Ring issues, as they most probably won't be prioritized by the team soon.

cxljs:
I'll take a look at the other Ring issues and see if there's anything I can help with.

PR #3398, opened 2025-06-08
fix: insert entry during iterating over a map

fukua95:
Inserting data while iterating over a map is not a good practice and may lead to undefined behavior.


ndyakov:
Looks good, thank you!

PR #3401, opened 2025-06-10
chore: remove a redundant method from `Ring`

fukua95:


ndyakov:
@fukua95, the tests should work with the `baseClient` anyway, let's keep them.

cxljs:
Got it.

ndyakov (approved):

PR #3408, opened 2025-06-16
[CAE-1028] fix(txpipeline): should return error on multi/exec on multiple slots

### LINKED ISSUE #3019: Why does go-redis allow cluster mode cross-shard transactions and pipelines?
Cross-shard/cross-slot batch operations in cluster mode-enabled clusters, like transactions and pipelines, are not supported in a Redis cluster, by specification.

* There is no server-side primitive that enables cross-shard transactional atomicity.
* A single pipeline cannot physically be applied to multiple targets, since it occurs over a single physical connection.

Other client libraries properly reject these operations with a cross-slot validation error. The go-redis library seems to instead (1) partition the batch commands by slot, (2) concurrently dispatch individual batch commands to each shard, and (3) aggregate the results client-side.

https://github.com/redis/go-redis/blob/v9.5.2/osscluster.go#L1237

As a simple reproducible example:

```
127.0.0.1:7000> set a aaaaa
-> Redirected to slot [15495] located at 127.0.0.1:7002
OK
127.0.0.1:7002> set b bbbbb
-> Redirected to slot [3300] located at 127.0.0.1:7000
OK
```

```go
ctx := context.Background()
pipe := client.TxPipeline()
pipe.Get(ctx, "a")
pipe.Get(ctx, "b")
fmt.Println(pipe.Exec(context.Background()))  // [aaaaa bbbbb]
```

This is both extremely surprising behavior, and the source of potentially severe correctness/consistency problems. I don't see any mechanism to opt out or otherwise disable the "smart virtual batching" exercised by the library.

Why does go-redis do this? Is it possible to implement a client option to disable this behavior?

## Expected Behavior

Cross-slot transactions and pipelines should return an error.

## Current Behavior

go-redis internally separates out commands by shard and executes them independently without error.

## Possible Solution

Provide a switch to allow the client to reject batch operations when they per-shard partitioning hashes the set of commands to multiple slots.

## Steps to Reproduce

1. Create a cluster with at least 2 nodes.
2. Write sample keys spanning both nodes.
3. Exercise the code snippet above.
4. Observe that no error is returned.

## Context (Environment)

Clients are expecting transactional atomicity which can only be guaranteed when the transaction affects keys that hash to the same slot.

## Detailed Description

See above.

## Possible Implementation

N/A

ndyakov:
`go-redis` should not try to do some magic and support multi slot transactions. Such cross slot operations are not supported by the database server for a reason and what the library can do is to report an error to the user as soon as possible. 

Althought this changes the behaviour it is considered a bugfix.

closes #3019

### DIFF
--- a/osscluster_test.go
+++ b/osscluster_test.go
@@ -598,6 +598,7 @@ var _ = Describe("ClusterClient", func() {
 				It("returns CrossSlot error", func() {
 					pipe.Set(ctx, "A{s}", "A_value", 0)
 					pipe.Set(ctx, "B{t}", "B_value", 0)
+					Expect(hashtag.Slot("A{s}")).NotTo(Equal(hashtag.Slot("B{t}")))
 					_, err := pipe.Exec(ctx)
 					Expect(err).To(MatchError(redis.ErrCrossSlot))
 				})
--- a/osscluster_test.go
+++ b/osscluster_test.go
@@ -462,8 +462,7 @@ var _ = Describe("ClusterClient", func() {
 		Describe("pipelining", func() {
 			var pipe *redis.Pipeline
 
-			assertPipeline := func() {
-				keys := []string{"A", "B", "C", "D", "E", "F", "G"}
+			assertPipeline := func(keys []string) {
 
 				It("follows redirects", func() {
 					if !failover {
@@ -482,13 +481,12 @@ var _ = Describe("ClusterClient", func() {
 					Expect(err).NotTo(HaveOccurred())
 					Expect(cmds).To(HaveLen(14))
 
-					_ = client.ForEachShard(ctx, func(ctx context.Context, node *redis.Client) error {
-						defer GinkgoRecover()
-						Eventually(func() int64 {
-							return node.DBSize(ctx).Val()
-						}, 30*time.Second).ShouldNot(BeZero())
-						return nil
-					})
+					// Check that all keys are set.
+					for _, key := range keys {
+						Eventually(func() string {
+							return client.Get(ctx, key).Val()
+						}, 30*time.Second).Should(Equal(key + "_value"))
+					}
 
 					if !failover {
 						for _, key := range keys {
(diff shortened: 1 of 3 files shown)


ndyakov:
@LINKIWI a transaction which is cross slot won't be a single transaction anyway and will break the contract of a "multi/exec". If users raise issues, we can safely suggest using pipelines and/or refactoring their code. The possibility to use it as it was, was allowing users to use the database and transactions incorrectly. I see your point about the backwards compatibility, but personally don't see a value in a flag and would prefer not to have one.

htemelski-redis (dismissed):
Looks good.

One minor nitpick: In the osscluster file, can we swap the places of the IFs? First check if the slice is empty and then check if it contains more than one element.

htemelski-redis (dismissed):
LGTM

osscluster_test.go
```diff
@@ -482,13 +481,12 @@ var _ = Describe("ClusterClient", func() {
...
-						Eventually(func() int64 {
-							return node.DBSize(ctx).Val()
-						}, 30*time.Second).ShouldNot(BeZero())
-						return nil
-					})
+					// Check that all keys are set.
+					for _, key := range keys {
+						Eventually(func() string {
+							return client.Get(ctx, key).Val()
+						}, 30*time.Second).Should(Equal(key + "_value"))
```
elena-kolevska, line 488: Nit: we can lower this timeout

osscluster_test.go
```diff
@@ -590,7 +589,24 @@ var _ = Describe("ClusterClient", func() {
...
+				keys := []string{"A{s}", "B{s}", "C{s}", "D{s}", "E{s}", "F{s}", "G{s}"}
+				assertPipeline(keys)
+
+				// make sure CrossSlot error is returned
+				It("returns CrossSlot error", func() {
+					pipe.Set(ctx, "A{s}", "A_value", 0)
+					pipe.Set(ctx, "B{t}", "B_value", 0)
+					_, err := pipe.Exec(ctx)
+					Expect(err).To(MatchError(redis.ErrCrossSlot))
+				})
```
elena-kolevska: I'd also add a line here showing that `s` and `t` hash to a different slot to begin with.
(I guess the func is `hashtag.Slot(x)`)

elena-kolevska (dismissed):

osscluster_test.go
ndyakov: This won't change anything if the test passes, the function will be executed each `10ms` by default, and the timeout will be `30 seconds`. It should be safe to drop it to 10 seconds or even 1s, but it won't change the execution when the tests pass and it won't speed up the tests. The 30 second value is used across the test, would you mind if we keep it as it is and once @htemelski-redis is looking at the tests we can decide if we would like to lower it.

elena-kolevska: Yes, of course, we would just have a faster test failure. 30sec sounds unreasonably long for a redis command, but I see your point in doing the change across all tests when we do the refactoring 👍

elena-kolevska (approved):

PR #3411, opened 2025-06-19
fix(txpipeline): keyless commands should take the slot of the keyed

ndyakov:
There is the possibility that there are keyless commands that can be executed on any node/slot. This should make sure that we choose the slot for the keyed commands when there is a tx pipelien request.

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -1583,15 +1583,15 @@ func (c *ClusterClient) processTxPipeline(ctx context.Context, cmds []Cmder) err
 func (c *ClusterClient) slottedKeyedCommands(cmds []Cmder) map[int][]Cmder {
 	cmdsSlots := map[int][]Cmder{}
 
-	prefferedRandomSlot := -1
+	preferredRandomSlot := -1
 	for _, cmd := range cmds {
 		if cmdFirstKeyPos(cmd) == 0 {
 			continue
 		}
 
-		slot := c.cmdSlot(cmd, prefferedRandomSlot)
-		if prefferedRandomSlot == -1 {
-			prefferedRandomSlot = slot
+		slot := c.cmdSlot(cmd, preferredRandomSlot)
+		if preferredRandomSlot == -1 {
+			preferredRandomSlot = slot
 		}
 
 		cmdsSlots[slot] = append(cmdsSlots[slot], cmd)


osscluster.go
```diff
@@ -1572,6 +1600,18 @@ func (c *ClusterClient) mapCmdsBySlot(cmds []Cmder) map[int][]Cmder {
...
+	keylessCmds := make([]Cmder, 0, len(cmds))
+	for _, cmd := range cmds {
+		if cmdFirstKeyPos(cmd) == 0 {
+			keylessCmds = append(keylessCmds, cmd)
+		} else {
+			keyedCmds = append(keyedCmds, cmd)
+		}
+	}
+	return keyedCmds, keylessCmds
+}
```
ndyakov: This can be optimized to just extract the `keyedCmds` since it doesn't look at the keyless cmds at the moment. Will rewrite if the CI passes.

ndyakov:
@LINKIWI bringing this to your attention.

LINKIWI:
> @LINKIWI bringing this to your attention. 

Thanks, good catch. Structurally this looks good to me. I can help run this patch through our internal correctness and performance test suite early next week.

command.go
```diff
@@ -17,6 +17,55 @@ import (
 	"github.com/redis/go-redis/v9/internal/util"
 )
 
+// keylessCommands contains Redis commands that have empty key specifications (9th slot empty)
+// Only includes core Redis commands, excludes FT.*, ts.*, timeseries.*, search.* and subcommands
+var keylessCommands = map[string]struct{}{
```
LINKIWI, line 22: Some other implementations ascertain this information dynamically with a call to https://redis.io/docs/latest/commands/command/ at initialization time. I don't necessarily think go-redis _should_ do that, since it would require a command to execute successfully just for the library to initialize its internal state correctly, but it could be an interesting angle to consider, which would allow you to avoid needing to hardcode this list and maintain this source of truth manually.

ndyakov: @LINKIWI I agree. We are considering this since there is other data in the Commands output that can and should be used. We will probably start with a static list and update it on initialization, but it will be part of a separate feature.

osscluster.go
```diff
@@ -1519,8 +1526,36 @@ func (c *ClusterClient) processTxPipeline(ctx context.Context, cmds []Cmder) err
 		return err
 	}
 
-	cmdsMap := c.mapCmdsBySlot(cmds)
+	cmdsMap := map[int][]Cmder{}
+	slot := -1
+	// get only the keyed commands
+	keyedCmds := c.keyedCmds(cmds)
```
htemelski-redis, line 1532: Something like that
```go
	slot := -1
	slottedCommands := c.slottedCommands(cmds)
	switch len(slottedCommands) {
	case 0:
		slot = hashtag.RandomSlot()
	case 1:
		for sl := range slottedCommands {
			slot = sl
			break
		}
	default:
		setCmdsErr(cmds, ErrCrossSlot)
		return ErrCrossSlot
	}
```

osscluster.go
```diff
@@ -1566,13 +1601,29 @@ func (c *ClusterClient) processTxPipeline(ctx context.Context, cmds []Cmder) err
...
+			preferredRandomSlot = slot
+		}
 		cmdsMap[slot] = append(cmdsMap[slot], cmd)
 	}
 	return cmdsMap
 }
 
+// keyedCmds returns all the keyed commands from the cmds slice
+// it determines keyed commands by checking if the command has a first key position
+func (c *ClusterClient) keyedCmds(cmds []Cmder) []Cmder {
```
htemelski-oss: We could combine `mapCmdsBySlot` and `keyedCmds` in a single function.
Something like
```go
func (c *ClusterClient) slottedCommands(cmds []Cmder) map[int][]Cmder {
	cmdsSlots := map[int][]Cmder{}

	prefferedRandomSlot := -1
	for _, cmd := range cmds {
		if cmdFirstKeyPos(cmd) == 0 {
			continue
		}

		slot := c.cmdSlot(cmd, prefferedRandomSlot)
		if prefferedRandomSlot == -1 {
			prefferedRandomSlot = slot
		}

		cmdsSlots[slot] = append(cmdsSlots[slot], cmd)
	}

	return cmdsSlots
}
```

It would reduce the number of iterations and it will allow us to simplify the newly added logic in `processTxPipeline`

osscluster.go
```diff
@@ -1519,8 +1526,36 @@ func (c *ClusterClient) processTxPipeline(ctx context.Context, cmds []Cmder) err
 		return err
 	}
 
-	cmdsMap := c.mapCmdsBySlot(cmds)
+	cmdsMap := map[int][]Cmder{}
+	slot := -1
```
htemelski-oss, line 1530: Something like that
```go
	slot := -1
	slottedCommands := c.slottedCommands(cmds)
	switch len(slottedCommands) {
	case 0:
		slot = hashtag.RandomSlot()
	case 1:
		for sl := range slottedCommands {
			slot = sl
			break
		}
	default:
		setCmdsErr(cmds, ErrCrossSlot)
		return ErrCrossSlot
	}
```

osscluster.go
```diff
@@ -1519,8 +1526,36 @@ func (c *ClusterClient) processTxPipeline(ctx context.Context, cmds []Cmder) err
...
+			return ErrCrossSlot
+		}
+		// get the slot, should be only one
+		for sl := range cmdsBySlot {
+			slot = sl
+			break
+		}
+	}
+	// slot was not determined, try random one
+	if slot == -1 {
```
htemelski-oss: Do we need this part of the code
```go
	if slot == -1 {
		slot = hashtag.RandomSlot()
	}
	cmdsMap[slot] = cmds

	// TxPipeline does not support cross slot transaction.
	// double check the commands are in the same slot
	if len(cmdsMap) > 1 {
		setCmdsErr(cmds, ErrCrossSlot)
		return ErrCrossSlot
	}
```

It seems a bit redundant.
Also there no point in the next for loop
```go
	for slot, cmds := range cmdsMap {
```
Since there's only one element in the map anyway and we already know the key and the value in it

ring_test.go
```diff
@@ -304,20 +304,20 @@ var _ = Describe("Redis Ring", func() {
...
+			err := ring.Set(ctx, "key", "test", 0).Err()
 			Expect(err).NotTo(HaveOccurred())
 
 			var stack []string
 
 			ring.AddHook(&hook{
 				processHook: func(hook redis.ProcessHook) redis.ProcessHook {
 					return func(ctx context.Context, cmd redis.Cmder) error {
-						Expect(cmd.String()).To(Equal("ping: "))
+						Expect(cmd.String()).To(Equal("get key: "))
```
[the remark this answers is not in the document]
ndyakov: the `Set` is executed before the hook and only once the hook is attached, the `Get` is executed. This is why the hook validates the `Get` execution.

ring_test.go
```diff
@@ -304,20 +304,20 @@ var _ = Describe("Redis Ring", func() {
 			ring = redis.NewRing(opt)
 		})
 		It("supports Process hook", func() {
-			err := ring.Ping(ctx).Err()
+			err := ring.Set(ctx, "key", "test", 0).Err()
```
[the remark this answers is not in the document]
ndyakov: the Set is executed before the hook and only once the hook is attached, the Get is executed. This is why the hook validates the Get execution.

htemelski-redis (approved):

PR #3412, opened 2025-06-20
feat: optimize connection pool waitTurn

twz915:
feat: optimize connection pool waitTurn


ndyakov (approved):
@twz915 thank you for your contribution

PR #3417, opened 2025-06-26
fix(ring): Pipelined returns error when shard is down

### LINKED ISSUE #3402: `Ring.Pipelined()` returns `nil` error despite `DialTimeout` and failed shard connection
When using `Ring.Pipelined()` with an unreachable Redis address and a low `DialTimeout`, no error is returned, even though none of the commands are actually sent or executed. In contrast, `Client.Pipelined()` correctly returns a `dial tcp ... i/o timeout` error in this case.

This inconsistency can lead to the false assumption that a pipeline was successfully executed when in fact it was silently discarded.

## Expected Behavior

The call to `ring.Pipelined()` should return an error if:
- No shard was available due to DialTimeout
- None of the commands in the pipeline were actually executed

Returning `nil` in this case gives the false impression that the operations succeeded.

## Current Behavior

When all configured shard addresses are unreachable (e.g., due to a `DialTimeout`), `Ring.Pipelined()` returns **no error**, even though the pipeline is **never sent** or **executed** on any shard. This is inconsistent with `Client.Pipelined()` behavior, which correctly returns an `i/o timeout error` under similar conditions.

## Possible Solution

`Ring.Pipelined()` should return an error if it fails to send the pipeline to any shard, either due to connection issues or internal routing failure.

## Steps to Reproduce

1. Configure `Ring` with an unreachable shard address (e.g., `10.255.255.1:6379`)
2. Set `DialTimeout` to a short value (e.g., `1s`)
3. Call `Ring.Pipelined()` with any Redis commands (e.g., `HSet`, `Expire`)
4. Observe that **no error is returned**

```go
package main

import (
	"context"
	"fmt"
	"time"

)

func main() {
	ring := redis.NewRing(&redis.RingOptions{
		Addrs: map[string]string{"shard1": "10.255.255.1:6379"}, // unreachable IP
		DialTimeout: 1 * time.Second,
	})

	ctx := context.Background()

	// Using Pipelined to execute multiple commands atomically
	_, err := ring.Pipelined(ctx, func(pipe redis.Pipeliner) error {
		pipe.HSet(ctx, "key", "field", "value")
		pipe.Expire(ctx, "key", time.Minute)
		return nil
	})

	if err != nil {
		fmt.Printf("pipeline error: %v\n", err)
	} else {
		fmt.Println("pipeline succeeded") // unexpected behavior
	}
}
```

## Context (Environment)

In some cases, transient network failures or misconfigured shard entries result in silent data loss, since `Pipelined()` returns no error. This creates difficult-to-detect consistency bugs in our application.

## Detailed 

`Ring.Pipelined()` currently swallows errors when it fails to obtain a client for a shard. Specifically, this happens in the following code block: https://github.com/redis/go-redis/blob/master/ring.go#L801-L803

kosiew:
Closes
#3402

### Problem
When a shard in a Ringis unreachable due to a network issue or a low DialTimeout、
•.'Ring.Pipelined () would not return an error. Instead, it failed silently, giving the false impression that the pipelined commands were successfully executed. This behavior was inconsistent with 'Client.Pipelined () and could lead to undetected data loss.

### Solution
This change modifies the `generalProcessPipeline` function to correctly handle and propagate errors.
- It now captures the first error that occurs when attempting to get a shard client or execute the pipeline on a shard.
- A `sync.Mutex` is used to ensure thread-safe error reporting| from the concurrent shard operations.
- The captured error is returned to the caller of Pipelined () , ensuring that failures are no longer silent.
A new test case has also been added to verify that Pipelined () 、returns a `dial` error when a shard is unreachable, preventing future regressions.

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/cespare/xxhash/v2"
 	"github.com/dgryski/go-rendezvous" //nolint
+
 	"github.com/redis/go-redis/v9/auth"
 
 	"github.com/redis/go-redis/v9/internal"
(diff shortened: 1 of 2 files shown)


ndyakov:
Hello @kosiew, isn't this a duplicate of https://github.com/redis/go-redis/pull/3403 ?

kosiew:
Aaa..... 🤔
 sorry, I did not know it was taken already.

PR #3422, opened 2025-06-30
Implement Request and Response Policy Based Routing in Cluster Mode

ofekshenawa:
This PR introduces support for Redis COMMAND-based request_policy and response_policy routing for Redis commands when used in OSS Cluster client.

### Key Additions:

**Command Policy Loader:** Parses and caches COMMAND metadata with routing/aggregation tips on first use. 
**Routing Engine Enhancements:**
Implements support for all request policies: default(keyless), default(hashslot), all_shards, all_nodes, multi_shard, and special.
**Response Aggregator:** Combines multi-shard replies based on response_policy:
all_succeeded, one_succeeded, agg_sum, special, etc.
Includes custom handling for special commands like FT.CURSOR.
Raw Command Support: Policies are enforced on Client.Do(ctx, args...).


.github/workflows/build.yml
```diff
@@ -2,9 +2,9 @@ name: Go
 
 on:
   push:
-    branches: [master, v9, v9.7, v9.8]
+    branches: [master, v9, v9.7, '*']
   pull_request:
-    branches: [master, v9, v9.7, v9.8]
```
ndyakov: Let's revert this only to version branches or branches namespaced with our handlers. We can have a pattern `ndyakov/*`, `ofekshenawa/*`, `htemelski/*` and let's include `ce/*`

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+	keyedResults := make(map[string]interface{})
+	var firstErr error
+
+	for result := range results {
+		if result.err != nil && firstErr == nil {
+			firstErr = result.err
+		}
+		if result.cmd != nil && result.err == nil {
+			// For MGET, extract individual values from the array result
+			if strings.ToLower(cmd.Name()) == "mget" {
```
htemelski-oss: Do we actually need this special case?

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+		return c.executeMultiShard(ctx, cmd, policy)
+	case policy != nil && policy.Request == routing.ReqSpecial:
+		return c.executeSpecialCommand(ctx, cmd, policy, node)
+	default:
+		return c.executeDefault(ctx, cmd, node)
+	}
+}
+
+// getCommandPolicy retrieves the routing policy for a command
+func (c *ClusterClient) getCommandPolicy(ctx context.Context, cmd Cmder) *routing.CommandPolicy {
```
htemelski-oss: ~It seems like this will introduce a big overhead for each command execution.~
We should fetch all policies during the connection handshake

htemelski-redis (changes requested):

osscluster_router.go
htemelski-oss: Note: for the first stage we should use hard-coded policy manager that can be extended in the future to take into account the `COMMAND` command output

bobymicroby: @htemelski-redis 💡 Consider implementing a `PolicyResolverConfig` type that users can override via the client options. This config should map module__command_name to metadata (policies, key requirements, etc.).

Set hardcoded defaults in the client options, but allow users to override policies per command as needed.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+// ResponseAggregator defines the interface for aggregating responses from multiple shards.
+type ResponseAggregator interface {
+	// Add processes a single shard response.
+	Add(result interface{}, err error) error
+
+	// AddWithKey processes a single shard response for a specific key (used by keyed aggregators).
+	AddWithKey(key string, result interface{}, err error) error
+
+	// Finish returns the final aggregated result and any error.
+	Finish() (interface{}, error)
```
ndyakov:
```suggestion
	Result() (interface{}, error)
```

I don't like `Finish` as a method name. In my mind, it should do some grand work and become unusable. That is not the case and although there is work being done, the name `Aggregator` should aggregate things and get you a `Result`.

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+	return c.finishAggregation(cmd, aggregator)
+}
+
+// createAggregator creates the appropriate response aggregator
+func (c *ClusterClient) createAggregator(policy *routing.CommandPolicy, cmd Cmder, isKeyed bool) routing.ResponseAggregator {
+	cmdName := strings.ToLower(cmd.Name())
+	// For MGET without policy, use keyed aggregator
+	if cmdName == "mget" {
+		return routing.NewDefaultAggregator(true)
+	}
```
ndyakov: Since we are passing the `cmd.Name()` in `routing.NewResponseAggregator` this can be handler by it. If `policy` is `nil` for `mget`, maybe the `NewResponseAggregator` can accept a `policy` and check the nil as well`.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+
+	if err != nil && a.firstErr == nil {
+		a.firstErr = err
+		return nil
+	}
+	if err == nil && !a.hasResult {
+		a.result = result
+		a.hasResult = true
+	}
+	return nil
```
ndyakov: Couple of questions here:
1. Should we return only the first observed error?
2. Why are we overwriting the result?
3. Can't we just have an atomic boolean `hasError`?
4. Same, if we can have atomic `hasResult` we can drop the mutex.

My questions and my idea is that if we are going to return on the first error, we can do this with atomics and skip the cpu cycle for the mutex.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+}
```
ndyakov, line 171: Same as with `AllSucceededAggregator`. Maybe we can use atomics here.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+		return nil
+	}
+	if err == nil {
+		val, err := toInt64(result)
+		if err != nil && a.firstErr == nil {
+			a.firstErr = err
+			return nil
+		}
+		if err == nil {
+			a.sum += val
```
ndyakov: Again, maybe we can use atomic.Int64

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+	defer a.mu.Unlock()
+
+	if a.firstErr != nil {
+		return nil, a.firstErr
+	}
+	return a.sum, nil
+}
+
+// AggMinAggregator returns the minimum numeric value from all shards.
+type AggMinAggregator struct {
```
ndyakov, line 291: Looking at https://github.com/haraldrudell/parl , there is atomic min and atomic max implementations that we can also use.

ndyakov: p.s. I do suggest copying only the needed implementation or using it as reference to reimplement, not including the whole dependency. Of course, mentioning the creator in the code.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+
+func (a *AggMinAggregator) Finish() (interface{}, error) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	if a.firstErr != nil {
+		return nil, a.firstErr
+	}
+	if !a.hasResult {
+		return nil, fmt.Errorf("redis: no valid results to aggregate for min operation")
```
ndyakov: Can we extract such errors in a separate file?

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+	for i, err := range a.errors {
+		if err == nil {
+			return a.results[i], nil
+		}
+	}
+	if len(a.errors) > 0 {
+		return nil, a.errors[0]
+	}
+	return nil, nil
+}
```
ndyakov: I do think we should be able to control the priority here. I assume for some commands the errors will be with higher priority, for others - the results.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,589 @@
...
+	SpecialAggregatorRegistry[cmdName] = fn
+}
+
+// NewSpecialAggregator creates a special aggregator with command-specific logic if available.
+func NewSpecialAggregator(cmdName string) *SpecialAggregator {
+	agg := &SpecialAggregator{}
+	if fn, exists := SpecialAggregatorRegistry[cmdName]; exists {
+		agg.SetAggregatorFunc(fn)
+	}
+	return agg
```
ndyakov: `SetAggregatorFunc` is only used internally in this package, I assume it can be private if needed at all, see next comment.

ndyakov, line 1001:
```suggestion
func NewSpecialAggregator(cmdName string) *SpecialAggregator {
    fn := SpecialAggregatorRegistry[cmdName]
	return &SpecialAggregator{aggregatorFunc: fn}
```

I do think this should be doable and we are going to remove the need for `SetAggregatorFunc` and therefore - locking and unlocking the mutex.

internal/routing/policy.go
```diff
@@ -0,0 +1,135 @@
...
+type CommandPolicy struct {
+	Request  RequestPolicy
+	Response ResponsePolicy
+	// Tips that are not request_policy or response_policy
+	// e.g nondeterministic_output, nondeterministic_output_order.
+	Tips map[string]string
+}
+
+func (p *CommandPolicy) CanBeUsedInPipeline() bool {
+	return p.Request != ReqAllNodes && p.Request != ReqAllShards && p.Request != ReqMultiShard
```
ndyakov, line 138: What about `special`? Can it be used in a pipeline?

internal/routing/shard_picker.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"math/rand"
+	"sync/atomic"
+)
+
+// ShardPicker chooses “one arbitrary shard” when the request_policy is
+// ReqDefault and the command has no keys.
+type ShardPicker interface {
+	Next(total int) int // returns an index in [0,total)
+}
```
ndyakov, line 12: Those are great, can we implement `StaticShardPicker` or `StickyShardPicker` that will always return the same shard. I do think this can be helpful for testing. This is not a blocker by any means.

command.go
```diff
@@ -14,9 +14,94 @@ import (
...
 
+// CmdTyper interface for getting command type
+type CmdTyper interface {
+	GetCmdType() CmdType
+}
+
+// CmdTypeGetter interface for getting command type without circular imports
+type CmdTypeGetter interface {
+	GetCmdType() CmdType
+}
```
ndyakov, line 79: Why do we need two interfaces that are the same in the same package?

command.go
```diff
@@ -876,7 +1070,7 @@ func (cmd *StringCmd) Bool() (bool, error) {
 	if cmd.err != nil {
 		return false, cmd.err
 	}
-	return strconv.ParseBool(cmd.val)
+	return strconv.ParseBool(cmd.Val())
```
ndyakov: why was this change needed?

command.go
```diff
@@ -3414,6 +4135,7 @@ type CommandInfo struct {
 	LastKeyPos  int8
 	StepCount   int8
 	ReadOnly    bool
+	Tips        *routing.CommandPolicy
```
ndyakov: In the CommandPolicy struct there is already a field named `Tips`. Isn't `Policy` or `CommandPolicy` / `RoutingPolicy` a better name for this field?

command.go
```diff
@@ -3540,9 +4263,34 @@ func (cmd *CommandsInfoCmd) readReply(rd *proto.Reader) error {
...
+					rawTips[tip] = ""
+					continue
+				}
+
+				// Handle normal key:value tips
+				k, v, ok := strings.Cut(tip, ":")
+				if !ok {
+					return fmt.Errorf("redis: unexpected tip %q in COMMAND reply", tip)
+				}
+				rawTips[k] = v
```
ndyakov:
```suggestion
				k, v, ok := strings.Cut(tip, ":")
				if !ok {
				    // Handle tips that don't have a colon (like "nondeterministic_output")
					rawTips[tip] = ""
				} else {
				    // Handle normal key:value tips
				    rawTips[k] = v
				}
```

command.go
```diff
@@ -3557,13 +4305,47 @@ func (cmd *CommandsInfoCmd) readReply(rd *proto.Reader) error {
...
+		for k, v := range cmd.val {
+			if v != nil {
+				newInfo := &CommandInfo{
+					Name:        v.Name,
+					Arity:       v.Arity,
+					FirstKeyPos: v.FirstKeyPos,
+					LastKeyPos:  v.LastKeyPos,
+					StepCount:   v.StepCount,
+					ReadOnly:    v.ReadOnly,
+					Tips:        v.Tips, // CommandPolicy can be shared as it's immutable
```
ndyakov: The `CommandPolicy` is not immutable the way it is implemented. I do agree it won't probably be changed and doesn't need to be cloned, but should we implement it in an immutable way?

command.go
```diff
@@ -3593,6 +4378,43 @@ func (c *cmdsInfoCache) Get(ctx context.Context) (map[string]*CommandInfo, error
...
+			}
+		}
+	}
+	tips := make(map[string]string, len(commandInfoTips))
+	for k, v := range commandInfoTips {
+		if k == requestPolicy || k == responsePolicy {
+			continue
+		}
+		tips[k] = v
+	}
```
ndyakov: can't we do both of those in a single range over `commandInfoTips`?

internal/routing/aggregator.go
htemelski-oss: Wouldn't this be achieved using the aggregatorFunc?

command.go
```diff
@@ -5615,3 +6836,292 @@ func (cmd *MonitorCmd) Stop() {
 	defer cmd.mu.Unlock()
 	cmd.status = monitorStatusStop
 }
+
+// ExtractCommandValue extracts the value from a command result using the fast enum-based approach
+func ExtractCommandValue(cmd interface{}) interface{} {
```
ndyakov: I assume all cases (types) for which `interface{ Val() interface{} }` is used for extracting the value can be combined together?

command.go
```diff
@@ -5615,3 +6836,292 @@ func (cmd *MonitorCmd) Stop() {
...
+			// For unknown command types, return nil
+			return nil
+		}
+	}
+
+	// If we can't get the command type, return nil
+	return nil
+}
+
+func (cmd *MonitorCmd) Clone() Cmder {
```
ndyakov: let's move this above the `ExtractCommandValue` function

json.go
```diff
@@ -274,6 +297,23 @@ func (cmd *IntPointerSliceCmd) readReply(rd *proto.Reader) error {
 	return nil
 }
 
+func (cmd *IntPointerSliceCmd) Clone() Cmder {
```
ndyakov, line 316: it's tricky here. do we need to return the same pointer or do we only want the value when cloning?

osscluster.go
```diff
@@ -1844,8 +1861,19 @@ func (c *ClusterClient) cmdsInfo(ctx context.Context) (map[string]*CommandInfo,
 	return nil, firstErr
 }
 
+// cmdInfo will fetch and cache the command policies after the first execution
 func (c *ClusterClient) cmdInfo(ctx context.Context, name string) *CommandInfo {
-	cmdsInfo, err := c.cmdsInfoCache.Get(ctx)
+	// Use a separate context that won't be canceled to ensure command info lookup
+	// doesn't fail due to original context cancellation
+	cmdInfoCtx := context.Background()
```
ndyakov: most of the time the cmdInfo should be cached already, why don't we just use the `c.context(ctx)` to determine if the original one (with it's timeout) be used or a Background context will be used when `c.opt.ContextTimeoutEnabled` is false.

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+	case policy != nil && policy.Request == routing.ReqAllShards:
+		return c.executeOnAllShards(ctx, cmd, policy)
+	case policy != nil && policy.Request == routing.ReqMultiShard:
+		return c.executeMultiShard(ctx, cmd, policy)
+	case policy != nil && policy.Request == routing.ReqSpecial:
+		return c.executeSpecialCommand(ctx, cmd, policy, node)
+	default:
+		return c.executeDefault(ctx, cmd, node)
+	}
+}
```
ndyakov, line 59:
```suggestion
func (c *ClusterClient) routeAndRun(ctx context.Context, cmd Cmder, node *clusterNode) error {
	policy := c.getCommandPolicy(ctx, cmd)
    if policy == nil {
        return c.executeDefault(ctx, cmd, node)
    }
	switch policy.Request {
	case routing.ReqAllNodes:
		return c.executeOnAllNodes(ctx, cmd, policy)
	case routing.ReqAllShards:
		return c.executeOnAllShards(ctx, cmd, policy)
	case routing.ReqMultiShard:
		return c.executeMultiShard(ctx, cmd, policy)
	case routing.ReqSpecial:
		return c.executeSpecialCommand(ctx, cmd, policy, node)
	default:
		return c.executeDefault(ctx, cmd, node)
	}
}
```

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+	}
+	return nil
+}
+
+// executeDefault handles standard command routing based on keys
+func (c *ClusterClient) executeDefault(ctx context.Context, cmd Cmder, node *clusterNode) error {
+	if c.hasKeys(cmd) {
+		// execute on key based shard
+		return node.Client.Process(ctx, cmd)
+	}
```
ndyakov: Do we know that this `node` servers the slot for the `key`?

internal/routing/aggregator.go
htemelski-oss: For the all succeed policy, we either return one of the replies if there is no error, or one of the errors if there's at least one
So
1. Yes, returning only the first error is sufficient
2. We are setting it only once
3/4. I feel that using atomics will overcomplicate the aggregators, plus there are some [caveats](https://pkg.go.dev/sync/atomic#pkg-notes) to using them. I think we should try to maximize the compatibility of the library

internal/routing/aggregator.go
htemelski-oss: Same as above

internal/routing/aggregator.go
htemelski-oss: -||-

internal/routing/aggregator.go
htemelski-oss: -||-

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+	return c.executeOnArbitraryShard(ctx, cmd)
```
ndyakov: since it doesn't matter and there is already some node selected, why not use it?

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+	case CmdTypeZSliceWithKey:
+		return NewZSliceWithKeyCmd(ctx, args...)
+	case CmdTypeFunctionList:
+		return NewFunctionListCmd(ctx, args...)
+	case CmdTypeFunctionStats:
+		return NewFunctionStatsCmd(ctx, args...)
+	case CmdTypeKeyFlags:
+		return NewKeyFlagsCmd(ctx, args...)
+	case CmdTypeDuration:
+		return NewDurationCmd(ctx, time.Second, args...)
```
ndyakov: Some `CmdTypeDuration` commands do use `time.Milisecond` as precision, see `PExpireTime` for example. Shouldn't we use it here so we don't lose precision?

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+// setCommandValue sets the aggregated value on a command using the enum-based approach
+func (c *ClusterClient) setCommandValue(cmd Cmder, value interface{}) error {
+	// If value is nil, it might mean ExtractCommandValue couldn't extract the value
+	// but the command might have executed successfully. In this case, don't set an error.
+	if value == nil {
+		// Check if the original command has an error - if not, the nil value is not an error
+		if cmd.Err() == nil {
+			// Command executed successfully but value extraction failed
+			// This is common for complex commands like CLUSTER SLOTS
+			// The command already has its result set correctly, so just return
```
ndyakov: I do not understand that comment here. Why the value extraction returned `nil`? Can we make sure the cmd has value set at least? If it doesn't, we may return a cmd with nil value and nil error, which doesn't make sense.

osscluster_router.go
```diff
@@ -0,0 +1,966 @@
...
+			if key, ok := value.(string); ok {
+				c.SetVal(key, []string{}) // Default empty values
+			}
+		}
+	case CmdTypeZSliceWithKey:
+		if c, ok := cmd.(*ZSliceWithKeyCmd); ok {
+			// ZSliceWithKeyCmd needs a key string and Z slice
+			if key, ok := value.(string); ok {
+				c.SetVal(key, []Z{}) // Default empty Z slice
+			}
```
ndyakov, line 785: why are we setting empty values here?

osscluster_router.go
htemelski-oss: We have two different ways of picking an arbitrary shard, either round robin or a random one

osscluster_router.go
htemelski-oss: Yes, the node should've been selected based on the slot osscluster.go:L1906 
```
func (c *ClusterClient) cmdNode(
```

osscluster_router.go
htemelski-oss: No idea tbh, will look into it

internal/routing/policy.go
htemelski-oss: My understanding is that `special` should be handled on a case-by-case basis

internal/routing/aggregator.go
ndyakov: We already do use atomics in the library so using them here is not a problem in terms of compatibility. As for increasing the complexity - that is rarely the case for such small units as the Aggregators. Let's try it out, I do agree it will be different than the mutex approach but in this specific case I do see the use of atomics as the correct approach for leveraging the cpu as much as possible without having the additional cpu cycle for the mutex.

For most aggregators we are either going to sum numbers, calculate min/max or check flags. This can be covered by atomics. For the cases where we are combining more complex response, it makes sense to lock and append to a slice.

internal/routing/aggregator.go
ndyakov: You are correct it can be handled by the `aggregatorFunc`.

osscluster_router.go
ndyakov: Yes, I understand that, but for some reason there is already a node selected here that may have been selected because `MOVED` OR normal key based selection. Why do we have to reselect the node? Shouldn't this selection of arbitrary node be done outside, so we do the node selection only one time and the node on line #52 is the one that should be used for this command?

command.go
htemelski-oss: Not sure that I completely understand the question

command.go
htemelski-oss: My understanding is that by "immutable" @ofekshenawa meant that it won't be modified once set

command.go
htemelski-oss: Not sure, for consistency maybe?

osscluster_router.go
htemelski-oss: Good catch, fixed

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+
+func (a *AllSucceededAggregator) BatchAdd(results map[string]interface{}) error {
+	for _, res := range results {
+		err := a.Add(res, nil)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
ndyakov: for this, why do we need to add any additional results? if we don't have errors, we do need just one result, correct?

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+}
+
+func (a *AllSucceededAggregator) BatchWithErrs(values []AggregatorResErr) error {
+	for _, val := range values {
+		err := a.Add(val.Result, val.Err)
+		if err != nil {
+			return err
+		}
+	}
```
ndyakov: the first error we observe here is sufficient to stop the execution, any following call to `Add` is not needed.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *OneSucceededAggregator) BatchAdd(results map[string]interface{}) error {
+	for _, res := range results {
+		err := a.Add(res, nil)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: same as previous, if we don't have errors, one result is sufficient here

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *OneSucceededAggregator) BatchWithErrs(values []AggregatorResErr) error {
+	for _, val := range values {
+		err := a.Add(val.Result, val.Err)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: same, if we don't have an error, we can skip adding the rest.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+
+func (a *OneSucceededAggregator) Result() (interface{}, error) {
+	res, e := a.res.Load(), a.err.Load()
+	if res != nil {
+		return nil, e.(error)
+	}
+
+	return res, nil
+}
```
ndyakov, line 214:
```suggestion
func (a *OneSucceededAggregator) Result() (interface{}, error) {
	res, e := a.res.Load(), a.err.Load()
	if res == nil {
		return nil, e.(error)
	}

	return res, nil
}
```

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *AggSumAggregator) BatchAdd(results map[string]interface{}) error {
+	for _, res := range results {
+		err := a.Add(res, nil)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: how about a little optimization, we can sum the values here and only do one Add for the atomic.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *AggSumAggregator) BatchWithErrs(values []AggregatorResErr) error {
+	for _, val := range values {
+		err := a.Add(val.Result, val.Err)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: we can stop on the first observed error.
Same as above, we can sum the numbers here.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+	if err != nil {
+		a.err.CompareAndSwap(nil, err)
+		return nil
+	}
+
+	intVal, e := toInt64(result)
+	if e != nil {
+		a.err.CompareAndSwap(nil, err)
+		return nil
+	}
```
ndyakov: shoudn't we return this error if the transform to int failed?

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *AggMinAggregator) BatchAdd(results map[string]interface{}) error {
+	for _, res := range results {
+		err := a.Add(res, nil)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: we can find the min and only use that for a.Add

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+	for _, val := range values {
+		err := a.Add(val.Result, val.Err)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
```
ndyakov: same, and maybe fail on first error?

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+	if err != nil {
+		a.err.CompareAndSwap(nil, err)
+		return nil
+	}
+
+	intVal, e := toInt64(result)
+	if e != nil {
+		a.err.CompareAndSwap(nil, err)
+		return nil
+	}
```
ndyakov: I was thinking we would like to return that err instead of storing it.
```suggestion
	intVal, e := toInt64(result)
	if e != nil {
		return err
	}
```

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *AggMaxAggregator) BatchAdd(results map[string]interface{}) error {
+	for _, res := range results {
+		err := a.Add(res, nil)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: find max before `s.Add`

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *AggMaxAggregator) BatchWithErrs(values []AggregatorResErr) error {
+	for _, val := range values {
+		err := a.Add(val.Result, val.Err)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: same, sounds reasonable to fail on first error

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+	if err != nil {
+		a.err.CompareAndSwap(nil, err)
+		return nil
+	}
+
+	val, e := toBool(result)
+	if e != nil {
+		a.err.CompareAndSwap(nil, e)
+		return nil
+	}
```
ndyakov, line 483: if it is not bool, return err?

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *AggLogicalAndAggregator) BatchAdd(results map[string]interface{}) error {
+	for _, res := range results {
+		err := a.Add(res, nil)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
ndyakov: As soon as we see res=false, we can stop adding results.

ndyakov: or even better, calculate the result of all res and do only one add.

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+
+	return a.add(result, err)
+}
+
+func (a *DefaultKeylessAggregator) BatchAdd(results map[string]interface{}) error {
+	for _, res := range results {
+		err := a.Add(res, nil)
+		if err != nil {
+			return err
+		}
```
ndyakov: let's have one lock here and use `a.add` (adding without a lock)

internal/routing/aggregator.go
```diff
@@ -0,0 +1,826 @@
...
+func (a *DefaultKeylessAggregator) AddWithKey(key string, result interface{}, err error) error {
+	return a.Add(result, err)
+}
+
+func (a *DefaultKeylessAggregator) BatchWithErrs(values []AggregatorResErr) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	for _, val := range values {
+		err := a.Add(val.Result, val.Err)
```
ndyakov:
```suggestion
		err := a.add(val.Result, val.Err)
```
or there will be a deadlock

internal/routing/aggregator.go
htemelski-oss: Modified the logic, we now add possible errors

osscluster.go
```diff
@@ -913,31 +921,32 @@ func (c *clusterStateHolder) ReloadOrGet(ctx context.Context) (*clusterState, er
...
 		nodes: newClusterNodes(opt),
 	}
 
-	c.state = newClusterStateHolder(c.loadState)
 	c.cmdsInfoCache = newCmdsInfoCache(c.cmdsInfo)
-	c.cmdable = c.Process
 
+	c.state = newClusterStateHolder(c.loadState)
+
+	c.cmdInfoResolver = c.NewDynamicResolver()
```
htemelski-oss: How should the resolver be configured initially?

internal/routing/aggregator.go
```diff
@@ -0,0 +1,979 @@
...
+		return nil, ErrMinAggregation
+	}
+	return val, nil
+}
+
+// AggMaxAggregator returns the maximum numeric value from all shards.
+type AggMaxAggregator struct {
+	err atomic.Value
+	res *util.AtomicMax
+}
```
ndyakov, line 382: general question, are those min,max aggregators only for ints?

internal/routing/policy.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (p *CommandPolicy) IsReadOnly() bool {
+	_, readOnly := p.Tips[ReadOnlyCMD]
+	fmt.Println(readOnly)
```
ndyakov: forgotten debug message i guess?

internal/routing/shard_picker.go
ndyakov: can we just add a todo for those, i think they can be useful in the future.

command.go
ndyakov: well, the more consistent thing would have been to keep it as `cmd.val` if it is not needed, feel free to return back to `cmd.val`

command.go
```diff
@@ -3593,6 +4399,43 @@ func (c *cmdsInfoCache) Get(ctx context.Context) (map[string]*CommandInfo, error
...
+			}
+		}
+	}
+	tips := make(map[string]string, len(commandInfoTips))
+	for k, v := range commandInfoTips {
+		if k == requestPolicy || k == responsePolicy {
+			continue
+		}
+		tips[k] = v
+	}
```
ndyakov:
```suggestion

	tips := make(map[string]string, len(commandInfoTips))
	for k, v := range commandInfoTips {
		if k == requestPolicy {
			if p, err := routing.ParseRequestPolicy(v); err == nil {
				req = p
			}
			continue
		}
		if k == responsePolicy {
			if p, err := routing.ParseResponsePolicy(v); err == nil {
				resp = p
			}
			continue
		}
		tips[k] = v
	}

```

json.go
ndyakov: still not sure where this type is used and if we would like the pointer of the value, cc @ofekshenawa , @htemelski-redis

main_test.go
```diff
@@ -105,6 +105,7 @@ var _ = BeforeSuite(func() {
 
 	if RedisVersion < 7.0 || RedisVersion > 9 {
 		panic("incorrect or not supported redis version")
+
```
ndyakov:
```suggestion
```

osscluster_router.go
```diff
@@ -0,0 +1,998 @@
...
+
+// executeDefault handles standard command routing based on keys
+func (c *ClusterClient) executeDefault(ctx context.Context, cmd Cmder, policy *routing.CommandPolicy, node *clusterNode) error {
+	if c.hasKeys(cmd) {
+		// execute on key based shard
+		return node.Client.Process(ctx, cmd)
+	}
+	if policy != nil {
+
+		fmt.Println(policy.Tips)
```
ndyakov: cut the println

osscluster_router.go
```diff
@@ -0,0 +1,998 @@
...
+		if c.readOnlyEnabled() && policy.IsReadOnly() {
+			fmt.Println("will execute on arbitrary node")
```
ndyakov: here as well

osscluster_router.go
```diff
@@ -0,0 +1,998 @@
...
+// executeCursorCommand handles FT.CURSOR commands with sticky routing
+func (c *ClusterClient) executeCursorCommand(ctx context.Context, cmd Cmder) error {
+	args := cmd.Args()
+	if len(args) < 4 {
+		return fmt.Errorf("redis: FT.CURSOR command requires at least 3 arguments")
+	}
+
+	cursorID, ok := args[3].(string)
+	if !ok {
+		return fmt.Errorf("redis: invalid cursor ID type")
```
ndyakov: if there are no additional data in those errors, feel free to predefine them as constant so we don't allocate them on each occurrence

osscluster_router.go
ndyakov: cc @ofekshenawa

osscluster_router.go
ndyakov: cc @ofekshenawa , @htemelski-redis

osscluster_router.go
ndyakov: cc @ofekshenawa , @htemelski-redis

osscluster_router.go
```diff
@@ -0,0 +1,998 @@
...
+	}
+
+	return nil
+}
+
+// setCommandValueReflection is a fallback function that uses reflection
+func (c *ClusterClient) setCommandValueReflection(cmd Cmder, value interface{}) error {
+	cmdValue := reflect.ValueOf(cmd)
+	if cmdValue.Kind() != reflect.Ptr || cmdValue.IsNil() {
+		return fmt.Errorf("redis: invalid command pointer")
```
ndyakov: here as well, preallocate

osscluster_router.go
```diff
@@ -0,0 +1,998 @@
...
+				args = append(args, reflect.ValueOf([]Z{}))
+			} else {
+				args = append(args, reflect.ValueOf([]string{}))
+			}
+		}
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
+			cmd.SetErr(fmt.Errorf("redis: failed to set command value: %v", r))
```
ndyakov, line 986: why don't we return the error as well? it will return nil, but the err will be set on the cmd.

.github/workflows/build.yml
```diff
@@ -2,9 +2,9 @@ name: Go
 
 on:
   push:
-    branches: [master, v9, v9.7, v9.8]
+    branches: [master, v9, v9.7, v9.8, 'ndyakov/*', 'ofekshenawa/*', 'htemelski/*', 'ce/*', '*']
   pull_request:
-    branches: [master, v9, v9.7, v9.8]
+    branches: [master, v9, v9.7, v9.8, 'ndyakov/*', 'ofekshenawa/*', 'htemelski/*', 'ce/*', '*']
```
ndyakov: fetch master for those

internal/routing/aggregator.go
htemelski-oss: That's a good question, the initial implementation was for ints only, but looking at the docs, we should support any numerical types

osscluster_router.go
htemelski-oss: Not sure, how good of a practice is to modify the return value from within recover

internal/routing/policy.go
ndyakov: makes sense, I think we are going to work towards removing the `special` keyword and be able to determine better if a command can be used in a pipeline.

osscluster_router.go
ndyakov: cc @ofekshenawa , @htemelski-redis

json.go
ofekshenawa: You're correct, done.

internal/routing/shard_picker.go
ofekshenawa: Great suggestion, added.

osscluster_router.go
ofekshenawa: The re-selection is required so keyless commands respect the configured ShardPicker strategy.
cmdNode() picks nodes by hash slot, which is correct for keyed commands but ignores the picker for keyless ones. Calling executeOnArbitraryShard() ensures keyless commands use the user’s picker

osscluster_router.go
ofekshenawa: You're absolutely right, I changed it

osscluster_router.go
ndyakov: But if the command is keyless, what will be the `node` here? Won't it be picked based on the ShardPolicy? Will it be just a random node? Can't we pick the correct node there?

osscluster_router.go
ndyakov: it is possible exactly for scenarios like this one.

htemelski-redis (approved):

go.mod
```diff
@@ -9,6 +9,8 @@ require (
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f
 )
 
+require go.uber.org/atomic v1.11.0
+
```
ndyakov, line 13: i personally do not like that we are pulling the whole package just for the `Float64` atomic type, since its implementation is ~150 lines (that we actually don't need and ca implement here with Uint64 and some math.

PR #3425, opened 2025-07-09
feat(proto): support scanning uuid.UUID in Scan()

### LINKED ISSUE #3139: Can't get UUID with Scan()

LinYuan201712:
### What this PR does

Adds support for scanning Redis string/bytes values into `uuid.UUID` via the `proto.Scan()` function.

### Why

Currently, scanning into `*uuid.UUID` fails even if the Redis string is a valid UUID. This PR enhances `proto.Scan()` to handle `*uuid.UUID` by parsing string/[]byte via `uuid.Parse`.

### How

- Extended the type switch in `proto.Scan` to detect `*uuid.UUID`
- Uses `util.BytesToString` to convert input to string
- Uses `uuid.Parse` to validate and assign
- Added unit test: `TestScanUUIDFromRedis`

### Related Issue

Fixes #3139 (or your own issue link)

### DIFF
--- a/internal/proto/scan.go
+++ b/internal/proto/scan.go
@@ -6,7 +6,7 @@ import (
 	"net"
 	"reflect"
 	"time"
-
+	"github.com/google/uuid"
 	"github.com/redis/go-redis/v9/internal/util"
 )
 
@@ -22,6 +22,14 @@ func Scan(b []byte, v interface{}) error {
 		return nil
 	case *[]byte:
 		*v = b
+		return nil
+	case *uuid.UUID:
+		str := util.BytesToString(b)
+		u, err := uuid.Parse(str)
+		if err != nil {
+			return fmt.Errorf("redis: failed to parse uuid: %w", err)
+		}
+		*v = u
 		return nil
 	case *int:
 		var err error
@@ -120,6 +128,7 @@ func Scan(b []byte, v interface{}) error {
 	case *net.IP:
 		*v = b
 		return nil
+
 	default:
 		return fmt.Errorf(
 			"redis: can't unmarshal %T (consider implementing BinaryUnmarshaler)", v)
--- a/internal/proto/test_scanuuid.go
+++ b/internal/proto/test_scanuuid.go
@@ -1,0 +1,20 @@
+package proto
+
+import (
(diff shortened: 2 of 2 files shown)


htemelski-oss:
Hello @LinYuan201712 
Thanks for opening the PR, it seems that you forgot to include the go.mod and go.sum files

internal/proto/scan.go
```diff
@@ -23,6 +23,14 @@ func Scan(b []byte, v interface{}) error {
 	case *[]byte:
 		*v = b
 		return nil
+	case *uuid.UUID:
```
justinhwang, line 26: Instead of adding support for a single uuid package, you could have individual uuid packages implement `encoding.BinaryUnmarshaler` to hit the case below. Not everyone who uses this package would use `github.com/google/uuid` as their uuid package

justinhwang: Given `github.com/google/uuid` [already implements it](https://github.com/google/uuid/blob/2d3c2a9cc518326daf99a383f07c4d3c44317e4d/marshal.go#L320), you're probably looking for #3144 instead

PR #3431, opened 2025-07-15
Add tracing hooks for existing connections

AllanGuigou:
Previously the redis/go-redis library would call newTracingHook with no connString as default tracing, but that was recently removed in 9.8.0 which has caused our clients to lose oteltracing if connections are created before tracing can be instrumented.


ndyakov:
Hello @AllanGuigou can you please cleanup the PRs, I can see two opened and the build is failing for both. Thank you.

PR #3432, opened 2025-07-15
Add tracing hooks for existing connections

AllanGuigou:
Previously `InstruementTracing` would call `newTracingHook` with no connString as a default tracing hook, but that was recently removed in v9.8.0 which has caused our clients to lose tracing if connections are created before tracing can be instrumented. In https://github.com/redis/go-redis/pull/3433, I've reverted the change to remove the default tracing hook, but a potential solution would be to use `ForEachShard` to create tracing hooks for any existing connections. This is a breaking change since it requires context, open to feedback if this PR should be moved forward instead of the revert.

### DIFF
--- a/extra/redisotel/tracing.go
+++ b/extra/redisotel/tracing.go
@@ -21,7 +21,7 @@ const (
 	instrumName = "github.com/redis/go-redis/extra/redisotel"
 )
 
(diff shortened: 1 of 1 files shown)


ndyakov:
Hello @AllanGuigou can you please cleanup the PRs, I can see two opened and the build is failing for both. Thank you.

PR #3435, opened 2025-07-16
chore: use go 1.20 idiomatic string<->byte conversion

justinhwang:
As of Go 1.20, this is the idiomatic way to convert from Bytes to String and vice-versa.

This updates `go.mod` to 1.20, builds have already been running on 1.23.x and 1.24.x since #3274.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/redis/go-redis/v9
 
-go 1.18
+go 1.20
 
 require (
 	github.com/bsm/ginkgo/v2 v2.12.0
--- a/internal/util/unsafe.go
+++ b/internal/util/unsafe.go
@@ -8,15 +8,10 @@ import (
 
 // BytesToString converts byte slice to string.
 func BytesToString(b []byte) string {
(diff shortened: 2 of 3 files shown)


justinhwang:
@ndyakov do you think you could help take a look? thanks!

ndyakov:
@justinhwang Thank you for your contribution. I do like the idea and would like to upgrade to a newer go version, but we discussed to put this one back a bit since it would be a breaking change and we would have to release a new major version. I will tag this PR and review / merge it once we get there (hopefully sooner and previously expected).

justinhwang:
@ndyakov It isn't a breaking change no? Semantically it does the same thing under the hood, just uses Go 1.20's safer way of doing it. Unless you're referring to the Go version bump as a breaking change

ndyakov:
@justinhwang yes, I am referring to bumping go to a new version as a "breaking change" since users will have to upgrade to a new version and this may actually not work well for all of them.

justinhwang:
ack yeah that makes sense. Please feel free to merge whenever you're ready for v10!

PR #3443, opened 2025-07-23
DOC-5472 time series doc examples

andy-stark-redis:
[DOC-5472](https://redislabs.atlassian.net/browse/DOC-5472)

Go versions of the Python examples in the [time series](https://redis.io/docs/latest/develop/data-types/timeseries/) doc page.

Couple of things to note:
- I've assumed that using 0 and `math.MaxInt64` instead of `-` and `+` for minimum/maximum timestamps is OK (eg, line 211), but if there's a better way to specify this then let me know.
- Some of the `TSMGet`/`TSMRange` examples return maps, so the ordering keys in nondeterministic. I've used code that's a bit more complicated than usual to print out the results in a deterministic way (eg, the example at line 438). If there's a simpler or more idiomatic way to handle this situation then I'll use that instead.

[DOC-5472]: https://redislabs.atlassian.net/browse/DOC-5472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ndyakov:
Hello @andy-stark-redis , I will review this till end of week. Thank you.

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+	}
+
+	fmt.Println(res5["retentionTime"]) // >>> 100
+	// STEP_END
+
+	// STEP_START create_labels
+	res6, err := rdb.TSAddWithArgs(
+		ctx,
+		"thermometer:3",
+		1, 10.4,
```
ndyakov:
```suggestion
		1, 
		10.4,
```

Maybe it is slightly more readable on separate lines, let me know what you think.

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+	}
+
+	fmt.Println(res3["totalSamples"]) // >>> 0
+	// STEP_END
+
+	// STEP_START create_retention
+	res4, err := rdb.TSAddWithArgs(
+		ctx,
+		"thermometer:2",
+		1, 10.8,
```
ndyakov: same suggestion to split the arguments on two separate lines.

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+		"thermometer:3",
+		1, 10.4,
+		&redis.TSOptions{
+			Labels: map[string]string{
+				"location": "UK",
+				"type":     "Mercury",
+			},
+		},
+	).Result()
+
```
ndyakov: you can remove the empty line before checking the err

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(res6) // >>> 1
+
+	res7, err := rdb.TSInfo(ctx, "thermometer:3").Result()
+
```
ndyakov: can be removed.

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+	rdb.FlushDB(ctx)
+	rdb.Del(ctx, "thermometer:1", "thermometer:2", "thermometer:3")
+	// REMOVE_END
+
+	// STEP_START create
+	res1, err := rdb.TSCreate(ctx, "thermometer:1").Result()
+
+	if err != nil {
+		panic(err)
+	}
```
ndyakov: Normally, when you are reading go code, the error check is something that happens 99% either on the same line as executing a method or on the next line. It would be better to remove those empty lines between the execution and `if err != nil`

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+
+	fmt.Println(res7)
+	// >>> [{1 14} {0 18}]
+	// STEP_END
+
+	// STEP_START range_filter
+	res8, err := rdb.TSRangeWithArgs(
+		ctx,
+		"rg:1",
+		0, math.MaxInt64,
```
ndyakov: split on two separate lines

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(res8) // >>> [{0 18} {2 22} {4 24}]
+
+	res9, err := rdb.TSRevRangeWithArgs(
+		ctx,
+		"rg:1",
+		0, math.MaxInt64,
```
ndyakov: split on two separate lines

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(res9) // >>> [{4 24} {2 22}]
+
+	res10, err := rdb.TSRevRangeWithArgs(
+		ctx,
+		"rg:1",
+		0, math.MaxInt64,
```
ndyakov: split on two separate lines

doctests/timeseries_tut_test.go
```diff
@@ -0,0 +1,1216 @@
...
+		panic(err)
+	}
+
+	fmt.Println(res43) // >>> [3 3 3 3]
+
+	// The result pairs contain the timestamp and the maximum sample value
+	// for the country at that timestamp.
+	res44, err := rdb.TSMRangeWithArgs(
+		ctx,
+		0, math.MaxInt64,
```
ndyakov: split on two separate lines

doctests/timeseries_tut_test.go
andy-stark-redis: @ndyakov Yeah, I'm not 100% sure about this. The reason I formatted them like this was that the timestamp and value together form a logical pair. It's a bit like the following code from the hash examples:

```go
hashFields := []string{
	"model", "Deimos",        // Key and value are just strings in the array but
	"brand", "Ergonom",      // this formatting emphasises the key-value pairs.
	"type", "Enduro bikes",
	"price", "4972",
}
```

But if it doesn't look like normal Go style to do it this way for the timestamps then I'll put them on separate lines.

doctests/timeseries_tut_test.go
andy-stark-redis: Fixed.

doctests/timeseries_tut_test.go
ndyakov: I see your point. Since this is supposed to be part of the documentation, I assume it would be better to look similar to other language that have similar APIs. I personally would prefer separate lines for the current code, but do see the benefit of having them combined in the hash example that you shared.

andy-stark-redis: OK, I'll use separate lines for this example. Thanks for clarifying this :-)

andy-stark-redis: Fixed.

Issue #3424, opened 2025-07-08
OpenTelemetry metrics create a memory leak when using ephemeral redis clients


rokn:
Issue tracker is used for reporting bugs and discussing new features. Please use
[stackoverflow](https://stackoverflow.com) for supporting issues.

## Expected Behavior

When closing a redis client the open telemetry should stop observing it. Another option is to provide a way to unregister the listener

## Current Behavior

Currently when instrumenting a redis client there is no way to stop observing it using the otel metrics. This also creates a hanging pointer to the redis client instance which can create memory leaks if the application creates many redis client instances.

## Possible Solution

1. Automatically stop observing the client when it's closed / destroyed in any way. (Prefferred option)
2. Provide a way for manual stoping of the instrumentation

## Steps to Reproduce

https://gist.github.com/rokn/8939cf0be0a70f873801f030218709a3
The gist provided is a minimal example which can demonstrate the issue.

1. Run the above gist
2. check localhost:8080/metrics `curl localhost:8080/metrics | grep db_client_connections_usage`

It has metrics for both clients.
3. Wait 10 secs

After the clients are closed the connections are 0 as expected but the metrics are still there(and the clients that provide those metrics)

## Context (Environment)

We are creating many redis connections (1000+) and are sometimes closing them and need to cleanly discard these instances from memory and the metrics they "provide"

htemelski-oss:
Hello @rokn 
Thanks for reporting the issue. Not sure how soon we would be able to prioritize this issue, but we will be grateful if you can contribute a fix.

PR #3444, opened 2025-07-23
Add closing support to otel metrics instrumentation

rokn:
Hey there, I'm opening this PR to try to address the problem from #3424.

This adds a new option when setting up the OpenTelemetry metric hooks which would allow the users to pass a close channel so they can notify us when we need to stop the metrics reporting and unregister all the needed callbacks.

The approach with a close channel through an option was chosen as to not introduce any breaking changes. Another possibility is to introduce a new function which receives a context as first param and waits for the context to close. I think both can work so it's up to you guys to decide what would be best.

I'll add some tests if you think this approach is fine

### DIFF
--- a/extra/redisotel/metrics.go
+++ b/extra/redisotel/metrics.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync"
 	"time"
 
 	"go.opentelemetry.io/otel"
@@ -12,6 +13,12 @@ import (
 
 	"github.com/redis/go-redis/v9"
 )
+
+type metricsState struct {
+	registrations []metric.Registration
+	closed        bool
+	mutex         sync.Mutex
+}
 
 // InstrumentMetrics starts reporting OpenTelemetry Metrics.
 //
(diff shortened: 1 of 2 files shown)


ndyakov:
Hello @rokn, thank you for the contribution. Will review the PR shortly. Pinging @monkey92t if he has interest to check this one as well.

extra/redisotel/metrics.go
```diff
@@ -30,49 +37,42 @@ func InstrumentMetrics(rdb redis.UniversalClient, opts ...MetricsOption) error {
...
-			if err := reportPoolStats(rdb, conf); err != nil {
-				otel.Handle(err)
+			state.mutex.Lock()
+			state.closed = true
+
+			for _, registration := range state.registrations {
+				if err := registration.Unregister(); err != nil {
+					otel.Handle(err)
+				}
 			}
```
ndyakov, line 58: Can this marking as `closed` before the `registration` is `Unregistered` result in a leak of registrations that we were thinking they would be unregistered, but leaked (errored on `Unregister()` ?

ndyakov (approved):
Overall LGTM

extra/redisotel/metrics.go
rokn: I don't think we can do much after an error on Unregister, the marking of closed is mostly so that new registrations don't come in after we have done the unregistration loop. Also I've checked and in our use case it seems that otel doesn't actually return any errors when Unregistering

PR #3445, opened 2025-07-24
fix: Prioritize TextMarshaler for argument serialization (#3139)

ostheperson:
Fixes: #3139 
* Ensures types implementing both TextMarshaler and BinaryMarshaler are marshaled as text.
* Also required reordering net.IP case to prevent string conversion due to added case.


ndyakov:
Hello @ostheperson and thank you for the contribution. This looks like it should work, but has the possibility to be a breaking change for users that are currently using with a type implemented by both TextMarshaler and BinaryMarshaler. I will label it for the next major version.

ostheperson:
understood

PR #3446, opened 2025-07-25
improve code readability

cxljs:
- replace two similar functions `appendUniqueNode` and `appendIfNotExists` with a generic function.

- simplify the implementation of the `get` method in `clusterNodes`.

- keep the member name `_generation` of `clusterNodes` consistent with other types.

- rename a data member `_masterAddr` to `masterAddr`.

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -342,8 +342,7 @@ type clusterNode struct {
 	failing    uint32 // atomic
 	loaded     uint32 // atomic
 
-	// last time the latency measurement was performed for the node, stored in nanoseconds
-	// from epoch
+	// last time the latency measurement was performed for the node, stored in nanoseconds from epoch
 	lastLatencyMeasurement int64 // atomic
 }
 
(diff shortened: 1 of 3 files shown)


ndyakov:
Hey @cxljs, thanks for your contribution.
Do you have a specific reason to prefer less linebrakes. I do think, sometimes multiline function declaration is better for readibility? 
Those function declarations and the generic function for append are the only one that I am not so happy about including, but I see the benefit of having that generic...

cxljs:
Hello @ndyakov , the reason for reducing line breaks is that there will be a comma after the last parameter, which often makes me think there are more parameters on the next line. e.g.:
```go
func masterReplicaDialer(
	failover *sentinelFailover,
) func(ctx context.Context, network, addr string) (net.Conn, error) {
```

But this is just my personal opinion, if you think multiline function declaration is better for readibility, let's keep it as it is.

ndyakov:
I do have a preference, but I am not the only person working on this :) @htemelski-redis , @ofekshenawa feel free to share your opinions.

htemelski-oss:
I am a bit on the fence about the function definitions.
I try to keep my lines shorter than 110-120 characters, and the proposed change is consistent with that.
On the other hand, for functions that accept more than a few arguments I preffer to separate them into new lines each
```go
func (c cmdable) GeoRadiusByMember(
    ctx context.Context, 
    key, member string, 
    query *GeoRadiusQuery,
) *GeoLocationCmd {
```

ndyakov:
I also do like what @htemelski-redis is suggesting.

cxljs:
I've reverted the function definitions to their original format.

PR #3447, opened 2025-07-25
[CAE-1072] Handle server maintenance notifications

ndyakov:
# Hitless Upgrades

Seamless Redis connection handoffs during cluster changes without dropping connections.

## Quick Start

```go
    Addr:     "localhost:6379",
    Protocol: 3, // RESP3 required
    HitlessUpgrades: &hitless.Config{
        Mode: hitless.MaintNotificationsEnabled,
    },
})
```

## Modes

- **`MaintNotificationsDisabled`** - Hitless upgrades disabled
- **`MaintNotificationsEnabled`** - Forcefully enabled (fails if server doesn't support)
- **`MaintNotificationsAuto`** - Auto-detect server support (default)

## Configuration

```go
&hitless.Config{
    Mode:                       hitless.MaintNotificationsAuto,
    EndpointType:               hitless.EndpointTypeAuto,
    RelaxedTimeout:             10 * time.Second,
    HandoffTimeout:             15 * time.Second,
    MaxHandoffRetries:          3,
    MaxWorkers:                 0,    // Auto-calculated
    HandoffQueueSize:           0,    // Auto-calculated
    PostHandoffRelaxedDuration: 0,    // 2 * RelaxedTimeout
    LogLevel:                   logging.LogLevelError,
}
```

### Endpoint Types

- **`EndpointTypeAuto`** - Auto-detect based on connection (default)
- **`EndpointTypeInternalIP`** - Internal IP address
- **`EndpointTypeInternalFQDN`** - Internal FQDN
- **`EndpointTypeExternalIP`** - External IP address
- **`EndpointTypeExternalFQDN`** - External FQDN
- **`EndpointTypeNone`** - No endpoint (reconnect with current config)

### Auto-Scaling

**Workers**: `min(PoolSize/2, max(10, PoolSize/3))` when auto-calculated
**Queue**: `max(20×Workers, PoolSize)` capped by `MaxActiveConns+1` or `5×PoolSize`

**Examples:**
- Pool 100: 33 workers, 660 queue (capped at 500)
- Pool 100 + MaxActiveConns 150: 33 workers, 151 queue

## How It Works

1. Redis sends push notifications about cluster changes
2. Client creates new connections to updated endpoints
3. Active operations transfer to new connections
4. Old connections close gracefully

## Supported Notifications

- `MOVING` - Slot moving to new node
- `MIGRATING` - Slot in migration state
- `MIGRATED` - Migration completed
- `FAILING_OVER` - Node failing over
- `FAILED_OVER` - Failover completed

## Hooks (Optional)

Monitor and customize hitless operations:

```go
type NotificationHook interface {
    PreHook(ctx, notificationCtx, notificationType, notification) ([]interface{}, bool)
    PostHook(ctx, notificationCtx, notificationType, notification, result)
}

// Add custom hook
manager.AddNotificationHook(&MyHook{})
```

### Metrics Hook Example

```go
// Create metrics hook
metricsHook := hitless.NewMetricsHook()
manager.AddNotificationHook(metricsHook)

// Access collected metrics
metrics := metricsHook.GetMetrics()
fmt.Printf("Notification counts: %v\n", metrics["notification_counts"])
fmt.Printf("Processing times: %v\n", metrics["processing_times"])
fmt.Printf("Error counts: %v\n", metrics["error_counts"])
```


hitless/config.go
```diff
@@ -0,0 +1,327 @@
...
+import (
+	"time"
+)
+
+// Config provides configuration options for hitless upgrades.
+type Config struct {
+	// Enabled controls whether hitless upgrades are enabled.
+	// Requires RESP3 protocol for push notifications.
+	// Default: false
+	Enabled bool
```
ndyakov: should be enabled by default?

hitless/config.go
```diff
@@ -0,0 +1,327 @@
...
+
+	// HandoffQueueSize is the size of the buffered channel used to queue handoff requests.
+	// If the queue is full, new handoff requests will be rejected.
+	// Default: 100
+	HandoffQueueSize int
+
+	// PostHandoffRelaxedDuration is how long to keep relaxed timeouts on the new connection
+	// after a handoff completes. This provides additional resilience during cluster transitions.
+	// Default: 10 seconds
+	PostHandoffRelaxedDuration time.Duration
```
ndyakov: default should be RelaxedTimeoutDuration*2 ?

hitless/config.go
```diff
@@ -0,0 +1,327 @@
...
+		ScaleDownDelay:             c.ScaleDownDelay,
+		LogLevel:                   c.LogLevel,
+	}
+}
+
+// applyWorkerDefaults calculates and applies worker defaults based on pool size
+func (c *Config) applyWorkerDefaults(poolSize int) {
+	// Calculate defaults based on pool size
+	if poolSize <= 0 {
+		poolSize = 100 // Default assumption if pool size unknown
```
ndyakov: If pool size is unknown, use the same calculation for poolsize that is currently done in the pool, i.e. `// default: 10 * runtime.GOMAXPROCS(0)`

hitless/config.go
```diff
@@ -0,0 +1,327 @@
...
+	isPrivate := isPrivateIP(addr)
+
+	var endpointType string
+
+	if tlsEnabled {
+		// TLS requires FQDN for certificate validation
+		if isPrivate {
+			endpointType = "internal-fqdn"
+		} else {
+			endpointType = "external-fqdn"
```
ndyakov: if currently fqdn is used, `isPrivate` will most likely be false

hitless/hitless_manager.go
```diff
@@ -0,0 +1,221 @@
...
+	if err := processor.RegisterHandler("MIGRATED", handler, true); err != nil {
+		return err
+	}
+	if err := processor.RegisterHandler("FAILING_OVER", handler, true); err != nil {
+		return err
+	}
+	if err := processor.RegisterHandler("FAILED_OVER", handler, true); err != nil {
+		return err
+	}
+
```
ndyakov: maybe rewrite in a loop? Try couple of different approaches and decide which works the best.

hitless/hitless_manager.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+// StartMovingOperationWithConnID starts a new MOVING operation with a specific connection ID.
+func (hm *HitlessManager) StartMovingOperationWithConnID(ctx context.Context, newEndpoint string, deadline time.Time, seqID int64, connID uint64) error {
```
ndyakov: initially this was starting the moving operation, but right now is just inserting a record for us to track the moving operation. It is only fair to actually rename the method to something like `TrackMovingOperationWithConnID` which will be more aligned with the fact that this method does actually nothing related to the actual handof.

hitless/hitless_manager.go
```diff
@@ -0,0 +1,221 @@
...
+		SeqID:  seqID,
+		ConnID: connID,
+	}
+
+	// Check for duplicate operation
+	if _, exists := hm.activeMovingOps[key]; exists {
+		// Duplicate MOVING notification, ignore
+		internal.Logger.Printf(ctx, "Duplicate MOVING operation ignored: %s", key.String())
+		return nil
+	}
```
ndyakov: add more comments why this is ignored and doesn't actually return an error. this should not happen in normal operation, i think...

hitless/hitless_manager.go
ndyakov: or `RecordMovingOperationWithConnID`

hitless/hitless_manager.go
```diff
@@ -0,0 +1,221 @@
...
+	// Create composite key
+	key := MovingOperationKey{
+		SeqID:  seqID,
+		ConnID: connID,
+	}
+
+	// Remove from active operations
+	delete(hm.activeMovingOps, key)
+
+	// Note: Atomic state updates removed (per-connection handoffs)
```
ndyakov: ✂️

hitless/hitless_manager.go
```diff
@@ -0,0 +1,221 @@
...
+	for key, op := range hm.activeMovingOps {
+		result[key] = &MovingOperation{
+			SeqID:       op.SeqID,
+			NewEndpoint: op.NewEndpoint,
+			StartTime:   op.StartTime,
+			Deadline:    op.Deadline,
+		}
+	}
+	return result
+}
```
ndyakov: this is not used and I would vote to remove it if it won't be needed in the final implementation. I can see how it can be beneficial for observability but we can find think about another way of logging without locking and copying

hitless/hitless_manager.go
```diff
@@ -0,0 +1,221 @@
...
+// GetCurrentState returns current state based on active operations.
+func (hm *HitlessManager) GetCurrentState() State {
+	hm.mu.RLock()
+	defer hm.mu.RUnlock()
+
+	if len(hm.activeMovingOps) > 0 {
+		return StateMoving
+	}
+	return StateIdle
+}
```
ndyakov: atomic flag instead of lock?

hitless/hitless_manager.go
```diff
@@ -0,0 +1,221 @@
...
+
+// IsHandoffInProgress returns true if any handoff is in progress.
+func (hm *HitlessManager) IsHandoffInProgress() bool {
+	hm.mu.RLock()
+	defer hm.mu.RUnlock()
+	return len(hm.activeMovingOps) > 0
+}
```
ndyakov: atomic flag.

adapters.go
```diff
@@ -0,0 +1,144 @@
...
+func (oa *optionsAdapter) GetPoolSize() int {
+	return oa.options.PoolSize
+}
+
+// NewDialer returns a new dialer function for the connection.
+func (oa *optionsAdapter) NewDialer() func(context.Context) (net.Conn, error) {
+	baseDialer := oa.options.NewDialer()
+	return func(ctx context.Context) (net.Conn, error) {
+		// Extract network and address from the options
+		network := "tcp"
```
ndyakov: extract the network from the options as well

options.go
```diff
@@ -650,16 +648,17 @@ func newConnPool(
...
-		PoolTimeout:     opt.PoolTimeout,
-		DialTimeout:     opt.DialTimeout,
-		MinIdleConns:    opt.MinIdleConns,
-		MaxIdleConns:    opt.MaxIdleConns,
-		MaxActiveConns:  opt.MaxActiveConns,
-		ConnMaxIdleTime: opt.ConnMaxIdleTime,
-		ConnMaxLifetime: opt.ConnMaxLifetime,
-		ReadBufferSize:  opt.ReadBufferSize,
-		WriteBufferSize: opt.WriteBufferSize,
+		PoolFIFO:                 opt.PoolFIFO || opt.HitlessUpgradeConfig.IsEnabled(), // Always FIFO with hitless upgrades
```
ndyakov: This can be changed, no need to be always FIFO

redis.go
```diff
@@ -451,16 +448,25 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
...
+		/*
+			pipe := conn.Pipeline()
+			hitlessHandshakeErr = pipe.ClientMaintNotifications(
+				ctx,
+				true,
+				endpointType.String(),
+			).Err()
+			pipe.Exec(ctx)
+
+		*/
```
ndyakov: enable hitless handshake

redis.go
```diff
@@ -451,16 +448,25 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
...
 		if hitlessHandshakeErr != nil {
+			if !isRedisError(hitlessHandshakeErr) {
+				// if not redis error, fail the connection
+				return hitlessHandshakeErr
+			}
```
ndyakov, line 450: network error, fine to fail

redis_test.go
```diff
@@ -627,6 +627,10 @@ var _ = Describe("Hook with MinIdleConns", func() {
 	BeforeEach(func() {
 		options := redisOptions()
 		options.MinIdleConns = 1
+		options.HitlessUpgradeConfig = &redis.HitlessUpgradeConfig{
+			Enabled: hitless.MaintNotificationsDisabled,
+		}
+		options.HitlessUpgradeConfig = nil
```
ndyakov: remove both

redis_test.go
```diff
@@ -652,6 +656,7 @@ var _ = Describe("Hook with MinIdleConns", func() {
 			processHook: func(hook redis.ProcessHook) redis.ProcessHook {
 				return func(ctx context.Context, cmd redis.Cmder) error {
 					res = append(res, "hook-2-process-start")
+					res = append(res, cmd.String())
```
ndyakov: remove

hitless/notification_handler.go
```diff
@@ -0,0 +1,244 @@
...
+	// If newEndpoint is empty, we should schedule a handoff to the current endpoint in timeS/2 seconds
+	if newEndpoint == "" || newEndpoint == internal.RedisNull {
+		// same as current endpoint
+		newEndpoint = snh.manager.options.GetAddr()
+		// delay the handoff for timeS/2 seconds to the same endpoint
+		// do this in a goroutine to avoid blocking the notification handler
+		go func() {
+			time.Sleep(time.Duration(timeS/2) * time.Second)
+			if err := snh.markConnForHandoff(poolConn, newEndpoint, seqID, deadline); err != nil {
+				// Log error but don't fail the goroutine
```
ndyakov: check conn existence before that, make sure it is not closed

hitless/notification_handler.go
```diff
@@ -0,0 +1,244 @@
...
+
+	deadline := time.Now().Add(time.Duration(timeS) * time.Second)
+	// If newEndpoint is empty, we should schedule a handoff to the current endpoint in timeS/2 seconds
+	if newEndpoint == "" || newEndpoint == internal.RedisNull {
+		// same as current endpoint
+		newEndpoint = snh.manager.options.GetAddr()
+		// delay the handoff for timeS/2 seconds to the same endpoint
+		// do this in a goroutine to avoid blocking the notification handler
+		go func() {
+			time.Sleep(time.Duration(timeS/2) * time.Second)
```
ndyakov: maybe we can use "last used at" for better calculation here? It won't be correct to determine when the connection was last used, but we can try.

Makefile
```diff
@@ -29,7 +29,7 @@ test.ci:
 	  (cd "$${dir}" && \
 	    go mod tidy -compat=1.18 && \
 	    go vet && \
-	    go test -v -coverprofile=coverage.txt -covermode=atomic ./... -race -skip Example); \
+	    go test -v -coverprofile=coverage.txt -covermode=atomic ./... -race -timeout 20m -skip Example); \
```
ndyakov: can be returned to 10m (by default)

hitless/config.go
```diff
@@ -0,0 +1,377 @@
...
+	// Default: 15 seconds (matches server-side eviction timeout)
+	HandoffTimeout time.Duration
+
+	// MaxWorkers is the maximum number of worker goroutines for processing handoff requests.
+	// Workers are created on-demand and automatically cleaned up when idle.
+	// If zero, defaults to min(10, PoolSize/3) to handle bursts effectively.
+	// If explicitly set, enforces minimum of 10 workers.
+	//
+	// Default: min(10, PoolSize/3), Minimum when set: 10
+	MaxWorkers int
```
ndyakov: maybe use GOMAXPROCS ? , default can be PoolSize/10?

internal/pool/hooks_test.go
```diff
@@ -0,0 +1,215 @@
...
+		t.Errorf("Expected removeHook.OnPutCalled to be 1, got %d", removeHook.OnPutCalled)
+	}
+	
+	// normalHook should not be called due to early return
+	if normalHook.OnPutCalled != 0 {
+		t.Errorf("Expected normalHook.OnPutCalled to be 0, got %d", normalHook.OnPutCalled)
+	}
+}
+
+
```
ndyakov: remove

internal/pool/pool_sticky.go
```diff
@@ -199,3 +199,11 @@ func (p *StickyConnPool) IdleLen() int {
...
 	return &Stats{}
 }
+
+func (p *StickyConnPool) AddPoolHook(hook PoolHook) {}
+
+func (p *StickyConnPool) RemovePoolHook(hook PoolHook) {}
+
+func (p *StickyConnPool) TrackConn(cn *Conn) {}
+
+func (p *StickyConnPool) UntrackConn(cn *Conn) {}
```
ndyakov: Are those needed now?

internal/pool/pool.go
```diff
@@ -57,29 +67,30 @@ type Pooler interface {
 	IdleLen() int
 	Stats() *Stats
 
+	AddPoolHook(hook PoolHook)
+	RemovePoolHook(hook PoolHook)
+
+	TrackConn(*Conn)
+	UntrackConn(*Conn)
+
```
ndyakov: now we use entirely new pool struct for pub sub, those can be dropped

hitless/notification_handler.go
ndyakov: last used at is not reliable for this calculation

internal/util/convert.go
```diff
@@ -28,3 +28,14 @@ func MustParseFloat(s string) float64 {
...
+// SafeIntToInt32 safely converts an int to int32, returning an error if overflow would occur.
+func SafeIntToInt32(value int, fieldName string) (int32, error) {
+	if value > math.MaxInt32 {
+		return 0, fmt.Errorf("redis: %s value %d exceeds maximum allowed value %d", fieldName, value, math.MaxInt32)
+	}
+	if value < math.MinInt32 {
+		return 0, fmt.Errorf("redis: %s value %d is below minimum allowed value %d", fieldName, value, math.MinInt32)
+	}
+	return int32(value), nil
+}
```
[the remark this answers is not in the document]
ndyakov: they are imported.

internal/pool/conn.go
```diff
@@ -64,23 +124,368 @@ func (cn *Conn) SetUsedAt(tm time.Time) {
...
+func (cn *Conn) SetInitConnFunc(fn func(context.Context, *Conn) error) {
+	cn.initConnFunc = fn
+}
+
+// ExecuteInitConn runs the stored connection initialization function if available.
+func (cn *Conn) ExecuteInitConn(ctx context.Context) error {
+	if cn.initConnFunc != nil {
+		return cn.initConnFunc(ctx, cn)
+	}
+	return fmt.Errorf("redis: no initConnFunc set for connection %d", cn.GetID())
```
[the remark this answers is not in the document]
ndyakov: invalid

push_notifications.go
```diff
@@ -4,24 +4,6 @@ import (
...
-	// FAILING_OVER indicates a failover is starting
-	PushNotificationFailingOver = "FAILING_OVER"
-
-	// FAILED_OVER indicates a failover has completed
-	PushNotificationFailedOver = "FAILED_OVER"
-)
-
 // NewPushNotificationProcessor creates a new push notification processor
 // This processor maintains a registry of handlers and processes push notifications
 // It is used for RESP3 connections where push notifications are available
```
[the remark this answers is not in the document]
ndyakov: invalid

hitless/notification_handler.go
```diff
@@ -0,0 +1,247 @@
...
+	return err
+}
+
+// handleMoving processes MOVING notifications.
+// ["MOVING", seqNum, timeS, endpoint] - per-connection handoff
+func (snh *NotificationHandler) handleMoving(ctx context.Context, handlerCtx push.NotificationHandlerContext, notification []interface{}) error {
+	if len(notification) < 3 {
+		return ErrInvalidNotification
+	}
+	seqIDStr, ok := notification[1].(string)
```
ofekshenawa: Why assumes sequence IDs are always strings? 
The server sends a push notification like this:
```
>4\r\n
+MOVING\r\n
:<seq_number>\r\n
:<time_s>\r\n
+<new_name_or_ip>:<port>\r\n
```

and we parsed it using readReply for each row (which means <seq_number> is int)

hitless/notification_handler.go
```diff
@@ -0,0 +1,247 @@
...
+		return ErrInvalidNotification
+	}
+
+	seqID, err := strconv.ParseInt(seqIDStr, 10, 64)
+	if err != nil {
+		return ErrInvalidNotification
+	}
+
+	// Extract timeS
+	timeSStr, ok := notification[2].(string)
```
ofekshenawa: Same here

hitless/notification_handler.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	deadline := time.Now().Add(time.Duration(timeS) * time.Second)
+	// If newEndpoint is empty, we should schedule a handoff to the current endpoint in timeS/2 seconds
+	if newEndpoint == "" || newEndpoint == internal.RedisNull {
+		// same as current endpoint
+		newEndpoint = snh.manager.options.GetAddr()
+		// delay the handoff for timeS/2 seconds to the same endpoint
+		// do this in a goroutine to avoid blocking the notification handler
+		go func() {
```
ofekshenawa: For every MOVING notification with an empty endpoint, the code starts a goroutine that does nothing but sleep for ~15 seconds. When notifications are frequent, these sleeping goroutines accumulate, unnecessarily consuming memory and overwhelming the Go scheduler.
Maybe replace `go func() + time.Sleep()` with `time.AfterFunc()` to use lightweight timers instead of goroutines?

hitless/pool_hook.go
```diff
@@ -0,0 +1,477 @@
...
+				conn.ClearHandoffState()
+			} else {
+				if err := ph.queueHandoff(conn); err != nil {
+					// Failed to queue handoff, remove the connection
+					internal.Logger.Printf(ctx, "Failed to queue handoff: %v", err)
+					return false, true, nil // Don't pool, remove connection, no error to caller
+				}
+
+				// Check if handoff was already processed by a worker before we can mark it as queued
+				if !conn.ShouldHandoff() {
```
ofekshenawa: There’s a TOCTOU race in 155–170: OnPut() checks conn.ShouldHandoff(), but a worker can finish the handoff and call conn.ClearHandoffState() before OnPut() marks it, causing conn.MarkQueuedForHandoff() to fail. 
Maybe for safety replace the check+mark sequence with a single atomic TryMarkQueuedForHandoff() using compare and swap to close the race and avoid connection loss under load.

ndyakov: MarkQueuedForHandoff does compare and swap, if it fails it will:
 check once again :
 ```
 					if !conn.ShouldHandoff() {
						// Handoff was processed - this is normal, pool the connection
						return true, false, nil
					}
```
and return this connection as a healthy one back to the pool (pool it, don't remove, no errors)
or if we should still handoff but was not able to mark it will remove the conn

Can you explain what is the case that is concerning to you, maybe I didn't get it?

File: hitless/notification_handler.go (hunk 30 above)
Comment (ndyakov) [reply hitless/notification_handler.go]: Great suggestion.

File: hitless/pool_hook.go (hunk 31 above)
Comment (ofekshenawa) [reply hitless/pool_hook.go]: You are correct

File: hitless/notification_handler.go (hunk 28 above)
Comment (ndyakov) [reply hitless/notification_handler.go]: You are correct, it should be safe to get them as ints here. Thank you.

File: hitless/pool_hook.go [hunk 32]
```diff
@@ -0,0 +1,477 @@
...
+		case <-shutdownCtx.Done():
+		}
+	}()
+
+	// Perform the handoff with cancellable context
+	err := ph.performConnectionHandoffWithPool(shutdownCtx, request.Conn, request.Pool)
+
+	// If handoff failed, restore the handoff state for potential retry
+	if err != nil {
+		request.Conn.RestoreHandoffState()
```
ofekshenawa: After a failed handoff we only restore the handoff flag and log “WILL RETRY,” but we don’t actually re-enqueue the connection. Since OnPut is the only place that queues and OnGet rejects ShouldHandoff conns, an idle conn can become stranded and reduce effective pool capacity.

hitless/pool_hook.go
```diff
@@ -0,0 +1,477 @@
...
+		if err != nil {
+			// If no port specified, assume default Redis port
+			host = endpoint
+			if port == "" {
+				port = "6379"
+			}
+		}
+
+		// Use the base dialer to connect to the new endpoint
+		return ph.baseDialer(ctx, ph.network, net.JoinHostPort(host, port))
```
ofekshenawa: When MOVING points to a different host/FQDN or an IP, we reuse the original TLS config (including ServerName) while dialing the new endpoint. That can cause hostname verification mismatches (e.g., cert valid for old host, not new host/IP).

ndyakov: Correct. Since there are no certificates transmitted as part of the notification, I am assuming there would be a root level configuration for tls which should cover the new endpoint.

hitless/pool_hook.go
ndyakov: Great observation! Thank you! I actually refactored the routine here and think the RestoreHandoffState won't be needed (and as you observed, won't work correctly).

hitless/notification_handler.go
ndyakov: should be resolved

hitless/pool_hook.go
ndyakov: Resolved, thank you.

hitless/notification_handler.go
ndyakov: Resolved.

hitless/notification_handler.go
ndyakov: Resolved

hitless/config.go
htemelski-oss: Until go 1.25, GOMAXPROCS wasn't container aware, which could lead to throttling
https://go.dev/blog/container-aware-gomaxprocs 
I suggest to use some sensible default and change the behaviour only when we bump the go.mod to a version >= 1.25

hitless/config.go
```diff
@@ -0,0 +1,422 @@
...
+
+	defaults := DefaultConfig()
+	result := &Config{}
+
+	// Apply defaults for enum fields (empty/zero means not set)
+	if c.Mode == "" {
+		result.Mode = defaults.Mode
+	} else {
+		result.Mode = c.Mode
+	}
```
htemelski-oss: Nitpick: Can we simplify these if-else blocks
```
result.Mode = defaults.Mode
if c.Mode != "" {
    result.Mode = c.Mode
}
```

hitless/config.go
```diff
@@ -0,0 +1,422 @@
...
+		// Configuration fields
+		MaxHandoffRetries: c.MaxHandoffRetries,
+	}
+}
+
+// applyWorkerDefaults calculates and applies worker defaults based on pool size
+func (c *Config) applyWorkerDefaults(poolSize int) {
+	// Calculate defaults based on pool size
+	if poolSize <= 0 {
+		poolSize = 10 * runtime.GOMAXPROCS(0)
```
htemelski-oss, line 365: As noted above, `GOMAXPROCS` can lead to throttling

hitless/config.go
ndyakov: Yeah, I saw that and I agree with you.

pubsub.go
```diff
@@ -73,10 +76,18 @@ func (c *PubSub) conn(ctx context.Context, newChannels []string) (*pool.Conn, er
 		return c.cn, nil
 	}
 
+	if c.opt.Addr == "" {
+		// TODO(hitless):
```
htemelski-oss, line 80: This should remain as a TODO, right?

hitless/hitless_manager.go
```diff
@@ -0,0 +1,310 @@
...
+		SeqID:       seqID,
+		NewEndpoint: newEndpoint,
+		StartTime:   time.Now(),
+		Deadline:    deadline,
+	}
+
+	// Use LoadOrStore for atomic check-and-set operation
+	if _, loaded := hm.activeMovingOps.LoadOrStore(key, movingOp); loaded {
+		// Duplicate MOVING notification, ignore
+		if hm.config.LogLevel >= 3 { // Warning level
```
htemelski-oss: Can we define the log levels as constants?

htemelski-redis, line 151: Can we define the log levels as constants?

htemelski-redis (commented):
Couldn't see any glaring issues

hitless/config.go
ndyakov: this is the default of the pool at the moment, keeping it here as well. I agree that for containerized this is not correct with older go version.

hitless/hitless_manager.go
ndyakov: Makes sense, will do.

pubsub.go
ndyakov: yes

htemelski-redis (approved):

PR #3449, opened 2025-07-28
Add VAMANA vector type to RediSearch

ofekshenawa:
Implements comprehensive support for the VAMANA vector index algorithm introduced in Redis 8.2+.

### Key Changes
-  **Core Implementation** : Added FTVamanaOptions struct with full parameter support (distance metrics, vector types, compression, advanced tuning)

- **Feature Set** : Supports L2/COSINE/IP distance metrics, FLOAT16/FLOAT32 vector types, LVQ8 compression, construction window size and graph max degree

- **Tests**: Includes test suite covering VAMANA functionality, distance metrics, compression scenarios, and parameter combinations


search_commands.go
```diff
@@ -1035,6 +1060,40 @@ func (c cmdable) FTCreate(ctx context.Context, index string, options *FTCreateOp
 				args = append(args, len(hnswArgs))
 				args = append(args, hnswArgs...)
 			}
+			if schema.VectorArgs.VamanaOptions != nil {
+				args = append(args, "VAMANA")
+				if schema.VectorArgs.VamanaOptions.Type == "" || schema.VectorArgs.VamanaOptions.Dim == 0 || schema.VectorArgs.VamanaOptions.DistanceMetric == "" {
+					panic("FT.CREATE: Type, Dim and DistanceMetric are required for VECTOR VAMANA")
```
ndyakov: Let's try not to panic in the code if we can simply return an error.

search_commands.go
```diff
@@ -985,8 +999,19 @@ func (c cmdable) FTCreate(ctx context.Context, index string, options *FTCreateOp
...
+				optionCount++
+			}
+			if schema.VectorArgs.HNSWOptions != nil {
+				optionCount++
+			}
+			if schema.VectorArgs.VamanaOptions != nil {
+				optionCount++
+			}
+			if optionCount != 1 {
+				panic("FT.CREATE: SCHEMA VectorArgs must have exactly one of FlatOptions, HNSWOptions, or VamanaOptions")
```
ndyakov: Let's try not to panic in the code if we can simply return an error.

ndyakov (changes requested):
@ofekshenawa thank you for contributing this. I can see there are panics that were in the code even before we added this change, but let's make sure we don't panic during the execution of a command. Maybe the arguments were dynamically build, and returning an error should be sufficient way of noting that they are wrong, panicking will stop the whole execution of the application (if not recovered) and we should not do that.

PR #3450, opened 2025-07-28
Added new stream commands

htemelski-redis:
Added support for the new XACKDEL and XDELEX commands.
Extended the existing XADD and XTRIM coomands.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@
 
 services:
   redis:
-    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:rs-7.4.0-v2}
+    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:8.2-RC1-pre}
     platform: linux/amd64
     container_name: redis-standalone
     environment:
@@ -23,7 +23,7 @@ services:
       - all
 
   osscluster:
-    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:rs-7.4.0-v2}
+    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:8.2-RC1-pre}
     platform: linux/amd64
     container_name: redis-osscluster
     environment:
@@ -40,7 +40,7 @@ services:
       - all
 
   sentinel-cluster:
-    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:rs-7.4.0-v2}
+    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:8.2-RC1-pre}
     platform: linux/amd64
     container_name: redis-sentinel-cluster
     network_mode: "host"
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 
 services:
   redis:
-    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:rs-7.4.0-v2}
+    image: ${CLIENT_LIBS_TEST_IMAGE:-redislabs/client-libs-test:8.2-RC1-pre}
     platform: linux/amd64
```
ndyakov: This is supposed to be the default for testing and I would assume it is the longest stable version (i.e. the oldest one that is supported), not the latest.

stream_commands.go
```diff
@@ -93,6 +106,16 @@ func (c cmdable) XAdd(ctx context.Context, a *XAddArgs) *StringCmd {
 	return cmd
 }
 
+func (c cmdable) XAckDel(ctx context.Context, stream string, group string, mode string, ids ...string) *SliceCmd {
+	args := []interface{}{"xackdel", stream, group, mode, "ids", strconv.Itoa(len(ids))}
```
ndyakov: no need to convert the int to string here, it is an `[]interface` and it should work. It will be handled by https://github.com/redis/go-redis/blob/37accb4b2807e1bec0d90079fd5e1506882cf6fe/internal/proto/writer.go#L62

stream_commands.go
```diff
@@ -103,6 +126,16 @@ func (c cmdable) XDel(ctx context.Context, stream string, ids ...string) *IntCmd
 	return cmd
 }
 
+func (c cmdable) XDelEx(ctx context.Context, stream string, mode string, ids ...string) *SliceCmd {
+	args := []interface{}{"xdelex", stream, mode, "ids", strconv.Itoa(len(ids))}
```
ndyakov: Same as above, no need to use `strconv.Itoa` here.

stream_commands.go
```diff
@@ -418,6 +451,42 @@ func (c cmdable) XTrimMinIDApprox(ctx context.Context, key string, minID string,
...
+	return c.xTrimMode(ctx, key, "maxlen", true, maxLen, limit, mode)
+}
+
+func (c cmdable) XTrimMinIDMode(ctx context.Context, key string, minID string, mode string) *IntCmd {
+	return c.xTrimMode(ctx, key, "minid", false, minID, 0, mode)
+}
+
+func (c cmdable) XTrimMinIDApproxMode(ctx context.Context, key string, minID string, limit int64, mode string) *IntCmd {
+	return c.xTrimMode(ctx, key, "minid", true, minID, limit, mode)
+}
```
ndyakov, line 489: Please add a comment to refactor this and expose similar api to the rest of the commands, i.e. `XTrimWithArgs`

ndyakov: Also can add it to our tech debt document to be addressed as part of the next major release.

docker-compose.yml
htemelski-oss: It seems that I've commited the file by mistake, will revert it

PR #3453, opened 2025-07-30
add configurable buffer sizes for Redis connections

### LINKED ISSUE #3438: Make pool.Conn reader/writer buffer size configurable
Hey team 👋 

We were recently looking at performance of a system that uses a lot of pipelines of moderate size (~100-200 commands) and noticed how based on the pipeline size its execution would take multiples of our normal latency (for example if normally a command has 5ms latency sometimes these pipelines would take 5/10/15/20ms).

I'm pretty sure this is happening because the `bufio.Writer` [`pool.Conn` uses](https://github.com/redis/go-redis/blob/0decfdc6ed1710b0abf5b41741bd02223c877c92/internal/pool/conn.go?plain=1#L36) is created using the default size of 4kb

https://github.com/redis/go-redis/blob/0decfdc6ed1710b0abf5b41741bd02223c877c92/internal/pool/conn.go?plain=1#L36

```golang
// from https://github.com/golang/go/blob/master/src/bufio/bufio.go#L18-L20

const (
  defaultBufSize = 4096
)

// ...

// from https://github.com/golang/go/blob/master/src/bufio/bufio.go#L607-L612

// NewWriter returns a new [Writer] whose buffer has the default size.
// If the argument io.Writer is already a [Writer] with large enough buffer size,
// it returns the underlying [Writer].
func NewWriter(w io.Writer) *Writer {
  return NewWriterSize(w, defaultBufSize)
}
```

When looking for what other defaults look like I found [redis/rueidis](https://github.com/redis/rueidis/tree/af266d4b3f8d82c6cbce3a7d80ebe8401295cf8d) defaults both read and write buffers to 0.5Mib:
[[permalink]](https://github.com/redis/rueidis/blob/af266d4b3f8d82c6cbce3a7d80ebe8401295cf8d/rueidis.go#L161-L164)

```golang
// ReadBufferEachConn is the size of the bufio.NewReaderSize for each connection, default to DefaultReadBuffer (0.5 MiB).
ReadBufferEachConn int
// WriteBufferEachConn is the size of the bufio.NewWriterSize for each connection, default to DefaultWriteBuffer (0.5 MiB).
WriteBufferEachConn int
```

I think having the ability to at least configure those sizes would be a big performance boost for users issuing many and/or "big" commands (scripts or long lists or args). What do you think? Also I'd be more than happy to contribute these changes, just wanted to see if this is something you agree with first. Thanks! ❤️

ofekshenawa:
- Set default buffer sizes to 0.5MiB (up from 4KB) for better performance
- Add ReadBufferSize and WriteBufferSize options to all client types

Fixes #3438


internal/pool/conn.go
```diff
@@ -28,12 +28,28 @@ type Conn struct {
 }
 
 func NewConn(netConn net.Conn) *Conn {
+	return NewConnWithBufferSize(netConn, proto.DefaultBufferSize, proto.DefaultBufferSize)
+}
```
ndyakov, line 32: This function is used in https://github.com/redis/go-redis/blob/a39a23a57b5b3c49bac660ee1cb497169961c90f/redis.go#L254 . Lets change it to use the buffer size from the options (if set).

options.go
```diff
@@ -130,6 +131,18 @@ type Options struct {
...
+	// ReadBufferSize is the size of the bufio.Reader buffer for each connection.
+	// Larger buffers can improve performance for commands that return large responses.
+	//
+	// default: 0.5MiB (524288 bytes)
+	ReadBufferSize int
+
+	// WriteBufferSize is the size of the bufio.Writer buffer for each connection.
+	// Larger buffers can improve performance for large pipelines and commands with many arguments.
+	//
+	// default: 0.5MiB (524288 bytes)
```
ndyakov: maybe we can add something like "Smaller buffers can improve memory usage for larger pools. "

internal/pool/conn.go
ndyakov: I was mistaken, this is `NewConn` on the pool object, which will use the correct buffer sizes.

ndyakov (approved):
Looks good, thank you!

PR #3463, opened 2025-08-05
Merges stale issues jobs into one job with two steps

elena-kolevska:


ndyakov (approved):
Looks good, will label  and merge this.

PR #3468, opened 2025-08-11
fix(options): Add buffer sizes to failover. Update README

### LINKED ISSUE #3465: Potential memory leaks on 9.12.0

## Expected Behavior

I would not expect this steep increase of memory consumption.

## Current Behavior

Soon after it got deployed, ECS memory consumption got increased from 2% to 10%.
A Lamba function with 128mb of memory also failed several invocations.
Might not have enough data points since service deployment has been reverted, but looks like it's also gradually increasing.

Image

## Possible Solution

Don't have one.

## Steps to Reproduce

Deployed our service which handles high throughput of event streams.

## Context (Environment)

go 1.24.5
ECS: ARM64, debian:12-slim
Lambda: x86_64, provided.al2023

Event stream Lambda handlers got behind, backlogs built, ECS memory spike or leak didn't cause a problem since it was reverted.

ndyakov:
- add missing buffer size to failover options
- half the default buffer size
- add section in readme related to buffer sizes

Fixes #3465


sentinel.go
```diff
@@ -90,6 +90,20 @@ type FailoverOptions struct {
 	WriteTimeout          time.Duration
 	ContextTimeoutEnabled bool
 
+	// ReadBufferSize is the size of the bufio.Reader buffer for each connection.
+	// Larger buffers can improve performance for commands that return large responses.
+	// Smaller buffers can improve memory usage for larger pools.
+	//
+	// default: 0.5MiB (524288 bytes)
```
htemelski-oss: Shouldn't this be 256KiB too?

sentinel.go
```diff
@@ -90,6 +90,20 @@ type FailoverOptions struct {
...
+	ReadBufferSize int
+
+	// WriteBufferSize is the size of the bufio.Writer buffer for each connection.
+	// Larger buffers can improve performance for large pipelines and commands with many arguments.
+	// Smaller buffers can improve memory usage for larger pools.
+	//
+	// default: 0.5MiB (524288 bytes)
```
htemelski-oss: Same as above

ndyakov: this is wrong

sentinel.go
ndyakov: this is wrong, should be 256kib

htemelski-redis (approved):
Looks, good there are two minor things to change but they are already noted

ofekshenawa (approved):
LGTM

cxljs:
Hello @ndyakov , Here's a small suggestion: How about keeping the read/write buffer size of the sentinel client at 4KB? Since there's no need for the sentinel client to handle large data transfers or pipeline scenarios.

ndyakov:
@cxljs great idea! Feel free to work on that or open a separate issue so we can pick it up when possible.

cxljs:
OK, I'll submit a PR later.

PR #3471, opened 2025-08-11
Removes dry run for stale issues policy

elena-kolevska:


ndyakov:
Let's merge after the release, so we don't need to update the release PR now.

PR #3476, opened 2025-08-14
Set the read/write buffer size of the sentinel client to 4KiB

cxljs:
Set the read/write buffer size of the sentinel client to 4KiB since there's no need for the sentinel client to handle large data transfers or pipeline scenarios.


ndyakov (approved):
Thank you. I don't think this needs to be addressed in the README, since it is a client that most users won't use directly.

PR #3478, opened 2025-08-15
Bumps test image to 8.2.1-pre

elena-kolevska:


htemelski-oss:
Could you update the image in the `workflows/build.yml` file too?

PR #3480, opened 2025-08-16
Support subscriptions against cluster slave nodes

LINKIWI:
Currently, `SUBSCRIBE` (and its sibling variant commands) establish connections only to masters, regardless of the `ReadOnly` option specified in client options.

This PR allows go-redis to respect the `ReadOnly` option for subscriptions in the pubsub client, by preferring slave nodes for new connections. Note that in both global and sharded pubsub modes, messages are replicated to slaves by Redis.

Note that `newConn` in the pubsub client is only used for creating connections for long-lived subscriptions, so it does not need to consider any per-command R/W-family properties as in the case for commands outside of the pubsub client.

I added a new unit test case for both global and sharded subscriptions with `ReadOnly` set to true, with an additional assertion on the value of `pubsub{shard}_channels` as returned by `INFO` to additionally verify that the subscription is served only by slave nodes.

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -1821,14 +1821,33 @@ func (c *ClusterClient) pubSub() *PubSub {
 			}
 
 			var err error
+
 			if len(channels) > 0 {
 				slot := hashtag.Slot(channels[0])
-				node, err = c.slotMasterNode(ctx, slot)
+
+				// newConn in PubSub is only used for subscription connections, so it is safe to
+				// assume that a slave node can always be used when client options specify ReadOnly.
+				if c.opt.ReadOnly {
+					state, err := c.state.Get(ctx)
+					if err != nil {
+						return nil, err
+					}
+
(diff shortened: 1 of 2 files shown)


ndyakov:
Hello @LINKIWI, this is a nice addition, I am working on something that touches the pubsub at the moment and would like to hold merging your PR once I am done. Do you have a need for this / is is urgent to merge?

LINKIWI:
> Hello @LINKIWI, this is a nice addition, I am working on something that touches the pubsub at the moment and would like to hold merging your PR once I am done. Do you have a need for this / is is urgent to merge? 

Thanks for the review. No worries, it is not urgent; happy to hold for now and rebase on top of the in-progress refactor once that is in. Thanks!

ndyakov (approved):
This looks good and instead of waiting for my work to be merged, I would prefer to include it in a version prior to releasing what I am working on (since it will be in a beta). 

Thank you @LINKIWI !

Issue #3479, opened 2025-08-15
Feature Request: Filter out specific commands for OTEL Tracing


cruzer200:
Hello everyone,
it would be nice if there was an option to exclude specific commands from being covered by traces.
The idea is that you can specify a slice of commands which you don't want traces to be created for, and can pass it as options when configuring the tracer.

If that's a feasible idea, I could make my hands dirty and start right away!

htemelski-oss:
Hello @cruzer200,
Thanks for raising this issue. Currently, we are in the process of rethinking and realigning the metrics and tracing across all of our client libraries.
We will take your feedback into consideration!

lucawol:
Hello @htemelski-redis  thank you for considering my proposal. I just opened a PR just in case with a first proposal.

Udhayarajan:
#3519 solved filter for only `ProcessHook` what about expanding support to `ProcessPipelineHook` and ability to disable dial hook, do you guys think will it be useful?

ndyakov:
@Udhayarajan definitely. Would you like to work on that ? If so, feel free to add the filters to all of the hooks.

zadil:
## Technical Proposal: Unified and Optimized Trace Filtering API

After reviewing the evolution from #3481 → #3519 → #3550, I'd like to propose a more comprehensive approach to the trace filtering API:

### Current State Analysis

**PR #3481**: Initial approach with `commandExclusions []string` using `slices.Contains` - O(n) lookup per command
**PR #3519**: Evolved to flexible filter function `WithCommandFilter(func(cmd redis.Cmder) bool)` - only for `ProcessHook`  
**PR #3550**: Extended filtering to `ProcessPipelineHook` and `DialHook` - maintains same pattern

### Identified Issues

1. **Performance**: Using slice lookups (`slices.Contains`) for command exclusions is O(n). For production systems excluding multiple commands (PING, INFO, SELECT, etc.), this creates unnecessary overhead on every traced operation.

2. **API Inconsistency**: The filter is applied inconsistently:
   - `ProcessHook`: filters individual commands
   - `ProcessPipelineHook`: filters pipelines (but doesn't filter individual commands within the pipeline)
   - `DialHook`: binary on/off via filter

3. **Limited Flexibility**: The current function signature `func(cmd redis.Cmder) bool` works for `ProcessHook` but doesn't generalize well to other hook types.

### Proposed Improvements

#### 1. Use Set-based Exclusion for Performance

[...]

PR #3481, opened 2025-08-17
First Proposal to filter out specific commands that shouldn't be traced via OTEL

lucawol:


lucawol:
This PR would resolve #3479

htemelski-oss:
Hello @lucawol,
We are going to take into account your proposal and this PR for our metrics stack

ndyakov:
Please see #3519 , which was an alternative that was just merged.

PR #3482, opened 2025-08-17
Allow users to enable read-write splitting in failover mode.

cxljs:
Currently, when users use failover mode and want to implement read-write splitting, a close alternative is to use NewFailoverClusterClient with RouteRandom=true, but this is not the best solution.

Two simple solutions:
1. Reuse FailoverOption.ReplicaOnly:
   When using NewFailoverClient(), ReplicaOnly=true means allowing all commands to be routed to replicas.
   When using NewFailoverClusterClient(), ReplicaOnly behaves the same as ClusterOption.ReadOnly.
2. Add a new configuration entry called ReadOnly.

@ndyakov @htemelski-redis @elena-kolevska Which solution do you think is better, or do you have another alternative?

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -25,7 +25,7 @@ runs:
         
         # Mapping of redis version to redis testing containers
         declare -A redis_version_mapping=(
-          ["8.2.x"]="8.2"
+          ["8.2.x"]="8.2.1-pre"
           ["8.0.x"]="8.0.2"
           ["7.4.x"]="rs-7.4.0-v5"
           ["7.2.x"]="rs-7.2.0-v17"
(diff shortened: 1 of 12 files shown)


ndyakov:
@cxljs I am not sure if `ReadOnly` will work correctly with failover cluster. Can we add couple of tests for it?

cxljs:
From the code, ReadOnly seems to work fine. I'll add tests later.

ndyakov:
@cxljs reminder for tests here, if we can merge this today we can release tomorrow / on monday

cxljs:
During my testing, I confirmed through the cluster logs that the ReadOnly option takes effect. However, the ClusterClient does not provide a corresponding API, so I couldn't write a test for it. Can I add a public API for test? Or do you have any idea? @ndyakov

cxljs:
@ndyakov Sorry, I didn't explain the purpose of this PR in detail before. I've now added more information.

ndyakov (approved):

PR #3485, opened 2025-08-18
fix UniversalOptions miss ReadBufferSize and WriteBufferSize options

cxljs:
refs: https://github.com/redis/go-redis/issues/3465#issuecomment-3196217612


ndyakov:
Thank you @cxljs. I already did that in a separate branch but will rebase.

PR #3491, opened 2025-08-27
fix(make test): Add default env in makefile

### LINKED ISSUE #3490: Unit / Integration Tests are Broken
## Expected Behavior

I should be able to run `make test` on a system with _either_ Docker Compose or Podman (and its equivalent) and have all the tests pass in `master`.

## Current Behavior

See [test-output.txt](https://github.com/user-attachments/files/21997348/test-output.txt) from a run on my system.

I'm willing to believe that this could be a compatibility problem with using Podman over Docker Compose, but it's hard to see how. 

I do note that there appears to be a setup that works enough for Github CI, but ... I don't think it's using this containerized environment at all.

## Possible Solution

1. The `--save ""` directive in `docker-compose.yml` seems to have the double-quotes escaped by the shell for three of the five environments fails; using `--save \"\"` seems to fix that so the Redis VMs start up.
2. Fixing the above still causes some tests to not be able to connect; uncommenting the commented out `ports:` and removing the `network: "host"` line seemed better, but still broken.

Here's the diff of what I did above [proposed-change-diff.txt](https://github.com/user-attachments/files/21997449/proposed-change-diff.txt), 
as well as output from running `make test` after those changes: [test-output-better.txt](https://github.com/user-attachments/files/21997436/test-output-better.txt).

## Steps to Reproduce

Just run `make test` as outlined in CONTRIBUTING.md. 🤷 

On my system, I'm using Podman. Output from `podman version`:

```
❯ podman version
Client:        Podman Engine
Version:       5.6.0
API Version:   5.6.0
Go Version:    go1.25.0
Built:         Fri Aug 15 09:42:35 2025
Build Origin:  brew
OS/Arch:       darwin/arm64

Server:       Podman Engine
Version:      5.4.2
API Version:  5.4.2
Go Version:   go1.23.7
Git Commit:   be85287fcf4590961614ee37be65eeb315e5d9ff
Built:        Tue Apr  1 20:00:00 2025
OS/Arch:      linux/arm64
```

## Context (Environment)

I'm just trying to develop patches and contribute to the project, and not having an test environment makes this riskier as I am unable to test my changes. This impacts all contributors to this project. 

## Detailed Description

Make the tests run clean. 

## Possible Implementation

I mean, fix it? (Or direct me to what is going on here so I can help?)

ndyakov:
- update to latest testing image
- add default env in makefile

Fixes #3490 for docker with host network available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,21 @@
 GO_MOD_DIRS := $(shell find . -type f -name 'go.mod' -exec dirname {} \; | sort)
+REDIS_VERSION ?= 8.2
+RE_CLUSTER ?= false
+RCE_DOCKER ?= true
+CLIENT_LIBS_TEST_IMAGE ?= redislabs/client-libs-test:8.2.1-pre
 
 docker.start:
+	export RE_CLUSTER=$(RE_CLUSTER) && \
+	export RCE_DOCKER=$(RCE_DOCKER) && \
+	export REDIS_VERSION=$(REDIS_VERSION) && \
+	export CLIENT_LIBS_TEST_IMAGE=$(CLIENT_LIBS_TEST_IMAGE) && \
 	docker compose --profile all up -d --quiet-pull
 
 docker.stop:
 	docker compose --profile all down
 
 test:
+	echo $(REDIS_VERSION)
 	$(MAKE) docker.start
 	@if [ -z "$(REDIS_VERSION)" ]; then \
 		echo "REDIS_VERSION not set, running all tests"; \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,12 +1,21 @@
...
+	export RCE_DOCKER=$(RCE_DOCKER) && \
+	export REDIS_VERSION=$(REDIS_VERSION) && \
+	export CLIENT_LIBS_TEST_IMAGE=$(CLIENT_LIBS_TEST_IMAGE) && \
 	docker compose --profile all up -d --quiet-pull
 
 docker.stop:
 	docker compose --profile all down
 
 test:
+	echo $(REDIS_VERSION)
```
ndyakov:
```suggestion
```

htemelski-redis (approved):

Issue #3494, opened 2025-08-30
UPDATE README: redisServerBin and redisServerConf are no longer present


gangcheng1030:
The redisServerBin and redisServerConf are no longer present in main_test.go, but they are mentioned in the README.

## Run the test

go-redis will start a redis-server and run the test cases.

The paths of redis-server bin file and redis config file are defined in `main_test.go`:

```go
var (
	redisServerBin, _  = filepath.Abs(filepath.Join("testdata", "redis", "src", "redis-server"))  // no longer present in main_test.go
	redisServerConf, _ = filepath.Abs(filepath.Join("testdata", "redis", "redis.conf"))  // no longer present in main_test.go
) 
```

cxljs:
Hello, I've submitted a PR #3495 to update the introduction about running tests in README.md. Hope it helps!

gangcheng1030:
> Hello, I've submitted a PR [#3495](https://github.com/redis/go-redis/pull/3495) to update the introduction about running tests in README.md. Hope it helps!

Thank you for the quick fix! I confirm that the issue has been resolved.

PR #3495, opened 2025-08-30
Update the introduction to running tests in README.md

cxljs:
ref: #3494


ndyakov (commented):
Ah, I missed that when updating the docs. Anyway, make test should be sufficient to run the tests. Would you like to suggest using that? This way we can control/change the actual way the tests are started and selected in the future (if needed)

cxljs:
Good idea!

ndyakov (approved):

PR #3496, opened 2025-08-31
Pipeliner expose queued commands

### LINKED ISSUE #3489: Expose pipeline command results via Cmds() method
## Motivation  
We are building a **transaction package manager** on top of `go-redis` that orchestrates pipelines and Postgres transactions together.  
For this integration, we need a way to **retrieve the list of commands** that were queued in a pipeline.  

At the moment, `pipeline.go` stores the commands in a private field. Because it is unexported, we cannot access or inspect the collected commands outside the package.  

## Proposal  
Introduce a method like:  
```go
func (p *Pipeline) Cmds() []Cmder

### PR DESCRIPTION
**Pull request by cxljs**
closes #3489

### CODE REVIEW

**Comment by ndyakov** [comment]
I think this will be okay to have. Will let @htemelski-redis review.

File: pipeline.go [hunk 1]
```diff
@@ -119,3 +122,7 @@ func (c *Pipeline) TxPipelined(ctx context.Context, fn func(Pipeliner) error) ([
 func (c *Pipeline) TxPipeline() Pipeliner {
 	return c
 }
+
```
ndyakov, line 125: @cxljs I know this is pretty much straight forward, but can we make sure we add a simple test for it? Maybe cover `nil`, zero elements and specific elements. `Pipeline` and `Pipelined` (if possible to test, can't think of a good way at the moment, but should be an interesting task to do :) )
Thank you!

ndyakov (commented):
Requested tests, not a blocker since this is straightforward getter.

pipeline.go
cxljs: Got it! I will update it later.

htemelski-redis (approved):

PR #3510, opened 2025-09-09
Added batch process method to the pipeline

htemelski-redis:
Addresses https://github.com/redis/go-redis/issues/3503

Changed `Process` to use `BatchProcess` since it's an edge case and DRY


ndyakov:
@htemelski-redis looks good to me, can we add a simple test to validate the new code and I can approve and merge this PR.

PR #3518, opened 2025-09-13
Feat/improve success rate of new connections

### LINKED ISSUE #3497: Proposal: Align context handling for connection creation with database/sql and net/http

cyningsun:
## Description

This PR implements a hybrid context pattern for connection creation to prevent premature cancellation caused by short-lived request timeouts, which previously led to connection pool instability under high concurrency.

## Solution
- Implemented a context approach for connection creation, similar to database/sql and net/http. The dialing operation now uses an independent context with a reasonable DialTimeout, while still listening for cancellation from the original request context for efficient cleanup
- Queue-based Fairness: Used a FIFO queue to ensure fairness among waiters when sharing a successful connection, preventing starvation.

## Changes
1. Implemented a FIFO queue for connection sharing.
2. Introduced a dedicated putIdleConn method to handle placement of newly created connections into the pool, bypassing the Put method’s hooks and idle connection checks.
> - **Avoids unnecessary hook executions** for newly established connections, improving performance and reducing overhead.
> - **Preserves resources by circumventing the MaxIdleConns constraint** during initial connection placement, preventing valid connections from being discarded prematurely when p.idleConnsLen.Load() >= p.cfg.MaxIdleConns.
> - **Defers enforcement of MaxIdleConns limits to the existing check during subsequent Put operations**, rather than immediately rejecting newly created connections. ( Avoids impacting connection acquisition latency by not adding cleanup overhead to `popIdle`)

## References

- Closes #3497 
- Discussion with @ndyakov in #3502 #3508

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -122,6 +122,14 @@ type wantConn struct {
 	result    chan wantConnResult // channel to deliver connection or error
 }
 
+// getCtxForDial returns context for dial or nil if connection was delivered or canceled.
+func (w *wantConn) getCtxForDial() context.Context {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	return w.ctx
+}
+
 func (w *wantConn) tryDeliver(cn *Conn, err error) bool {
 	w.mu.Lock()
 	defer w.mu.Unlock()
@@ -576,12 +584,13 @@ func (p *ConnPool) asyncNewConn(ctx context.Context) (*Conn, error) {
 		defer w.cancelCtx()
 		defer func() { <-p.dialsInProgress }() // Release connection creation permission
 
-		cn, cnErr := p.newConn(w.ctx, true)
+		dialCtx := w.getCtxForDial()
+		cn, cnErr := p.newConn(dialCtx, true)
 		delivered := w.tryDeliver(cn, cnErr)
 		if cnErr == nil && delivered {
 			return
 		} else if cnErr == nil && !delivered {
-			p.Put(w.ctx, cn)
+			p.Put(dialCtx, cn)
 		} else { // freeTurn after error
 			p.freeTurn()
 		}
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -91,6 +91,7 @@ type Options struct {
 
 	PoolFIFO                 bool
 	PoolSize                 int32
+	MaxConcurrentDials       int
 	DialTimeout              time.Duration
 	PoolTimeout              time.Duration
 	MinIdleConns             int32
@@ -113,13 +114,65 @@ type lastDialErrorWrap struct {
 	err error
 }
 
+type wantConn struct {
+	mu        sync.Mutex      // protects ctx, done and sending of the result
+	ctx       context.Context // context for dial, cleared after delivered or canceled
+	cancelCtx context.CancelFunc
+	done      bool                // true after delivered or canceled
+	result    chan wantConnResult // channel to deliver connection or error
+}
+
+func (w *wantConn) tryDeliver(cn *Conn, err error) bool {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	if w.done {
+		return false
+	}
+
+	w.done = true
+	w.ctx = nil
+
+	w.result <- wantConnResult{cn: cn, err: err}
+	close(w.result)
+
+	return true
+}
+
+func (w *wantConn) cancel(ctx context.Context, p *ConnPool) {
+	w.mu.Lock()
+	var cn *Conn
+	if w.done {
+		select {
+		case result := <-w.result:
+			cn = result.cn
+		default:
+		}
+	} else {
+		close(w.result)
+	}
+
+	w.done = true
+	w.ctx = nil
+	w.mu.Unlock()
+
+	if cn != nil {
+		p.Put(ctx, cn)
+	}
+}
+
+type wantConnResult struct {
+	cn  *Conn
+	err error
+}
+
 type ConnPool struct {
 	cfg *Options
 
 	dialErrorsNum uint32 // atomic
 	lastDialError atomic.Value
 
-	queue chan struct{}
+	queue           chan struct{}
+	dialsInProgress chan struct{}
 
 	connsMu   sync.Mutex
 	conns     map[uint64]*Conn
@@ -145,9 +198,10 @@ func NewConnPool(opt *Options) *ConnPool {
 	p := &ConnPool{
 		cfg: opt,
 
-		queue:     make(chan struct{}, opt.PoolSize),
-		conns:     make(map[uint64]*Conn),
-		idleConns: make([]*Conn, 0, opt.PoolSize),
+		queue:           make(chan struct{}, opt.PoolSize),
+		conns:           make(map[uint64]*Conn),
+		dialsInProgress: make(chan struct{}, opt.MaxConcurrentDials),
+		idleConns:       make([]*Conn, 0, opt.PoolSize),
 	}
 
 	// Only create MinIdleConns if explicitly requested (> 0)
@@ -473,9 +527,8 @@ func (p *ConnPool) getConn(ctx context.Context) (*Conn, error) {
 
 	atomic.AddUint32(&p.stats.Misses, 1)
 
-	newcn, err := p.newConn(ctx, true)
+	newcn, err := p.asyncNewConn(ctx)
 	if err != nil {
-		p.freeTurn()
 		return nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


ndyakov:
@cyningsun thank you for opening this pr, looking forward to continuing the discussion.

internal/pool/pool.go
```diff
@@ -512,6 +517,97 @@ func (p *ConnPool) getConn(ctx context.Context) (*Conn, error) {
...
+
+func (p *ConnPool) putIdleConn(ctx context.Context, cn *Conn) {
+	for {
+		w, ok := p.dialsQueue.dequeue()
+		if !ok {
+			break
+		}
+		if w.tryDeliver(cn, nil) {
+			return
+		}
```
ndyakov, line 595: if `putIdleConn` returns here, `p.freeTurn()` won't be called

internal/pool/pool.go
```diff
@@ -512,6 +517,97 @@ func (p *ConnPool) getConn(ctx context.Context) (*Conn, error) {
...
+	}
+
+	cn.SetUsable(true)
+
+	p.connsMu.Lock()
+	defer p.connsMu.Unlock()
+
+	if p.closed() {
+		_ = cn.Close()
+		return
```
ndyakov, line 605: here as well, `p.freeTurn()` won't be called. We can argue it doesn't matter in this case anyway.

cyningsun:
@ndyakov My apologies for the delayed response. The PR is now updated and ready for your next review.

htemelski-redis (approved):
Looks good to me too

PR #3519, opened 2025-09-14
Add support for filtering traces for certain commands

Sovietaced:
Closes https://github.com/redis/go-redis/issues/3516

`redisotel` can generate traces for Redis commands that include sensitive data (ie. hello command). There currently exists an option to disable the `db.statement` attribute in traces but this is a global setting. 

This pull request adds support for a configurable command filter that will omit traces for commands that filter returns true for. It also adds a basic command filter to filter out sensitive data by default which I feel like should be the correct behavior to prevent security incidents.


ndyakov:
Hello @Sovietaced and thank you for your contribution. Maybe @vmihailenco can take a look at this. If he doesn't I will around the end of the month.

ndyakov (dismissed):

ndyakov:
@Sovietaced I can see we have #3481 opened as well, I personally prefer the filtering approach in this PR, but let's discuss with @vmihailenco.

@htemelski-redis feel free to decide which one you prefer when you are back from vacation.

htemelski-oss:
I also do like this approach.

My only concern is the introduction of the `defaultCommandFilter`. It changes the tracing behaviour and some people may overlook it when upgrading the go-redis version.

A better approach, in my opinion, would be to rename it to `BasicCommandFilter` and leave the default filter to be `nil`

Sovietaced:
> My only concern is the introduction of the `defaultCommandFilter`. It changes the tracing behaviour and some people may overlook it when upgrading the go-redis version.
> 
> A better approach, in my opinion, would be to rename it to `BasicCommandFilter` and leave the default filter to be `nil`

Ultimately it is up to you all so I'm fine with updating the pull request with whatever decision you make. 

My preference is to provide safe defaults that don't cause security incidents since many folks will not spend time customizing these things.

ndyakov:
I do agree with @Sovietaced that we would like to hide the hello and acl commands by default. This may be considered a bugfix, not a breaking change. cc @htemelski-redis

PR #3550, opened 2025-10-20
feat(tracing): add trace filter for process pipeline and dial operation

Udhayarajan:
This PR is to address #3479.

While #3519 solves the issue partially (for Process Hook).

In this, I've added similar filters for dial hook and process pipeline hook


Udhayarajan:
Looks like the CI failed due to Docker pull issues on test images, not related to the code changes. 
Can someone please re-run the pipeline? Or should I push empty commit to retrigger it?

ndyakov:
Hello @Udhayarajan, I just reran the pipeline for you.

ndyakov:
@Udhayarajan what do you think about having the BasicCommandFilter renamed to `DefaultCommandFilter` and applied for all possible hooks, so we are sure by default the `auth` command is filtered as well as `hello` when there is a password in it?

Udhayarajan:
> @Udhayarajan what do you think about having the BasicCommandFilter renamed to `DefaultCommandFilter` and applied for all possible hooks, so we are sure by default the `auth` command is filtered as well as `hello` when there is a password in it?


that's really sounds good,  I'll make the changes

Udhayarajan:
@ndyakov  Pushed the changes, taking reference from one of the older PRs.
Keeping `DefaultCommandFilter` exported so users can chain them with custom ones if needed.

I also saw @htemelski-redis  had a different view earlier https://github.com/redis/go-redis/pull/3519#issuecomment-3345619980

ndyakov:
Yes, but it make sense to have it as default. Just because we had already released a version, keep the old name as well, so we don't break the contract.

Udhayarajan:
`DefaultCommandFilter` definitely sounds more appropriate than `BasicCommandFilter`. Since a version with the old name is already released, I’d suggest keeping both for now,  make `BasicCommandFilter` just a wrapper around `DefaultCommandFilter` and mark it as deprecated for backward compatibility.

PR #3551, opened 2025-10-21
docs: explain why MaxRetries is disabled for ClusterClient

### LINKED ISSUE #2249: Default MaxRetries on cluster clients is -1 and not 3
## Expected Behavior

I believe the `MaxRetries` [default value](https://github.com/go-redis/redis/blob/master/cluster.go#L113-L115) should be the same for clusters and standalone mode, unless it's intended?

## Current Behavior

[This commit](https://github.com/go-redis/redis/commit/b657760cca08d30dea76767e361bf9e06de0bfef) added a specific disable to `MaxRetries` in `ClusterOptions` as opposed to `Options` which defaults to 3.

## Possible Solution

If it's intended, we should add documentation in the code at least. If it's not, then it's a simple fix :)

Thanks

Pika-Gopher:
Fixes #2249 
This PR adds a comment to the ClusterOptions.MaxRetries field to explain why retries are disabled for ClusterClient. Since the cluster client handles all retries internally, this is intentional and different from the standalone client.

I noticed some confusion about this in the issues, so hopefully this makes things clearer for everyone.

### DIFF
--- a/extra/redisotel/metrics.go
+++ b/extra/redisotel/metrics.go
@@ -275,7 +275,7 @@ func (mh *metricsHook) DialHook(hook redis.DialHook) redis.DialHook {
 		attrs = append(attrs, mh.attrs...)
 		attrs = append(attrs, statusAttr(err))
 
-		mh.createTime.Record(ctx, milliseconds(dur), metric.WithAttributes(attrs...))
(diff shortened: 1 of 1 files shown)


ndyakov (approved):
Thank you @Pika-Gopher! Looks good to me, will add this.

PR #3552, opened 2025-10-21
chore(redisotel): use metric.WithAttributeSet to avoid copy

boekkooi-impossiblecloud:
Good day,

In order to improve performance this PR replaces `WithAttributes` with `WithAttributeSet`. This avoids the slice allocation and copy that is done in `WithAttributes`.

For more information see https://github.com/open-telemetry/opentelemetry-go/blob/v1.38.0/metric/instrument.go#L357-L376

Thanks for reviewing and have a great day!


ndyakov:
@boekkooi-impossiblecloud thank you for contributing!

PR #3555, opened 2025-10-22
fix(init): conn init should be thread safe

ndyakov:
conn init should be thread safe


ndyakov:
closed in favour of https://github.com/redis/go-redis/pull/3559

Issue #3558, opened 2025-10-24
Feature Request: ability to provide a logger as an Options setting


ccoVeille:
Right now, the logger is global to the module.

This could help to solve things like this:

- we may want a VoidLogger for one client, or for the sentinel client but not the others
- provide better context error with each client, for example when code handles connections to multiple Redis instances
- it can be challenging when dealing with unit tests.

Here this issue is here to get feedback about the idea.

I feel like a client option for the logger of the NewClient, NewSentinel and NewDialer seems an easy way to provide this
https://pkg.go.dev/github.com/redis/go-redis/v9#Options

Then we would fall back on the global logger.

I can implement this if you are fine with the idea.

ndyakov:
Hello @ccoVeille , we were thinking about adding a more configurable logger in `v10` because it would break the current logging. Your suggestion to use the global logger IF a local one is not specified makes sense and may enable us to ship this even in `v9.x`. Feel free to take a shot at it!

ccoVeille:
Great then, I will code it when I will have a computer with me, maybe next week.

Feel free to assign it to me.

Is there an issue I can follow about the v10 logger ? or is it just an idea/plan ?

Thanks

ndyakov:
@ccoVeille it's just one of the topics we marked as tech dept that should be addressed, I do have some ideas, maybe we can work on putting them in writing. 

The logger:
- should be able to support different log levels
- should be able to print structured log messages, at the moment I used json for https://github.com/redis/go-redis/blob/master/internal/maintnotifications/logs/log_messages.go , you can take a look at it
- in the future, we may want to integrate https://opentelemetry.io/docs/specs/otel/logs/ , so this should be doable
- for the tests, it is useful to be able to filter given logs by pattern (blacklist or whitelist)

ccoVeille:
This would go beyond what I thought, but why not.

I feel like logger could be a [slog.Handler](https://pkg.go.dev/log/slog#Handler)

https://github.com/golang/example/blob/master/slog-handler-guide/README.md

I could provide a logger like this, and refactoring everything and make it retro compatible. But the scope would be wider than I wanted to work on.

PR #3560, opened 2025-10-24
feat: add optional logger wherever possible

ccoVeille:
This commit introduces an optional logger parameter to various structs.
This enhancement allows users to provide custom logging implementations.

This is a naive implementation of #3558

It is provided to iterate on the changes, there are many things that are missing:
- tests
- WithLogger should be added at multiple places

Also, the current discussion in #3558 let me think the current logging interface should be reconsidered, so everything will have to be refactored.


ndyakov:
@ccoVeille thank you for opening this PR. I am willing to work with you on improving the logging and do agree that the current interface lacks some useful features. If you have a current need to just be able to set logger per client, we can improve this initially and then continue on the whole logging interface.

ccoVeille:
I found a workaround by injecting a dedicated logger in the context and restoring from the context in the current Printf interface.

So I don't have a need right now. We can work together on the target solution you planned initially for the v10 with a fallback for the v9

ndyakov:
so @ccoVeille would you consider this pr ready for review ?

ccoVeille:
Yes, you can review. I'm curious about your feedback.

I dislike the idea it adds a configuration option in each struct that satisfies the current internal.Logging interface. 

I feel like we shouldn't add option with interface we may change for supporting the log level and other things you listed.

So here I'm open to suggestions.

ccoVeille:
> so @ccoVeille would you consider this pr ready for review ?

Ping @ndyakov 

Maybe I should have pinged you

ndyakov:
@ccoVeille just moved it to Ready for Review and will work on reviewing it this week, thank you.

internal/pool/pool.go
```diff
@@ -115,6 +115,9 @@ type Options struct {
 	// DialerRetryTimeout is the backoff duration between retry attempts.
 	// Default: 100ms
 	DialerRetryTimeout time.Duration
+
+	// Optional logger for connection pool operations.
+	Logger *internal.Logging
```
ndyakov: Why is this a pointer to an `interface` ? I the type should be the interface itself.

internal/pool/pool.go
```diff
@@ -961,3 +964,11 @@ func (p *ConnPool) isHealthyConn(cn *Conn, now time.Time) bool {
 	}
 	return true
 }
+
+func (p *ConnPool) logf(ctx context.Context, format string, args ...any) {
+	logger := internal.Logger
+	if p.cfg.Logger != nil {
+		logger = *p.cfg.Logger
+	}
+	logger.Printf(ctx, format, args...)
```
ndyakov:
```suggestion
	logger := p.cfg.Logger
	if logger == nil {
		logger = internal.Logger
	}
	logger.Printf(ctx, format, args...)
```

I think this shows the precedence more clearly. We could go as far as 
```
	if p.cfg.Logger != nil {
		p.cfg.Logger.Printf(ctx, format, args...)
	} else {
		internal.Logger.Printf(ctx, format, args...)
	}
```
Which would be the more widely used approach.

maintnotifications/circuit_breaker.go
```diff
@@ -351,3 +359,11 @@ func (cbm *CircuitBreakerManager) Reset() {
...
 	})
 }
+
+func (cbm *CircuitBreakerManager) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if cbm.config != nil && cbm.config.Logger != nil {
+		logger = *cbm.config.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
```
ndyakov, line 361: Consider changing this similarly to the previous note about `logf` in the pool

maintnotifications/config.go
```diff
@@ -128,6 +128,9 @@ type Config struct {
 	// After this many retries, the connection will be removed from the pool.
 	// Default: 3
 	MaxHandoffRetries int
+
+	// Logger is an optional custom logger for maintenance notifications.
+	Logger *internal.Logging
```
ndyakov:
```suggestion

	// Logger is an optional custom logger for maintenance notifications.
	Logger internal.Logging
```

maintnotifications/handoff_worker.go
```diff
@@ -478,15 +478,23 @@ func (hwm *handoffWorkerManager) closeConnFromRequest(ctx context.Context, reque
...
 	}
 }
+
+func (hwm *handoffWorkerManager) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if hwm.config != nil && hwm.config.Logger != nil {
+		logger = *hwm.config.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
```
ndyakov, line 503: same as the other `logf` comments

maintnotifications/manager.go
```diff
@@ -318,3 +318,11 @@ func (hm *Manager) AddNotificationHook(notificationHook NotificationHook) {
...
 	hm.hooks = append(hm.hooks, notificationHook)
 }
+
+func (hm *Manager) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if hm.config != nil && hm.config.Logger != nil {
+		logger = *hm.config.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
```
ndyakov, line 320: same, maybe this can be extracted in a separate type/struct and composed ?  You can just address the functions themselves and we can work on this in the future we will have to see if having this in something like `Logger` but composed withint the `Manager`. Anyway, let's address only the functions at this stage.

maintnotifications/pool_hook.go
```diff
@@ -182,3 +182,11 @@ func (ph *PoolHook) OnRemove(_ context.Context, _ *pool.Conn, _ error) {
...
 	return ph.workerManager.shutdownWorkers(ctx)
 }
+
+func (ph *PoolHook) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if ph.config != nil && ph.config.Logger != nil {
+		logger = *ph.config.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
```
ndyakov, line 190: same as other `logf` definitions.

maintnotifications/push_notification_handler.go
```diff
@@ -257,26 +257,34 @@ func (snh *NotificationHandler) handleFailedOver(ctx context.Context, handlerCtx
...
 	return nil
 }
+
+func (snh *NotificationHandler) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if snh.manager != nil && snh.manager.config != nil && snh.manager.config.Logger != nil {
+		logger = *snh.manager.config.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
```
ndyakov, line 283: same.

options.go
```diff
@@ -267,6 +268,10 @@ type Options struct {
 	// transitions seamlessly. Requires Protocol: 3 (RESP3) for push notifications.
 	// If nil, maintnotifications are in "auto" mode and will be enabled if the server supports it.
 	MaintNotificationsConfig *maintnotifications.Config
+
+	// Logger is the logger used by the client for logging.
+	// If none is provided, the global logger [internal.Logger] is used.
+	Logger *internal.Logging
```
ndyakov: this should not be a pointer to an interface.

osscluster.go
```diff
@@ -148,6 +148,9 @@ type ClusterOptions struct {
 	// If nil, maintnotifications upgrades are in "auto" mode and will be enabled if the server supports it.
 	// The ClusterClient does not directly work with maintnotifications, it is up to the clients in the Nodes map to work with maintnotifications.
 	MaintNotificationsConfig *maintnotifications.Config
+
+	// Logger is an optional logger for logging cluster-related messages.
+	Logger *internal.Logging
```
ndyakov:
```suggestion

	// Logger is an optional logger for logging cluster-related messages.
	Logger internal.Logging
```

osscluster.go
```diff
@@ -703,6 +708,14 @@ func (c *clusterNodes) Random() (*clusterNode, error) {
...
 }
 
+func (c *clusterNodes) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if c.opt.Logger != nil {
+		logger = *c.opt.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
+
```
ndyakov: same as other `logf`

osscluster.go
```diff
@@ -2126,6 +2139,16 @@ func (c *ClusterClient) context(ctx context.Context) context.Context {
...
+func (c *ClusterClient) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+
+	if c.opt.Logger != nil {
+		logger = *c.opt.Logger
+	}
+
+	logger.Printf(ctx, format, args...)
+}
+
```
ndyakov: same as other `logf` definitions

pubsub.go
```diff
@@ -141,6 +141,17 @@ func mapKeys(m map[string]struct{}) []string {
...
+//
+// it uses the client logger if set, otherwise it uses the global logger.
+func (c *PubSub) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if c.opt.Logger != nil {
+		logger = *c.opt.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
+
```
ndyakov: same as the rest of the `logf` definitions.

pubsub.go
```diff
@@ -631,9 +642,18 @@ func WithChannelSendTimeout(d time.Duration) ChannelOption {
...
+		c.Logger = logger
+	}
+}
+
 type channel struct {
 	pubSub *PubSub
 
+	// Optional logger for logging channel-related messages.
+	Logger *internal.Logging
+
```
ndyakov: those should not be pointers to an interface

pubsub.go
```diff
@@ -787,13 +807,24 @@ func (c *channel) initAllChan() {
...
+func (c *channel) logf(ctx context.Context, format string, args ...any) {
+	logger := internal.Logger
+	switch {
+	case c.pubSub.opt.Logger != nil:
+		logger = *c.pubSub.opt.Logger
+	case c.Logger != nil:
+		logger = *c.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
```
ndyakov, line 819: this is a bit more tricky, but again, we don't have to dereference the interface (what we are doing at the moment) and the default should be the last fallback.

redis.go
```diff
@@ -228,6 +228,9 @@ type baseClient struct {
 
 	// streamingCredentialsManager is used to manage streaming credentials
 	streamingCredentialsManager *streaming.Manager
+
+	// Logger is the logger used by the client for logging.
+	logger *internal.Logging
```
ndyakov:
```suggestion

	// Logger is the logger used by the client for logging.
	logger internal.Logging
```

redis.go
```diff
@@ -672,6 +676,17 @@ func (c *baseClient) context(ctx context.Context) context.Context {
...
+// it uses the client logger if set, otherwise it uses the global logger.
+func (c *baseClient) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+	if c.logger != nil {
+		logger = *c.logger
+	}
+
+	logger.Printf(ctx, format, args...)
+}
+
```
ndyakov: same as rest of `logf` definitions.

ring.go
```diff
@@ -154,6 +154,8 @@ type RingOptions struct {
 	DisableIdentity bool
 	IdentitySuffix  string
 	UnstableResp3   bool
+
+	Logger *internal.Logging
```
ndyakov:
```suggestion

	Logger internal.Logging
```

ring.go
```diff
@@ -559,6 +561,14 @@ func (c *ringSharding) Close() error {
...
 }
 
+func (c *ringSharding) logf(ctx context.Context, format string, args ...any) {
+	logger := internal.Logger
+	if c.opt.Logger != nil {
+		logger = *c.opt.Logger
+	}
+	logger.Printf(ctx, format, args...)
+}
+
```
ndyakov: same.

sentinel.go
```diff
@@ -148,6 +148,9 @@ type FailoverOptions struct {
 	// If nil, maintnotifications upgrades are disabled.
 	// (however if Mode is nil, it defaults to "auto" - enable if server supports it)
 	//MaintNotificationsConfig *maintnotifications.Config
+
+	// Optional logger for logging
+	Logger *internal.Logging
```
ndyakov:
```suggestion

	// Optional logger for logging
	Logger internal.Logging
```

sentinel.go
```diff
@@ -1103,6 +1112,16 @@ func (c *sentinelFailover) listen(pubsub *PubSub) {
...
+func (c *sentinelFailover) logf(ctx context.Context, format string, args ...interface{}) {
+	logger := internal.Logger
+
+	if c.opt.Logger != nil {
+		logger = *c.opt.Logger
+	}
+
+	logger.Printf(ctx, format, args...)
+}
+
```
ndyakov: and this as well

internal/pool/pool.go
ccoVeille: I would have preferred to do not.

I'm unsure why I came to this.

I'll check.

internal/pool/pool.go
ccoVeille: I'll update to a pattern like this, here and everywhere.

maintnotifications/manager.go
ccoVeille: Yes, it could be done differently. It was uneasy for me to know how far I could go with this rewrite. I decided to limit the changes, but then I felt disapointed by the code quality

ndyakov: There is no rush to do it in one PR. As long as we can keep the solution backwards compatible, we can iterate over the intervals in multiple PRs. Let me know if you need any assistance. You can also reach me on discord.

ccoVeille: Of course, but I would prefer to do not provide an exported option everywhere if the design is bad 😅

ccoVeille: > You can also reach me on discord. 

i find you on Slack's Gophers workspace, but not on Discord Gophers

https://discord.gg/golang

I'm on it with ccoVeille.

ccoVeille:
@ndyakov  I applied your suggestions to the first commit

Then I have added another commit where I apply something that seems to me to be better.

Let me know what you think.

ccoVeille:
I found a panic in the code, I made a fix for this branch, but you may want me to fix it separately

The bug is here, because err can be nil
https://github.com/redis/go-redis/blob/c1766729ec1826a1bfcdcd1d506d4ce44c1734ea/maintnotifications/handoff_worker.go#L484-L490

But here the call to reason.Error() will panic

https://github.com/redis/go-redis/blob/c1766729ec1826a1bfcdcd1d506d4ce44c1734ea/internal/maintnotifications/logs/log_messages.go#L298-L303

Here is the code that spotted it because of my refactoring to always call the logger, and so remove the test on WarnOrAbove
https://github.com/redis/go-redis/actions/runs/19151645751/job/54743116032#step:3:1025

internal/pool/pool.go
```diff
@@ -961,3 +964,11 @@ func (p *ConnPool) isHealthyConn(cn *Conn, now time.Time) bool {
 	}
 	return true
 }
+
+func (p *ConnPool) logger() internal.LoggerWithLevel {
+	if p.cfg.Logger != nil {
+		return p.cfg.Logger
+	}
+
+	return &internal.LegacyLoggerAdapter{}
```
ndyakov: this is stateless, so why don't we just create a global LegacyLoggerAdapterInstance (can definitely be named better) and use it? This will instantiate a new LegacyLoggerAdapter for each logger() call, right? We don't need (or want) that.

ndyakov: p.s. I saw you already have `LegacyLoggerWithLevel` which should work here.

ndyakov (approved):
Nice approach! Thank you for fixing the panic as well! There is just one comment and I think we can merge this (please sync with master, there is another PR that may interfere with this one, but I will figure out the merging order so we can release next week.

maintnotifications/handoff_worker.go
```diff
@@ -219,20 +211,17 @@ func (hwm *handoffWorkerManager) processHandoffRequest(request HandoffRequest) {
...
 				// Get current retry count for better logging
 				currentRetries := request.Conn.HandoffRetries()
 				maxRetries := 3 // Default fallback
 				if hwm.config != nil {
 					maxRetries = hwm.config.MaxHandoffRetries
 				}
-				hwm.logf(context.Background(), logs.HandoffFailed(request.ConnID, request.Endpoint, currentRetries, maxRetries, err))
-			}
+				hwm.logger().Infof(context.Background(), logs.HandoffFailed(request.ConnID, request.Endpoint, currentRetries, maxRetries, err))
+
```
ndyakov: I would've kept the `if internal.LogLevel.InfoOrAbove() {` so the calculation can be skipped when logging is not needed.

internal/pool/pool.go
ccoVeille: I apparently forgot this one

internal/maintnotifications/logs/log_messages.go
```diff
@@ -295,11 +295,10 @@ func RemovingConnectionFromPool(connID uint64, reason error) string {
...
 }
 
-func NoPoolProvidedCannotRemove(connID uint64, reason error) string {
-	message := fmt.Sprintf("conn[%d] %s due to: %v", connID, NoPoolProvidedMessageCannotRemoveMessage, reason)
+func NoPoolProvidedCannotRemove(connID uint64) string {
+	message := fmt.Sprintf("conn[%d] %s", connID, NoPoolProvidedMessageCannotRemoveMessage)
 	return appendJSONIfDebug(message, map[string]interface{}{
 		"connID": connID,
-		"reason": reason.Error(),
 	})
```
ndyakov, line 303: it will still be nice to have the reason included if it is `!=nil`

ccoVeille:
I plan to squash the first 2 commits

internal/maintnotifications/logs/log_messages.go
ccoVeille: Please suggest an error message you want me to use when there is no error in the code calling this

maintnotifications/handoff_worker.go
ccoVeille: it should be better now, please review

internal/pool/pool.go
```diff
@@ -835,7 +838,7 @@ func (p *ConnPool) removeConn(cn *Conn) {
 		// this can be idle conn
 		for idx, ic := range p.idleConns {
 			if ic.GetID() == cid {
-				internal.Logger.Printf(context.Background(), "redis: connection pool: removing idle conn[%d]", cid)
+				p.logger().Errorf(context.Background(), "redis: connection pool: removing idle conn[%d]", cid)
```
ccoVeille: I'm unsure about the error level here.

Should we decrease to a Warn or an Info ?

ndyakov: That is an Info actually.

internal/log.go
```diff
@@ -77,3 +77,66 @@ func (l LogLevelT) InfoOrAbove() bool {
...
+	Infof(ctx context.Context, format string, v ...interface{})
+
+	// Warnf logs a warning level message
+	Warnf(ctx context.Context, format string, v ...interface{})
+
+	// Debugf logs a debug level message
+	Debugf(ctx context.Context, format string, v ...interface{})
+
+	// Errorf logs an error level message
+	Errorf(ctx context.Context, format string, v ...interface{})
```
ndyakov: can we just order them by error level

internal/log.go
```diff
@@ -77,3 +77,66 @@ func (l LogLevelT) InfoOrAbove() bool {
...
+
+	// Enabled reports whether the given log level is enabled in the logger
+	Enabled(ctx context.Context, level LogLevelT) bool
```
ndyakov: I am not sure about needing a context here even in the future. Can you think of an example in which it will be useful?

internal/maintnotifications/logs/log_messages.go
ndyakov: it is fine to have "due to: nil" and "reason": nil
I am still confused how this happened and will have to investigate further once we have this merged.

ccoVeille: @ndyakov please let me know what you want here

ndyakov (dismissed):
This is starting to look better on each iterations, congrats @ccoVeille.

I have one note / suggestion that we can either implement now and do it in a separate PR:

- Right now we have one official way of setting the log level `SetLogLevel` but at the end this is only needed when there is no custom logger (granted, a user can use the global log level in their own custom logger implementation, but this is not important).

What I would like to suggest is to have a "constructor" function to create a new logger for a given level and then suggest (in the readme and in the documentation) that the users should create custom loggers and pass them in the `options` if they would like to change the log level. Of course, we won't remove the `SetLogLevel` function to be backwards compatible, and it will continue to work, but we can mark it as Deprecated and this can help us transition easily to custom loggers in the future.

internal/log.go
ccoVeille: I hesitated. But as I found one on slog.Logger

https://pkg.go.dev/log/slog#Logger.Enabled

ccoVeille:
> This is starting to look better on each iterations, congrats @ccoVeille.

You are welcome.

> I have one note / suggestion that we can either implement now and do it in a separate PR:
> 
> - Right now we have one official way of setting the log level `SetLogLevel` but at the end this is only needed when there is no custom logger (granted, a user can use the global log level in their own custom logger implementation, but this is not important).
> 
> What I would like to suggest is to have a "constructor" function to create a new logger for a given level and then suggest (in the readme and in the documentation) that the users should create custom loggers and pass them in the `options` if they would like to change the log level. Of course, we won't remove the `SetLogLevel` function to be backwards compatible, and it will continue to work, but we can mark it as Deprecated and this can help us transition easily to custom loggers in the future. 

We can of course do it in another PR.

But I feel like the current MR should target having code that accepts a logger which signature would match the one of slog.Logger

I feel like your lib should move to structured logger. Not in this PR. Take a look at the hack you have with the metadata you add via json.Encode.

I dislike the idea of merging the PR like this.

I read again slog and log. I have ideas. I will try to find a solution.

ndyakov:
@ccoVeille I agree, but don't see a clear way to do it without breaking changes (if we include this PR). Let me think about it a bit. I won't have much time next week, but will get back to it as soon as possible.

ccoVeille:
> @ccoVeille I agree, but don't see a clear way to do it without breaking changes (if we include this PR). Let me think about it a bit. I won't have much time next week, but will get back to it as soon as possible.

I refactored everything.

Please take a took. I added structured logging support.

There might be some iterations to validate everything works as expected. I also need to add tests.
But now, it matches what I would expect from the package.

internal/log.go
ccoVeille: I removed the need to use Enabled in my latest changes

maintnotifications/handoff_worker.go
```diff
@@ -477,16 +463,20 @@ func (hwm *handoffWorkerManager) closeConnFromRequest(ctx context.Context, reque
...
-		}
+		hwm.logger().Warnf(ctx, logs.RemovingConnectionFromPool(conn.GetID(), err))
 	} else {
 		err := conn.Close() // Close the connection if no pool provided
 		if err != nil {
-			internal.Logger.Printf(ctx, "redis: failed to close connection: %v", err)
-		}
-		if internal.LogLevel.WarnOrAbove() {
-			internal.Logger.Printf(ctx, logs.NoPoolProvidedCannotRemove(conn.GetID(), err))
+			hwm.logger().Errorf(ctx, "redis: failed to close connection: %v", err)
```
ndyakov, line 491: can you please add a TODO here for me to extract this error as well?

logging/legacy.go
```diff
@@ -0,0 +1,91 @@
...
+
+	for i := 0; i < len(v); i += 2 {
+		if i+1 >= len(v) {
+			break
+		}
+		format += " %v=%v"
+		args = append(args, v[i], v[i+1])
+	}
+
+	return format, args
```
ndyakov, line 31: the logger may miss the last v if len(v) is not even

ndyakov (approved):
overall looks good to me, i will change the messages / calling in the maintnotif package to have more wraps with `Enabled`, but the logger itself looks good. What do you think @ccoVeille, are you happy with this iteration?

logging/legacy.go
ccoVeille: I can adjust. I feel like I can use something like !BADITEM or any marker like sprintf does.

Or simply let sprintf handle it, it would result an error like the one I describe

ccoVeille:
I'll try to work on this soon

ccoVeille:
this is not funny :grin: 

image

logging/legacy.go
ccoVeille: Here is the system in the Go code when a slog.Hander is defined as log.SetDefault

https://cs.opensource.google/go/go/+/master:src/log/slog/record.go;l=160-182;drc=8c41a482f9b7a101404cd0b417ac45abd441e598

ccoVeille:
I'm facing a panic, I will try to identify its root cause

But I noticed something strange

redis: 2025/11/19 15:42:03 legacy.go:35: conn[16] connection marked for handoff
redis: 2025/11/19 15:42:03 legacy.go:35: conn[17] connection marked for handoff
redis: 2025/11/19 15:42:03 legacy.go:35: conn[18] connection marked for handoff
redis: 2025/11/19 15:42:03 legacy.go:35: conn[19] connection marked for handoff

The default logger is now reporting legacy.go:35 as the emitter of all error.

I can try to fix it, but it means it would have consequence on people who implemented their own logger, using stack in their logger

ccoVeille:
But I feel the issue also exists with the filterLogger that introduces another level in the stack when calling Printf

ccoVeille:
I used calldepth:4 in the default logger, but it's a bit imperfect according to me as there a still a few calls to `internal.Logger.Printf` in the code

One solution could be to create a internal.LegacyLoggerPrintf method, that will call internal.Logger by adding two level to the stack, it can be done.

Also, filteredLogger could be fixed the same way. But I would appreciate if it could be merged like this, and fixed later

@ndyakov

internal/pool/pool.go
```diff
@@ -257,7 +261,7 @@ func (p *ConnPool) checkMinIdleConns() {
 						p.idleConnsLen.Add(-1)
 
 						p.freeTurn()
-						internal.Logger.Printf(context.Background(), "addIdleConn panic: %+v", err)
+						p.logger().Errorf(context.Background(), "checkMinIdleConns panic: %+v", err)
```
ndyakov: this is in a goroutine that executes `addIdleConn`, so prtobably the panic is in `addIdleConn` not in `checkMinIdleConns`

ndyakov (commented):
overall looks good to me, i can handle the rest of the internal.Logger and do some testing to verify the output is as expected (we do have some e2e tests that assert based on the logs).

internal/pool/pool.go
ccoVeille: I apparently mixed up things with a conflict

ccoVeille:
The others requires a larger refactoring, something I wanted to avoid to make things easier to review. I'm glad you consider to do them anyway.

Next step in another PR, might be to start refectory to use a non-sprintf logger.

Something that could be a Info(ctx, msg, keyVals...)

But here, it's a large refactoring

ndyakov:
@ccoVeille if you don't mind, I will use your branch to update some of the logger calls that are left

ccoVeille:
Sure.

If you want to push to my repository, you may face issues because it belongs to an organization. A place where there is no way to enable the classic "allow maintainers to ...".

So if you want this, please let me know. I can either move out my fork from the repository or grant as a maintainer in my fork

PR #3563, opened 2025-10-24
internal/proto/peek_push_notification_test : Refactor test helpers to…

12ya:
… use fmt.Fprintf for buffers

Replaced buf.WriteString(fmt.Sprintf(...)) with fmt.Fprintf or fmt.Fprint in test helper functions for improved clarity and efficiency. This change affects push notification and RESP3 test utilities.

### DIFF
--- a/doctests/cmds_hash_test.go
+++ b/doctests/cmds_hash_test.go
@@ -79,7 +79,7 @@ func ExampleClient_hset() {
 
 	keys := make([]string, 0, len(res6))
 
-	for key, _ := range res6 {
+	for key := range res6 {
 		keys = append(keys, key)
 	}
 
@@ -186,7 +186,7 @@ func ExampleClient_hgetall() {
 
 	keys := make([]string, 0, len(hGetAllResult2))
 
-	for key, _ := range hGetAllResult2 {
+	for key := range hGetAllResult2 {
 		keys = append(keys, key)
 	}
 
--- a/hset_benchmark_test.go
+++ b/hset_benchmark_test.go
@@ -86,7 +86,7 @@ func benchmarkHSETOperations(b *testing.B, rdb *redis.Client, ctx context.Contex
 				b.Fatalf("HSET operation failed: %v", err)
 			}
 		}
-		totalTimes = append(totalTimes, time.Now().Sub(startTime))
+		totalTimes = append(totalTimes, time.Since(startTime))
 	}
 
 	// Stop the timer to calculate metrics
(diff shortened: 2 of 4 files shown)


internal/proto/peek_push_notification_test.go
```diff
@@ -295,25 +295,26 @@ func createValidPushNotification(notificationName, data string) *bytes.Buffer {
...
 	if data == "" {
 
 		// Single element notification
 		buf.WriteString(">1\r\n")
 		if simpleOrString {
-			buf.WriteString(fmt.Sprintf("+%s\r\n", notificationName))
+			fmt.Fprint(buf, defMsg)
 		} else {
-			buf.WriteString(fmt.Sprintf("$%d\r\n%s\r\n", len(notificationName), notificationName))
+			fmt.Fprint(buf, defMsg)
```
ndyakov, line 307: Those are two different cases that should be covered.

internal/proto/peek_push_notification_test.go
```diff
@@ -295,25 +295,26 @@ func createValidPushNotification(notificationName, data string) *bytes.Buffer {
 	buf := &bytes.Buffer{}
 
 	simpleOrString := rand.Intn(2) == 0
+	defMsg := fmt.Sprintf("$%d\r\n%s\r\n", len(notificationName), notificationName)
 
 	if data == "" {
 
 		// Single element notification
 		buf.WriteString(">1\r\n")
```
ndyakov: should we use Fprint for those as well (there is one more around line 311

internal/proto/peek_push_notification_test.go
```diff
@@ -350,7 +351,7 @@ func createPushNotificationWithArgs(notificationName string, args ...string) *by
 func createSingleElementPushNotification(notificationName string) *bytes.Buffer {
 	buf := &bytes.Buffer{}
 	buf.WriteString(">1\r\n")
```
ndyakov: here as well?

internal/proto/peek_push_notification_test.go
12ya: 62f8bdb0b284318aa4811b083f2eeace0c84ecd0

internal/proto/peek_push_notification_test.go
12ya: 46523627927893dd5fa27bcd956dfee9bb7856d4

12ya:
> @12ya thank you for your contribution. left couple of comments for things that should be addressed. lets keep this consistent if we decide to use `fmt.Fprint` instead of `buf.WriteString`

hey @ndyakov, i agree - replaced everywhere for consistency

PR #3567, opened 2025-10-25
feat(otel): add unified filtering API with O(1) command exclusion

### LINKED ISSUE #3479: Feature Request: Filter out specific commands for OTEL Tracing

zadil:
Add comprehensive filtering capabilities with performance-optimized command exclusion and unified filtering API.

- WithExcludedCommands(): O(1) command exclusion using map lookup
- WithProcessFilter(): Custom filtering for individual commands
- WithProcessPipelineFilter(): Custom filtering for pipeline operations
- WithDialFilterFunc(): Custom filtering for dial operations
- Backward compatible with existing filtering APIs
- Comprehensive test coverage and updated examples

Fixes #3479


ndyakov:
@zadil thank you for your contribution, will review it next week.

ndyakov:
pinging @Udhayarajan and @lucawol for comments here as they were active in the discussion.

Udhayarajan:
I've a few things to mark on.

## 0. `excludedCommands`

The goal is to filter in O(1). Anyone who want to achieve O(1) is still possible via existing methods. They can create a custom stuct and still can achieve the same (as I mentioned in discussion). Afaik go inline methods, when I tested with same I proposed in discussion I can see the go compiler says both `FilterPipeline` and `FilterCommand` can inline. The O(1) map lookup logic is now as efficient as possible since the filtering methods are inlined. 

This `excludeCommands` just offers a way so that people no need to create their own struct to achieve O(1).

### Trade off:
IMO, We are trading off library code quality with pre-cooked fields in config struct. I love the way you reduced double maintenance with simple helper functions, everything looks like over-kill, at lease to me ,as `func(cmds []redis.Cmder) bool` and `func(cmd redis.Cmder) bool` can do what `excludedCommands` does. 

## 1. `unified*` vs `filter*` (excluding `unifiedDialFilter`)

everything is exactly the same do we really need to duplicate this fields? Though the version is released these are non-exported filed and shouldn't affect any user, may be we just have one field and use them in 2 different option functions, what's your opinion?

## 2. unifiedDialFilter

I designed dial filter assuming, user won't filter the trace based on network and address thus kept is simple. Current proposed approach gives better flexibility to filter the dial with help of address and port. 

### Trade off:
I don't think there is any critical trade-off, but in rare case, say user configured to filter from their current address and port but later if redis server is changed then they suddenly starting seeing trace. Again, I'm not sure is this trade-off or can this be even labeled as careless update by user. If I'm that user for sure I won't be remembering this part where I configured to ignore dial from certain address and port. **But it offers great flexibility to users for filtering out dial for specific server in cluster mode.** 

#### I'm in favor of this `unifiedDialFilter`


@ndyakov pls add your inputs too. 

cc: @htemelski-redis

Udhayarajan:
I would like to pull in one more person, since we're just adding toppings to the base baked by @Sovietaced

ndyakov:
Here are my 2 cents:
Let's think about the main usecase of those filter:
- Developers would like to filter traces/logs related to AUTH commands
- Developers may want to filter traces including specific keys/ key prefixes.

Although not the most performant approach, the way to address both would be to search not only the command name, but the body of the command itself. We could exclude both `HELLO` and `AUTH` for the `AUTH` case, but what about the keys? 

I haven't benchmarked this, so take my suggestion with a grain of salt:
- I would have two filters, exclude command and exclude keys
- - exclude command would be the "faster" approach
- - exclude keys is tricky to implement, so maybe even we would like to skip it? If we do a string match we can end up matching the `value` and not a key.

Both filter should be applicable to both pipelines and normal commands.

As for the dialer, I do not think we need a filter there in general, don't have a strong opinion if we should keep one, if you think it is going to be useful, sure, let's keep it. 

The map lookup may limit the flexibility of the filter, but at the same time gives the users a single way to filter a command, which should be less prone to errors from unexpected or wrong use of the filters. 

I would vote for the map lookup. even, would store both uppercase and lowercase cmd names in the map and skip the strings.ToUpper on each execution.

ndyakov:
@Udhayarajan , @zadil any final decision we can work with?

zadil:
Thank you for the thorough feedback, @Udhayarajan and @ndyakov. I really appreciate the in-depth analysis. Here's my proposal to move forward:

## On the key points:

**1. `excludedCommands` - Let's simplify**

You're absolutely right, @Udhayarajan. If developers can already achieve O(1) with custom structs and inlined methods, then `excludedCommands` doesn't add that much value. I propose to **remove `excludedCommands`** and let advanced users use the filter callbacks if needed. This reduces API complexity.

**2. `unified*` vs `filter*` - Consolidate**

Excellent observation, @Udhayarajan. These fields are identical and non-exported. I agree to **have a single field and reuse them in both option functions**. This improves maintainability with no user impact.

**3. `unifiedDialFilter` - I'm keeping it**

@ndyakov, I'm convinced by your point: it's useful in cluster mode. Even if there's a theoretical risk that a user forgets their configuration after a server change, it's a very rare edge case and the flexibility it provides is worth it. I propose to **keep `unifiedDialFilter`** but add a note in the documentation mentioning this edge case.

**4. On removing `excludedCommands`**

If we remove `excludedCommands` entirely, this point becomes moot. The map lookup optimization was specific to that API. Users would instead use `WithProcessFilter()` or `WithProcessPipelineFilter()` callbacks for custom filtering logic, giving them full flexibility without pre-baked solutions.

Should I proceed with these changes? I can have an updated PR in a few days.

Udhayarajan:
Sorry Stuck on a busy schedule.  Just now check the discussion.

I wasn’t trying to push back on the addition of `excludedCommands`, I just wanted to share my point of view so we could reach a solid conclusion. Glad you accepted the feedback! If there are still any mismatched opinions, maybe we could do a benchmark so it stands as future proof. My schedule is tight this week, but I’ll try to set aside some time this weekend if possible.

Totally agree, it’s really good to keep unifiedDialFilter.

And yes, you’re right on the last point.

ndyakov:
@Udhayarajan @zadil would you like to proceed with this one? @ofekshenawa can help you with feedback, he worked on a newer approach for reporting telemetry and may have meaningful feedback on this one. After all, I do think the team will focus on supporting the new `redisotel-native` package and the old `redisotel` will be mostly supported by the community. For this reason, I will leave you to decide the correct approach as long as there are no breaking changes for current user, I am willing to help you with merging this.

Udhayarajan:
I would like to add one point,

I’d prefer that we only move forward with changes to

```go
filterDial            bool
```

by turning it into

```go
filterDial            func(network, addr string) bool
```


### Reason:
In most of the OTEL SDKs I’ve used, filters are exposed as functions that receive all available parameters and return a bool, letting users decide based on full context. For example, both [`otelgin`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/33922e3723d4f1a2b62bd34b1a3e1bfece478983/instrumentation/github.com/gin-gonic/gin/otelgin/config.go#L57) and [`otelgrpc`](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/33922e3723d4f1a2b62bd34b1a3e1bfece478983/instrumentation/google.golang.org/grpc/otelgrpc/config.go#L30) follow this pattern.
​

Adopting the same shape here would keep this filter consistent with other OTEL SDKs. 

What do you think?

PR #3568, opened 2025-10-26
Return error instead of panic from commands

dragneelfps:
Instead of panic in few commands, we can return an error to avoid unexpected panics in application code.

Resolves https://github.com/redis/go-redis/issues/2834

### DIFF
--- a/bitmap_commands.go
+++ b/bitmap_commands.go
@@ -141,7 +141,9 @@ func (c cmdable) BitPos(ctx context.Context, key string, bit int64, pos ...int64
 		args[3] = pos[0]
 		args[4] = pos[1]
 	default:
-		panic("too many arguments")
+		cmd := NewIntCmd(ctx)
+		cmd.SetErr(errors.New("too many arguments"))
+		return cmd
 	}
 	cmd := NewIntCmd(ctx, args...)
 	_ = c(ctx, cmd)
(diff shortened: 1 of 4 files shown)


dragneelfps:
There is also [ClientSetInfo command](https://github.com/redis/go-redis/blob/a3a369b2f5bb7cde200af748e548bfef04481d56/commands.go#L318), but I was not sure if returning a error instead will mess up existing app logic using the pipelines. Happy to to take discuss on how to handle it.

ndyakov:
Hello @dragneelfps and thank you for your contribution! This was on our calendar for a while, glad you were able to address it. As for `ClientSetInfo`, this is validating the input of the command, and since this command is used when initialising the connection, if the info is not correct, the developer should fix it in the application. In my mind the initialisation of the client is the only place where `panic` makes sense and this is part of it. What do you think?

dragneelfps:
@ndyakov Honestly, I would prefer if go-redis can avoid panics altogether. Most clients usages don't handle panic well in my experience. Also, client initialization doesn't always need to be done in the initial app startup. For example, there could be scenarios where client is initialized dynamically(maybe on API request in case of server). There most servers don't expect and handle panics. Moreover, the ClientSetInfo already returns a response(StatusCmd) which can embed errors and I would assume the users will expect error there only, not a panic in case they provided invalid configuration or something.

In short, as a user of the library, I would prefer an error in the response, instead of panic even for ClientSetInfo.

By the way, this PR doesn't include any changes to ClientSetInfo so we can further discuss on it separately and have a future PR handle it. What do you think?

ndyakov:
@dragneelfps yes, let's merge this PR and we can continue a discussion in an issue, feel free to open one.

ndyakov (approved):
looks good to me, commands should not panic

PR #3573, opened 2025-10-29
feat(command): Add hybrid search command

htemelski-oss:
Added support for the new ft.hybrid command.
Test are adapted version of the redis-py tests https://github.com/redis/redis-py/pull/3813

### DIFF
--- a/string_commands.go
+++ b/string_commands.go
@@ -85,6 +85,9 @@ type DelExArgs struct {
 // Compare-and-delete with flexible conditions.
 //
 // Returns the number of keys that were removed (0 or 1).
+//
+// NOTE DelExArgs is still experimental
+// it's signature and behaviour may change
 func (c cmdable) DelExArgs(ctx context.Context, key string, a DelExArgs) *IntCmd {
 	args := []interface{}{"delex", key}
 
@@ -118,6 +121,9 @@ func (c cmdable) DelExArgs(ctx context.Context, key string, a DelExArgs) *IntCmd
 // example/digest-optimistic-locking/
 //
 // Redis 8.4+. See https://redis.io/commands/digest/
+//
+// NOTE Digest is still experimental
+// it's signature and behaviour may change
 func (c cmdable) Digest(ctx context.Context, key string) *DigestCmd {
 	cmd := NewDigestCmd(ctx, "digest", key)
 	_ = c(ctx, cmd)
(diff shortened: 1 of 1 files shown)


search_commands.go
```diff
@@ -344,6 +346,85 @@ type FTSearchOptions struct {
...
+}
+
+// FTHybridWithCursor represents cursor configuration for hybrid search
+type FTHybridWithCursor struct {
+	Count   int // Number of results to return per cursor read
+	MaxIdle int // Maximum idle time in milliseconds before cursor is automatically deleted
+}
+
+// FTHybridOptions hold options that can be passed to the FT.HYBRID command
+type FTHybridOptions struct {
```
ndyakov, line 416:
```suggestion
type FTHybridArgs struct {
```

htemelski-oss: For FTSearch the struct is called `FTSearchOptions` so lets keep them consistent

search_commands.go
```diff
@@ -2191,3 +2473,211 @@ func (c cmdable) FTTagVals(ctx context.Context, index string, field string) *Str
...
+			}
+		}
+
+		// Add LOAD (projected fields)
+		if len(options.Load) > 0 {
+			args = append(args, "LOAD", len(options.Load))
+			for _, field := range options.Load {
+				// Redis requires field names in LOAD to be prefixed with '@' (or '$' for JSON paths).
+				// Tests pass plain field names (e.g. "description"), so add the '@' prefix here.
+				args = append(args, "@"+field)
```
ndyakov: shouldn't the user decide if he/she would like to use plain fields or json paths? hence we should not prefix, but the user should pass `@field` or `$path.name` ?

search_commands.go
```diff
@@ -2191,3 +2473,211 @@ func (c cmdable) FTTagVals(ctx context.Context, index string, field string) *Str
...
+
+		// Add SORTBY
+		if len(options.SortBy) > 0 {
+			sortByOptions := []interface{}{}
+			for _, sortBy := range options.SortBy {
+				// Redis requires field names in SORTBY to be prefixed with '@' (or '$' for JSON paths)
+				fieldName := sortBy.FieldName
+				if fieldName != "" && fieldName[0] != '@' && fieldName[0] != '$' {
+					fieldName = "@" + fieldName
+				}
```
ndyakov: similar here, we are checking if there is a prefix, we can do the same for load. yes, it makes some sense to autoprefix `@` if there is no prefix, but this is debatable. how are other clients handling this?

search_commands.go
```diff
@@ -344,6 +346,85 @@ type FTSearchOptions struct {
...
+	Scorer       string
+	ScorerParams []interface{}
+	YieldScoreAs string
+}
+
+// FTHybridVectorExpression represents a vector expression in hybrid search
+type FTHybridVectorExpression struct {
+	VectorField  string
+	VectorData   Vector
+	Method       string // KNN or RANGE
```
ndyakov: should we have those `KNN` and `RANGE` defined?

ndyakov (dismissed):
left couple comments, other than that lgtm.

search_commands.go
htemelski-oss: The other libraries just pass the args directly, without ensuring that there's a prefix. 
But considering that the feature is still experimental and the autoprefixing adds a bit of ease-of-use, I think it would be better to keep it for now
We can always remove it if we decide

PR #3578, opened 2025-11-02
Add support for XReadGroup CLAIM argument

ofekshenawa:
This PR implements support for the new CLAIM argument in the XREADGROUP command, introduced in Redis 8.4+. This feature allows developers to consume both incoming entries and idle pending entries from streams in a single XREADGROUP call, simplifying stream processing with consumer groups.


ndyakov (approved):
looks good to me

PR #3580, opened 2025-11-03
Add support for MSetEX command

ofekshenawa:
Implements the MSETEX command for go-redis, which atomically sets multiple key-value pairs with expiration options and conditional setting modes. This command is an extension of MSETNX that adds expiration and XX options.


ndyakov (approved):

ndyakov (approved):
Looks good.

PR #3583, opened 2025-11-05
feat(cmd): Add CAS/CAD commands

ndyakov:
Adding CAS/CAD commands available in redis 8.4:
```
SET key value [NX | XX
  | IFEQ match-value | IFNE match-value
  | IFDEQ match-digest | IFDNE match-digest] [GET]
# return changes:
# - if GET specified, returns previous value (regardless of IFEQ etc)
# - if GET not specified, nil reply if operation aborted due to IFEQ etc
DELEX key [IFEQ match-value | IFNE match-value
  | IFDEQ match-digest | IFDNE match-digest]
# returns integer of the number of keys that were removed (so: 0 or 1)
DIGEST key
# returns hex of the specified key (pending "null if not exists?")
```

I did change the methods and added `*Get` methods for each `SetIF*` method, since the responses mean different things (i.e. Set will return status `OK` but if it is  `set ... ifneq .. get`  it still can return OK if the previous value of the key was `OK`. This way it is more clear, the returned commands are StatusCmd or StringCmd and they both mean different things (should) to the user.
What is still a `StatusCmd` that can actually hold the previous value, is the return of `SetArgs` if we have `Get=True` in the args. Any feedback on how to handle this better is welcome.

### DIFF
--- a/digest_test.go
+++ b/digest_test.go
@@ -2,13 +2,35 @@ package redis_test
 
 import (
 	"context"
+	"os"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/redis/go-redis/v9"
 )
 
+func init() {
+	// Initialize RedisVersion from environment variable for regular Go tests
+	// (Ginkgo tests initialize this in BeforeSuite)
+	if version := os.Getenv("REDIS_VERSION"); version != "" {
+		if v, err := strconv.ParseFloat(strings.Trim(version, "\""), 64); err == nil && v > 0 {
+			RedisVersion = v
+		}
+	}
+}
+
+// skipIfRedisBelow84 checks if Redis version is below 8.4 and skips the test if so
+func skipIfRedisBelow84(t *testing.T) {
+	if RedisVersion < 8.4 {
+		t.Skipf("Skipping test: Redis version %.1f < 8.4 (DIGEST command requires Redis 8.4+)", RedisVersion)
+	}
+}
+
 // TestDigestBasic validates that the Digest command returns a uint64 value
 func TestDigestBasic(t *testing.T) {
+	skipIfRedisBelow84(t)
+
 	ctx := context.Background()
 	client := redis.NewClient(&redis.Options{
 		Addr: "localhost:6379",
@@ -51,6 +73,8 @@ func TestDigestBasic(t *testing.T) {
 
 // TestSetIFDEQWithDigest validates the SetIFDEQ command works with digests
 func TestSetIFDEQWithDigest(t *testing.T) {
+	skipIfRedisBelow84(t)
+
 	ctx := context.Background()
 	client := redis.NewClient(&redis.Options{
 		Addr: "localhost:6379",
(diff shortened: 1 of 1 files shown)


string_commands.go
```diff
@@ -25,6 +28,14 @@ type StringCmdable interface {
...
 	SetArgs(ctx context.Context, key string, value interface{}, a SetArgs) *StatusCmd
 	SetEx(ctx context.Context, key string, value interface{}, expiration time.Duration) *StatusCmd
+	SetIFEQ(ctx context.Context, key string, value interface{}, matchValue string, expiration time.Duration) *StatusCmd
+	SetIFEQGet(ctx context.Context, key string, value interface{}, matchValue string, expiration time.Duration) *StringCmd
+	SetIFNE(ctx context.Context, key string, value interface{}, matchValue string, expiration time.Duration) *StatusCmd
+	SetIFNEGet(ctx context.Context, key string, value interface{}, matchValue string, expiration time.Duration) *StringCmd
+	SetIFDEQ(ctx context.Context, key string, value interface{}, matchDigest string, expiration time.Duration) *StatusCmd
+	SetIFDEQGet(ctx context.Context, key string, value interface{}, matchDigest string, expiration time.Duration) *StringCmd
+	SetIFDNE(ctx context.Context, key string, value interface{}, matchDigest string, expiration time.Duration) *StatusCmd
+	SetIFDNEGet(ctx context.Context, key string, value interface{}, matchDigest string, expiration time.Duration) *StringCmd
```
ndyakov: maybe the matchValue should be `interface{}`

ndyakov:
I am considering changing the digest to uint64 to match https://github.com/zeebo/xxh3 and add documentation to suggest using https://github.com/zeebo/xxh3 for generating digests on the application level.

digest_test.go
```diff
@@ -0,0 +1,265 @@
...
+	if version := os.Getenv("REDIS_VERSION"); version != "" {
+		if v, err := strconv.ParseFloat(strings.Trim(version, "\""), 64); err == nil && v > 0 {
+			RedisVersion = v
+		}
+	}
+}
+
+// skipIfRedisBelow84 checks if Redis version is below 8.4 and skips the test if so
+func skipIfRedisBelow84(t *testing.T) {
+	if RedisVersion < 8.4 {
```
ofekshenawa, line 25: It looks like the Redis version in 8.4-RC1-pre.2 is 8.3.224, so maybe consider changing it

commands_test.go
```diff
@@ -1796,6 +1796,200 @@ var _ = Describe("Commands", func() {
 			Expect(get.Err()).To(Equal(redis.Nil))
 		})
 
+		It("should DelExArgs when value matches", func() {
+			SkipBeforeRedisVersion(8.4, "CAS/CAD commands require Redis >= 8.4")
```
ofekshenawa, line 1800: Same here, It looks like the Redis version in 8.4-RC1-pre.2 is 8.3.224, so maybe consider changing it

ndyakov: Yeah, I have observed that the reported version is 8.3.xxx before the GA, but the version we are testing here is what we are telling (from the workflow) that the redis version is. So it works the way it is.

digest_test.go
ndyakov: This `RedisVersion` is going to be populated from the env variable, so at the end it depends what the `workflow` things the version is, not what it is reported by redis. But it sounds like a good suggestion for improvement, to use the redis version reported from the server to populate the env.... anyway, let's keep it 8.4 for now as we are passing this as 8.4.

PR #3585, opened 2025-11-05
feat: Add support for certain slowlog commands

destinyoooo:
- add SLOWLOG{LEN, RESET} commands
- add slowlog tests


destinyoooo:
the same as https://github.com/redis/go-redis/pull/3584#issuecomment-3491543205, failed with redis-ee, but local test is ok

ndyakov:
@destinyoooo maybe we can try to figure out what are the other two slow commands reported? In general, you can skip the test for redis enterprise if you would wish, see my comment to the other PR.

PR #3589, opened 2025-11-07
pubsub error handling

17neverends:
hi! I noticed strange error ignore in pubsub section. i think we need to change method signature for (*PubSub, error) for move error on top call stack

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -1166,7 +1166,11 @@ func (c *Client) pubSub() *PubSub {
 func (c *Client) Subscribe(ctx context.Context, channels ...string) *PubSub {
 	pubsub := c.pubSub()
 	if len(channels) > 0 {
-		_ = pubsub.Subscribe(ctx, channels...)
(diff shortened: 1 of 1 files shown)


ndyakov:
Good catch @17neverends , sadly we cannot change the signature right now (it will be a breaking change). We should track this and address it in `v10`. If you would like, you can open an issue and ping me to triage it.

p.s. feel free to fix the build and we can review / approve / merge the pr.

17neverends:
> Good catch @17neverends , sadly we cannot change the signature right now (it will be a breaking change). We should track this and address it in `v10`. If you would like, you can open an issue and ping me to triage it.
> 
> 
> 
> p.s. feel free to fix the build and we can review / approve / merge the pr.

ok, thank for feedback! i didn't understand what it meant "fix the build"... you mean failed benchmarks?

ndyakov:
Yes, by "the build," I’m referring to the workflows defined in the repository.

PR #3602, opened 2025-11-18
feat(errors): Introduce typed errors

ndyakov:
This PR introduces typed errors for the library, utilising errors.As for error handling instead of the string checks that were used before. This way, errors can be wrapped and set to the command in hooks without breaking the library functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -527,28 +527,33 @@ func (h PipelineLoggingHook) ProcessHook(next redis.ProcessHook) redis.ProcessHo
 func (h PipelineLoggingHook) ProcessPipelineHook(next redis.ProcessPipelineHook) redis.ProcessPipelineHook {
     return func(ctx context.Context, cmds []redis.Cmder) error {
         start := time.Now()
-        err := next(ctx, cmds)
-        duration := time.Since(start)
 
-        // Log pipeline execution
+        // Execute the pipeline
+        err := next(ctx, cmds)
+
+        duration := time.Since(start)
         log.Printf("Pipeline executed %d commands in %v", len(cmds), duration)
 
-        // Check for errors in individual commands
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -429,6 +429,76 @@ vals, err := rdb.Eval(ctx, "return {KEYS[1],ARGV[1]}", []string{"key"}, "hello")
...
+            // Wrap with custom error type
+            wrappedErr := &AppError{
+                Code:      "REDIS_ERROR",
+                RequestID: getRequestID(ctx),
+                Err:       err,
+            }
+            cmd.SetErr(wrappedErr)
+            return wrappedErr
+        }
+        return nil
```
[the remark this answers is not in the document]
ndyakov: You **MUST** return the wrapped error from your hook - if you return the original error, the client will call cmd.SetErr() again and overwrite your wrapped error, making it lost. The pattern is: cmd.SetErr(wrappedErr); return wrappedErr; - both calls must use the same wrapped error.

htemelski-redis (approved):
Seems like a great and much needed addition
Well done

internal/proto/redis_errors.go
```diff
@@ -198,6 +198,11 @@ func IsLoadingError(err error) bool {
 	if errors.As(err, &loadingErr) {
 		return true
 	}
+	// Check if wrapped error is a RedisError with LOADING prefix
+	var redisErr RedisError
+	if errors.As(err, &redisErr) && strings.HasPrefix(string(redisErr), "LOADING ") {
```
htemelski-oss: can we use `redisErr.Error()` instead of casting it to string

ndyakov: done

PR #3611, opened 2025-11-24
fix(time): remove cached time optimization

ndyakov:
We had a report for higher cpu usage with 9.17 that I assume is related to the constant waking up of the background goroutine to update the cached time. This PR is opened to continue the investigation of the issue and provide a potential fix. If this "optimization" of the hot path is found to be the underlying issue, it may be implemented under a configuration option that can be enabled only from people that are aware of the potential tradeoffs (higher idle cpu usage for slightly better throughput)

Potentially fixes https://github.com/redis/go-redis/issues/3608

### DIFF
--- a/scripts/tag.sh
+++ b/scripts/tag.sh
@@ -4,9 +4,25 @@ set -e
 
 DRY_RUN=1
 
+helps() {
+    cat <<- EOF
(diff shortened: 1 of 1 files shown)


ndyakov:
@ofekshenawa feel free to check the discussion in the linked issue.

PR #3615, opened 2025-11-24
add wait to keyless commands list

marcoferrer:
Proposed fix for https://github.com/redis/go-redis/issues/3607


ndyakov:
Hello @marcoferrer , this does fix the specific issue and I am willing to include it as it is. we can investigate if there are other commands missing. 
Thank you for opening this PR.

PR #3618, opened 2025-11-25
feat: optional logger

ndyakov:
Continue the work on https://github.com/redis/go-redis/pull/3560 opened by @ccoVeille

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       # First step: Handle regular issues (excluding needs-information)
       - name: Mark regular issues as stale
-        uses: actions/stale@v9
+        uses: actions/stale@v10
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
@@ -64,7 +64,7 @@ jobs:
 
       # Second step: Handle needs-information issues with accelerated timeline
       - name: Mark needs-information issues as stale
-        uses: actions/stale@v9
+        uses: actions/stale@v10
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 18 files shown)


logging/custom.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	cl.logger.ErrorContext(ctx, msg, args...)
+}
+
+func (cl *LoggerWrapper) Errorf(ctx context.Context, format string, v ...any) {
+	if cl == nil || cl.logger == nil {
+		legacyLoggerWithLevel.Errorf(ctx, format, v...)
+		return
+	}
+	cl.logger.ErrorContext(ctx, format, v...)
```
wzy9607, line 57: Use `ErrorContext` etc. will cause inaccurate source log attr.

Should use `cl.logger.Handler().Handle`instead when wrapping, see slog godoc [Example (Wrapping)](https://pkg.go.dev/log/slog#example-package-Wrapping).

ndyakov: @ccoVeille Maybe you would like to check this.

@wzy9607 thank you for pointing this out

ccoVeille: I feel like there is an error here, yes

it should be

```go
	cl.logger.ErrorContext(cl.printfToStructured(ctx, format, v...))
```

I will fix it.

I created the printToStructured to be able to handle the right format

ccoVeille: I have fixed the issue in https://github.com/redis/go-redis/pull/3622

ccoVeille:
I feel like this PR can be closed now, as https://github.com/redis/go-redis/pull/3622 replaces it

ndyakov:
closed in favour of https://github.com/redis/go-redis/pull/3622

logging/custom.go
ndyakov: maybe the fix was lost when you force pushed? Can we just add a test with slog that it will work as expected?

PR #3622, opened 2025-11-27
feat: add optional logger wherever possible

### LINKED ISSUE #3558: Feature Request: ability to provide a logger as an Options setting

ccoVeille:
This commit introduces an optional logger parameter to various structs.
This enhancement allows users to provide custom logging implementations.

Fixes #3558

- #3558

> [!NOTE]
> This is a follow-up of #3560 that was merged in a branch (not `master`) and #3618 that was updated by @ndyakov to add changes on it.
>
> The MR is now ready to be merged.

- #3560 
- #3618

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -58,7 +58,7 @@ func ExampleParseURL() {
 }
 
 func ExampleNewFailoverClient() {
-	// See http://redis.io/topics/sentinel for instructions how to
+	// See https://redis.io/docs/latest/operate/oss_and_stack/management/sentinel for instructions how to
 	// setup Redis Sentinel.
 	rdb := redis.NewFailoverClient(&redis.FailoverOptions{
 		MasterName:    "master",
@@ -68,7 +68,7 @@ func ExampleNewFailoverClient() {
 }
 
 func ExampleNewClusterClient() {
-	// See http://redis.io/topics/cluster-tutorial for instructions
+	// See https://redis.io/docs/latest/operate/oss_and_stack/management/scaling for instructions
 	// how to setup Redis Cluster.
 	rdb := redis.NewClusterClient(&redis.ClusterOptions{
 		Addrs: []string{":7000", ":7001", ":7002", ":7003", ":7004", ":7005"},
--- a/internal/hashtag/hashtag.go
+++ b/internal/hashtag/hashtag.go
@@ -11,7 +11,7 @@ const slotNumber = 16384
 // CRC16 implementation according to CCITT standards.
 // Copyright 2001-2010 Georges Menie (www.menie.org)
 // Copyright 2013 The Go Authors. All rights reserved.
-// http://redis.io/topics/cluster-spec#appendix-a-crc16-reference-implementation-in-ansi-c
+// https://redis.io/docs/latest/operate/oss_and_stack/reference/cluster-spec#appendix-a-crc16-reference-implementation-in-ansi-c.
 var crc16tab = [256]uint16{
 	0x0000, 0x1021, 0x2042, 0x3063, 0x4084, 0x50a5, 0x60c6, 0x70e7,
 	0x8108, 0x9129, 0xa14a, 0xb16b, 0xc18c, 0xd1ad, 0xe1ce, 0xf1ef,
--- a/internal/hashtag/hashtag_test.go
+++ b/internal/hashtag/hashtag_test.go
@@ -15,7 +15,7 @@ func TestGinkgoSuite(t *testing.T) {
 }
 
 var _ = Describe("CRC16", func() {
-	// http://redis.io/topics/cluster-spec#keys-distribution-model
+	// https://redis.io/docs/latest/operate/oss_and_stack/reference/cluster-spec#key-distribution-model.
 	It("should calculate CRC16", func() {
 		tests := []struct {
 			s string
--- a/osscluster.go
+++ b/osscluster.go
@@ -1152,7 +1152,7 @@ type ClusterClient struct {
 }
 
 // NewClusterClient returns a Redis Cluster client as described in
-// http://redis.io/topics/cluster-spec.
+// https://redis.io/docs/latest/operate/oss_and_stack/reference/cluster-spec.
 func NewClusterClient(opt *ClusterOptions) *ClusterClient {
 	opt.init()
 
--- a/pipeline.go
+++ b/pipeline.go
@@ -49,7 +49,7 @@ type Pipeliner interface {
 var _ Pipeliner = (*Pipeline)(nil)
 
 // Pipeline implements pipelining as described in
-// http://redis.io/topics/pipelining.
+// https://redis.io/docs/latest/develop/using-commands/pipelining.
 // Please note: it is not safe for concurrent use by multiple goroutines.
 type Pipeline struct {
 	cmdable
--- a/pubsub.go
+++ b/pubsub.go
@@ -16,7 +16,7 @@ import (
 )
 
 // PubSub implements Pub/Sub commands as described in
-// http://redis.io/topics/pubsub. Message receiving is NOT safe
+// https://redis.io/docs/latest/develop/pubsub. Message receiving is NOT safe
 // for concurrent use by multiple goroutines.
 //
 // PubSub automatically reconnects to Redis Server and resubscribes
--- a/tx.go
+++ b/tx.go
@@ -11,7 +11,7 @@ import (
 const TxFailedErr = proto.RedisError("redis: transaction failed")
 
 // Tx implements Redis transactions as described in
-// http://redis.io/topics/transactions. It's NOT safe for concurrent use
+// https://redis.io/docs/latest/develop/using-commands/transactions. It's NOT safe for concurrent use
 // by multiple goroutines, because Exec resets list of watched keys.
 //
 // If you don't need WATCH, use Pipeline instead.


ccoVeille:
I apparently failed to keep your GPG signature @ndyakov, it sounds logic as I used history rewritting to fix an issue in one of the commit I did earlier. Feel free to pull, sign your commits, and push again. My fork is now all set for maintainers edits

logging/custom.go
```diff
@@ -0,0 +1,142 @@
...
+	"context"
+	"fmt"
+)
+
+// LoggerWrapper is a slog.Logger wrapper that implements the Lgr interface.
+type LoggerWrapper struct {
+	logger        LoggerWithLevelI
+	loggerLevel   *LogLevelT
+	printfAdapter PrintfAdapter
+}
```
ccoVeille, line 13: Side note about wording

Does it make sense to have something that would be referred as "logging.LoggerWrapper" outside the package

What do you think about `logging.Wrapper` ?


```suggestion
// Wrapper is a slog.Logger wrapper that implements the Lgr interface.
type Wrapper struct {
	logger        LoggerWithLevelI
	loggerLevel   *LogLevelT
	printfAdapter PrintfAdapter
}
```

logging/custom.go
```diff
@@ -0,0 +1,142 @@
...
+	cl := &LoggerWrapper{
+		logger: logger,
+	}
+	for _, opt := range opts {
+		opt(cl)
+	}
+	return cl
+}
+
+type LoggerWrapperOption func(*LoggerWrapper)
```
ccoVeille, line 28: same question

https://github.com/redis/go-redis/pull/3622/files#r2568588557

it could be


```suggestion
type WrapperOption func(*Wrapper)
```

logging/logging.go
```diff
@@ -89,3 +89,22 @@ func (l *filterLogger) Printf(ctx context.Context, format string, v ...interface
...
+// Lgr is a logger interface with leveled logging methods.
+// It is implemented by LoggerWrapper and legacyLoggerAdapter.
+// If you would like to use `slog.Logger` from the standard library, you can use LoggerWrapper.
+//
+//		logger := slog.New(slog.NewTextHandler(os.Stderr))
+//	    db := redis.NewClient(&redis.Options{
+//			Logger: logging.NewLoggerWrapper(logger),
+//		})
+//
+// This will NOT handle all logging at the moment, singe there is still a global logger in use.
```
ccoVeille: typo spotted

```suggestion
// This will NOT handle all logging at the moment, since there is still a global logger in use.
```

logging/logging.go
```diff
@@ -89,3 +89,22 @@ func (l *filterLogger) Printf(ctx context.Context, format string, v ...interface
...
+// that will be deprecated in the future.
+type Lgr interface {
+	Errorf(ctx context.Context, format string, v ...any)
+	Warnf(ctx context.Context, format string, v ...any)
+	Infof(ctx context.Context, format string, v ...any)
+	Debugf(ctx context.Context, format string, v ...any)
+	Enabled(ctx context.Context, level LogLevelT) bool
+}
```
ccoVeille: The name of this interface is a bit awkward to me.

ccoVeille: There is something I don't get here with this interface

Either `LegacyLogger` and `LoggerWrapper` also support

```go
Error(ctx context.Context, msg string, v ...any)
Warn(ctx context.Context, msg string, v ...any)
Info(ctx context.Context, msg string, v ...any)
Debug(ctx context.Context, msg string, v ...any)
```

Why do you restrict the interface to simili printf logger ?

logging/custom.go
ndyakov: i thought about this, but defaulted to LoggerWrapper, since in the `logging` package, a `Wrapper` may refer to both logger and a log message itself.

logging/custom.go
ndyakov: same answer :)

logging/logging.go
ndyakov: I am open to suggestions.

ndyakov: those were the only one used, i don't mind adding the non format methods. feel free to add them.

ndyakov (dismissed):
Overall looks good to me, can we just make sure this one is addressed (preferably add a simple test):
https://github.com/redis/go-redis/pull/3618#discussion_r2565332445

ccoVeille:
> Overall looks good to me, can we just make sure this one is addressed (preferably add a simple test):
> https://github.com/redis/go-redis/pull/3618#discussion_r2565332445

Here the difference I applied to an old commit.

```diff
func (cl *LoggerWrapper) Errorf(ctx context.Context, format string, v ...any) { 
         if cl == nil || cl.logger == nil { 
                 legacyLoggerWithLevel.Errorf(ctx, format, v...) 
                 return 
         } 
-         cl.logger.ErrorContext(ctx, format, v...)
+         cl.logger.ErrorContext(cl.printfToStructured(ctx, format, v...))
 }
```

I can add a test, but adding a test about a logger is never fun.

Maybe it can be done in a separate PR. I'm suggesting that, but if you are OK to wait for me to add the tests a few more days, maybe a week. I can do it.

logging/logging.go
ccoVeille: In fact, I feel like we should consider inverting the logic here.

The custom logger available in the various config should only mention

```go
Error(ctx context.Context, msg string, v ...any)
Warn(ctx context.Context, msg string, v ...any)
Info(ctx context.Context, msg string, v ...any)
Debug(ctx context.Context, msg string, v ...any)
Enabled(ctx context.Context, level LogLevelT) bool
```

So no `Whateverf` involved, then the code use the wrapper to transform things.

I feel like asking people to implement Errorf, Infof and co is an error.

ndyakov:
@ccoVeille No problem, I will prepare a beta without the logger, you can write tests and resolve conflicts in the next week and then we will include it.

logging/logging.go
ndyakov: The point of those logger methods is to print a formatted log message, so `*f` makes sense to me and it is consistent. Users don't have to implement those methods, they do however have to wrap their logger with the provided wrapper if their logger follows the `*Context` interface like `slog`.

ccoVeille: I wasn't suggesting to rename things such as:

Infof -> Info

My problem was the fact people will have to add a struct that would satisfy the legacy Printf methods

While we could add this in a way the logger only have to satistify the method that would be used by a structured logger

So the InfoContext one and so on.

The code could adapt itself to use the LoggerWrapper if needed.

Said otherwise people could provide directly a slog.Handler with the v9 as the logger of each config.

The logger wrapper is something we need right now because the code needs the Infof

But I would prefer the magic to be hidden from user. People seeing the interface would think "oh I need to implement Errorf, Warnf and co it's boring and not convenient"

I will add a commit to explain what I have in mind. But later, either this weekend or next week. I will also add tests for all that.

ccoVeille: I have refactored in the way I found the best for future maintainability

ccoVeille:
@ndyakov please review again.

I feel the system is better now.

~~The only thing I'm unsure is the fact the LoggerWrapper should be exported (like in the current PR) or moved to internal/logging~~

EDIT: this is needed if a LoggerWrapper is provided directly

ccoVeille:
Apparently, I broke something.

It's too late here, but the code is close to what is needed.

logging/custom.go
```diff
@@ -0,0 +1,146 @@
+package logging
+
+import (
+	"context"
+	"fmt"
+)
+
+// LoggerWrapper is a slog.Logger wrapper that implements the Lgr interface.
```
ccoVeille, line 8:
```suggestion
// LoggerWrapper is a wrapper for logger, it provides compatibility with legacy Printf logger and new structured logger
```

logging/custom.go
```diff
@@ -0,0 +1,146 @@
...
+type LoggerWrapper struct {
+	logger        LoggerWithLevelI
+	loggerLevel   *LogLevelT
+	printfAdapter PrintfAdapter
+}
+
+func NewLoggerWrapper(logger LoggerWithLevelI, opts ...LoggerWrapperOption) *LoggerWrapper {
+	cl := &LoggerWrapper{
+		logger: logger,
+	}
```
ccoVeille: Check if logger is already a LoggerWrapper, this way it can be used directly

ccoVeille:
I feel like I have to move out the change about fixing the panic and sprintf issue

ccoVeille:
> I feel like I have to move out the change about fixing the panic and sprintf issue

Moved to 
- #3632
- #3633

ccoVeille:
I squashed everything in one commit even yours @ndyakov.

I was done to deal with conflicts when someone was merging something :roll_eyes:

ccoVeille:
The PR is still in DRAFT mode because I have no clear status on what is the minimal Go version for go-redis

As stated in #3631, right now the `go.mod` mentions Go 1.18+, but the current code can only work with Go 1.23+

- #3631 

Except that the code can be reviewed.

But if the minimal version is moved to something at least Go 1.21, my implementation could be simplified a lot as I would be able to delete everything about `//go:build` tags

ccoVeille:
Thanks for taking back my PR

PR #3625, opened 2025-11-28
fix(pool): double freeturn bug

ndyakov:
A regression was reported where with aggressive timeout configuration the pool was stuck. Traced back to https://github.com/redis/go-redis/pull/3518

This PR fixes #3619 and #3623

PR #3626, opened 2025-11-29
fix(pool): correct turn management in putIdleConn to prevent connection leaks

cyningsun:
### Bug Description  
A critical defect was identified in the connection pool's turn management mechanism within the `putIdleConn` method, breaking the fundamental 1:1 correspondence between turns and connections. ( Introduced by #3518)

### Root Cause Analysis  

**Turn Release Scenarios (Expected Behavior):**
```
- new connection failed
  → freeTurn: allow other goroutines to create connections
- pool.Put 
  → freeTurn: allow acquiring existing idle connections  
- pool.putIdleConn
  ├── deliver connection to waiting goroutine
  │   → ❌ DEFECT: turn should be released by receiving goroutine via Put
  └── place connection in idle pool
      → freeTurn: allow acquiring existing idle connections
```

**The Defect:**  In `putIdleConn`, when delivering connections to waiting goroutines, the turn was incorrectly released immediately instead of being transferred to the receiving goroutine for proper release via `Put`.

**Test Gap:** The test case "should not leak turn when delivering connection via putIdleConn" contained flawed assertions due to misunderstanding of turn ownership mechanism in the context of connection delivery, failing to detect the leakage.

**Relationship to #3625:**  
This PR addresses the same turn management issue identified in #3625. I want to acknowledge and thank @ndyakov for:

- The comprehensive test cases that clearly reproduce the issue 
- The prompt response in providing a solution
- The valuable implementation in #3625

This PR offers an additional perspective on the solution, focusing specifically on turn ownership semantics. It's intended to provide more options for review and discussion as we work towards the best resolution for this issue.

**Note to Reviewers:**  
This PR is optional and can be closed if #3625 covers all needs. It's just here to offer another perspective.

### DIFF
--- a/internal/pool/double_freeturn_simple_test.go
+++ b/internal/pool/double_freeturn_simple_test.go
@@ -1,0 +1,158 @@
+package pool_test
+
+import (
+	"context"
+	"net"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/redis/go-redis/v9/internal/pool"
+)
+
+// TestDoubleFreeTurnSimple tests the double-free bug with a simple scenario.
+// This test FAILS with the OLD code and PASSES with the NEW code.
+//
+// Scenario:
+// 1. Request A times out, Dial A completes and delivers connection to Request B
+// 2. Request B's own Dial B completes later
+// 3. With the bug: Dial B frees Request B's turn (even though Request B is using connection A)
+// 4. Then Request B calls Put() and frees the turn AGAIN (double-free)
+// 5. This allows more concurrent operations than PoolSize permits
+//
+// Detection method:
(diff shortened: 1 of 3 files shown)


ndyakov:
I do personally like this approach better, it is easier to follow and read. Thank you @cyningsun. Will let @ofekshenawa decide which one to include. Could you please sync the rest of the changes (the context calculation and the tests).

cyningsun:
@ndyakov, thank you! I'm glad you like the approach. I've gone ahead and synced the context calculation and tests from #3625. All tests are passing.

cyningsun:
@ndyakov, thank you! I'm glad you like the approach. I've gone ahead and synced the context calculation and tests from #3625. All tests are passing.

PR #3632, opened 2025-12-02
fix: bad sprintf syntax

ccoVeille:
There is no Stringer on the channel struct, so using %s causes an error.
Using %v to print the default representation instead.


ccoVeille:
> that seems correct, but i am wondering what information this will give us? Should we think about some identifier for those channels in the future? 

I agree. The current error message was simply broken.

I provided something that would work. But it doesn't mean it would be meaningful.

I feel like it has to be updated to provide a clear error message, but this was out of the scope of my changes. And it would require some discussion

PR #3633, opened 2025-12-02
fix: panic on nil error in handoffWorkerManager closeConnFromRequest

ccoVeille:
Idenfied while working on #3622

### DIFF
--- a/internal/maintnotifications/logs/log_messages.go
+++ b/internal/maintnotifications/logs/log_messages.go
@@ -288,19 +288,29 @@ func OperationNotTracked(connID uint64, seqID int64) string {
 
 // Connection pool functions
 func RemovingConnectionFromPool(connID uint64, reason error) string {
-	message := fmt.Sprintf("conn[%d] %s due to: %v", connID, RemovingConnectionFromPoolMessage, reason)
-	return appendJSONIfDebug(message, map[string]interface{}{
+	metadata := map[string]interface{}{
 		"connID": connID,
-		"reason": reason.Error(),
-	})
+		"reason": "unknown", // this will be overwritten if reason is not nil
+	}
+	if reason != nil {
+		metadata["reason"] = reason.Error()
+	}
+
+	message := fmt.Sprintf("conn[%d] %s due to: %v", connID, RemovingConnectionFromPoolMessage, reason)
+	return appendJSONIfDebug(message, metadata)
 }
 
 func NoPoolProvidedCannotRemove(connID uint64, reason error) string {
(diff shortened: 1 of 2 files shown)


internal/maintnotifications/logs/log_messages.go
```diff
@@ -295,11 +295,10 @@ func RemovingConnectionFromPool(connID uint64, reason error) string {
 	})
 }
 
-func NoPoolProvidedCannotRemove(connID uint64, reason error) string {
-	message := fmt.Sprintf("conn[%d] %s due to: %v", connID, NoPoolProvidedMessageCannotRemoveMessage, reason)
+func NoPoolProvidedCannotRemove(connID uint64) string {
+	message := fmt.Sprintf("conn[%d] %s", connID, NoPoolProvidedMessageCannotRemoveMessage)
 	return appendJSONIfDebug(message, map[string]interface{}{
 		"connID": connID,
-		"reason": reason.Error(),
```
ndyakov, line 302: can we support the `nil` but at the same time keep reporting the `reason` if there is one?

maintnotifications/handoff_worker.go
```diff
@@ -506,7 +506,7 @@ func (hwm *handoffWorkerManager) closeConnFromRequest(ctx context.Context, reque
 			internal.Logger.Printf(ctx, "redis: failed to close connection: %v", err)
 		}
 		if internal.LogLevel.WarnOrAbove() {
-			internal.Logger.Printf(ctx, logs.NoPoolProvidedCannotRemove(conn.GetID(), err))
```
ndyakov, line 509: lets keep the `err` but support `nil` when logging.

ccoVeille: The error we might log is about the closed connection ?

Shouldn't we log the warning only when there is no error?

I mean the code could return on error, and then the error is not passed to the NoPoolProvidedCannotRemove

I can do the change you request, but I would like to challenge you on the need to log something that sounds unrelated to the warning message ?

I might be wrong. I don't mind 😁

Just tell me.

ndyakov: actually, you are not wrong, the err is logged as part of the previous log message, that should be completly fine.

ndyakov (dismissed):

maintnotifications/handoff_worker.go
ccoVeille: No you are right, there was a problem in the code, but not with the `err` variable I thought.

I will fix it: closing the connection shadows the err variable provided as argument

ccoVeille:
@ndyakov please review again, I feel like I have fixed the root cause.

I have also applied a safeguard against any future issue when dealing with reason.Error() in `internal/maintnotifications/logs`

PR #3637, opened 2025-12-03
Add OpenTelemetry Native Metrics Support

ofekshenawa:
This PR implements Observability (Metrics) for go-redis, adding comprehensive OpenTelemetry metrics support following the [OpenTelemetry Database Client Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/database/database-metrics/).

The implementation uses a Bridge Pattern to keep the core library dependency-free while providing optional,  metrics instrumentation through the new extra/redisotel-native package.

***What's Included:***
Metric groups covering aspects of Redis operations:
- Command metrics: Operation duration with retry tracking
- Connection basic: Connection count and creation time
- Resiliency: Errors, handoffs, timeout relaxation
- Connection advanced: Wait time and use time
- Pubsub metrics: Published and received messages 
- Stream metrics: Processing duration and maintenance notifications


internal/pool/pool.go
```diff
@@ -32,6 +32,43 @@ var (
...
+	// Parameters: ctx, duration, cn
+	connectionUseTimeCallback func(ctx context.Context, duration time.Duration, cn *Conn)
+
+	// Global callback for connection timeouts (set by otel package)
+	// Parameters: ctx, cn, timeoutType
+	connectionTimeoutCallback func(ctx context.Context, cn *Conn, timeoutType string)
+
+	// Global callback for connection closed (set by otel package)
+	// Parameters: ctx, cn, reason, err
+	connectionClosedCallback func(ctx context.Context, cn *Conn, reason string, err error)
```
ndyakov: This global callbacks are not thread safe. Someone may set (or set nil - remove) the callback while a check on its existence is happening. In the best case scenario - we will end up having undefined behaviour, worst case - nil pointer panics. We should fix this with either a global callback mutex or atomic values here. I do think a `RWMutex` will be an easier and cleaner approach.

internal/pool/pool.go
```diff
@@ -419,6 +549,12 @@ func (p *ConnPool) dialConn(ctx context.Context, pooled bool) (*Conn, error) {
 			cn.expiresAt = noExpiration
 		}
 
+		// Record connection creation time
+		if connectionCreateTimeCallback != nil {
+			duration := time.Since(startTime)
+			connectionCreateTimeCallback(ctx, duration, cn)
```
ndyakov: do we want to only record the dial time or the whole connection creation. The connection creation includes the handshake (i assume?) and not just the tcp socket initiation.

internal/pool/pool.go
```diff
@@ -480,9 +616,26 @@ func (p *ConnPool) getConn(ctx context.Context) (*Conn, error) {
...
 	}
 
+	// Track pending requests in pool stats
+	// NOTE: We only track in stats, not via callback. The AsyncGauge reads stats directly.
+	atomic.AddUint32(&p.stats.PendingRequests, 1)
+	defer func() {
+		if err != nil {
+			// Failed to get connection, decrement pending requests
+			atomic.AddUint32(&p.stats.PendingRequests, ^uint32(0)) // -1
+		}
```
ndyakov, line 698: I think this `err` here is not the same as the one in line#631 for example. If we would like to be able to do such check for the err, we should define it in the return types initially (named returns) and the use the same variable - e.g. https://play.golang.com/p/-wC3trYIiDi

internal/pool/pool.go
```diff
@@ -365,10 +487,18 @@ func (p *ConnPool) newConn(ctx context.Context, pooled bool) (*Conn, error) {
...
+		connectionStateChangeCallback(ctx, cn, "", "idle")
+	}
+
 	return cn, nil
 }
 
 func (p *ConnPool) dialConn(ctx context.Context, pooled bool) (*Conn, error) {
+	// Start measuring connection creation time
+	startTime := time.Now()
+
```
ndyakov: we are going to use this only if `connectionCreateTimeCallback != nil`, let's call it only if it is not nil. Also, let's move it after the guards for closed pool and dial errors.

internal/pool/pool.go
```diff
@@ -480,9 +616,26 @@ func (p *ConnPool) getConn(ctx context.Context) (*Conn, error) {
...
+	atomic.AddUint32(&p.stats.PendingRequests, 1)
+	defer func() {
+		if err != nil {
+			// Failed to get connection, decrement pending requests
+			atomic.AddUint32(&p.stats.PendingRequests, ^uint32(0)) // -1
+		}
+	}()
+
+	// Track wait time
+	waitStart := time.Now()
```
ndyakov: can we make some abstraction here. calculate those timers only if we have the callbacks set?

internal/pool/pool.go
```diff
@@ -762,6 +947,14 @@ func (p *ConnPool) putConnWithoutTurn(ctx context.Context, cn *Conn) {
 
 // putConn is the internal implementation of Put that optionally frees a turn.
 func (p *ConnPool) putConn(ctx context.Context, cn *Conn, freeTurn bool) {
+	// Record use time (time between checkout and return)
+	checkoutAtNs := cn.checkoutAt.Load()
+	if checkoutAtNs > 0 && connectionUseTimeCallback != nil {
+		useTime := time.Duration(time.Now().UnixNano() - checkoutAtNs)
+		connectionUseTimeCallback(ctx, useTime, cn)
+		cn.checkoutAt.Store(0) // Reset checkout time
```
ndyakov: instead of checkoutAt. can't we use `UsedAtNs` ?

internal/pool/pool.go
```diff
@@ -876,6 +1079,14 @@ func (p *ConnPool) RemoveWithoutTurn(ctx context.Context, cn *Conn, reason error
...
 // removeConnInternal is the internal implementation of Remove that optionally frees a turn.
 func (p *ConnPool) removeConnInternal(ctx context.Context, cn *Conn, reason error, freeTurn bool) {
+	// Record use time if connection was checked out
+	checkoutAtNs := cn.checkoutAt.Load()
+	if checkoutAtNs > 0 && connectionUseTimeCallback != nil {
+		useTime := time.Duration(time.Now().UnixNano() - checkoutAtNs)
+		connectionUseTimeCallback(ctx, useTime, cn)
+		cn.checkoutAt.Store(0) // Reset checkout time
+	}
+
```
ndyakov: same, can we use `UsedAtNs` ?

ndyakov (commented):
submitting partial review

internal/pool/pool.go
vladvildanov: Does this time includes connection handshake?

internal/pool/pool.go
```diff
@@ -419,6 +549,12 @@ func (p *ConnPool) dialConn(ctx context.Context, pooled bool) (*Conn, error) {
 			cn.expiresAt = noExpiration
 		}
 
+		// Record connection creation time
+		if connectionCreateTimeCallback != nil {
+			duration := time.Since(startTime)
```
vladvildanov: The variable is basically not needed AFAIU, it's not used anywhere anymore

internal/pool/pool.go
```diff
@@ -480,9 +616,26 @@ func (p *ConnPool) getConn(ctx context.Context) (*Conn, error) {
...
+			atomic.AddUint32(&p.stats.PendingRequests, ^uint32(0)) // -1
+		}
+	}()
+
+	// Track wait time
+	waitStart := time.Now()
 	if err := p.waitTurn(ctx); err != nil {
+		// Record timeout if applicable
+		if err == ErrPoolTimeout && connectionTimeoutCallback != nil {
+			connectionTimeoutCallback(ctx, nil, "pool")
```
vladvildanov: Should we also record general error metric here?

redis.go
```diff
@@ -559,6 +560,12 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 				// enabled mode, fail the connection
 				c.optLock.Unlock()
 				cn.GetStateMachine().Transition(pool.StateClosed)
+
+				// Record handshake failure metric
+				if errorCallback := pool.GetErrorCallback(); errorCallback != nil {
+					errorCallback(ctx, "HANDSHAKE_FAILED", cn, "HANDSHAKE_FAILED", true, 0)
```
vladvildanov: What does this HANDSHAKE_FAILED means? We have a defined set of error categories in the HLD

internal/otel/metrics.go
```diff
@@ -0,0 +1,238 @@
...
+	FullName() string
+	Args() []interface{}
+	Err() error
+}
+
+// Recorder is the interface for recording metrics.
+type Recorder interface {
+	// RecordOperationDuration records the total operation duration (including all retries)
+	// dbIndex is the Redis database index (0-15)
+	RecordOperationDuration(ctx context.Context, duration time.Duration, cmd Cmder, attempts int, cn *pool.Conn, dbIndex int)
```
vladvildanov: This one misses error attributes, operation duration should be recorded even if operation failed

maintnotifications/push_notification_handler.go
```diff
@@ -40,14 +40,44 @@ func (snh *NotificationHandler) HandlePushNotification(ctx context.Context, hand
 	var err error
 	switch notificationType {
 	case NotificationMoving:
+		// Record maintenance notification metric
```
vladvildanov: You can do it once after the switch instead of handling each case

internal/pool/pool.go
```diff
@@ -32,6 +32,43 @@ var (
...
+	// Parameters: ctx, delta (+1/-1), cn, poolName, notificationType
+	connectionRelaxedTimeoutCallback func(ctx context.Context, delta int, cn *Conn, poolName, notificationType string)
+
+	// Global callback for connection handoff (set by otel package)
+	// Parameters: ctx, cn, poolName
+	connectionHandoffCallback func(ctx context.Context, cn *Conn, poolName string)
+
+	// Global callback for error tracking (set by otel package)
+	// Parameters: ctx, errorType, cn, statusCode, isInternal, retryAttempts
+	errorCallback func(ctx context.Context, errorType string, cn *Conn, statusCode string, isInternal bool, retryAttempts int)
```
vladvildanov: I see that there's only two places where this metric is exposed. The metric should expose ALL errors in a client including command execution failure.

extra/redisotel-native/config.go
```diff
@@ -0,0 +1,264 @@
...
+func (c *config) isCommandIncluded(command string) bool {
+	if c.excludeCommands != nil && c.excludeCommands[command] {
+		return false
+	}
+
+	if c.includeCommands != nil {
+		return c.includeCommands[command]
+	}
+
+	return true
```
ndyakov, line 67: should we make sure the `command` is lowercase for example?

extra/redisotel-native/config.go
```diff
@@ -0,0 +1,264 @@
...
+
+// WithExcludeCommands sets a command deny-list for metrics.
+// Commands in this list will not have metrics recorded.
+func (c *Config) WithExcludeCommands(commands []string) *Config {
+	c.ExcludeCommands = make(map[string]bool)
+	for _, cmd := range commands {
+		c.ExcludeCommands[cmd] = true
+	}
+	return c
+}
```
ndyakov, line 233: let's make sure the strings are the same case as the ones in `isCommandIncluded`

extra/redisotel-native/redisotel.go
```diff
@@ -0,0 +1,540 @@
...
+// After initialization, all Redis clients will automatically collect and export
+// metrics without needing any additional configuration.
+func (o *ObservabilityInstance) Init(cfg *Config) error {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	// If already initialized, shutdown previous instance
+	if o.initialized {
+		o.shutdownLocked()
+	}
```
ndyakov: why did we decide to reinitialise instead of return error?

otel.go
```diff
@@ -0,0 +1,206 @@
...
+	recorder OTelRecorder
+}
+
+// toConnInfo converts internal pool.Conn to public ConnInfo interface
+func toConnInfo(cn *pool.Conn) ConnInfo {
+	if cn != nil {
+		return cn
+	}
+	return nil
+}
```
ndyakov: This is not needed.

maintnotifications/push_notification_handler.go
```diff
@@ -40,14 +40,44 @@ func (snh *NotificationHandler) HandlePushNotification(ctx context.Context, hand
 	var err error
```
ndyakov, line 41: can we add:
```
		// Record maintenance notification metric
		if maintenanceCallback := pool.GetMaintenanceNotificationCallback(); maintenanceCallback != nil {
			if conn, ok := handlerCtx.Conn.(*pool.Conn); ok {
				maintenanceCallback(ctx, conn, notificationType)
			}
		}
```
		
once, instead in any of the switch/case-es

ndyakov: hm... instead of the case when this is a pubsub...

internal/pool/pool.go
```diff
@@ -51,6 +87,151 @@ var (
...
+	connectionStateChangeCallback = fn
+	callbackMu.Unlock()
+}
+
+func getConnectionStateChangeCallback() func(ctx context.Context, cn *Conn, fromState, toState string) {
+	callbackMu.RLock()
+	cb := connectionStateChangeCallback
+	callbackMu.RUnlock()
+	return cb
+}
```
ndyakov: there is a slight chance that `connectionStateChangeCallback` can be changed after the runlick but before the use. If we agree that this timeframe is something we are willing to tolerate, this is fine.

internal/pool/pool.go
```diff
@@ -51,6 +87,151 @@ var (
...
+
+// SetConnectionTimeoutCallback sets the global callback for connection timeouts.
+// This is called by the otel package to register metrics recording.
+func SetConnectionTimeoutCallback(fn func(ctx context.Context, cn *Conn, timeoutType string)) {
+	callbackMu.Lock()
+	connectionTimeoutCallback = fn
+	callbackMu.Unlock()
+}
```
ndyakov: if we are setting multiple callbacks (as we should in most cases) wouldn't it be better to have a single lock -> set all callbacks -> unlock block for them?

internal/pool/pool.go
```diff
@@ -366,6 +548,11 @@ func (p *ConnPool) newConn(ctx context.Context, pooled bool) (*Conn, error) {
 		}
 	}
 
+	// Notify metrics: new connection created and idle
+	if cb := getConnectionStateChangeCallback(); cb != nil {
+		cb(ctx, cn, "", "idle")
+	}
```
ndyakov, line 549: since those callbacks are only ment to be used for metric collection, should we use some `Metric` prefix in their names? Just to make sure future developers don't try to use them for anything else.

extra/redisotel-native/config.go
```diff
@@ -0,0 +1,261 @@
...
+	bucketsConnectionCreateTime     []float64
+	bucketsConnectionWaitTime       []float64
+}
+
+func (c *config) isMetricGroupEnabled(group MetricGroup) bool {
+	return c.enabledMetricGroups[group]
+}
+
+func (c *config) isCommandIncluded(command string) bool {
+	if c.excludeCommands != nil && c.excludeCommands[command] {
```
ndyakov, line 59:
```suggestion
    command = strings.ToLower(command)
	if c.excludeCommands != nil && c.excludeCommands[command] {
```

extra/redisotel-native/redisotel.go
```diff
@@ -0,0 +1,516 @@
...
+					}
+					if serverAddr != "" {
+						baseAttrs = append(baseAttrs, attribute.String("server.address", serverAddr))
+					}
+					if serverPort != "" && serverPort != "6379" {
+						baseAttrs = append(baseAttrs, attribute.String("server.port", serverPort))
+					}
+
+					// Add pool name
+					baseAttrs = append(baseAttrs, attribute.String("db.client.connection.pool.name", poolInfo.name))
```
elena-kolevska: I'd prefer to see all attributes defined as constants rather than using them as strings across multiple places/files.

extra/redisotel-native/metrics.go
```diff
@@ -0,0 +1,910 @@
...
+		return false
+	}
+
+	// Check for net.Error with Timeout() method (standard way)
+	if netErr, ok := err.(net.Error); ok {
+		return netErr.Timeout()
+	}
+
+	return false
+}
```
ndyakov, line 376: this is the internal implementation that we have for detecting timeout errors:
```
func isTimeoutError(err error) (isTimeout bool, hasTimeoutFlag bool) {
	// Check for timeoutError interface (works with wrapped errors)
	var te timeoutError
	if errors.As(err, &te) {
		return true, te.Timeout()
	}

	// Check for net.Error specifically (common case for network timeouts)
	var netErr net.Error
	if errors.As(err, &netErr) {
		return true, netErr.Timeout()
	}

	return false, false
}
```
It could be other type of error (not only net.Error) that has `Timeout()` method.

extra/redisotel-native/metrics.go
```diff
@@ -0,0 +1,910 @@
...
+	}
+
+	// Check for network errors (transport/DNS/socket issues)
+	if isNetworkError(err) || isTimeoutError(err) ||
+		strings.Contains(errStr, "connection refused") ||
+		strings.Contains(errStr, "connection reset") ||
+		strings.Contains(errStr, "no route to host") ||
+		strings.Contains(errStr, "network is unreachable") ||
+		strings.Contains(errStr, "dns") ||
+		strings.Contains(errStr, "lookup") ||
```
ndyakov: if the actual error message is something like `dns lookup` let's have them together. I can see how `lookup` can be part of another error type.

extra/redisotel-native/metrics.go
```diff
@@ -0,0 +1,910 @@
...
+	// Check for Redis server errors (including cluster errors)
+	if prefix := extractRedisErrorPrefix(err); prefix != "" {
+		return "server"
+	}
+
+	// Uncategorized errors
+	return "other"
+}
+
+func getErrorCategoryFromType(errorType string) string {
```
ndyakov: are both `getErrorCategoryFromType` and `getErrorCategory` needed? Can we combine them since they are doing pretty much the same thing?

extra/redisotel-native/redisotel.go
```diff
@@ -0,0 +1,516 @@
...
+			metric.WithDescription("The time it took to create a new connection"),
+			metric.WithUnit("s"),
+		)
+		if cfg.histAggregation == HistogramAggregationExplicitBucket {
+			connectionCreateTimeOpts = append(connectionCreateTimeOpts,
+				metric.WithExplicitBucketBoundaries(cfg.bucketsConnectionCreateTime...),
+			)
+		}
+		connectionCreateTime, err = meter.Float64Histogram(
+			"db.client.connection.create_time",
```
ndyakov: should we have this as const?

extra/redisotel-native/redisotel.go
```diff
@@ -0,0 +1,516 @@
...
+			connectionCreateTimeOpts...,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create connection create time histogram: %w", err)
+		}
+
+		connectionRelaxedTimeout, err = meter.Int64UpDownCounter(
+			"redis.client.connection.relaxed_timeout",
```
ndyakov: should we have this as const?

extra/redisotel-native/redisotel.go
```diff
@@ -0,0 +1,516 @@
...
+			metric.WithDescription("How many times the connection timeout has been increased/decreased (after a server maintenance notification)"),
+			metric.WithUnit("{relaxation}"),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create connection relaxed timeout metric: %w", err)
+		}
+
+		connectionHandoff, err = meter.Int64Counter(
+			"redis.client.connection.handoff",
```
ndyakov: should we have this as const?

extra/redisotel-native/redisotel.go
```diff
@@ -0,0 +1,516 @@
...
+			metric.WithDescription("The time it took to obtain a connection from the pool"),
+			metric.WithUnit("s"),
+		)
+		if cfg.histAggregation == HistogramAggregationExplicitBucket {
+			connectionWaitTimeOpts = append(connectionWaitTimeOpts,
+				metric.WithExplicitBucketBoundaries(cfg.bucketsConnectionWaitTime...),
+			)
+		}
+		connectionWaitTime, err = meter.Float64Histogram(
+			"db.client.connection.wait_time",
```
ndyakov: should we have this as const?

extra/redisotel-native/redisotel.go
```diff
@@ -0,0 +1,516 @@
...
+
+	return recorder, nil
+}
+
+// parsePoolName extracts server address, port, and database index from a pool name.
+// Pool name format: "host:port/db" or "host/db" or "host:port" or "host"
+// Returns: (serverAddr, serverPort, dbIndex)
+func parsePoolName(poolName string) (string, string, string) {
+	// Handle special pool names
+	if poolName == "main" || poolName == "pubsub" {
```
ndyakov: let's extract those as const.

maintnotifications/handoff_worker.go
```diff
@@ -434,6 +434,11 @@ func (hwm *handoffWorkerManager) performHandoffInternal(
 		deadline := time.Now().Add(hwm.config.PostHandoffRelaxedDuration)
 		conn.SetRelaxedTimeoutWithDeadline(relaxedTimeout, relaxedTimeout, deadline)
 
+		// Record relaxed timeout metric (post-handoff)
+		if relaxedTimeoutCallback := pool.GetMetricConnectionRelaxedTimeoutCallback(); relaxedTimeoutCallback != nil {
+			relaxedTimeoutCallback(ctx, 1, conn, "main", "HANDOFF")
```
ndyakov: `"main"` as const

maintnotifications/handoff_worker.go
```diff
@@ -462,6 +467,11 @@ func (hwm *handoffWorkerManager) performHandoffInternal(
 	internal.Logger.Printf(ctx, logs.HandoffSucceeded(connID, newEndpoint))
 
 	// successfully completed the handoff, no retry needed and no error
+	// Notify metrics: connection handoff succeeded
+	if handoffCallback := pool.GetMetricConnectionHandoffCallback(); handoffCallback != nil {
+		handoffCallback(ctx, conn, "main")
```
ndyakov: same, let's have `"main"` as const

maintnotifications/push_notification_handler.go
```diff
@@ -191,6 +198,12 @@ func (snh *NotificationHandler) handleMigrating(ctx context.Context, handlerCtx
 		internal.Logger.Printf(ctx, logs.RelaxedTimeoutDueToNotification(conn.GetID(), "MIGRATING", snh.manager.config.RelaxedTimeout))
 	}
 	conn.SetRelaxedTimeout(snh.manager.config.RelaxedTimeout, snh.manager.config.RelaxedTimeout)
+
+	// Record relaxed timeout metric
+	if relaxedTimeoutCallback := pool.GetMetricConnectionRelaxedTimeoutCallback(); relaxedTimeoutCallback != nil {
+		relaxedTimeoutCallback(ctx, 1, conn, "main", "MIGRATING")
```
ndyakov: `"main"` as const

maintnotifications/push_notification_handler.go
```diff
@@ -249,6 +262,12 @@ func (snh *NotificationHandler) handleFailingOver(ctx context.Context, handlerCt
 		internal.Logger.Printf(ctx, logs.RelaxedTimeoutDueToNotification(connID, "FAILING_OVER", snh.manager.config.RelaxedTimeout))
 	}
 	conn.SetRelaxedTimeout(snh.manager.config.RelaxedTimeout, snh.manager.config.RelaxedTimeout)
+
+	// Record relaxed timeout metric
+	if relaxedTimeoutCallback := pool.GetMetricConnectionRelaxedTimeoutCallback(); relaxedTimeoutCallback != nil {
+		relaxedTimeoutCallback(ctx, 1, conn, "main", "FAILING_OVER")
```
ndyakov: `"main"` as const

options.go
```diff
@@ -21,6 +23,15 @@ import (
...
 
+// generateUniqueID generates a short unique identifier for pool names.
+func generateUniqueID() string {
+	b := make([]byte, 4)
+	if _, err := rand.Read(b); err != nil {
+		return ""
+	}
+	return hex.EncodeToString(b)
+}
+
```
ndyakov: does it make sense to use global auto increment so we know which pool is it?

ndyakov (approved):
overall looks good, can merge as is, left couple of improvements that we can address

options.go
elena-kolevska: I aligned redis-py to use the same rand id, so let me know if you change it.

ndyakov:
@ofekshenawa let's resolve the comments here and work on merging this

options.go
ndyakov: Actually @elena-kolevska, I think we should rework this in the near future. Both random and counters will be shared ids between instances. Let's think of a id of the client (maybe unixtimestamp) and then have the pool id as `client_id-counter` ? 
@ofekshenawa what do you think? Maybe we can address this separately?

PR #3638, opened 2025-12-03
Remove metrics with empty data

fengve:
Because empty data cannot be considered an error.

### DIFF
--- a/extra/redisotel/metrics.go
+++ b/extra/redisotel/metrics.go
@@ -329,7 +329,10 @@ func milliseconds(d time.Duration) float64 {
 }
 
 func statusAttr(err error) attribute.KeyValue {
-	if err != nil && err != redis.Nil {
+	if err != nil {
+		if  err == redis.Nil {
+			return attribute.String("status", "nil")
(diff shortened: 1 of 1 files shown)


ndyakov:
@fengve thank you for contributing, do you think it makes sense to have this as a separate status - `nil`?

fengve:
@ndyakov 



Generally, we consider an empty value to be a correct state because it is not an error in the Redis protocol.

When we use Redis for caching(pseudocode):
```
data, err := redis.Get(key)
if err != nil && err != redis.ErrNil {
	return nil, err
}

data = getData()

redis.Set(key,data)
```

ndyakov:
@fengve i agree with you that `redis.Nil` should not be considered an error response here, but what i would like to discuss is - if it should be `attribute.String("status", "nil")` instead of `attribute.String("status", "ok")`, since even in your case it shows cache miss and not a cache hit.

fengve:
@ndyakov 

You're right, this makes it easier to distinguish states. I've added the nil state.

PR #3640, opened 2025-12-04
chore(go): update go version to 1.21

ndyakov:
Thanks to the work done by @ccoVeille it was determined that we use features not available in go `1.18` and we decided to gradually move towards `1.23`. Initially we will move to `1.21` and with the release of `1.26` we will bump to `1.23`. After that, with each new golang version we will jump one version, which means that at the end, our minimal go version will be three versions before the latest release - just a reminder, go officially supports the last two versions (`stable` and `oldstable`).

- Added ci run with min version ( so we can observe incompatible changes with it )
- update version in examples

### DIFF
--- a/.github/wordlist.txt
+++ b/.github/wordlist.txt
@@ -76,3 +76,4 @@ oauth
 entraid
 MiB
 KiB
+oldstable


internal/routing/aggregator.go
```diff
@@ -482,10 +482,21 @@ func (a *AggLogicalAndAggregator) Add(result interface{}, err error) error {
...
-		a.res.And(0)
+		newVal = 0
+	}
+	for {
+		old := a.res.Load()
+		desired := old & newVal
+		if a.res.CompareAndSwap(old, desired) {
+			break
+		}
 	}
```
ccoVeille: zero value of int64 is 0

```go
         var newVal int64 
         if val { 
                 newVal = 1 
         } 
         for { 
                 old := a.res.Load() 
                 desired := old & newVal 
                 if a.res.CompareAndSwap(old, desired) { 
                         break 
                 } 
          }
```

internal/routing/aggregator.go
```diff
@@ -566,10 +577,21 @@ func (a *AggLogicalOrAggregator) Add(result interface{}, err error) error {
...
-		a.res.Or(0)
+		newVal = 0
+	}
+	for {
+		old := a.res.Load()
+		desired := old | newVal
+		if a.res.CompareAndSwap(old, desired) {
+			break
+		}
 	}
```
ccoVeille: Isn't a risk the for loop never ends?

I mean trying with a timeout or limit the number of retry

ccoVeille: Why res is an int64 and not a atomic.Bool ?

internal/routing/aggregator.go
ccoVeille: But is there a point to use LoadCompare when val is false ?

Couldn't we return out the method?

ndyakov: we need a.res to be `false` , if `val` is `false` what can be done is just a simple `Store` `false`. As to why the aggregator is using `int64` instead of `bool` - I assume it was done because the atomic `And` and `Or` are available for integer types. But I do think this can be refactored with bool types and even skip the loops. Let me rewrite this.

ccoVeille: Yes. exactly

.github/workflows/build.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           - "8.2.x" # Redis CE 8.2
           - "8.0.x" # Redis CE 8.0
         go-version:
+          - "1.21.x"
           - "1.23.x"
           - "1.24.x"
```
ccoVeille: Please add 


```suggestion
          - "1.24.x"
          - "1.25.x"
```

Or simply use


```suggestion
          - oldstable
          - stable
```

.github/workflows/build.yml
```diff
@@ -78,6 +79,7 @@ jobs:
             - "8.2.x" # Redis CE 8.2
             - "8.0.x" # Redis CE 8.0
           go-version:
+            - "1.21.x"
             - "1.23.x"
             - "1.24.x"
```
ccoVeille: See

https://github.com/redis/go-redis/pull/3640#discussion_r2590297796

internal/routing/aggregator.go
ndyakov: i was thinking the same, but overall this woudn't have been a problem. anyway, it is MUCH cleaner with atomic.bool here

ndyakov: addressed.

.github/workflows/build.yml
ndyakov: addressed

PR #3649, opened 2025-12-13
fix(queuedNewConn): protect against nil context

ndyakov:
fixes https://github.com/redis/go-redis/issues/3630


cyningsun:
**LGTM.** The nil-check guard added in `newConn` is a **straightforward, safe fix** for the panic condition identified. It cleanly addresses the immediate issue.

Issue #3652, opened 2025-12-19
FailoverClient does not retry getMasterAddr on timeouts


SamuelGD:
FailoverClient (master replica dialer) times out on unavailable Sentinel and terminates before searching for a live one.
`MasterAddr` does not retry `getMasterAddr` calls on a different sentinel if one of the first two calls times out.

If `getMasterAddr` times out, this will result in a `context.DeadlineExceeded` error, which will skip retries on other sentinels: https://github.com/redis/go-redis/blob/v9.17.2/sentinel.go#L830 and https://github.com/redis/go-redis/blob/v9.17.2/sentinel.go#L848.

This bug looks similar to https://github.com/redis/go-redis/issues/3439 on a different code path.

## Context (Environment)

We run redis sentinel in Kubernetes. When the pod hosting the sentinel container is deleted, we hit a race condition at high scale where our clients open a few connections right after the container is terminated - but probably before the pod is deleted. These connections don't get an RST and the connections keep hanging until timing out (DialTimeout). Because the go client doesn't retry on a live sentinel, commands fail with a `dial tcp <old sentinel pod ip>:<sentinel port>: i/o timeout` error.

## Expected Behavior

We would expect MasterAddr to try calling getMasterAddr on the remaining sentinels: https://github.com/redis/go-redis/blob/v9.17.2/sentinel.go#L879.

## Current Behavior

During the race condition, MasterAddr terminates earlier in https://github.com/redis/go-redis/blob/v9.17.2/sentinel.go#L830 and https://github.com/redis/go-redis/blob/v9.17.2/sentinel.go#L848. As a result it never tries getting the master address from the remaining sentinels.

ndyakov:
Hello @SamuelGD, thank you for reporting this. We will probably take a look at it after the holidays

cxljs:
This appears to be resolved in #3654 , so I think we can close this issue, WDYT @ndyakov @SamuelGD ?

ndyakov:
fixed in #3654 , @SamuelGD feel free to reopen if you are still experiencing this issue.

PR #3654, opened 2025-12-21
When the connection to a sentinel times out, attempt to connect to other sentinel

### LINKED ISSUE #3439: FailoverClient (with ReplicaOnly) times out on unavailable Sentinel and terminates search for a live one
FailoverClient (with ReplicaOnly) times out on unavailable Sentinel and terminates search for a live one.

## Expected Behavior

FailoverClient successfully finds an available Sentinel and establishes a connection.

## Current Behavior

FailoverClient.replicaAddrs terminates immediately if it tries to connect to a Sentinel that is network-unavailable.
(see Detailed Description)

## Possible Solution

Need to check outer context expiration instead of checking if the returned error matches context.DeadlineExceeded.

## Steps to Reproduce

1. Run several Redis-server+Sentinel instances locally (6379+26379 <- leader, 6380+26380, 6381+26381)
2. Simulate network timeout for connection to one of them (block in firewall via packet drop, not reject)
   blocking via pfctl (reproducing on mac os x)
   ```
   block drop in proto tcp from any to any port = 6381
   block drop in proto tcp from any to any port = 26381
   ```
3. Create FailoverClient and execute a command (Ping) (check [main_sentinel.go](https://gist.github.com/perekalov/8f710337a3dd6ba8873886c7198c610a))
4. During the first command, discovery occurs, and if [shuffle](https://github.com/redis/go-redis/blob/master/sentinel.go#L407) places the problematic sentinel first, the connection establishment will timeout and the command may fail with an error without trying to connect to other live Sentinels.

Log file containing multiple executions (`main_sentinel.log`) is attached. In certain cases, the connection can still be established (see Detailed Description).

## Detailed Description

Under the hood, the driver creates a network connection via `net.Dialer`.
The `DialContext` method in case of network timeout can return an error that can be interpreted as `context.DeadlineExceeded`: https://github.com/golang/go/issues/51428
For example, this happens when the Timeout expires during connection (this parameter is [set](https://github.com/redis/go-redis/blob/master/sentinel.go#L467) by go-redis).

That is, `Dialer.DialContext` can return an error (comparable to `context.DeadlineExceeded`), not because the context passed from outside has expired, but due to an internal connection timeout.
The go-redis code, as far as I can see, expects that the `context.DeadlineExceeded` error means the expiration of the context passed from outside.

Thus `sentinelFailover.replicaAddrs` fails with `context.DeadlineExceeded` error when encountering an unavailable Sentinel, even if the context passed to the command has no timeout set.
This breaks the failover mechanism's core purpose of finding available sentinels.

Check [main_deal.go](https://gist.github.com/perekalov/bc90689409952dcc7af702415ee5af9e) for reproduction. Again, packet drop needs to be configured on port 26381.
A log file from one of the runs is attached too (`main_deal.log`). Interestingly, _sometimes_ the function doesn't wrap `i/o timeout` in `context.DeadlineExceeded`, and `replicaAddrs` proceeds to check the next Sentinel and ultimately finds a live one.
But according to the mentioned change in the standard library code, we should assume that it should behave exactly this way.

## Possible Implementation

As I mentioned above, comparing the error returned by `net.Dialer` with `context.DeadlineExceeded` seems incorrect.
Perhaps instead, we could check whether the externally passed context has not yet expired.
```
diff --git a/sentinel.go b/sentinel.go
index 04c0f72..61a6f6b 100644
--- a/sentinel.go
+++ b/sentinel.go
@@ -833,7 +833,7 @@ func (c *sentinelFailover) replicaAddrs(ctx context.Context, useDisconnected boo
                replicas, err := sentinel.Replicas(ctx, c.opt.MasterName).Result()
                if err != nil {
                        _ = sentinel.Close()
-                      if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+                     if ctx.Err() != nil {
                                return nil, err
                        }
                        internal.Logger.Printf(ctx, "sentinel: Replicas master=%q failed: %s",
```

cxljs:
closes #3439 and #3652 

According to Go's commit https://github.com/golang/go/commit/85b5f86584686677c554b2538a7edee96d684aea, the error value returned by dial timeout is also a context error, so we need to use ctx.Err() to distinguish whether it is a dial timeout or a user-set deadline.


ndyakov:
Thank you @cxljs , will review next week

ndyakov (approved):

PR #3663, opened 2026-01-08
fix(maintnotif): fix smigrated parser and add cluster state reload interval option

ndyakov:


ndyakov:
After an offline discussion it was decided to set the default interval to `10s` for backwards compatibility.

PR #3667, opened 2026-01-12
osscluster_test: use strings.Cut instead of manually statLine[callsSt…

12ya:
…art+6:]


12ya:
> Can you motivate this change? I actually think this reduces readability in this case.

Yeap, instead of 2 operation (indexing and slicing) we have 1 (Cut) and the API is actually pretty straightforward - returns us before and after separator, also instead of returning -1 if not found returns us boolean which is clearer

ndyakov:
@12ya as we discussed in the other PR that is introducing the cut, the cut actually is 3 operations -> indexing and slicing both before and after. I would actually argue this (indexing and only if the index is found -> slicing) is cleaner as it is a standard approach in most programming languages,  than calling a Cut helper method which is go specific and discarding part of the results.

Cut has its place if you are going to work on both returned values, but in my mind, if you are going to discard one of them - it is a waste and it won't actually improve readability.

PR #3668, opened 2026-01-12
internal/hashtag/hashtag: use strings.Cut instead of manual key[s+1:]

12ya:


internal/hashtag/hashtag.go
```diff
@@ -60,8 +60,9 @@ func Present(key string) bool {
 	if key == "" {
 		return false
 	}
-	if s := strings.IndexByte(key, '{'); s > -1 {
-		if e := strings.IndexByte(key[s+1:], '}'); e > 0 {
+
+	if _, after, found := strings.Cut(key, "{"); found {
+		if e := strings.IndexByte(after, '}'); e > 0 {
```
ndyakov, line 65: Since `Cut` underlying implementation is:
```
func Cut(s, sep string) (before, after string, found bool) {
	if i := Index(s, sep); i >= 0 {
		return s[:i], s[i+len(sep):], true
	}
	return s, "", false
}
```
Can you please benchmark both and see if it makes sense to change it to `Cut` instead of using `key[s+1]`

12ya: yeap sure

12ya: @ndyakov , on average 1.35 ns/op difference. but since its only 1 file i would agree to leave it as it is. thanks for reviewing!

ndyakov: @12ya what 1.35 ns/op difference do you observe? can you provide the benchmark and its output? Is it 1.35 ns faster or slower? What is the whole time, this 1.35 ns can be 50% or 5% of the execution time of this operation.

12ya: @ndyakov , 
Screenshot 2026-01-13 at 9 10 02 PM

NoTag | 3.28 ns | 5.37 ns 
TagAtStart | 6.52 ns | 7.18 ns
TagInMiddle | 6.84 ns | 8.21 ns
TagAtEnd | 7.04 ns | 7.70 ns
EmptyTag | 6.27 ns | 7.39 ns
ShortKey | 6.23 ns | 7.20 ns
LongKey | 5.57 ns | 7.31 ns
NoBraceClose | 5.85 ns | 7.53 ns
NoBraceOpen | 3.54 ns | 5.40 ns

12ya: attached the examples i used

ndyakov: Hello @12ya, can you please just note what does the first and what does the second number means? Which is the normal `indexOf` and which is the `cut` ?

12ya: > Hello @12ya, can you please just note what does the first and what does the second number means? Which is the normal `indexOf` and which is the `cut` ?

first is `Index` and second is `Cut`, in my opinion the `Cut` API helps DX, but i've read your review on it in another PR

ndyakov: Based on this benchmark I do not think this degradation is justified, `IndexOf` and slicing is a standard practice and I would argue - it is even easier to read

12ya: > Based on this benchmark I do not think this degradation is justified, `IndexOf` and slicing is a standard practice and I would argue - it is even easier to read

got it, i'll close all related PRs :)

PR #3673, opened 2026-01-13
chore(cmd): add deprecation warnings

ndyakov:
based on information from https://github.com/redis/jedis/pull/4397

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -8,7 +8,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
       - name: Check Spelling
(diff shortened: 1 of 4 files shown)


commands.go
```diff
@@ -665,6 +668,9 @@ func (c cmdable) ShutdownNoSave(ctx context.Context) *StatusCmd {
 	return c.shutdown(ctx, "nosave")
 }
 
+// SlaveOf sets a Redis server as a replica of another, or promotes it to being a master.
+//
+// Deprecated: Use ReplicaOf instead as of Redis 5.0.0.
```
moredure, line 673: but there is no ReplicaOf

ndyakov: Thank you @moredure, would you mind submitting an issue, we will address it as soon as possible.

PR #3687, opened 2026-01-22
Support NaN values for timeseries

ofekshenawa:
This PR implements support for NaN (Not a Number) values in Redis time series commands, enabling developers to insert and retrieve NaN-valued measurements for timestamps where data is unknown, invalid, or unavailable (e.g., broken sensors, unreported values).

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -24,6 +24,7 @@ runs:
 
         # Mapping of redis version to redis testing containers
         declare -A redis_version_mapping=(
+          ["8.6.x"]="custom-21183968220-debian-amd64"
           ["8.4.x"]="8.4.0"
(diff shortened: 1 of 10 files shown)


ofekshenawa:
The failures in tests are unrelated to NaN support (They are related to FT.HYBRID and XINFO STREAM)

ndyakov:
@ofekshenawa the test that I can see failing are related to the `NaN` changes (actually, the tests you added), since it is not available with the older redis versions in our matrix. We should run those tests only for 8.6+ , maybe you can rebase on top of https://github.com/redis/go-redis/pull/3685 to get the new image and matrix for 8.6 ?

ofekshenawa:
@ndyakov Yes, you're right. I meant the 8.6 test suite that will be implemented in #3685
I'll fix this.

PR #3689, opened 2026-01-25
pre-allocate capacity to prevent multiple capacity expansions

feelshu:
pre-allocate capacity to prevent multiple capacity expansions


ndyakov:
Thank you @feelshu

PR #3693, opened 2026-01-26
Add support for Redis Streams Idempotent Production

ofekshenawa:
This PR adds support for Redis 8.6+ Streams Idempotent Production feature, enabling at-most-once production semantics to prevent duplicate entries. 
The implementation:
- Extends XAddArgs with three new fields (ProducerID, IdempotentID, IdempotentAuto) to support both manual idempotent IDs (IDMP) and auto-generated IDs based on message content (IDMPAUTO)
-  Adds a new XCFGSET command via XCfgSet method to configure idempotent production settings (duration and maxsize),

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -8,7 +8,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
       - name: Check Spelling
-        uses: rojopolis/spellcheck-github-actions@0.57.0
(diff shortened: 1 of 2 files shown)


elena-kolevska:
@ofekshenawa  @ndyakov  we're missing the new fields in the `XINFO STREAM` response

ndyakov:
@elena-kolevska they are addressed in the base branch.

PR #3695, opened 2026-01-28
[feat] Add support for Redis HOTKEYS commands

ofekshenawa:
Implements support for the Redis HOTKEYS feature, which provides a method for identifying hot keys in a database based on CPU consumption and network utilization.

### DIFF
--- a/internal_maint_notif_test.go
+++ b/internal_maint_notif_test.go
@@ -1,0 +1,96 @@
+package redis
+
+import (
+	"bufio"
+	"context"
+	"net"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/redis/go-redis/v9/internal/pool"
+)
+
+// TestInitConnNilMaintNotificationsConfig is a regression test for
+// https://github.com/redis/go-redis/issues/3675
+//
+// initConn previously accessed MaintNotificationsConfig.EndpointType
+// unconditionally, even though the preceding line correctly nil-checked
+// MaintNotificationsConfig. This caused a nil pointer dereference panic
+// and left optLock.RLock held, leading to a subsequent deadlock.
+func TestInitConnNilMaintNotificationsConfig(t *testing.T) {
+	// Start a minimal TCP server that speaks enough RESP to let
+	// initConn get past the HELLO / AUTH / pipeline phases and reach
+	// the MaintNotificationsConfig code path.
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("failed to listen: %v", err)
+	}
+	defer ln.Close()
+
+	// mockRedis responds to every RESP command with a Redis-protocol
+	// error. This lets initConn fall through HELLO (Redis errors are
+	// not fatal when there is no password) and the empty pipeline
+	// succeeds trivially.
+	go func() {
+		for {
+			conn, err := ln.Accept()
+			if err != nil {
+				return
+			}
+			go func(c net.Conn) {
+				defer c.Close()
+				scanner := bufio.NewScanner(c)
+				for scanner.Scan() {
+					line := scanner.Text()
+					if strings.HasPrefix(line, "*") {
+						_, _ = c.Write([]byte("-ERR unknown command\r\n"))
+					}
+				}
+			}(conn)
+		}
+	}()
+
+	opt := &Options{
+		Addr: ln.Addr().String(),
+	}
+	opt.init()
+
+	// Force MaintNotificationsConfig to nil after init() to reproduce
+	// the scenario from issue #3675.
+	opt.MaintNotificationsConfig = nil
+
+	c := &baseClient{
+		opt: opt,
+	}
+	c.initHooks(hooks{
+		dial:       c.dial,
+		process:    c.process,
+		pipeline:   c.processPipeline,
+		txPipeline: c.processTxPipeline,
+	})
+
+	// Dial a real connection to the mock server.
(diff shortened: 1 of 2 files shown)


ndyakov:
on hold for now, response type should change.

hotkeys_commands.go
```diff
@@ -0,0 +1,113 @@
...
+
+// HotKeysStartArgs contains the arguments for the HOTKEYS START command.
+// This command is only available on standalone clients due to its stateful nature
+// requiring session affinity. It must NOT be used on cluster or pooled clients.
+type HotKeysStartArgs struct {
+	// Metrics to track. At least one must be specified.
+	Metrics []HotKeysMetric
+	// Count is the number of top keys to report.
+	// Default: 10, Min: 10, Max: 64
+	Count int64
```
elena-kolevska: We can cover the 10-64 range with a `uint8`

hotkeys_commands.go
```diff
@@ -0,0 +1,113 @@
...
+	// Duration is the auto-stop tracking after this many seconds.
+	// Default: 0 (no auto-stop)
+	Duration int64
+	// Sample is the sample ratio - track keys with probability 1/sample.
+	// Default: 1 (track every key), Min: 1
+	Sample int64
+	// Slots specifies specific hash slots to track.
+	// All specified slots must be hosted by the receiving node.
+	// If not specified, all slots are tracked.
+	Slots []int64
```
elena-kolevska: Slots range from 0 to 16383, so can be covered with `uint16`

command.go
```diff
@@ -4910,6 +4912,243 @@ func (cmd *LatencyCmd) Clone() Cmder {
...
+type HotKeysKeyEntry struct {
+	Key   string
+	Value interface{} // Can be int64 or string
+}
+
+// HotKeysResult represents the response data from HOTKEYS GET command.
+// Field names match the Redis response format.
+type HotKeysResult struct {
+	TrackingActive                       bool
+	SampleRatio                          int64
```
elena-kolevska: Can be `uint8`

command.go
```diff
@@ -4910,6 +4912,243 @@ func (cmd *LatencyCmd) Clone() Cmder {
...
+			TrackingActive:                       cmd.val.TrackingActive,
+			SampleRatio:                          cmd.val.SampleRatio,
+			SampledCommandSelectedSlotsUs:        cmd.val.SampledCommandSelectedSlotsUs,
+			AllCommandsSelectedSlotsUs:           cmd.val.AllCommandsSelectedSlotsUs,
+			AllCommandsAllSlotsUs:                cmd.val.AllCommandsAllSlotsUs,
+			NetBytesSampledCommandsSelectedSlots: cmd.val.NetBytesSampledCommandsSelectedSlots,
+			NetBytesAllCommandsSelectedSlots:     cmd.val.NetBytesAllCommandsSelectedSlots,
+			NetBytesAllCommandsAllSlots:          cmd.val.NetBytesAllCommandsAllSlots,
+			CollectionStartTimeUnixMs:            cmd.val.CollectionStartTimeUnixMs,
+			CollectionDurationMs:                 cmd.val.CollectionDurationMs,
```
elena-kolevska: Do we usually return the timestamp and duration in `ms` or `us`, without converting to Go time types, time.Duration and time.Time? Pls check how we do it in other commands, for consistency.

hotkeys_commands.go
```diff
@@ -0,0 +1,113 @@
...
+func (c *Client) HotKeysStart(ctx context.Context, args *HotKeysStartArgs) *StatusCmd {
+	cmdArgs := make([]interface{}, 0, 16)
+	cmdArgs = append(cmdArgs, "hotkeys", "start")
+
+	if len(args.Metrics) > 0 {
+		cmdArgs = append(cmdArgs, "metrics", len(args.Metrics))
+		for _, metric := range args.Metrics {
+			cmdArgs = append(cmdArgs, strings.ToLower(string(metric)))
+		}
+	}
```
elena-kolevska: as per the spec, at least one value for `metrics` should be specified, so we might just as well do that validation on the client side instead of letting it slip to the server, just so it can return an error.

hotkeys_commands_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+			stop := client.HotKeysStop(ctx)
+			Expect(stop.Err()).NotTo(HaveOccurred())
+			Expect(stop.Val()).To(Equal("OK"))
+
+			get2 := client.HotKeysGet(ctx)
+			Expect(get2.Err()).NotTo(HaveOccurred())
+			result2 := get2.Val()
+			Expect(result2).NotTo(BeNil())
+			Expect(result2.TrackingActive).To(BeFalse())
```
elena-kolevska, line 82: Can we please add some more assertions on the response type? Just a basic validation that confirms we're not just returning the default values, but that something has actually been set.

elena-kolevska (approved):

command.go
```diff
@@ -4910,6 +4912,243 @@ func (cmd *LatencyCmd) Clone() Cmder {
...
+				for _, sr := range s {
+					if val, ok := sr.(int64); ok {
+						slotRange = append(slotRange, val)
+					}
+				}
+				result.SelectedSlots = append(result.SelectedSlots, slotRange)
+			}
+		}
+	}
+	if v, ok := data["sampled-command-selected-slots-us"].(int64); ok {
```
elena-kolevska:
```suggestion
	if v, ok := data["sampled-commands-selected-slots-us"].(int64); ok {
```
This was renamed, right?

ndyakov: Yes, I already addressed it and pushed a new commit

PR #3700, opened 2026-01-30
Make Script helper work in strict FIPS mode by avoiding client-side SHA-1

chaitanyabodlapati:
go-redis `Script` currently computes the script digest using SHA-1 in `NewScript()` and then uses that digest for `EVALSHA`.

In strict FIPS environments, client-side SHA-1 can be blocked, which causes applications using `Script`/`EVALSHA` to fail.

This PR adds a FIPS-safe path where go-redis does not compute SHA-1 in Go:

--> Default behavior is unchanged for existing users: we still compute the digest locally and use `EVALSHA`.
--> In FIPS mode, we obtain the digest from Redis instead:
  --> call `SCRIPT LOAD` to load the script
  --> Redis returns the script digest
  --> cache the returned digest in the `Script` object
  --> use that digest for `EVALSHA` / `EVALSHA_RO`
--> If Redis returns `NOSCRIPT` (script cache flushed), we reload once and retry.
--> If loading fails, we fall back to `EVAL` / `EVALRO` to avoid hard failures.

FIPS mode is enabled when either:

--> `GODEBUG` includes `fips140=only` or `fips140=on`, or
--> `GO_REDIS_DISABLE_CLIENT_SHA1=true` is set (manual override).

This keeps the common path identical, while allowing strict FIPS deployments to use scripts without client-side SHA-1.

---

> [!NOTE]
> **Medium Risk**
> Adds a new execution path for Redis scripts that changes how `Script` hashes are populated/cached and introduces retry logic on `NOSCRIPT`; while mostly opt-in via `NewScriptServerSHA`, it touches frequently used script helpers and adds locking that could affect concurrency behavior.
> 
> **Overview**
> Adds an opt-in `NewScriptServerSHA` mode for `Script` that **avoids client-side SHA-1** by using `SCRIPT LOAD` to obtain and cache the digest, then running via `EVALSHA`/`EVALSHA_RO`.
> 
> Updates `Script` to be concurrency-safe (mutex-protected hash) and to **reload+retry once** on `NOSCRIPT` for the server-hash path, with fallback to `EVAL`/`EVALRO` if `SCRIPT LOAD` fails; `Load` now also caches the returned digest.
> 
> Adds unit + integration tests covering digest caching and `SCRIPT FLUSH`/`NOSCRIPT` retry behavior, and tidies `go.mod` by moving `go.uber.org/atomic` into the main `require` block.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 36f2731f0c172a8cffb236dc69752a0097a7a225. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,9 +8,10 @@ require (
 	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f
 	github.com/zeebo/xxh3 v1.1.0
+	go.uber.org/atomic v1.11.0
 )
 
-go.uber.org/atomic v1.11.0
+
 
 require (
 	github.com/klauspost/cpuid/v2 v2.2.10 // indirect


ndyakov:
@chaitanyabodlapati thank you for opening this pr. I am no a fan of using init here. Would prefer the application to control this with an option. What do you think?

chaitanyabodlapati:
Thanks for the feedback — that makes sense.
We used `init()` because in our packaging we ship client as a prebuilt binary, so customers can't realistically change application code just to enable this. We need something that can be turned on through deployment configuration.
Would you be okay if we remove the `init()` logic, and instead support a simple environment variable switch (and/or an explicit option) so this can be enabled without code changes?
Default behavior would stay the same unless the env var/option is set.

ndyakov:
@chaitanyabodlapati If you are shipping the client as a prebuilt binary I would suggest to do a wrapper or a patch on your end. Such environment variables make sense in tests and applications, rarely in libraries and only for such that are specifically made to run in a given setup. Let's discuss another way to facilitate this. Explicit option would be a breaking change. Let me see if I understand your usecase correctly:
- your users do not have access to the initialisation of the client (e.g. Options)
- do they have access to the methods called? If so, can you have your own (or maybe it can be provided in the library) "constructor" for Script that uses server side SHA?

chaitanyabodlapati:
Thanks, that’s a fair point — let me give a bit more context on our use case.

We ship Traefik (which uses go-redis v9) as part of a bigger product as a single prebuilt binary. Our customers do not write Go code or construct `redis.Client` / `redis.Script` themselves — that all happens inside Traefik and our packaging. What they can change is only the runtime environment (for example, whether they run with strict FIPS settings).

In strict FIPS environments we run with `GODEBUG=fips140=only`. In that mode, calling `crypto/sha1` from user code is not allowed. Since `NewScript` always uses `crypto/sha1` today, creating a script will fail in FIPS mode, and Traefik’s Redis rate limiting stops working. We can’t ask customers to patch Traefik or go-redis, and we also don’t want to ship separate FIPS and non-FIPS binaries — it’s one build, and the environment decides.

What we would like is a way to use server-side SHA for scripts when we are in FIPS mode, without changing the behavior for everyone else:

 Non-FIPS: keep current behavior (`NewScript` does SHA-1 in Go and calls `EVALSHA`, as today).
 FIPS: avoid SHA-1 in Go, and instead use `SCRIPT LOAD` to get the hash from Redis and then call `EVALSHA` with that hash (with a `NOSCRIPT` reload).

We understand your concerns about `init()` and env vars inside the library, and we agree that changing `NewScript`’s signature would be a breaking change. So instead of the env-toggle version, we’d be happy with a small API addition, for example:

 keep `NewScript(src string)` exactly as it is today, and
 add a new constructor like `NewScriptFromServer(src string)` that implements the “server-side SHA” behavior.

On our side, we can then do the runtime decision in Traefik:

 non-FIPS: call `NewScript(...)`
 FIPS: call `NewScriptFromServer(...)`

That way the library stays clean (no env vars, no `init()` logic), existing users see no change.

Would that approach work for you? If yes, I can update the PR to drop the env-based toggle and instead implement the separate “server-SHA” constructor.

robbiet480:
Funny enough, we at @CampusTech are suddenly in need of a FIPS-140 compliant Redis library too (and must run `GODEBUG=fips140=only` as well), so will be keeping an eye on this PR and happy to jump in to help out if needed @chaitanyabodlapati. I'm shocked by your impeccable timing! Thanks!

ndyakov:
@chaitanyabodlapati this additional method that will directly use Server for hashing is the correct approach for a library in my mind. Feel free to implement it. This will provide the tools to the application to decide how to create the `Script`. Thank you!

chaitanyabodlapati:
Thanks for the feedback .
I updated this PR per your suggestion:
  • Removed the init()-based switching logic.
  • Kept NewScript(src) unchanged (still computes SHA-1 client-side and uses EVALSHA as today).
  • Added an opt-in constructor NewScriptServerSHA(src) which avoids client-side SHA-1 by:
           - using SCRIPT LOAD to obtain the digest from Redis (server-side hashing),
           - then executing via EVALSHA,
           -and reloading/retrying once on NOSCRIPT.
This keeps the default behavior the same for everyone, while allowing applications running in strict FIPS mode to explicitly choose the server-side digest path.

go.mod
```diff
@@ -10,10 +10,14 @@ require (
 	github.com/zeebo/xxh3 v1.1.0
 )
 
-require go.uber.org/atomic v1.11.0
+require (
+	github.com/alicebob/miniredis/v2 v2.37.0
```
ndyakov: Hey @chaitanyabodlapati, in the CI we do have redis working on the default port, you do not need miniredis and I would prefer to not include this dependency if not absolutely needed.

go.mod
```diff
@@ -10,10 +10,14 @@ require (
...
+	go.uber.org/atomic v1.11.0
+)
 
 require (
 	github.com/klauspost/cpuid/v2 v2.2.10 // indirect
+	github.com/yuin/gopher-lua v1.1.1 // indirect
 	golang.org/x/sys v0.30.0 // indirect
 )
```
ndyakov:
```suggestion
go.uber.org/atomic v1.11.0

require (
	github.com/klauspost/cpuid/v2 v2.2.10 // indirect
	golang.org/x/sys v0.30.0 // indirect
)
```

go.sum
```diff
@@ -1,3 +1,5 @@
+github.com/alicebob/miniredis/v2 v2.37.0 h1:RheObYW32G1aiJIj81XVt78ZHJpHonHLHW7OLIshq68=
+github.com/alicebob/miniredis/v2 v2.37.0/go.mod h1:TcL7YfarKPGDAthEtl5NBeHZfeUQj6OXMm/+iu5cLMM=
```
ndyakov:
```suggestion
```

go.sum
```diff
@@ -14,6 +16,8 @@ github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZb
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+github.com/yuin/gopher-lua v1.1.1 h1:kYKnWBjvbNP4XLT3+bPEwAXJx262OhaHDWDVOPjL46M=
+github.com/yuin/gopher-lua v1.1.1/go.mod h1:GBR0iDaNXjAgGg9zfCvksxSRnQx76gclCIb7kdAd1Pw=
```
ndyakov:
```suggestion
```

script_server_sha_integration_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	rdb := redis.NewClient(&redis.Options{Addr: mr.Addr()})
+	defer rdb.Close()
+
+	s := redis.NewScriptServerSHA("return 1")
+
+	if err := s.RunRO(ctx, rdb, nil).Err(); err != nil {
+		t.Fatalf("RunRO failed: %v", err)
+	}
+}
```
ndyakov:
```suggestion
package redis_test

import (
	. "github.com/bsm/ginkgo/v2"
	. "github.com/bsm/gomega"
	"github.com/redis/go-redis/v9"
)

var _ = Describe("NewScriptServerSHA", func() {
	var client *redis.Client

	BeforeEach(func() {
		client = redis.NewClient(redisOptions())
		Expect(client.FlushDB(ctx).Err()).NotTo(HaveOccurred())
	})

	AfterEach(func() {
		Expect(client.Close()).NotTo(HaveOccurred())
	})

	It("should reload script after SCRIPT FLUSH", func() {
		s := redis.NewScriptServerSHA("return 1")

		// First run should succeed (loads script via SCRIPT LOAD)
		err := s.Run(ctx, client, nil).Err()
		Expect(err).NotTo(HaveOccurred())

		// Flush all scripts from Redis
		err = client.Do(ctx, "SCRIPT", "FLUSH").Err()
		Expect(err).NotTo(HaveOccurred())

		// Second run should still succeed (reloads script after NOSCRIPT)
		err = s.Run(ctx, client, nil).Err()
		Expect(err).NotTo(HaveOccurred())
	})

	It("should run script with RunRO", func() {
		s := redis.NewScriptServerSHA("return 1")

		err := s.RunRO(ctx, client, nil).Err()
		Expect(err).NotTo(HaveOccurred())
	})

	It("should run script with EvalSha", func() {
		s := redis.NewScriptServerSHA("return KEYS[1]")

		result, err := s.EvalSha(ctx, client, []string{"mykey"}).Result()
		Expect(err).NotTo(HaveOccurred())
		Expect(result).To(Equal("mykey"))
	})

	It("should run script with EvalShaRO", func() {
		s := redis.NewScriptServerSHA("return ARGV[1]")

		result, err := s.EvalShaRO(ctx, client, nil, "myarg").Result()
		Expect(err).NotTo(HaveOccurred())
		Expect(result).To(Equal("myarg"))
	})

	It("should cache hash after first load", func() {
		s := redis.NewScriptServerSHA("return 42")

		// Hash should be empty before first run
		Expect(s.Hash()).To(Equal(""))

		// Run the script
		result, err := s.Run(ctx, client, nil).Result()
		Expect(err).NotTo(HaveOccurred())
		Expect(result).To(Equal(int64(42)))

		// Hash should be populated after run
		Expect(s.Hash()).NotTo(Equal(""))
	})
})
```

panga:
@chaitanyabodlapati thanks for pushing the fix 🙌

i’m seeing the same issue on my side and noticed a few subtle concerns in the PR:

1. `script load` is only needed once for the lifetime of the script to compute the hash and store it for later use. it won’t negatively impact non-FIPS mode. in fact, it could be the main way we load scripts, so no need to branch into a server-side implementation.
2. if `script load` errors, we shouldn’t swallow it and then run `eval` under the hood. that’s risky, can hide subtle code issues, and may force the unoptimized path every time.
3. in cluster mode, `script load` runs on a random node, so it can’t be used to fix `NOSCRIPT`. instead, just call `eval` so it gets cached on the correct shard node based on the key, and subsequent calls will have it.

TLDR: `script load` only needs to run once and we should store the hash (sha1) to avoid computing it client-side.

ndyakov:
@panga 1) and 2) sound like application logic, not client logic. In my opinion, since this is achievable with what the client provides it can be implemented in the application. 3) is something that can be addressed by getting a specific node client with `MasterForKey` as well.

panga:
My point on 3) is that `EVAL` encapsulates the full workflow: 1. locate the node for the key, 2. load the script (and cache), and 3. execute it.

This avoids introducing custom coordination logic and eliminates an additional Redis round-trip.

chaitanyabodlapati:
> left the required changes as suggestions
> 
> i was planning to address this, since my review was what delayed the merge of this PR, but the `maintainer_can_modify` is `false`, so @chaitanyabodlapati, feel free to address this and ping me when it is done.

@ndyakov  This PR is coming from a fork, so I don't have the ability to enable "maintainer can modify" on my side.
I've pushed the requested changes. When you get a chance, could you please take another look and let me know if there's anything else you'd like me to update?

panga:
@ndyakov I noticed my comment about cluster mode doesn’t seem to have been addressed in the merged code. Is this something you plan to look at before releasing?

ndyakov:
@panga feel free to submit an issue. It is a known problem for the team but with lower priority compared to other work we are doing at this time. If someone from the community would like to take a shot at it that would be great.

PR #3702, opened 2026-02-04
fix(redisotel): use per-node config in cluster/ring metrics

akash329d:
In cluster/ring mode, `registerClient` is called once per node via `OnNewNode`. The shared `*config` pointer was being mutated (`conf.poolName` set, `conf.attrs` appended to), so every node ended up reporting the first node's address as `pool.name`.

Shallow-copy the config per node and build a separate attrs slice so each node gets its own `pool.name` attribute.

### DIFF
--- a/extra/redisotel/metrics.go
+++ b/extra/redisotel/metrics.go
@@ -93,29 +93,19 @@ func registerClient(rdb *redis.Client, conf *config, state *metricsState) error 
 		}
 	}
 
-	// Determine pool name for this specific node. Use the node's address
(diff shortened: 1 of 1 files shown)


ndyakov:
hello @akash329d , can you check https://github.com/redis/go-redis/pull/3699 . I do think it solves the same problem?

Jesse-Bonfire:
Looks like it solves the same problem to me :)

akash329d:
Can close if it solves the same problem :)

PR #3713, opened 2026-02-18
Add DoRaw and DoRawWriteTo for raw RESP protocol access

### LINKED ISSUE #3698: Feature Request: allow custom commands via Cmder interface
I would like the ability to implement an HGetAll command with custom deserialization logic to unmarshal Redis responses directly into structs, avoiding the intermediate `map[string]string` allocation (as used in `MapStringStringCmd`).

Currently, this is not possible because `redis.Cmder` contains unexported methods, specifically `readReply(*proto.Reader)`, which prevents implementing custom command types.

We're currently using [radix](https://github.com/mediocregopher/radix), which allows custom deserialization and saves us significant allocations in hot paths. We would like to migrate to go-redis, but the inability to implement custom deserialization is a blocker for adoption.

ofekshenawa:
Adds two new methods for accessing raw RESP protocol bytes:

- `DoRaw(ctx, args...)` - Returns raw RESP bytes, useful for custom parsing
- `DoRawWriteTo(ctx, writer, args...)` - Streams RESP bytes directly to an io.Writer without intermediate allocations

Add RawCmd and RawWriteToCmd types that return raw RESP protocol bytes
without parsing, allowing users to implement custom deserialization.

Closes #3698

---

> [!NOTE]
> **Medium Risk**
> Adds new raw-reply execution paths and modifies core retry behavior (single and pipelined, including cluster/ring) to honor a new `NoRetry` contract, which could affect failure handling if misapplied.
> 
> **Overview**
> Adds `DoRaw` and `DoRawWriteTo` APIs plus new command types (`RawCmd`, `RawWriteToCmd`) to let callers receive raw RESP replies as bytes or stream them directly to an `io.Writer`.
> 
> Implements raw-reply reading in `proto.Reader` (buffered and streaming, including nested types and RESP3 attributes) and introduces a `Cmder.NoRetry()` flag; retry loops for standalone commands and pipelines (including cluster/ring) now skip retries when any command is marked non-retryable to avoid partial-write corruption. Tests are expanded to validate raw outputs across common reply types and the new no-retry behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4150e198b1cf3c5b7b3b8fe5e45d76b567693d55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -1300,7 +1300,7 @@ func (c *ClusterClient) process(ctx context.Context, cmd Cmder) error {
 			continue
 		}
 
-		if shouldRetry(lastErr, cmd.readTimeout() == nil) {
+		if shouldRetry(lastErr, cmd.readTimeout() == nil) && !cmd.NoRetry() {
 			// First retry the same node.
 			if attempt == 0 {
 				continue
--- a/ring.go
+++ b/ring.go
@@ -802,7 +802,7 @@ func (c *Ring) process(ctx context.Context, cmd Cmder) error {
 		}
 
 		lastErr = shard.Client.Process(ctx, cmd)
-		if lastErr == nil || !shouldRetry(lastErr, cmd.readTimeout() == nil) {
+		if lastErr == nil || !shouldRetry(lastErr, cmd.readTimeout() == nil) || cmd.NoRetry() {
 			return lastErr
 		}
 	}


command_test.go
```diff
@@ -94,3 +95,201 @@ var _ = Describe("Cmd", func() {
...
+	It("streams raw RESP bytes for HGETALL", func() {
+		client.HSet(ctx, "rawhash", "f1", "v1", "f2", "v2")
+		var buf bytes.Buffer
+		cmd := client.DoRawWriteTo(ctx, &buf, "HGETALL", "rawhash")
+		written, err := cmd.Result()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(written).To(BeNumerically(">", 0))
+		raw := buf.String()
+		Expect(raw).To(ContainSubstring("f1"))
+		Expect(raw).To(ContainSubstring("v1"))
```
ndyakov, line 455: can we test the whole response here?

internal/proto/reader.go
```diff
@@ -646,3 +646,144 @@ func IsNilReply(line []byte) bool {
...
+
+// ReadRawReplyWriteTo streams the next RESP reply directly to w without intermediate allocations.
+// Returns the number of bytes written and any error encountered.
+func (r *Reader) ReadRawReplyWriteTo(w io.Writer) (int64, error) {
+	return r.readRawReplyWriteTo(w)
+}
+
+func (r *Reader) readRawReplyWriteTo(w io.Writer) (int64, error) {
+	line, err := r.readLine()
+	if err != nil {
```
ndyakov, line 747: maybe we should not provide any guarantee of the response and can directly use `return io.Copy(w, r.rd)` ?

ndyakov: `io.Copy` has optimizations:
`io.Copy` checks if the source implements `io.WriterTo`. Since `bufio.Reader` (`r.rd`) implements `WriterTo`, `io.Copy` will call `WriteTo()` instead of repeatedly calling `Read()` (what we do).
The `bufio.Reader.WriteTo()` method:
First writes any already-buffered data to the destination
Then copies directly from the underlying reader to the writer, bypassing the buffer for efficiency

ofekshenawa: Using io.Copy(w, r.rd) directly won't work because Redis uses persistent connections with no EOF between responses. When you pipeline commands like PING, GET foo, and INCR counter, the wire contains all responses concatenated: +PONG\r\n$5\r\nhello\r\n:42\r\n. Since io.Copy reads until EOF (which only occurs when the connection closes), the first command would "steal" all 23 bytes, leaving nothing for subsequent commands—causing data corruption in pipelines. For single commands, it would block forever waiting for an EOF that never comes.

ndyakov: Ok, you are correct. The case for pipelines won't be a problem for Raw commands, as if the raw command is a pipeline - responding with everything in the buffer is what it is expected. As for if the raw command is a part of a pipeline - we can forbid this by not exposing the `DoRaw...` in the `Pipeliner`. The issue, however, is that it will block until EOF and we cannot use `CopyN` easily, since we don't know the length of the whole response. Good finding! I do think there may be a room for improvement here if we decide to introduce our own Buffer Reader which can optimize this a bit further. Thank you!

ndyakov (dismissed):

ndyakov:
@fdimuccio please take a look if this covers your usecase. I am willing to proceed with this PR and merge it.

fdimuccio:
@ndyakov it looks perfect, DoRawWriteTo is exactly what I was looking for. Thank you!

redis.go
```diff
@@ -703,7 +703,9 @@ func (c *baseClient) process(ctx context.Context, cmd Cmder) error {
 		if cn != nil {
 			lastConn = cn
 		}
-		if err == nil || !retry {
+		// Don't retry if command explicitly disables retries (e.g., RawWriteToCmd
+		// which writes directly to an io.Writer and cannot undo partial writes)
+		if err == nil || !retry || cmd.NoRetry() {
```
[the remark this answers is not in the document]
ndyakov: cc @ofekshenawa maybe just take a look at that.

PR #3717, opened 2026-02-23
feat: add ShardedPubSub for multi-shard channel subscriptions

### LINKED ISSUE #3133: Sharded pubsub subscriber cannot get messages from all subscribed channels

## Expected Behavior

Sharded pubsub subscriber should get messages from all channels that it's subscribed to.

## Current Behavior

Sharded pubsub subscriber can only get messages from the shard that the first subscribed channel belongs.

## Possible Solution

Have sharded pubsub subscriber listen to all shards to which its subscribed channels belong.

## Steps to Reproduce

1. Start Redis Server with cluster node on.
2. Run the go script given below.
3. Observe that the ssubscriber can only receive messages from a subset of channels that it has subscribed to.

## Context (Environment)

This happens across different environments as long as a Redis cluster is used.

## Detailed Description

Code to reproduce this bug:

```
package main

import (
    "context"
    "flag"
    "fmt"
    "os"

    "github.com/google/logger"
)

const logPath = "./ssub.log"

var verbose = flag.Bool("verbose", true, "print info level logs to stdout")

func main() {
    lf, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0660)
    if err != nil {
        logger.Fatalf("Failed to open log file: %v", err)
    }
    defer lf.Close()
    defer logger.Init("SsubLogger", *verbose, true, lf).Close()
    logger.Info("")

    // Create a new RedisCluster client
    rdb := redis.NewClusterClient(&redis.ClusterOptions{
        Addrs: []string{
            "localhost:6379",
        },
    })

    // Ping the Redis Cluster to ensure connectivity
    if err := rdb.Ping(context.Background()).Err(); err != nil {
        logger.Infof("Could not connect to Redis Cluster: %v\n", err)
    }
    logger.Info("RedisCluster client is ready")

    pubsub := rdb.SSubscribe(context.Background())
    defer pubsub.Close()

    // Ssubscribe to ch 1 - 9
    for i := 1; i <= 9; i++ {
        ch := fmt.Sprintf("ch%d", i)

        pubsub.SSubscribe(context.Background(), ch)
        logger.Info("Subscribed to ", ch)

        rdb.SPublish(context.Background(), ch, "Hello")
    }

    chss, _ := rdb.PubSubShardChannels(context.Background(), "ch*").Result()
    logger.Info("Currently subscribed channels: ", chss)

    for {
        // logger.Info("Waiting for sharded Redis pub-sub messages")
        // ReceiveMessage will crash when SSubscribe is used.
        // msg, err := pubsub.ReceiveMessage(context.Background())
        // if err != nil {
        //  logger.Info(err)
        // }
        msg := <-pubsub.Channel()
        logger.Infof("Received %s %s", msg.Channel, msg.Payload)
    }
}
```

Output:
```
INFO : 2024/09/25 15:35:14.774884 ssub.go:24:
INFO : 2024/09/25 15:35:14.781693 ssub.go:37: RedisCluster client is ready
INFO : 2024/09/25 15:35:14.782690 ssub.go:47: Subscribed to ch1
INFO : 2024/09/25 15:35:14.783934 ssub.go:47: Subscribed to ch2
INFO : 2024/09/25 15:35:14.784396 ssub.go:47: Subscribed to ch3
INFO : 2024/09/25 15:35:14.784838 ssub.go:47: Subscribed to ch4
INFO : 2024/09/25 15:35:14.785478 ssub.go:47: Subscribed to ch5
INFO : 2024/09/25 15:35:14.785714 ssub.go:47: Subscribed to ch6
INFO : 2024/09/25 15:35:14.785889 ssub.go:47: Subscribed to ch7
INFO : 2024/09/25 15:35:14.786079 ssub.go:47: Subscribed to ch8
INFO : 2024/09/25 15:35:14.786246 ssub.go:47: Subscribed to ch9
INFO : 2024/09/25 15:35:14.786489 ssub.go:53: Currently subscribed channels: []
INFO : 2024/09/25 15:35:14.786744 ssub.go:63: Received ch1 Hello
INFO : 2024/09/25 15:35:14.786808 ssub.go:63: Received ch2 Hello
INFO : 2024/09/25 15:35:14.887868 ssub.go:63: Received ch5 Hello
INFO : 2024/09/25 15:35:14.888170 ssub.go:63: Received ch6 Hello
INFO : 2024/09/25 15:35:14.989075 ssub.go:63: Received ch9 Hello
```

veeceey:
Fixes #3133

When using sharded pub/sub in a Redis cluster, channels hash to different slots served by different nodes. The current `SSubscribe` creates a single `PubSub` connection that only connects to the shard of the first channel, so messages from channels on other shards are never received.

This adds a `ShardedPubSub` type and a `ClusterClient.SSubscribeSharded` method that properly handles this. It:

- Groups channels by their hash slot to determine which cluster node serves each
- Maintains one `PubSub` connection per shard node
- Multiplexes messages from all shards into a single `Channel()` for consumption

```go
sps := rdb.SSubscribeSharded(ctx, "ch1", "ch2", "ch3")
defer sps.Close()

// Add more channels later - routed to correct shard automatically
sps.SSubscribe(ctx, "ch4", "ch5")

for msg := range sps.Channel() {
    fmt.Println(msg.Channel, msg.Payload)
}
```

The existing `SSubscribe` returning `*PubSub` is unchanged for backward compatibility. Added a doc comment noting the single-shard limitation and pointing to `SSubscribeSharded`.

Tests confirm ch1-ch9 hash to 9 different slots (matching the bug report).

---

> [!NOTE]
> **Medium Risk**
> New cluster connection lifecycle and topology migration logic can affect long-lived subscribers (brief duplicate messages during failover); core PubSub SSUBSCRIBE behavior changes for multi-slot batches but existing single-shard SSubscribe API is preserved.
> 
> **Overview**
> Adds **`ShardedPubSub`** and **`ClusterClient.SSubscribeSharded`** so cluster clients can subscribe to shard channels that hash to **different slots/nodes**, multiplexing messages into one `Channel()`. Existing **`SSubscribe`** is unchanged; docs now warn about the single-shard limitation.
> 
> **`PubSub`** shard subscribe/unsubscribe now **groups channels by hash slot** (avoids CROSSSLOT) and only updates local `schannels` for slot groups that succeeded on the server.
> 
> **`ClusterClient`** registers sharded pub/subs, runs a **coalesced async notifier** on cluster state reload to migrate subscriptions, adds **`pubSubForAddr`** for pinned per-node connections, and **closes registered `ShardedPubSub` instances** on client close.
> 
> Includes **`example/sharded-pubsub`**, integration test for multi-shard delivery, and extensive unit tests for error paths and topology notification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abb1138cc09f9288d31cbef0a032925f1a42fb1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


veeceey:
**Test results:**

```
$ go build ./...
(clean, no errors)

$ go vet ./...
(clean)

$ go test -run "TestShardedPubSub" -count=1 -v .
=== RUN   TestShardedPubSubChannelGrouping
    sharded_pubsub_test.go:25: channels hash to 9 different slots:
    map[1548:[ch7] 1672:[ch3] 5677:[ch6] 5801:[ch2] 9806:[ch5] 9930:[ch1] 10178:[ch9] 13935:[ch4] 14307:[ch8]]
--- PASS: TestShardedPubSubChannelGrouping (0.00s)
=== RUN   TestShardedPubSubNewAndClose
--- PASS: TestShardedPubSubNewAndClose (0.00s)
=== RUN   TestShardedPubSubSSubscribeWhenClosed
--- PASS: TestShardedPubSubSSubscribeWhenClosed (0.00s)
PASS

$ go test -run "TestUnit" -count=1 ./...
(all existing tests pass)
```

The slot grouping test confirms that ch1-ch9 all hash to different slots, matching the original bug report. With `ShardedPubSub`, each gets its own connection to the correct node.

ndyakov:
Thank you @veeceey, we planned to work on the pubsub soon and this comes at just the right time, will review your PR later this week!

sharded_pubsub.go
```diff
@@ -0,0 +1,205 @@
...
+		groups[addr] = append(groups[addr], ch)
+		s.chanShard[ch] = addr
+	}
+
+	// Subscribe on the correct shard for each group.
+	for addr, chs := range groups {
+		ps := s.getOrCreateShard(addr)
+		if err := ps.SSubscribe(ctx, chs...); err != nil {
+			return err
+		}
```
ndyakov: this error can be related to a network issue, do you think it will be better to find a way to reinitialise the pubsub if that is the case.

sharded_pubsub.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return node.Client.opt.Addr, nil
+}
+
+// getOrCreateShard returns an existing PubSub for the given node address,
+// or creates a new one.
+func (s *ShardedPubSub) getOrCreateShard(addr string) *PubSub {
+	if ps, ok := s.shards[addr]; ok {
+		return ps
+	}
```
ndyakov, line 136: Without any type of healthcheck here we cannot be sure if this pubsub connection is still opened.

sharded_pubsub.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/redis/go-redis/v9/internal/pool"
+)
+
+// ShardedPubSub wraps multiple PubSub connections to support sharded pub/sub
+// across different cluster nodes. In a Redis cluster, channels hash to
+// different slots which may be served by different nodes. A single connection
+// can only receive messages for channels on the shard it is connected to.
+// ShardedPubSub transparently manages one PubSub per shard and multiplexes
+// all messages into a single Go channel.
+type ShardedPubSub struct {
```
ndyakov, line 56: Currently I don't see how this `ShardedPubSub` will react on topology changes (e.g. nodes are moved => shards are on a different address / new nodes are added => shards now split between nodes). Any plans or ideas how to address that with the current implementation?

ndyakov (commented):
@veeceey I did leave a short review and would like for us to discuss and improve this implementation a bit. Let me know if you need any help in doing so.

sharded_pubsub.go
veeceey: Good point — you're right that a network error here could be transient and we shouldn't just surface it and give up.

I think the approach would be to catch network-level errors (connection refused, timeouts, EOF, etc.) during `SSubscribe` on the underlying `PubSub`, and when that happens:

1. Close the failed `PubSub` for that shard address
2. Remove it from the `shards` map so the next attempt gets a fresh connection
3. Re-resolve the node address for the affected channels (since the cluster state may have changed)
4. Retry the subscribe with a backoff, reusing the cluster's `retryBackoff` settings for consistency

This way transient network blips get retried transparently, and if the node is genuinely down the cluster state refresh in step 3 should route us to the new primary.

For the `forwardMessages` goroutine, I'd add similar logic — if the channel read returns an error/closes unexpectedly, attempt to re-establish that shard's `PubSub` and re-subscribe to the channels that were on it, rather than silently dropping messages.

Happy to implement this — I'll model the retry logic after how the existing `PubSub.reconnect` works internally so it stays consistent with the rest of the library.

sharded_pubsub.go
veeceey: That's a valid concern. Right now we create the `PubSub` via `s.cluster.pubSub()` and hand it off, but there's no ongoing verification that the connection is still alive.

I think the right approach is to add a `Ping` method on `ShardedPubSub` that iterates over all shard connections and pings each one. Something like:

```go
func (s *ShardedPubSub) Ping(ctx context.Context) error {
    s.mu.Lock()
    defer s.mu.Unlock()

    for addr, ps := range s.shards {
        if err := ps.Ping(ctx); err != nil {
            // connection is dead — trigger reconnect for this shard
            return fmt.Errorf("shard %s health check failed: %w", addr, err)
        }
    }
    return nil
}
```

On top of that, I could add an optional background health check loop (controlled by an option, disabled by default to avoid surprises) that periodically pings each shard and triggers reinitialization for any that have gone stale. This would pair well with the reconnect logic from the error handling comment — if a health check fails, we close the dead `PubSub`, re-resolve the node, and re-subscribe.

Makes sense to have at least the manual `Ping` method in this PR. Want me to add both the manual and the optional background health check, or keep the background one for a follow-up?

sharded_pubsub.go
veeceey: Great question — this is the piece I knew would come up and it's definitely the trickiest part.

Currently the `shards` map is keyed by the node address resolved at subscribe time, and that mapping is completely static after that. If a slot migrates to a different node, we'd keep reading from the old node (which would stop receiving messages for those channels) and never know about it.

Here's my plan for handling this:

**1. Hook into cluster state refresh**
The `ClusterClient` already periodically reloads cluster state via `state.Get()`. After each topology refresh, `ShardedPubSub` should re-resolve every channel's node address and compare against the current mapping. For any channel whose node has changed:
- Subscribe to the channel on the new node's `PubSub`
- Unsubscribe from the old node's `PubSub`
- If an old node's `PubSub` has no remaining channels, close it and remove it from the map

**2. Listen for MOVED responses**
If any operation returns a `MOVED` error, that's an immediate signal that topology has shifted. This could trigger an on-demand re-resolve for the affected shard rather than waiting for the next periodic refresh.

**3. Notification channel for resharding events**
This might be useful for consumers who want to know when migrations happened, but that feels like it could be a follow-up.

I think items 1 and 2 are important to get right in this PR. The implementation would look something like adding a `refreshTopology(ctx)` method that gets called either periodically (piggy-backing on the cluster's reload interval) or reactively on errors. I'll work on this — would appreciate your input on whether you'd prefer it as a periodic goroutine or event-driven off the cluster state reload.

ndyakov: @veeceey I do think that we should have the ClusterClient as the source of truth for the topology and that ShardedPubSub should update (or just recalculate, if an update is not needed) its map when ClusterClient updates its topology. Hence, something like an event driven approach would fit best here. 

Since the `ShardedPubSub` is a specific tool used only in the `ClusterClient` feel free to have a set (slice) of all `ShardedPubSubs` somehowe registered on the `ClusterClient` and trigger their topology changes when the ClusterClient updates its topology. It would be way simpler, somehowe limiting (to just covering this case, instead of generic event channel), but a good balance in my mind. 

Let me know if you find any problems with this approach or thing another one fits better, I am willing to review alternatives as well.

veeceey: that makes a lot of sense — having the ClusterClient own the topology and push updates to registered ShardedPubSubs is way cleaner than each pubsub polling independently. I like the simplicity of a slice of registered instances on the ClusterClient.

I'm thinking something like:

- add a `shardedPubSubs []*ShardedPubSub` field on ClusterClient (protected by the existing mutex)
- ShardedPubSub registers itself on creation, deregisters on Close
- when ClusterClient refreshes topology (in `reloadState` or wherever that happens), it calls a `onTopologyChange()` method on each registered ShardedPubSub
- that method re-resolves channels to slots/nodes and migrates subscriptions as needed

I don't see any issues with this approach — it keeps things scoped to exactly this use case without over-engineering a generic event system. I'll start working on it and push an update. appreciate the guidance!

veeceey:
friendly ping on this PR

ndyakov:
@veeceey Thank you for pinging me. I do think the discussed approaches sound good. Would you like to implement them in this PR? Maybe I am missing something, I am not seeing them committed here?

sharded_pubsub.go
ndyakov: I think we can skip the health check for now and fail (change the underlying connection) if `SSUBSCRIBE` fails. In practice, making the "healtcheck" reactive than proactive.

veeceey:
Hey @ndyakov, just pushed the implementation! Here's what's in:

**Topology change handling:**
- ShardedPubSub registers itself with the ClusterClient on creation, deregisters on Close
- Added an `onReload` callback to `clusterStateHolder` — after each successful state reload, ClusterClient calls `onTopologyChange()` on all registered ShardedPubSubs
- `onTopologyChange()` re-resolves every channel's node address and migrates subscriptions when slots have moved (unsubscribe from old node, subscribe on new, close empty shards)

**Reactive reconnect (no proactive health check, as you suggested):**
- If `SSubscribe` fails on a shard, it closes the failed connection, re-resolves the channels, and retries on a fresh connection
- No background pinging — we only reconnect when something actually fails

**Tests:**
- Added `TestShardedPubSubRegistration` — verifies register on create, deregister on close, and correct bookkeeping

All existing tests pass, `go vet` clean. Let me know if you'd like anything adjusted!

sharded_pubsub.go
veeceey: makes sense -- went with the reactive approach. if SSubscribe fails, we close the broken shard, re-resolve the channels, and retry on a fresh connection. no background pinging.

veeceey:
Good catches from the bugbot review — just pushed fixes for all three issues:

**Close race (send-on-closed-channel):** Added a `sync.WaitGroup` to track forwarder goroutines. `Close()` now waits for all forwarders to exit before closing `msgCh`, preventing the panic.

**Channel opts ignored:** `Channel(opts...)` now extracts the configured `chanSize` from the options and forwards them to each underlying PubSub shard's `Channel()` call. Options are also stored so newly created shards get the same config.

**Deadlock in SSubscribe/onTopologyChange:** Moved `nodeAddrForChannel` calls outside the mutex in both `SSubscribe` and `onTopologyChange`. Address resolution can trigger `state.Get` -> `Reload` -> `onReload` -> `notifyShardedPubSubs` -> `onTopologyChange`, which needs `s.mu` — holding it during resolution would deadlock. Both methods now resolve addresses first, then acquire the lock for state mutations.

All existing tests pass, `go vet` clean.

osscluster.go
```diff
@@ -1185,6 +1194,38 @@ func NewClusterClient(opt *ClusterOptions) *ClusterClient {
...
+func (c *ClusterClient) deregisterShardedPubSub(sps *ShardedPubSub) {
+	c.spsMu.Lock()
+	defer c.spsMu.Unlock()
+	for i, s := range c.shardedPubSubs {
+		if s == sps {
+			c.shardedPubSubs = append(c.shardedPubSubs[:i], c.shardedPubSubs[i+1:]...)
+			return
+		}
+	}
+}
```
ndyakov, line 1238: a slight optimization with a map (or even sync.Map) can be done here, but I think this is fine for now. I don't think we can expect such number of sharded pubsubs that looping over them may be a problem.

ndyakov (dismissed):
@veeceey Thank you! Overall this PR looks good. I will do a more in-depth review in the following days.

sharded_pubsub.go
```diff
@@ -0,0 +1,371 @@
...
+	for ch, addr := range addrMap {
+		s.chanShard[ch] = addr
+	}
+
+	// Subscribe on the correct shard for each group.
+	for addr, chs := range groups {
+		ps := s.getOrCreateShard(addr)
+		if err := ps.SSubscribe(ctx, chs...); err != nil {
+			// Reactive reconnect: close failed shard, re-resolve, and retry once.
+			if reconnErr := s.resubscribeShard(ctx, addr); reconnErr != nil {
```
ndyakov, line 130: `resubscribeShard` is called while holding `s.mu`, but internally it calls `nodeAddrForChannel`, which calls `slotMasterNode`, which calls `state.Get`.
- If the cluster state is `nil`, Get calls `Reload` which will eventually call `notifyShardedPubSubs`, which calls `onTopologyChange` trying to acquire `s.mu`.

Release the lock before the retry, similar to how the main SSubscribe flow already does it for address resolution.

sharded_pubsub.go
```diff
@@ -0,0 +1,371 @@
...
+		}
+	}
+
+	if len(migrations) == 0 {
+		return
+	}
+
+	// Apply migrations under the lock.
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
ndyakov, line 262: since we will hold the lock for applying the migration, does it make sense to verify the channels still exist when subscribing?

```
    if _, exists := s.chanShard[ch]; exists {
        s.chanShard[ch] = newAddr
    }
```

sharded_pubsub.go
```diff
@@ -0,0 +1,371 @@
...
+
+	if s.closed {
+		return pool.ErrClosed
+	}
+	s.closed = true
+	close(s.exit)
+
+	// Deregister from the cluster client.
+	s.cluster.deregisterShardedPubSub(s)
```
ndyakov: I think we can deregister before acquiring the `	s.mu.Lock()`

osscluster.go
```diff
@@ -1059,6 +1061,9 @@ func (c *clusterStateHolder) Reload(ctx context.Context) (*clusterState, error)
 		return nil, err
 	}
 	c.state.Store(state)
+	if c.onReload != nil {
+		c.onReload()
+	}
```
ndyakov, line 1066: Can this be async? maybe with validating that it is not triggered multiple times. If you look at `LazyReload` it has a mechanism for executing the latest reload, maybe something similar here would help.

sharded_pubsub.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/redis/go-redis/v9/internal/pool"
+)
+
+// ShardedPubSub wraps multiple PubSub connections to support sharded pub/sub
+// across different cluster nodes. In a Redis cluster, channels hash to
+// different slots which may be served by different nodes. A single connection
+// can only receive messages for channels on the shard it is connected to.
+// ShardedPubSub transparently manages one PubSub per shard and multiplexes
+// all messages into a single Go channel.
+type ShardedPubSub struct {
```
ndyakov, line 19: The existing `PubSub` supports `Ping`, `ReceiveMessage` and `ChannelWithSubscriptions`. ShardedPubSub only exposes `Channel()`. Can you either add them for feature parity or document that they are intentionally not supported? I would like to have a `PubSub` interface in the future (for v10 for example), where we can return the same Interface for the pubsub and sharded pubsub.

sharded_pubsub.go
```diff
@@ -0,0 +1,371 @@
...
+		state, err := s.cluster.state.Get(ctx)
+		if err != nil {
+			return "", err
+		}
+		node, err := s.cluster.slotReadOnlyNode(state, slot)
+		if err != nil {
+			return "", err
+		}
+		return node.Client.opt.Addr, nil
+	}
```
ndyakov, line 60: While theoretically we should connect to the replica on read only, I am wondering if we want to? The master will receive the message first. I will validate if this is needed at all or should be changed, but for now let's keep it as it is.

sharded_pubsub.go
```diff
@@ -0,0 +1,371 @@
...
+			}
+
+			// Subscribe on new shard.
+			ps := s.getOrCreateShard(newAddr)
+			_ = ps.SSubscribe(ctx, channels...)
+
+			// Update channel->shard mapping.
+			for _, ch := range channels {
+				s.chanShard[ch] = newAddr
+			}
```
ndyakov, line 288: If SSubscribe fails (error is swallowed with _), the channel is now:

- Unsubscribed from the old shard
- NOT subscribed on the new shard
- BUT chanShard is updated to point to the new shard
The channel is effectively lost - no messages will be received, and the code thinks everything is fine.

Can we update the mapping ONLY on successful `SSubscribe` here?

ndyakov (commented):
@veeceey Thank you for working on this and being patient about my review. It took some time to get my head around what we would like to do for `v10` and how this will fit. I did leave some comments (few of which I do consider blockers). Let me know what you think and if you would like to continue working on it.

sharded_pubsub.go
veeceey: good catch — you're right, resubscribeShard holding s.mu while calling nodeAddrForChannel creates the same deadlock path as the original SSubscribe issue. I'll release the lock before the retry and re-acquire after, same pattern as the main SSubscribe flow.

sharded_pubsub.go
veeceey: yep, makes total sense — if someone unsubscribed while we were resolving, we shouldn't silently re-add it. will add the existence check before updating chanShard.

sharded_pubsub.go
veeceey: agreed — deregistering doesn't need the lock and doing it first avoids holding s.mu longer than necessary. will move it up.

osscluster.go
veeceey: yeah, making it async would avoid blocking the reload path. I'll look at LazyReload's dedup mechanism and model it similarly — signal a goroutine to do the migration instead of doing it inline, with a check to skip if one's already in-flight.

sharded_pubsub.go
veeceey: that's a fair point for v10 interface parity. I'll add Ping and ReceiveMessage methods on ShardedPubSub — Ping iterates over shards and pings each, ReceiveMessage reads from the merged channel. ChannelWithSubscriptions I can stub for now with a TODO since the subscription tracking is a bit different with sharded channels. Will document what's supported vs not.

sharded_pubsub.go
veeceey: sounds good, I'll leave it as-is for now then. makes sense to validate whether replica reads are actually useful for pubsub before changing it.

sharded_pubsub.go
veeceey: you're right, that's a nasty silent failure. I'll restructure so we only update chanShard[ch] = newAddr after SSubscribe succeeds. if it fails, we keep the old mapping so at least the channel isn't orphaned — and we can log or return the error so it's visible.

veeceey:
pushed all the fixes from the review:

- **resubscribeShard deadlock**: releases s.mu before nodeAddrForChannel, re-acquires after — same pattern as main SSubscribe flow
- **chanShard race in migration**: added existence check before updating, and only update on successful SSubscribe
- **Close ordering**: deregister from ClusterClient before acquiring s.mu
- **async topology notification**: notifyShardedPubSubs now runs in a goroutine with a CAS guard so it doesn't block the reload path
- **interface parity**: added Ping, ReceiveMessage, and ReceiveTimeout methods on ShardedPubSub

all tests pass, go vet clean. let me know if there's anything else!

sharded_pubsub.go
```diff
@@ -0,0 +1,442 @@
...
+				_ = ps.SUnsubscribe(ctx, channels...)
+			}
+
+			// Subscribe on new shard.
+			ps := s.getOrCreateShard(newAddr)
+			if err := ps.SSubscribe(ctx, channels...); err != nil {
+				// SSubscribe failed — keep the old mapping so the channel
+				// isn't orphaned with an incorrect chanShard entry.
+				continue
+			}
```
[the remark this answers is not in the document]
ndyakov: @veeceey i haven't personally checked the code yet, but maybe those make sense to be checked?

ndyakov:
@veeceey Hey, do you expect to continue the work on this one or we should take over?

osscluster.go
```diff
@@ -2248,6 +2337,54 @@ func (c *ClusterClient) pubSub() *PubSub {
...
+// needs every (re)connection to land exactly on the node its bookkeeping is
+// keyed by — the channel-derived resolution in pubSub() could pick a
+// different node (e.g. another replica) on reconnect.
+func (c *ClusterClient) pubSubForAddr(addr string) *PubSub {
+	var node *clusterNode
+	pubsub := &PubSub{
+		opt: c.opt.clientOptions(),
+		newConn: func(ctx context.Context, _ string, channels []string) (*pool.Conn, error) {
+			if node != nil {
+				panic("node != nil")
```
ofekshenawa, line 2357: Panic feels heavy here. I think returning an error (or logging and dialing fresh) seems safer.

PR #3719, opened 2026-02-24
fix: prevent non-timeout errors from being treated as timeouts (Fixes #3664)

### LINKED ISSUE #3664: Changed behaviour when wrapped net.Error is returned in a ClusterClient Watch function
I am not sure if this really is a bug or not. Either way it was a change in behaviour that took me by surprise so I figured I would post it to see what you think.

## Expected Behavior

In go-redis versions before v9.17.0 when returning a wrapped net.Error inside a redis.ClusterClient Watch function it was simply propagated out to the surrounding code.

## Current Behavior

Since v9.17.0 when a redis.ClusterClient Watch function returns a wrapped net.Error the Watch code will retry and run the function again. 

This behaviour does not appear if using a regular redis.Client, if the example is changed to use a redis.Client it will pass both on v9.16.0 and v9.17.0.

## Possible Solution

Perhaps it would be possible to only trigger retries on net.Errors caused by Redis operations?

## Steps to Reproduce

The following test illustrates the difference when running with v9.16.0 and v9.17.0. 

```
package main

import (
	"context"
	"net"
	"testing"

	"github.com/alicebob/miniredis/v2"
)

func TestNetErrorInClusterClientWatch(t *testing.T) {
	r := miniredis.RunT(t)

	cl := redis.NewClusterClient(&redis.ClusterOptions{
		Addrs: []string{r.Addr()},
	})

	count := 0
	txFn := func(tx *redis.Tx) error {
		count++
		return fmt.Errorf("%w", net.ErrClosed)
	}

	_ = cl.Watch(context.Background(), txFn, "somekey")

	if count != 1 {
		t.Errorf("txFn called %d times", count)
	}
}
```

## Context (Environment)

When doing network calls unrelated to Redis inside a Watch function they can result in a net.Error. To avoid excessive retries I would have to avoid wrapping the error which makes error-handling harder.

BenediktusG:
## Summary
Fixes an architectural bug where isTimeoutError falsely categorized non-timeout network errors (such as net.ErrClosed or custom user HTTP errors) as transient timeouts.

## Problem
Previously, isTimeoutError used errors.As(err, &netErr) to catch any error implementing the net.Error interface. However, because errors like net.ErrClosed implement this interface but return false for .Timeout(), they were incorrectly allowed to pass through the timeout safety checks in shouldRetry.

If a user's Watch transaction function returned one of these non-timeout network errors, shouldRetry mistakenly returned true, causing the client to unnecessarily retry the transaction until it completely exhausted the `MaxRedirects` limit. (as reported in #3664)

## Solution
Refactored isTimeoutError to strictly evaluate the boolean return value of .Timeout() inline.

Added `&& te.Timeout()` and `&& netErr.Timeout()` to the `errors.As` blocks.

This ensures that only actual timeouts are retried, while cleanly aborting the loop for safely closed connections or custom application network errors.

Related Issues:
Fixes #3664

### DIFF
--- a/error.go
+++ b/error.go
@@ -63,13 +63,13 @@ func isContextError(err error) bool {
 func isTimeoutError(err error) (isTimeout bool, hasTimeoutFlag bool) {
 	// Check for timeoutError interface (works with wrapped errors)
 	var te timeoutError
-	if errors.As(err, &te) {
+	if errors.As(err, &te) && te.Timeout() {
 		return true, te.Timeout()
 	}
 
 	// Check for net.Error specifically (common case for network timeouts)
 	var netErr net.Error
-	if errors.As(err, &netErr) {
+	if errors.As(err, &netErr) && netErr.Timeout() {
 		return true, netErr.Timeout()
 	}
 
--- a/error_test.go
+++ b/error_test.go
@@ -61,7 +61,7 @@ var _ = Describe("error", func() {
 		Expect(redis.ShouldRetry(t1, false)).To(Equal(false))
 
 		t2 := testTimeout{timeout: false}
-		Expect(redis.ShouldRetry(t2, true)).To(Equal(true))
-		Expect(redis.ShouldRetry(t2, false)).To(Equal(true))
+		Expect(redis.ShouldRetry(t2, true)).To(Equal(false))
+		Expect(redis.ShouldRetry(t2, false)).To(Equal(false))
 	})
 })


ndyakov:
Hello @BenediktusG, I would argue that the method is correct as it is, since the second return value should be checked if the error is a timeout error or not. The first return shows if the error implements `Timeout()` method.

BenediktusG:
That makes sense, you are right. `isTimeoutError` already works well. After looking at the code again, I noticed that the `shouldRetry` function (in the timeout check block) returns true when `hasTimeoutFlag` is false. Is this intended behavior, or is this where the bug for `net.ErrClosed` is actually happening?

ndyakov:
@BenediktusG i will double check but this should be the case until max retries - you should retry if encounter any error, the timeout errors are configurable - sometimes you don't want to retry. Closing this PR for now.

PR #3735, opened 2026-03-09
fix: connection closure metrics and enable all metric groups by default

ofekshenawa:
This commit addresses three issues in the redisotel-native metrics implementation:

1. Fix connection closure metrics not appearing in dashboards
   - Root cause: CloseConn() in internal/pool/pool.go was not calling the metricConnectionClosedCallback when closing stale connections
   - Added callback invocation in CloseConn() to record 'stale' closures

2. Verify tracing and metrics compatibility
   - Added TestTracingAndMetricsCompatibility to document that redisotel (tracing via AddHook) and redisotel-native (metrics via SetOTelRecorder) can coexist without conflicts

3. Enable all metric groups by default
   - Changed default MetricGroups from connection-basic+resiliency to MetricGroupAll for comprehensive observability out of the box

---

> [!NOTE]
> **Medium Risk**
> Updates the `Pooler.CloseConn` API and connection lifecycle paths to emit new close/state-change metrics, which could affect pool implementations and metric volume/cardinality. Also changes redisotel-native defaults and bumps OTel deps, so dashboards/alerting may see new or renamed series and higher emission rates.
> 
> **Overview**
> **Fixes missing connection-close metrics and standardizes pool close semantics.** `Pooler.CloseConn` is changed to accept `ctx`, a close `reason`, and `fromState`, with new constants (`CloseReason*`, `MetricState*`) and updated call sites (stale/hook/auth failures) to ensure `metricConnectionClosedCallback` and state transitions are recorded.
> 
> **Makes redisotel-native more “on-by-default” for observability and adds coverage.** `NewConfig()` now defaults `MetricGroups` to `MetricGroupAll`, adds semconv metric-name assertions plus Redis-backed stress/compatibility tests, and updates `extra/redisotel-native` dependencies to OTel `v1.40.0` (including `sdk/metric`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dfbc5d7b68aebaab1059076fe715f5e70a5589d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extra/redisotel-native/redisotel.go
+++ b/extra/redisotel-native/redisotel.go
@@ -7,10 +7,12 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/redis/go-redis/v9"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/semconv/v1.38.0/dbconv"
+
+	"github.com/redis/go-redis/v9"
 )
 
 // Metric name constants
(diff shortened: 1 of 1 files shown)


redis.go
```diff
@@ -347,7 +347,10 @@ func (c *baseClient) onAuthenticationErr() func(poolCn *pool.Conn, err error) {
 		if err != nil {
 			if isBadConn(err, false, c.opt.Addr) {
 				// Close the connection to force a reconnection.
-				err := c.connPool.CloseConn(poolCn)
+				// Re-auth happens on connections that were idle in the pool (the pool hook
+				// waits for IDLE state before transitioning to UNUSABLE for re-auth).
+				// From metrics perspective, the connection was never "used" by a client.
+				err := c.connPool.CloseConn(poolCn, "auth_error", "idle")
```
ndyakov: can we extract the reasons as consts? also the states are defined as consts, you can use them.

internal/pool/pool.go
```diff
@@ -13,6 +13,38 @@ import (
...
+// not the internal state machine state (ConnState).
+const (
+	// MetricStateIdle indicates the connection is idle in the pool,
+	// ready to be acquired.
+	MetricStateIdle = "idle"
+
+	// MetricStateUsed indicates the connection is currently being used
+	// by a client operation.
+	MetricStateUsed = "used"
+)
```
ndyakov, line 46: can we use `ConnState` instead? It won't be a breaking change if we decide to do so in the future, since this is an internal packaged and even if exported, we can drop this and switch to ConnState

PR #3738, opened 2026-03-16
Typed error for `NOSCRIPT` replies from Redis

LINKIWI:
This PR proposes exposing a typed error `ErrNoScript` that indicates a `NOSCRIPT` reply from Redis during script execution.

This allows library consumers who directly use the `EvalSha` API (as opposed to the `*redis.Script` abstraction) to explicitly detect and handle this case, without needing to roll their own error introspection logic:

```go
ctx := ...
rdb := redis.NewClient(...)

r, err := rdb.EvalSha(ctx, sha1, []string{...}).Result()
if errors.Is(err, redis.ErrNoScript) {
    ...
}

...
```

This change also makes the internal implementation of `*redis.Script` less fragile, in that the `EvalSha` and `EvalShaRO` callsites can check for a typed error rather than performing error string prefix matching.

I have added two new unit tests to exercise the behavior.

---

> [!NOTE]
> **Medium Risk**
> Changes the error value returned from `EvalSha`/`EvalShaRO` when Redis replies `NOSCRIPT`, which could affect consumers that compare error strings or types. Scope is limited to scripting command error handling plus new tests.
> 
> **Overview**
> Adds a new exported typed error `redis.ErrNoScript` to represent Redis `NOSCRIPT` replies for missing script digests.
> 
> Updates the scripting command path so `EvalSha`/`EvalShaRO` normalize `NOSCRIPT` responses to `ErrNoScript`, and updates `(*Script).Run`/`RunRO` to use `errors.Is` against this typed error instead of string-prefix matching. Adds unit tests asserting `NOSCRIPT` is propagated as `redis.ErrNoScript` for unknown digests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 27fd86fe2205201c5e1de5630c95578ad7fde8e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/error.go
+++ b/error.go
@@ -28,6 +28,11 @@ var ErrPoolTimeout = pool.ErrPoolTimeout
 // is used on a ClusterClient with keys in different slots.
 var ErrCrossSlot = proto.RedisError("CROSSSLOT Keys in request don't hash to the same slot")
 
+// ErrNoScript is returned when EVALSHA is requested for a script digest that
+// is not available in the script cache. Note that this error text is reproduced
+// literally from that used by Redis.
+var ErrNoScript = proto.RedisError("NOSCRIPT No matching script. Please use EVAL.")
+
 // HasErrorPrefix checks if the err is a Redis error and the message contains a prefix.
 func HasErrorPrefix(err error, prefix string) bool {
 	var rErr Error
--- a/commands_test.go
+++ b/commands_test.go
@@ -2,6 +2,8 @@ package redis_test
 
 import (
 	"context"
+	"crypto/rand"
+	"crypto/sha1"
 	"encoding/json"
 	"fmt"
 	"reflect"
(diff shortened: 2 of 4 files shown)


error.go
```diff
@@ -28,6 +28,11 @@ var ErrPoolTimeout = pool.ErrPoolTimeout
 // is used on a ClusterClient with keys in different slots.
 var ErrCrossSlot = proto.RedisError("CROSSSLOT Keys in request don't hash to the same slot")
 
+// ErrNoScript is returned when EVALSHA is requested for a script digest that
+// is not available in the script cache. Note that this error text is reproduced
+// literally from that used by Redis.
+var ErrNoScript = proto.RedisError("NOSCRIPT No matching script. Please use EVAL.")
```
LINKIWI, line 34: Source: https://github.com/redis/redis/blob/8.6.1/src/server.c#L2126

This also maintains backwards compatibility with behavior expected from any existing clients that currently detect this condition from a full literal error string match.

PR #3748, opened 2026-03-23
feat(csc): add standalone client-side caching

ofekshenawa:
Adds a thread-safe, in-memory local cache with LRU eviction as the foundation for client-side caching 

---

> [!NOTE]
> **High Risk**
> Large correctness-sensitive surface: stale reads, invalidation races, multi-identity cache isolation, and connection lifecycle interact with the hot command path; misbehavior would be subtle and data-related rather than a simple API break.
> 
> **Overview**
> Implements **end-to-end client-side caching (CSC)** on RESP3 clients: a shared local cache backed by per-pool-connection `CLIENT TRACKING`, `invalidate` push handling, and a background idle-connection drainer so buffered invalidations are not missed.
> 
> **Configuration and strategy:** `ClientSideCache` / `ClientSideCacheConfig` attach via **SharedTracking** (default). Attachment is gated on DB 0, fixed credentials (dynamic providers disable CSC), `ConnOwnedCache`, pool idle draining, and lifecycle hooks. Cache keys are namespaced by DB and hashed ACL username; shared caches stay isolated across identities.
> 
> **Command path:** Allow-listed read-only commands go through reserve/fulfill caching (including negative `redis.Nil` hits). Entries are attributed to serving connection IDs with generation checks to avoid publishing after handoff, removal, or drainer shutdown. **Guards** reject `SELECT`, `AUTH`, protocol-changing `HELLO`, `RESET`, pooled `SUBSCRIBE`, and manual `CLIENT TRACKING` when built-in CSC is on (pipelines included). Typed **`ClientTracking`** APIs are added for dedicated `Conn` use, with client-side option validation.
> 
> **Pool/transport:** Connection hooks (`onCscClose`, `onCscReinit`, read-pending probes) support eviction on close/reinit; health checks avoid unwrapping buffered TLS during pool probes while CSC draining unwraps safely. `Client.CSCStats()` exposes hit/miss counters when the cache supports it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c57922467e480e83eddfd537e9a888656c3eb63a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ofekshenawa:
bugbot run verbose=true

local_cache.go
```diff
@@ -0,0 +1,476 @@
...
+	// CacheEntryValid marks an entry that contains a value that can be returned.
+	CacheEntryValid
+)
+
+// CacheEntry represents a cached command reply and its Redis-key associations.
+type CacheEntry struct {
+	CacheKey  string
+	RedisKeys []string
+	Value     []byte
+	State     CacheEntryState
```
ndyakov, line 26: do we need those to be public? 
In general, do we need the whole CacheEntry to be public? What if we provide an interface to the client and put this as a default implementation in `internal` ? Later on we can ever support other cache implementations if a user would like to do its own.

internal/cache/cache.go
```diff
@@ -0,0 +1,476 @@
...
+
+		value := cloneBytes(entry.Value)
+		c.mu.RUnlock()
+
+		c.mu.Lock()
+		// Touch under write lock keeps LRU metadata consistent with concurrent deletes/updates.
+		if current, exists := c.entries[cacheKey]; exists && current == entry && current.State == CacheEntryValid {
+			c.touchEntryLocked(current)
+		}
+		c.mu.Unlock()
```
ndyakov: can this counter be atomic? It may be approximate (if the entry changes), but would be **significantly** better to not try to acquire the mutex on read.

local_cache.go
```diff
@@ -0,0 +1,476 @@
...
+	staleTimeout   time.Duration
+	nextToken      uint64
+}
+
+const defaultCacheEntryOverhead int64 = 96
+
+func defaultCacheSizer(cacheKey string, redisKeys []string, value []byte) int64 {
+	size := defaultCacheEntryOverhead + int64(len(cacheKey)+len(value))
+	for _, key := range redisKeys {
+		size += int64(len(key)) + 16
```
ndyakov, line 265: why do we have those offsets?

ndyakov (commented):
I would prefer to see this in an `internal` package and for us to expose only an interface to the client. Other than that, great work @ofekshenawa!

local_cache.go
ofekshenawa: The offsets were an attempt to account for Go's runtime memory overhead beyond the raw byte lengths of the key strings. I was trying to make MemoryUsage() a closer estimate of actual heap usage rather than just the size of the user payload. I changed it in the last commit because I understood that this can be confusing.

internal/cache/cache.go
```diff
@@ -0,0 +1,484 @@
...
+	defer c.mu.RUnlock()
+	return len(c.entries)
+}
+
+func (c *localCache) MemoryUsage() int64 {
+	return c.usedBytes.Load()
+}
+
+func (c *localCache) computeSizeLocked(cacheKey string, redisKeys []string, value []byte) int64 {
+	size := c.sizer(cacheKey, redisKeys, value)
```
ndyakov: `sizer` comes from the `Config.Sizer` and it is part of the users input. Please add a comment that this function will be executed while holding a lock and if the user references the cache in it this may deadlock.

internal/cache/cache.go
```diff
@@ -0,0 +1,484 @@
...
+
+		if e.state == inProgress {
+			waitCh := e.waitCh
+			c.mu.RUnlock()
+			// Wait for the in-flight fetch to either publish (Fulfill) or abort (Cancel/Delete/Flush).
+			if waitCh != nil {
+				select {
+				case <-waitCh:
+				case <-ctx.Done():
+					return nil, false
```
ndyakov: does it make sense to also have a timeout here? can the fetch not close the `waitCh` for any reason?

internal/cache/cache.go
```diff
@@ -0,0 +1,484 @@
...
+			return nil, false
+		}
+
+		if e.state == inProgress {
+			waitCh := e.waitCh
+			c.mu.RUnlock()
+			// Wait for the in-flight fetch to either publish (Fulfill) or abort (Cancel/Delete/Flush).
+			if waitCh != nil {
+				select {
+				case <-waitCh:
```
ndyakov: should we directly read here instead of looping? Can the code theoretically get stuck in this loop? Should we have some counter as well that will stop the loop if it iterates n times?

ndyakov (approved):
Overall looks good to me, left couple of questions and a documentation suggestion.

internal/cache/cache.go
ofekshenawa: Added in ec0b1ec9 , the doc now warns Sizer runs under the cache lock and must not re-enter the cache.

ndyakov (approved):
Round-3 fixes verified: the Get wait is now bounded by StaleTimeout with no timer re-arm issue, and the Sizer doc landed. The remaining comments (approximate-LRU wording, Reserve (0,true) contract, oversized Set behavior) are doc-level and do not block. Approving.

PR #3756, opened 2026-03-29
validate FT.HYBRID vector input type

DengY11:
See #3755

With this change:
- `FT.HYBRID` only accepts `VectorFP32`
- unsupported types like `VectorValues` and `VectorRef` now return a clear client-side error
- added unit tests for valid and invalid vector inputs

---

> [!NOTE]
> **Medium Risk**
> Changes `FT.HYBRID` argument building to validate/normalize vector blobs client-side and to reject some previously-accepted `Vector` implementations (e.g. `VectorValues`/`VectorRef`), which could be a behavior-breaking change for callers.
> 
> **Overview**
> `FT.HYBRID` now extracts and validates the raw vector *blob* via a new `hybridVectorBlob` helper, returning clear client-side errors for missing/empty vectors and explicitly rejecting unsupported `Vector` implementations (notably `VectorValues` and `VectorRef`) instead of sending malformed args.
> 
> Adds new blob-backed vector wrappers (`VectorFloat16`, `VectorBFloat16`, `VectorFloat64`, `VectorInt8`, `VectorUint8`) and unit tests covering accepted/rejected vector types and ensuring the command uses the raw blob in the generated args.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c973763626a3759b529a486d80b4b2969af08a6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/search_commands.go
+++ b/search_commands.go
@@ -2853,6 +2853,23 @@ func (c cmdable) FTHybrid(ctx context.Context, index string, searchExpr string, 
 	return c.FTHybridWithArgs(ctx, index, options)
 }
 
+func hybridVectorBlob(v Vector) ([]byte, error) {
+	if v == nil {
+		return nil, fmt.Errorf("FT.HYBRID: vector data is required")
+	}
+
+	vectorFP32, ok := v.(*VectorFP32)
+	if !ok {
+		return nil, fmt.Errorf("FT.HYBRID: unsupported vector type %T; use VectorFP32 with encoded vector blob", v)
+	}
+
+	if len(vectorFP32.Val) == 0 {
+		return nil, fmt.Errorf("FT.HYBRID: vector blob is required")
+	}
+
+	return vectorFP32.Val, nil
+}
+
 // FTHybridWithArgs - Executes a hybrid search with advanced options
 // FTHybridWithArgs is still experimental, the command behaviour and signature may change
 func (c cmdable) FTHybridWithArgs(ctx context.Context, index string, options *FTHybridOptions) *FTHybridCmd {
@@ -2879,16 +2896,11 @@ func (c cmdable) FTHybridWithArgs(ctx context.Context, index string, options *FT
 		for _, vectorExpr := range options.VectorExpressions {
 			args = append(args, "VSIM", "@"+vectorExpr.VectorField)
 
-			// For FT.HYBRID, we need to send just the raw vector bytes, not the Value() format
-			// Value() returns [format, data] but FT.HYBRID expects just the blob
-			vectorValue := vectorExpr.VectorData.Value()
-			var vectorBlob interface{}
-			if len(vectorValue) >= 2 {
-				// vectorValue is [format, data, ...] - we only want the data part
-				vectorBlob = vectorValue[1]
-			} else {
-				// Fallback for unexpected format
-				vectorBlob = vectorValue
+			vectorBlob, err := hybridVectorBlob(vectorExpr.VectorData)
+			if err != nil {
+				cmd := newFTHybridCmd(ctx, options, args...)
+				cmd.SetErr(err)
+				return cmd
 			}
 
 			// If VectorParamName is provided, use PARAMS mechanism (required for Redis 8.6+)
--- a/search_commands.go
+++ b/search_commands.go
@@ -2853,21 +2853,40 @@ func (c cmdable) FTHybrid(ctx context.Context, index string, searchExpr string, 
 	return c.FTHybridWithArgs(ctx, index, options)
 }
 
-func hybridVectorBlob(v Vector) ([]byte, error) {
+func hybridVectorBlob(v Vector) (interface{}, error) {
 	if v == nil {
 		return nil, fmt.Errorf("FT.HYBRID: vector data is required")
 	}
 
-	vectorFP32, ok := v.(*VectorFP32)
-	if !ok {
-		return nil, fmt.Errorf("FT.HYBRID: unsupported vector type %T; use VectorFP32 with encoded vector blob", v)
+	switch vector := v.(type) {
+	case *VectorFP32:
+		return hybridVectorBytes(vector.Val)
+	case *VectorFloat16:
+		return hybridVectorBytes(vector.Val)
+	case *VectorBFloat16:
+		return hybridVectorBytes(vector.Val)
+	case *VectorFloat64:
+		return hybridVectorBytes(vector.Val)
+	case *VectorInt8:
+		return hybridVectorBytes(vector.Val)
+	case *VectorUint8:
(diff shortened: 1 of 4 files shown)


ndyakov (dismissed):
Looks good, thank you @DengY11.

ndyakov:
@DengY11 after a more detailed check, ft.hybrid should accept any vector type defined in the index, right? If that is the case, let's not limit the command and just return the error from the server. What do you think about that? Maybe it would be better to actually implement additional vector types for indexes in Redis - https://redis.io/docs/latest/develop/ai/search-and-query/vectors/#flat-index ?

DengY11:
> @DengY11 after a more detailed check, ft.hybrid should accept any vector type defined in the index, right? If that is the case, let's not limit the command and just return the error from the server. What do you think about that? Maybe it would be better to actually implement additional vector types for indexes in Redis - https://redis.io/docs/latest/develop/ai/search-and-query/vectors/#flat-index ?

Yes I don't think FT.HYBRID should only accept fp32. The other existing Vector types are more likely to be designed for vectorset command and do not map correctly to hybrid search query input. So instead of limiting it to fp32, it is better to improve/add proper vector input types for FT.HYBRID, matching the actual index vector encodings.
For example, support  types like:
- `FLOAT16`
- `BFLOAT16`
- `FLOAT32`
- `FLOAT64`
- `INT8`
- `UINT8`

ndyakov:
@DengY11 I do think we should add those as vector types and since the API is already in a released version, we cannot change the FTHybrid signature at this stage. Let's improve by adding the rest of the vector types and check if the type is supported by the command (either ft hybrid or vector set command) when executing. However, I still think it is more than OK to let the server return an error. This way if there is a new vector type we don't need to alter the code. Since this is an interface, the user can implement their own `Vector` and use it in the commands.

DengY11:
> 我确实认为我们应该把这些作为向量类型添加，而且由于API已经发布，目前我们无法更改FTHybrid签名。我们通过添加其他向量类型来改进，并在执行时检查该类型是否被命令（无论是 ft hybrid 还是 vector set 命令）支持。不过，我仍然认为让服务器返回错误是完全可以接受的。这样如果有新的向量类型，我们就不需要修改代码。由于这是一个界面，用户可以实现自己的接口并在命令中使用。`Vector`

I understand the point about keeping the interface open, but I still think there are 2 problems here.

1. Returning server error is not always strong enough. Some wrong input can still be syntactically valid, so the server may not reject it, and then we get semantically wrong behavior instead of a clear failure.
For example, if the index field is `FLOAT16 DIM 8`, Redis expects a blob with `8 * 2 = 16` bytes. If client sends another 16-byte blob with different encoding by mistake, the request can still look valid from server side: command syntax is valid, argument position is valid, blob length is also valid
Then Redis may not reject it, and just interpret the bytes as `FLOAT16 DIM 8`. In that case it is not a clear server error, but a semantically wrong request.

2. The `Vector` interface is also not very clear for users. They only see `Value() []any`, but dont really know what `Value()` is supposed to return for `FT.HYBRID` or other commands.

So I think maybe better to do one of these:
1. document the `Vector` contract more clearly
2. or add a more explicit API for `FT.HYBRID` input, without changing the current signature, for example by adding something like `FTQueryVector` and `FTHybridWithQueryVector(...)`

DengY11:
> @DengY11 I do think we should add those as vector types and since the API is already in a released version, we cannot change the FTHybrid signature at this stage. Let's improve by adding the rest of the vector types and check if the type is supported by the command (either ft hybrid or vector set command) when executing. However, I still think it is more than OK to let the server return an error. This way if there is a new vector type we don't need to alter the code. Since this is an interface, the user can implement their own `Vector` and use it in the commands.

I understand the point about keeping the interface open, but I still think there are 2 problems here.

1. Returning server error is not always strong enough. Some wrong input can still be syntactically valid, so the server may not reject it, and then we get semantically wrong behavior instead of a clear failure.
For example, if the index field is `FLOAT16 DIM 8`, Redis expects a blob with `8 * 2 = 16` bytes. If client sends another 16-byte blob with different encoding by mistake, the request can still look valid from server side: command syntax is valid, argument position is valid, blob length is also valid
Then Redis may not reject it, and just interpret the bytes as `FLOAT16 DIM 8`. In that case it is not a clear server error, but a semantically wrong request.

2. The `Vector` interface is also not very clear for users. They only see `Value() []any`, but dont really know what `Value()` is supposed to return for `FT.HYBRID` or other commands.

So I think maybe better to do one of these:
1. document the `Vector` contract more clearly
2. or add a more explicit API for `FT.HYBRID` input, without changing the current signature, for example by adding something like `FTQueryVector` and `FTHybridWithQueryVector(...)`

ndyakov:
@DengY11 documenting the `Vector` contract and introducing a new one for FTHybrid makes sense. However, transitioning from the `Vector` to the `FTQueryVector` will be a tricky process as we cannot change the already released api. We can however deprecate it, but I am not sure if this won't be confusing as well.

I fully understand the concern that the server may accept wrong input if its format is correct. This however is not something the client can guard against (for example  you can set any value and later if you want to read it and unmarshal it - it may not be the correct type). 

How would sound documenting the `Vector` contract better, introducing the types that are currently missing and documenting their usage as well - some will be used for the vector sets, others for the fthybrid.

DengY11:
@ndyakov Hi, I updated the PR to add the missing vector blob types, and also kept a legacy fallback for known types  .For known types I kept the old behavior and still use `Value()[1]`, so this should not break users who already implemented their own `Vector`.
I keep reject `VectorValues` and `VectorRef` for `FT.HYBRID`,these are invalid inputs and don't really map to hybrid search query blob. Also added unit tests for the new types and the legacy fallback.

ndyakov (commented):
Thank you @DengY11! Overall I do like this approach. What do you think about - instead having `Val` exported on the specific types - add it as part of the interface? Technically we are extending the interface, but if someone has implemented their own vector type this may be breaking. I do not think this is likely, but would like to take your input on it since you invested quite a lot of time on this one.

DengY11:
Thanks for checking this. I would prefer to keep vector interface unchanged for now. Even if custom implementations are probably rare, adding a new method to the interface still feels like a stronger breaking change.
So I think the current approach is a bit safer and better: handle the built-in blob types directly, and keep the old `Value()[1]` fallback for custom vectors. And maybe discuss a cleaner vector API separately in future issue.

ndyakov:
@DengY11 I agree and accept this approach as a safer one to include now. Thank you for your work on this!

ndyakov (approved):

PR #3762, opened 2026-04-01
ring: inline rendezvous hashing and remove external dependency

bigsk05:
## Why not keep the dependency?

The original implementation is small (~50 LOC) and stable. Inlining it:
- avoids pulling an extra dependency into go-redis
- removes license management overhead
- keeps behavior fully deterministic and test-covered

Additionally, avoiding a single-purpose dependency helps keep the
dependency graph minimal and easier to maintain.

## Changes

- Inline a minimal rendezvous hashing implementation
- Remove dependency on github.com/dgryski/go-rendezvous
- Keep algorithm identical to upstream:
  - same hash combination (khash ^ nodehash)
  - same xorshift* mixing function
  - same iteration/order semantics

## Compatibility

No behavior changes are intended.

## Notes

The implementation is intentionally minimal and mirrors the original
algorithm to ensure deterministic behavior.

## Checklist

- [x] No public API changes
- [x] No behavior changes

---

> [!NOTE]
> **Medium Risk**
> Changes the ring sharding hash implementation and removes an external dependency; any subtle mismatch could alter key-to-shard mapping and affect routing in `Ring` deployments.
> 
> **Overview**
> Replaces the `Ring` consistent-hash implementation with an in-repo rendezvous (HRW) hash (`internal/hashtag/RendezvousHash`) and updates `ring.go` to use it instead of `github.com/dgryski/go-rendezvous`.
> 
> Removes the `go-rendezvous` dependency from the root module plus all `example/*` and `extra/*` modules, and adds a new test suite (`internal/hashtag/rendezvous_test.go`) to validate determinism, distribution, and basic stability characteristics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 08832cc9a32f8e3899bb6a4a89e44fd49f7088db. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/hashtag/rendezvous_test.go
+++ b/internal/hashtag/rendezvous_test.go
@@ -46,28 +46,27 @@ var _ = Describe("RendezvousHash", func() {
 		}
 	})
 
-	It("should handle node ordering consistently (tie-breaker semantics)", func() {
-		// Test that node order doesn't affect determinism
-		nodesA := []string{"node-a", "node-b", "node-c"}
-		nodesB := []string{"node-c", "node-a", "node-b"}
-		nodesC := []string{"node-b", "node-c", "node-a"}
-
-		hA := NewRendezvousHash(nodesA)
-		hB := NewRendezvousHash(nodesB)
-		hC := NewRendezvousHash(nodesC)
+	It("should be consistent within a single instance (per-instance determinism)", func() {
+		// This test verifies per-instance determinism: the same RendezvousHash instance
+		// returns consistent results for the same key across multiple calls.
+		// Note: Node order in the input may affect results due to first-match tie-breaking
+		// when multiple nodes produce equal scores (uses > not >= in score comparison).
+		nodes := []string{"node-a", "node-b", "node-c"}
+		h := NewRendezvousHash(nodes)
 
 		testKeys := []string{"key-1", "key-2", "key-3", "test", "foo", "bar"}
 
 		for _, testKey := range testKeys {
-			resultA := hA.Get(testKey)
-			resultB := hB.Get(testKey)
-			resultC := hC.Get(testKey)
+			// Same instance should consistently return the same node for the same key
+			result1 := h.Get(testKey)
+			result2 := h.Get(testKey)
+			result3 := h.Get(testKey)
 
-			// HRW should map to the same logical node regardless of input order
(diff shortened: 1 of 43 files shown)


ndyakov (dismissed):
@bigsk05 I like this approach and am happy to include it. Yes, removing the dependency is great benefit from having this piece in the library code. Thank you!

internal/hashtag/rendezvous_test.go
```diff
@@ -0,0 +1,162 @@
...
+		for i := 0; i < 100; i++ {
+			node := h.Get("key-" + strconv.Itoa(i))
+			Expect(node).To(BeElementOf(nodes))
+			results[node]++
+		}
+
+		// Verify distribution is roughly even (allowing some variance)
+		for _, node := range nodes {
+			count := results[node]
+			Expect(count).To(BeNumerically(">", 0), "node %s should receive at least one key", node)
```
ndyakov: Since there are 100 keys and 3 nodes, maybe we can check for `>25` here?

ndyakov (commented):
Just wondering, can we test better for (mostly) even distribution of the keys?

internal/hashtag/rendezvous_test.go
bigsk05:
```
// Verify distribution is roughly even (allowing some variance)
// With 100 keys and 3 nodes, expect roughly 33 keys per node.
// Check that each node receives more than 25 keys to verify fairly even distribution.
for _, node := range nodes {
	count := results[node]
	Expect(count).To(BeNumerically(">", 25),
		"node %s should receive more than 25 keys (got %d)", node, count)
}
```

How about this?

bigsk05:
```
=======================================================================
Random Seed: 1775122596

Will run 11 of 11 specs
•••••••••••

Ran 11 of 11 Specs in 0.009 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestGinkgoSuite (0.01s)
PASS
ok      github.com/redis/go-redis/v9/internal/hashtag   (cached)
```

ndyakov (approved):
Looks good now, thank you @bigsk05!

PR #3764, opened 2026-04-04
fix(pool): two fixes for closed connection handling

### LINKED ISSUE #3703: Zombie connection in ConnPool.idleConns
Connections closed on +switch-master notification from Sentinel remains in ConnPool.idleConns forever

## Expected Behavior

Idle connections with state "CLOSED" should be cleaned up from ConnPool.idleConns and new connections should be reestablished.

## Current Behavior

Idle connections remains in ConnPool.idleConns forever.
Currently we use v9.17.2.

## Possible Solution

On v9.16.0 it works as expected.
First, on +switch-master event for every connection in pool p.closeConn(cn) is called. The difference between versions is a new line `cn.stateMachine.Transition(StateClosed)` in Conn.Close() method, which changes the state to "CLOSED".
Second, ConnPool.popIdle() behavior was changed.
In v9.16.0 the following code returns cn, which later is being tested in p.isHealthyConn(cn, now). As a result closed connection is being removed from pool.

```

		if cn.CompareAndSwapUsed(false, true) {
			if cn.IsUsable() {
				p.idleConnsLen.Add(-1)
				break
			}
			cn.SetUsed(false)
		}

```

In v9.17.2 the code above was replaced with the following:

```

		// Hot path optimization: try IDLE → IN_USE or CREATED → IN_USE transition
		// Using inline TryAcquire() method for better performance (avoids pointer dereference)
		if cn.TryAcquire() {
			// Successfully acquired the connection
			p.idleConnsLen.Add(-1)
			break
		}

```

cn.TryAcquire() will give True if and only if the stated is in "IDLE" or "CREATED". We have state "CLOSED" here, so as a result we have cn=nil. There are no remaining available connection in pool, so we go through the branch of code where for every upcoming commands a new connection is being created.

I believe we need to return closed cn from popIdle method to be cleared in isHealthyConn as before, or remove from pool here.

## Steps to Reproduce

1. Run redis in sentinel mode
2. Create a Failover Client with poolSize=1

```

        client = goredis.NewFailoverClient(&goredis.FailoverOptions{
		MasterName:      sentinelName,
		SentinelAddrs:   strings.Split(sentinels, ","),
		PoolSize:        1,
		DisableIdentity: true,
	})

```

3. Send at least one command to fill the pool
4. Trigger switch master
5. To easier catch the bug we can use multi-exec 

```
	
	set, err := client.Set(ctx, "testRadishKeyMulti2", "5", 0).Result()
	assert.NoError(t, err)
	assert.Equal(t, "OK", set)

	multi, err := client.Do(ctx, "MULTI").Result()
	assert.NoError(t, err)
	assert.Equal(t, "OK", multi.(string))

	incr, err := client.Do(ctx, "INCR", "testRadishKeyMulti2").Result()
	assert.NoError(t, err)
	assert.Equal(t, "QUEUED", incr.(string))

	decr, err := client.Do(ctx, "DECR", "testRadishKeyMulti2").Result()
	assert.NoError(t, err)
	assert.Equal(t, "QUEUED", decr.(string))

	exec, err := client.Do(ctx, "EXEC").Result()
	assert.NoError(t, err)
	execArray, ok := exec.([]any)
	assert.True(t, ok)
	assert.EqualValues(t, int64(6), execArray[0])
	assert.EqualValues(t, int64(5), execArray[1])

```

## Context (Environment)

We moved from v9.16.0 due to that maintenance bug

cxljs:
Closes #3703 

1. In putConn(): prevent CLOSED state conns from being added to idleConns

2. In Filter(): ensure closed conns are removed from idleConns during failover

---

> [!NOTE]
> **Medium Risk**
> Changes connection pool close/removal behavior and metrics emission paths, which can affect connection lifecycle correctness under concurrency (especially during failover). Risk is moderated by added targeted sentinel failover tests and mostly localized changes to pooling/metrics codepaths.
> 
> **Overview**
> Improves closed-connection handling in the pool, especially around failover and concurrent closes.
> 
> `Conn.Close()` is now idempotent and ensures `onClose` runs at most once. The pool introduces a per-conn atomic `closeReason` (plus new `CloseReasonFailover`) so an in-use conn closed by another goroutine can later emit the correct `ConnectionClosed` metric reason when it’s returned.
> 
> `ConnPool.Filter()` is updated to properly distinguish *idle* vs *in-use* conns: idle conns are removed from `idleConns`/pool and closed immediately with failover metrics, while in-use conns are marked with `closeReason` and closed without corrupting idle lists. Metrics emission for `CloseConn` is refactored into `recordConnectionMetrics()`.
> 
> Adds a new serial/ordered sentinel failover test that asserts exactly one `failover` close metric across different `MaxActiveConns` settings, and marks the `UniversalClient` test suite as `Serial` to reduce inter-test interference.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7500d2911e3a26e1255d28c8b49e2e3463d660a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -1345,6 +1345,19 @@ func (p *ConnPool) putConn(ctx context.Context, cn *Conn, freeTurn bool) {
 		}
 	} else {
 		shouldCloseConn = true
+		removedFromPool = p.removeConnWithLock(cn)
+
+		// Only emit if we actually removed it from the map (not already taken by Close()).
+		if removedFromPool {
+			// Notify metrics: connection removed (used -> nothing)
+			if cb := getMetricConnectionStateChangeCallback(); cb != nil {
+				cb(ctx, cn, MetricStateUsed, "")
+			}
+			// Record connection count decrement (connection removed while in used state)
+			if cb := getMetricConnectionCountCallback(); cb != nil {
+				cb(ctx, -1, cn, "used", false)
+			}
+		}
 	}
 
 	if freeTurn {
@@ -1365,6 +1378,8 @@ func (p *ConnPool) putConn(ctx context.Context, cn *Conn, freeTurn bool) {
 		}
 		_ = p.closeConn(cn)
 	}
+
+	cn.SetLastPutAtNs(getCachedTimeNs())
 }
 
 func (p *ConnPool) Remove(ctx context.Context, cn *Conn, reason error) {
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -30,6 +30,9 @@ const (
 
 	// CloseReasonTest is used in tests when closing connections.
 	CloseReasonTest = "test"
+
+	// CloseReasonFailover indicates the connection was closed due to a failover event.
+	CloseReasonFailover = "failover"
 )
 
 // Metric state constants for connection state tracking.
@@ -1366,7 +1369,11 @@ func (p *ConnPool) putConn(ctx context.Context, cn *Conn, freeTurn bool) {
 		// If removedFromPool is false, Close() already emitted connectionClosed for this conn.
 		if removedFromPool {
 			if cb := getMetricConnectionClosedCallback(); cb != nil {
-				cb(ctx, cn, "conn_pool_close", nil)
+				reason := "conn_pool_close"
+				if cn.closeReason != "" {
+					reason = cn.closeReason
+				}
+				cb(ctx, cn, reason, nil)
 			}
 		}
 		_ = p.closeConn(cn)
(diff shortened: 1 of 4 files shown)


internal/pool/pool.go
```diff
@@ -1212,26 +1214,23 @@ func (p *ConnPool) putConn(ctx context.Context, cn *Conn, freeTurn bool) {
...
 	}
 
 	if freeTurn {
 		p.freeTurn()
 	}
 
 	if shouldCloseConn {
-		_ = p.closeConn(cn)
+		_ = p.CloseConn(ctx, cn, CloseReasonStale, MetricStateUsed)
+		return
```
ndyakov: 👍

ndyakov:
@cxljs other than the notes from the CI, I don't have anything against this. We haven't tested the new metrics collector with sentinel (where this reason will be used), have you tested it?

cxljs:
> the notes from the CI

I saw the CI comment. Since these past few days are a special holiday in my country, I haven’t had time to address it yet. I’ll take care of it as soon as possible.

>  We haven't tested the new metrics collector with sentinel (where this reason will be used), have you tested it?

Thanks for the reminder. I’ll test it.

universal_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/redis/go-redis/v9"
 )
 
-var _ = Describe("UniversalClient", func() {
+var _ = Describe("UniversalClient", Serial, func() {
```
ndyakov, line 10: @cxljs any specific reason why we would like those to be ran one by one? After e brief look it looks like they should be fine running in parallel?

ndyakov (dismissed):
overall looks good to me.

universal_test.go
cxljs: Hello, because a failover may occur during the execution of the test case https://github.com/redis/go-redis/blob/ca74fb89ed80ea14f86ca09f3d50bc46f27a9c84/universal_test.go#L73, it can cause the test to fail. Originally, the probability of this happening was very low. However, in the new test case "Sentinel Failover with Conns", we manually trigger a failover by executing the SHUTDOWN command, which increases the likelihood of failure for the former test. Therefore, running these two tests sequentially can prevent them from interfering with each other. After carefully reviewing the related test cases, it appears that only these two need to be executed sequentially.

cxljs: @ndyakov Even with Serial execution, the universal tests are still failing. I suspect a race condition: the client checks the role (sees it's a slave), a failover happens, the node becomes a master, and then the SET command succeeds unexpectedly, which breaks the test. I can’t tell if it’s an issue with my new tests or just an unstable test environment. WDYT?
https://github.com/redis/go-redis/actions/runs/24659104606/job/72100104473?pr=3764

cxljs: In the current solution, after the manual failover test is completed, sleep for 10 seconds to restore the state of the Sentinel cluster to normal.

internal/pool/pool.go
```diff
@@ -1552,13 +1561,33 @@ func (p *ConnPool) closed() bool {
...
+				// Idle connection - remove from pool and close.
+				p.removeConn(cn)
+				p.recordConnectionMetrics(ctx, cn, CloseReasonFailover, MetricStateIdle)
+				err = p.closeConn(cn)
+			} else {
+				// Used connection - set closeReason and close the connection.
+				// The connection remains in p.conns. When putConn() is called later,
+				// it will close the connection instead of pooling it.
+				cn.closeReason.Store(CloseReasonFailover)
+				err = cn.Close()
```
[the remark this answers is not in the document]
cxljs: The Filter() function is currently only used in this one place, and it is not part of the Pooler API, so I prefer to keep it as it is and avoid over-engineering.

universal_test.go
ndyakov: @cxljs will review this in-depth tomorrow.

ndyakov: I do think testing the failover in the CI is tricky, we will have to sync somehow and make sure other tests do not touch the instance at this time (do not start before the instance is in a predefined required state). For now, lets leave it as it is, we have the need to spend some time on the tests in general.

ndyakov (approved):
Looks good to me, thank you @cxljs!

PR #3768, opened 2026-04-08
Fix issue 3139 feat(hscan): add support for encoding.BinaryUnmarshaler

### LINKED ISSUE #3139: Can't get UUID with Scan()

Aaditya-dubey1:
Fixes #3139.
Added support for encoding.BinaryUnmarshaler in hscan.Scan to allow scanning into types like UUID.
Included unit test in hscan_test.go to verify the functionality.

---

> [!NOTE]
> **Medium Risk**
> Touches the core `hscan` struct field scanning path, which is widely used and could subtly change decoding precedence for custom types. Change is small and covered by a focused unit test.
> 
> **Overview**
> Adds support for scanning Redis hash values into struct fields whose types implement `encoding.BinaryUnmarshaler`, calling `UnmarshalBinary([]byte)` during `hscan` decoding.
> 
> Extends the `hscan` test suite with a new case that verifies a `BinaryUnmarshaler` implementation is invoked and receives the raw value bytes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a2d65b4611d429056c4b9d8f1df040a534c0f54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/hscan/hscan_test.go
+++ b/internal/hscan/hscan_test.go
@@ -48,6 +48,19 @@ type TimeData struct {
 	Time *TimeRFC3339Nano `redis:"login"`
 }
 
+type binaryUnmarshaler struct {
+	s string
+}
+
+func (b *binaryUnmarshaler) UnmarshalBinary(data []byte) error {
+	b.s = string(data)
+	return nil
+}
+
+type BinaryData struct {
+	Binary *binaryUnmarshaler `redis:"binary"`
+}
+
 type i []interface{}
 
 func TestGinkgoSuite(t *testing.T) {
@@ -216,5 +229,10 @@ var _ = Describe("Scan", func() {
 		var tt TimeTime
 		Expect(Scan(&tt, i{"time"}, i{now.Format(time.RFC3339Nano)})).NotTo(HaveOccurred())
 		Expect(now.Unix()).To(Equal(tt.Time.Unix()))
+	})
+	It("Implements BinaryUnmarshaler", func() {
+		var bd BinaryData
+		Expect(Scan(&bd, i{"binary"}, i{"hello"})).NotTo(HaveOccurred())
+		Expect(bd.Binary.s).To(Equal("hello"))
 	})
 })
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
Hello @Aaditya-dubey1. I do like this, but not sure at this state if we should incorporate it withing v9, let me think about a possible breaking change this can introduce for users. If I can't produce one, I will merge it, if I can - I will hold it for v10.

Aaditya-dubey1:
@ndyakov  Thanks for the review!

I completely understand the concern around v9 stability. I’ve verified that the implementation preserves existing precedence by prioritizing `encoding.TextUnmarshaler` before `encoding.BinaryUnmarshaler`. Since standard types like `time.Time` and `net.IP` implement both, their current behavior remains unchanged.

In effect, this only adds support for custom types that previously fell through as unsupported in hscan, so it acts as a backward-compatible feature expansion.

I’ve also added a local test to verify precedence (ensuring `UnmarshalText` is always preferred when both are implemented), and I’m happy to include it in the PR if that helps increase confidence for a v9 merge.

ndyakov:
Hello @Aaditya-dubey1 I do agree it is backwards compatible in terms of it won't break the current support and the case where this would be a problem does not refer to directly using `Scan` so it should be fine, thank you!

Aaditya-dubey1:
@ndyakov  Thank you for reviewing it again. Glad we could clarify the compatibility concerns. Appreciate your time and feedback!

PR #3778, opened 2026-04-15
fix: suppress pool Close() errors for stale connections

### LINKED ISSUE #3775: Pool close errors are not suppressed when ConnMaxIdleTime expired
When using `ConnMaxIdleTime`, when a pool connection is expired and then it fails to close, the error is suppressed [here](https://github.com/redis/go-redis/blob/462843a2755010f3b6b6d0d2bcbd9011b38f9647/internal/pool/pool.go#L817) (if I understand correctly). I expect the error to be suppressed in the same manner when calling `pool.Close()`, but instead the error is returned. This makes it seem like something has gone wrong in a normal state where the server is expected to silently close idle connections (e.g. ElasticCache).

## Expected Behavior

The Pool `Close()` method returns `nil` when connections fail to close where `ConnMaxIdleTime` is exceeded.

## Current Behavior

`Close()` returns an error: `tls: failed to send closeNotify alert (but connection was closed anyway): write tcp [redacted]: write: connection timed out`

## Possible Solution

If `isHealthyConn` is true, drop the error, similar to [here](https://github.com/redis/go-redis/blob/462843a2755010f3b6b6d0d2bcbd9011b38f9647/internal/pool/pool.go#L817).

## Steps to Reproduce

1. Connect to a server that drops idle connections after a timeout
2. Configure a pool with `ConnMaxIdleTime` less than the server idle timeout
3. Use the pool
4. Wait for a connection to expire from the server
5. Call `pool.Close()`

ofekshenawa:
Fixes #3775

`pool.Close()` now checks `isHealthyConn` before propagating connection close errors, aligning its behavior with the existing error suppression in `Get()` where stale connections are silently discarded. 
This prevents spurious errors (e.g., TLS `closeNotify` timeouts) from being returned when closing a pool whose connections have already been dropped by the server due to idle timeout.

---

> [!NOTE]
> **Medium Risk**
> Changes `ConnPool.Close()` error propagation semantics by suppressing close errors for connections deemed stale via `isHealthyConn`, which could mask some real shutdown-close failures if the health check misclassifies a connection.
> 
> **Overview**
> `ConnPool.Close()` now evaluates each connection with `isHealthyConn` *before* closing it and only propagates the first `Close()` error when the connection is still considered healthy; close errors from stale/expired connections are suppressed to match `Get()`’s stale-connection handling.
> 
> Tests add a `failCloseConn` helper to simulate TLS/network close failures and verify that `Close()` returns no error for idle-timeout-stale conns, but still returns the close error for healthy ones.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9329719791e4dfbd20d9557936e09ac3be708511. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/pool/main_test.go
+++ b/internal/pool/main_test.go
@@ -130,6 +130,10 @@ type failCloseConn struct {
 }
 
 func (f *failCloseConn) Close() error {
+	// Close the underlying raw conn so that connCheck sees a closed fd,
+	// matching real-world behavior (e.g., TLS closeNotify failure after
+	// the OS-level socket is torn down).
+	f.dummyConn.rawConn.Close()
 	return f.closeErr
 }
 
--- a/internal/pool/main_test.go
+++ b/internal/pool/main_test.go
@@ -121,3 +121,18 @@ func (d *dummyRawConn) Close() {
 	d.closed = true
 	d.mu.Unlock()
 }
+
+// failCloseConn wraps a dummyConn but returns a configured error on Close().
+// This is used to simulate TLS closeNotify errors or network timeouts during connection close.
+type failCloseConn struct {
+	*dummyConn
+	closeErr error
+}
+
+func (f *failCloseConn) Close() error {
+	return f.closeErr
+}
+
+func (f *failCloseConn) SyscallConn() (syscall.RawConn, error) {
+	return f.dummyConn.SyscallConn()
+}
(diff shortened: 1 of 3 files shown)


ndyakov (dismissed):
Looks good to me!

ndyakov (approved):

PR #3782, opened 2026-04-20
feat(ft.aggregate): Add Steps for query building

ndyakov:
Redis's `FT.AGGREGATE` command allows `SORTBY` to appear multiple times at any stage of the aggregation pipeline. 

This PR also changes the Builder behaviour which is still experimental and should not be considered a breaking change.

---

> [!NOTE]
> **Medium Risk**
> Touches `FT.AGGREGATE` argument construction and the experimental `AggregateBuilder`, so mis-ordered/invalid step validation or serialization bugs could change query behavior at runtime.
> 
> **Overview**
> Adds a new step-based `FT.AGGREGATE` pipeline API via `FTAggregateOptions.Steps`, allowing `LOAD`, `APPLY`, `GROUPBY`, and `SORTBY` (with per-step `MAX`) to be repeated and interleaved in arbitrary order.
> 
> Updates query argument generation (`FTAggregateQuery`/`FTAggregateWithArgs`) to serialize `Steps`, validates that each step sets exactly one operation, and rejects mixing `Steps` with the legacy `Load`/`Apply`/`GroupBy`/`SortBy`/`SortByMax` fields (which are now marked deprecated).
> 
> Reworks the experimental `AggregateBuilder` to emit ordered `Steps`, adds builder-side validation (errors returned from `Run`), and changes `SortBy` behavior to merge only consecutive `SortBy` calls into a single `SORTBY` step. Docs/tests are updated and a new runnable example (`example/search-aggregate-steps`) is added to demonstrate multi-stage pipelines and shard-level trimming (`SORTBY ... MAX` before `GROUPBY`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d13b6c00c25c1c35dd0899a8ca22112fb1949c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


search_commands.go
```diff
@@ -258,22 +258,42 @@ type FTAggregateWithCursor struct {
...
 	Scorer string
 	// AddScores is available in Redis CE 8
-	AddScores         bool
-	Apply             []FTAggregateApply
+	AddScores bool
+
+	// Steps is the ordered sequence of aggregation pipeline operations.
+	// It can contain LOAD, APPLY, GROUPBY and SORTBY in any order, multiple times.
+	// When Steps is non-empty, the deprecated Load, Apply, GroupBy, SortBy and
+	// SortByMax fields are ignored.
```
elena-kolevska: I don't think we should silently fail on this. We should error out with something like "Steps cannot be combined with Load/GroupBy..."

search_commands.go
```diff
@@ -656,53 +748,66 @@ func FTAggregateQuery(query string, options *FTAggregateOptions) (AggregateQuery
...
 		}
 
-		for _, apply := range options.Apply {
-			queryArgs = append(queryArgs, "APPLY", apply.Field)
-			if apply.As != "" {
-				queryArgs = append(queryArgs, "AS", apply.As)
+		if len(options.Steps) > 0 {
+			for _, step := range options.Steps {
+				if options.LoadAll && step.Load != nil {
+					continue
```
elena-kolevska: We should return the same error as the legacy code: `FT.AGGREGATE: LOADALL and LOAD are mutually exclusive`

search_commands.go
```diff
@@ -617,6 +648,67 @@ func (c cmdable) FTAggregate(ctx context.Context, index string, query string) *M
 	return cmd
 }
 
+// appendFTAggregateStep appends the Redis command arguments for a single
+// aggregation pipeline step (LOAD/APPLY/GROUPBY/SORTBY). Only the first
+// non-nil field on the step is emitted.
+func appendFTAggregateStep(args []interface{}, step FTAggregateStep) ([]interface{}, error) {
+	switch {
```
elena-kolevska, line 694: We shouldn't rely on the docs here, we need to validate there's no more than one element in each step.

```suggestion
	set := 0
	if step.Load != nil    { set++ }
	if step.Apply != nil   { set++ }
	if step.GroupBy != nil { set++ }
	if step.SortBy != nil  { set++ }
	if set != 1 {
		return args, fmt.Errorf(
			"FT.AGGREGATE: each step must set exactly one of Load, Apply, GroupBy, SortBy (got %d)", set)
	}
	
	switch {
```

search_builders.go
```diff
@@ -259,62 +279,65 @@ func (b *AggregateBuilder) Timeout(ms int) *AggregateBuilder {
...
+	g.Reduce = append(g.Reduce, FTAggregateReducer{Reducer: fn, Args: args, As: alias})
 	return b
 }
 
-// SortBy adds SORTBY <field> ASC|DESC.
+// SortBy adds SORTBY <field> ASC|DESC. Consecutive SortBy calls (with no
+// other step in between) are merged into a single SORTBY clause so fields
+// act as tiebreakers. A SortBy call after a non-SortBy step starts a new
+// SORTBY step.
 func (b *AggregateBuilder) SortBy(field string, asc bool) *AggregateBuilder {
```
elena-kolevska, line 324: We should make it clear in the release docs that we have a semantics change here (which is ok, since the builder is experimental).

search_builders.go
```diff
@@ -241,15 +241,35 @@ func (b *AggregateBuilder) LoadAll() *AggregateBuilder {
...
+	for i := len(b.options.Steps) - 1; i >= 0; i-- {
+		if b.options.Steps[i].GroupBy != nil {
+			return b.options.Steps[i].GroupBy
+		}
+	}
+	return nil
+}
+
+// lastSortByStep returns a pointer to the last SORTBY step, or nil if none.
+func (b *AggregateBuilder) lastSortByStep() *FTAggregateSortByStep {
```
elena-kolevska: I'm not sure if we should be doing this. I think we should just check if the last step before Max is a sortBy and return an error if not.
It's also a performance thing - I prefer developers getting the error while developing, instead of paying extra cost in production every time the command is executed.

elena-kolevska (changes requested):
Looks mostly good, with smaller comments. Maybe we can clean up some general repetition.

Also, we need an example in `query_agg_test.go`.

search_commands.go
ndyakov: resolved.

search_commands.go
ndyakov: addressed.

search_commands.go
ndyakov: addressed.

search_builders.go
ndyakov: altered the comment and fixed the implementation.

search_builders.go
ndyakov: altered the implementation to check the last step, this should work better. I will let @ofekshenawa review the builder if it is correct. We can fix the builder later as well, as it is not yet officially the recommended approach of building ft queries.

elena-kolevska (approved):

ndyakov:
@ofekshenawa please review and merge if there are no blockers from your side.

ofekshenawa (approved):
LGTM

PR #3784, opened 2026-04-22
redis: stop chaining baseClient.onClose on every initConn with StreamingCredentialsProvider

### LINKED ISSUE #3772: Memory leak: wrappedOnClose builds unbounded closure chain per initConn when using StreamingCredentialsProvider
## Versions

- **go-redis:** v9.18.0
- **Redis:** 6.x (reproduced with `redis:6` Docker image; bug is in go-redis, not Redis)

## What did you do?

Used `StreamingCredentialsProvider` with Azure EntraID token rotation (~45–90 min lifetime).
With `PoolSize=100`, the service OOM-killed every 18–22 hours in production.

## What did you expect?

Heap stays flat regardless of how many token rotations occur.

## What did you see instead?

Heap grows ~600 MB per rotation and kills the process.

Reproduction (PoolSize=10, 50 workers, 15s rotation interval):

```
heap=   2 MB  conns=10  stale=0      rotations=0
heap= 603 MB  conns=10  stale=8711   rotations=1
heap=1038 MB  conns=10  stale=15082  rotations=2
heap=3343 MB  conns=10  stale=47059  rotations=11   (3 minutes)
```

## Root cause

In `redis.go:486`, inside `initConn`:

```go
c.onClose = c.wrappedOnClose(unsubscribeFromCredentialsProvider)  // line 486 — BUG
cn.SetOnClose(unsubscribeFromCredentialsProvider)                  // line 487 — correct
```

`wrappedOnClose` captures the current `c.onClose` in a new closure and returns a wrapper
that calls both. Called on every `initConn`, this builds an unbounded linked list on the
**shared `baseClient`**:

```
initConn #1: c.onClose = wrap(unsub_1, nil)
initConn #2: c.onClose = wrap(unsub_2, wrap(unsub_1, nil))
initConn #3: c.onClose = wrap(unsub_3, wrap(unsub_2, wrap(unsub_1, nil)))
...
initConn #N: chain of N closures, each retaining a dead *pool.Conn (~70 KB TLS buffers)
```

This chain lives on the shared `baseClient` and is only walked at `Close()` — never trimmed
during normal operation. The leak is amplified by `ReAuthPoolHook`: during a token rotation
the hook rejects all pooled connections simultaneously, firing a storm of
`queuedNewConn → initConn` calls that rapidly extends the chain.

`cn.SetOnClose` on line 487 already correctly handles per-connection unsubscription when
individual connections close. Line 486 is redundant and harmful.

## Fix

```diff
-        c.onClose = c.wrappedOnClose(unsubscribeFromCredentialsProvider)
         cn.SetOnClose(unsubscribeFromCredentialsProvider)
```

With the fix applied, heap stays flat at ~3 MB for the full 3-minute run.

## Minimal reproduction

```bash
docker run --rm -p 6399:6379 redis:6 redis-server --requirepass test-token
go run main.go
```

Self-contained ~130-line reproduction: https://gist.github.com/fxsml/0191f9a91b62078ddecf84cd94160e9f

## Related

- #3542 — unexpected `used` connection count spike with `StreamingCredentialsProvider` (same `ReAuthPoolHook` path)
- #3582 — hook rejection errors after v9.16 upgrade (same path)

SAY-5:
## Summary

Fixes #3772.

`initConn` attached `unsubscribeFromCredentialsProvider` to the connection two ways:

```go
c.onClose = c.wrappedOnClose(unsubscribeFromCredentialsProvider)
cn.SetOnClose(unsubscribeFromCredentialsProvider)
```

`cn.SetOnClose` on the per-connection `pool.Conn` is the one that actually runs when that connection leaves the pool. `wrappedOnClose` wraps the prior `c.onClose` in a new closure, so every `initConn` extends a linked list on the shared `baseClient`. A token rotation storm (`ReAuthPoolHook` rejects the whole pool simultaneously and forces `initConn` for every replacement Conn) grows that list by one link per rotation, each retaining a dead `*pool.Conn` via the captured unsubscribe callback (~70 KB of TLS buffers per link). The chain is only walked at `baseClient.Close` — never trimmed during normal operation. Production heap grew ~600 MB per rotation and the process OOM'd every 18-22h.

## Fix

Drop the `c.onClose` assignment; `cn.SetOnClose` already covers the per-connection unsubscribe. The `wrappedOnClose` helper itself is retained — `internal_test.go`'s `Ring` shard-close counter test still uses it for the intended fan-out pattern, so no public API changes.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -short -run "TestStreamingCredentialsProvider|TestReAuthCredentialsListener|TestInitConnNilMaintNotificationsConfig"` green
- [ ] Repro from #3772 (PoolSize=10, 50 workers, 15s rotation): heap should stay flat at ~3 MB instead of growing ~600 MB per rotation.

Retains behaviour for non-streaming credential providers entirely — the changed branch only fires when `c.opt.StreamingCredentialsProvider != nil`.

---

> [!NOTE]
> **Medium Risk**
> Adjusts connection close-hook wiring for streaming credential subscriptions; while scoped to `StreamingCredentialsProvider`, it affects connection lifecycle and cleanup behavior under rotation/reconnect scenarios.
> 
> **Overview**
> Fixes a leak/retention issue when `StreamingCredentialsProvider` is enabled by **removing the per-`initConn` wrapping of `baseClient.onClose`** and relying only on the owning `pool.Conn`’s `SetOnClose(unsubscribe)` callback.
> 
> This prevents the shared `baseClient.onClose` chain from growing without bound during frequent re-auth/token-rotation events, avoiding retention of dead connections via captured unsubscribe closures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8a2ce9089ec3f54962a75b74fe1cbcaf6ac6689. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -496,7 +496,14 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 			return fmt.Errorf("failed to subscribe to streaming credentials: %w", initErr)
 		}
 
-		c.onClose = c.wrappedOnClose(unsubscribeFromCredentialsProvider)
+		// Attaching the unsubscribe callback to the per-connection
+		// SetOnClose hook below is sufficient: it runs exactly when the
+		// connection that owns this subscription goes away. Wrapping
+		// c.onClose here as well rebuilds the shared baseClient.onClose
+		// chain on every initConn, which grows unbounded under token
+		// rotation storms and retains each dead *pool.Conn via the
+		// captured closure (#3772). baseClient.Close walks the chain
+		// only once at shutdown, so it never trimmed live entries.
 		cn.SetOnClose(unsubscribeFromCredentialsProvider)
 
 		username, password = credentials.BasicAuth()


ndyakov:
@SAY-5 thank you for opening this. I actually did a slightly different approach ( the same for this specific issue, but covered the baseClient case as well) last night but opened the PR today - https://github.com/redis/go-redis/pull/3785 I hope you don't mind me closing this one in favor of #3785

PR #3794, opened 2026-04-24
csc: Support standalone client

ofekshenawa:
> [!NOTE]
> **High Risk**
> Large correctness-critical surface (cache hits, invalidations, auth/DB namespacing, and connection lifecycle races); stale reads or cross-tenant leakage would be severe if any guard fails.
> 
> **Overview**
> Adds **end-to-end client-side caching (CSC)** for standalone `Client` on RESP3 using the **SharedTracking** strategy: `CLIENT TRACKING`, RESP3 `invalidate` push handling, a per-client **idle-connection drainer**, and **per-connection cache attribution** so entries are evicted when connections close, reinit, or lose tracking.
> 
> **Configuration & API:** `Options.ClientSideCache` / `ClientSideCacheConfig`, strategy defaulting to SharedTracking, `Client.CSCStats()`, and typed **`CLIENT TRACKING`** helpers (with client-side validation). Built-in CSC **rejects** manual `CLIENT TRACKING` and other commands that would desync pool state (`SELECT`, `AUTH`, protocol-changing `HELLO`, `RESET`, pooled `SUBSCRIBE`).
> 
> **Correctness hardening:** Cache keys and invalidation indexes use **wire-faithful** RESP encoding; several commands are dropped from the allow-list (`XREAD`, `XPENDING`, `SORT_RO` with BY/GET, streaming `RawWriteTo`). Namespaces separate **DB 0 only**, fixed ACL username (not passwords), and disable CSC for **dynamic credential providers**. CSC disables itself when tracking fails or RESP3 falls back to RESP2.
> 
> **Pool / lifecycle:** New conn hooks (`onCscClose`, `onCscReinit`, read-pending probes), safer TLS health checks, RESP3 read-buffer clamping, GC cleanup for forgotten drainers, and removal of `internal/cache` in favor of the public local cache with `ConnOwnedCache`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1b3d8942fc944ecf4ee0db7db85c28599ed078d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


csc_integration.go
```diff
@@ -0,0 +1,210 @@
...
+		seen[cn.GetID()] = struct{}{}
+		borrowed = append(borrowed, cn)
+		if err := c.peekAndProcessPushNotifications(ctx, cn); err != nil {
+			internal.Logger.Printf(ctx, "csc: error draining invalidations on cache hit: %v", err)
+		}
+	}
+	for _, cn := range borrowed {
+		c.connPool.Put(ctx, cn)
+	}
+}
```
ndyakov: this will be quite consuming for larger pools, let's discuss different approaches here.

ndyakov (commented):
took a brief look at this as well.

local_cache.go
```diff
@@ -330,14 +341,17 @@ func (c *localCache) Flush() int {
...
+	// DeleteByRedisKey: the in-flight reply is post-invalidation.
+	removed := 0
+	for cacheKey, entry := range c.entries {
+		if entry.State == CacheEntryInProgress {
+			continue
+		}
+		if c.removeEntryLocked(cacheKey) {
+			removed++
+		}
 	}
```
ndyakov: what will happen if the flush was triggered from the connection that was fetching the entry?

redis.go
```diff
@@ -1259,6 +1298,18 @@ func NewClient(opt *Options) *Client {
...
+	// delivered as push notifications, so RESP2 clients cannot use CSC.
+	if opt.Protocol == 3 {
+		var cache Cache
+		if explicit := opt.ClientSideCache; explicit != nil {
+			cache = explicit
+		} else if cfg := opt.ClientSideCacheConfig; cfg != nil {
+			cache = NewLocalCache(*cfg)
+		}
+		c.baseClient.attachCSC(context.Background(), cache)
+	}
```
ndyakov, line 1504: as soon as the pool is created it will start to initiate idle connections (if minIdle is set). if we attach the csc later than initConn, the tracking on the connections won't be enabled. However, processing commands on those connections will be considered cacheable (based on the command, of course) and data will be stored in the cache without a connection tracking its invalidation.

redis.go
```diff
@@ -539,6 +544,7 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 		}
 	}
 
+	trackingEnabled := c.csc != nil && c.opt.Protocol == 3
```
ndyakov: wouldn't it be better if we check the options here instead of the csc ?

csc_integration.go
```diff
@@ -0,0 +1,209 @@
...
+	borrowed := make([]*pool.Conn, 0, remaining)
+	for i := 0; i < remaining; i++ {
+		cn, err := c.connPool.Get(ctx)
+		if err != nil {
+			break
+		}
+		if _, dup := seen[cn.GetID()]; dup {
+			c.connPool.Put(ctx, cn)
+			break
+		}
```
ndyakov: can't me introduce a mechanism to get a connection with a blacklist of ids? i think if we do the check in the pool during holding the lock , instead of locking and unlocking twice (on get and put) this should be better. We have to validate this assumption, as we are going to hold the lock a bit longer.

csc_integration.go
```diff
@@ -0,0 +1,209 @@
...
+	key := dbNamespacedKey(db, rawKey)
+	nsRedisKeys := make([]string, len(redisKeys))
+	for i, k := range redisKeys {
+		nsRedisKeys[i] = dbNamespacedKey(db, k)
+	}
+
+	if _, ok := c.csc.Get(ctx, key); ok {
+		// Drain pending invalidations and re-check so the hit path cannot
+		// serve a value the server has already marked stale.
+		c.drainPendingInvalidations(ctx)
```
ndyakov: here it will make sense to check ONLY the owner of the cached data, not all connections

options.go
```diff
@@ -312,6 +312,18 @@ type Options struct {
 	// transitions seamlessly. Requires Protocol: 3 (RESP3) for push notifications.
 	// If nil, maintnotifications are in "auto" mode and will be enabled if the server supports it.
 	MaintNotificationsConfig *maintnotifications.Config
+
+	// ClientSideCacheConfig enables client-side caching when non-nil.
+	// Requires Protocol: 3 (RESP3) so that invalidation messages are delivered
+	// as out-of-band push notifications. If ClientSideCache is also set, it
+	// takes precedence over this config.
+	ClientSideCacheConfig *ClientSideCacheConfig
```
ndyakov: let's add a log if this is enabled but the protocol is 2, i think we should do the same for maintnotifications (if it is missing).

options.go
```diff
@@ -312,6 +312,18 @@ type Options struct {
...
+
+	// ClientSideCache is an explicit Cache implementation used for client-side
+	// caching. When set, it overrides ClientSideCacheConfig. Intended for
+	// advanced users that want to share a cache across clients or supply a
+	// custom implementation.
```
ndyakov, line 332: sharing across clients is safe only if they use the same database underneath, right?

csc_commands.go
```diff
@@ -1,224 +1,159 @@
...
-	"EXISTS": {}, "TYPE": {}, "SORT_RO": {}, "LCS": {},
+	"exists": {}, "type": {}, "sort_ro": {}, "lcs": {},
 	// Geo commands
-	"GEODIST": {}, "GEOHASH": {}, "GEOPOS": {}, "GEOSEARCH": {},
-	"GEORADIUSBYMEMBER_RO": {}, "GEORADIUS_RO": {},
+	"geodist": {}, "geohash": {}, "geopos": {}, "geosearch": {},
+	"georadiusbymember_ro": {}, "georadius_ro": {},
 	// Stream commands
-	"XLEN": {}, "XPENDING": {}, "XRANGE": {}, "XREAD": {}, "XREVRANGE": {},
+	"xlen": {}, "xpending": {}, "xrange": {}, "xread": {}, "xrevrange": {},
```
ndyakov: are stream commands cacheable? I wouldn't think so. Can we validate that.

csc_integration.go
```diff
@@ -0,0 +1,209 @@
...
+func (h *invalidateHandler) HandlePushNotification(
+	_ context.Context, _ push.NotificationHandlerContext, notification []interface{},
+) error {
+	if h.cache == nil || len(notification) < 2 {
+		return nil
+	}
+
+	switch payload := notification[1].(type) {
+	case nil:
+		h.cache.Flush()
```
ndyakov: 👍

ndyakov (changes requested):
The potential race between `initConn` and `attachCSC` is a blocker, the rest are potential improvements (that may have been addressed in #3851 ). 

I did not see invalidation on connection drop/change but I assume it is implemented in #3851

csc_integration.go
ofekshenawa: Agreed.  #3851 removes this hit-path walk entirely: hits are served straight from cache and a background drainer applies invalidations instead.

csc_integration.go
ofekshenawa: fixed in #3851

options.go
ofekshenawa: Done in 884beb2c . Options.init() warns when CSC is configured with Protocol: 2

options.go
ofekshenawa: Correct, doc note added, and CSC is now restricted to DB 0 (disabled with a warning otherwise).

ofekshenawa:
> The potential race between `initConn` and `attachCSC` is a blocker, the rest are potential improvements (that may have been addressed in #3851 ).
> 
> I did not see invalidation on connection drop/change but I assume it is implemented in #3851

The pool warm-up only dials, CLIENT TRACKING ON runs in initConn on first acquisition, which is always after attachCSC completes in NewClient. Also reconnect re-runs initConn too, so every conn that can serve a cacheable command has tracking.

local_cache.go
ofekshenawa: We safe here. Replies and pushes are ordered per connection, so a reply after the flush push is post-flush state, and one before it is deleted when the push is processed.

redis.go
ofekshenawa: c.csc is nil when attachCSC declines (DB != 0, registration failure), so checking options would enable tracking with no cache attached. This is why checking c.csc covers more than checking options.

csc_integration.go
ofekshenawa: Fixed in #3851 .

csc_multidb_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/redis/go-redis/v9"
+)
+
+// cscMultiDBAddr is the address used by the standalone CSC regression tests
+// in this file. We bypass the Ginkgo harness (which assumes a CI cluster
+// fixture) and connect directly so these tests run with just a single Redis
+// instance available.
+func cscMultiDBAddr() string {
+	if v := os.Getenv("REDIS_ADDR"); v != "" {
```
ndyakov, line 25: The problem: these tests can wipe a Redis instance the test suite was never pointed at.

How it happens: the tests read REDIS_ADDR and fall back to localhost:6379, but the rest of the suite is configured through REDIS_PORT. TestCSCReadYourWrites calls FlushDB. So on a machine where only REDIS_PORT is set and something else runs on 6379 (a local dev instance, another CI service), the test flushes that instead of the suite's server.

Fix: read REDIS_PORT too, or reuse the suite's address helper. Raising it again because the head has not moved since I first flagged it.

ndyakov (changes requested):
Requesting changes on two grounds: the inline test issue below, and the standing constraint that this PR must only ship together with #3851 (the connection-loss staleness and the buffered-invalidation miss reproduce at this head and are addressed up the stack).

csc_commands.go
```diff
@@ -1,224 +1,165 @@
...
+}
+
+// sortROHasByGet reports whether a SORT_RO invocation uses BY or GET
+// (case-insensitive), scanning past the command name and key.
+func sortROHasByGet(args []interface{}) bool {
+	if len(args) < 3 {
+		return false
+	}
+	for _, a := range args[2:] {
+		s, ok := a.(string)
```
ndyakov: isn't it valid for this to be a `[]byte`? if so, won't this check fail?

redis.go
```diff
@@ -552,13 +629,32 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
 			pipe.ClientSetName(ctx, c.opt.ClientName)
 		}
 
+		if trackingEnabled {
+			// Must run before any cacheable command is issued on this conn.
+			pipe.ClientTrackingOn(ctx, nil)
+		}
```
ndyakov, line 653: what if the client tracking on command fails? instead of silently continuing to work without csc, it will fail the initConn pipeline / handshake

csc_integration.go
```diff
@@ -0,0 +1,713 @@
...
+			c.csc.Flush()
+		}
+	})
+}
+
+// applyCachedReply populates cmd from a previously captured raw RESP reply by
+// replaying it through the command's own readReply.
+func applyCachedReply(cmd Cmder, raw []byte) error {
+	return cmd.readReply(proto.NewReader(bytes.NewReader(raw)))
+}
```
ndyakov, line 767: proto.NewReader allocates a 32 KiB bufio reader, so every cache hit pays a
32 KiB allocation just to replay a reply that is usually a few hundred bytes.
the capture path in redis.go line 1003 pays the same again on every fetch.
at the request rates this feature targets that is significant GC pressure on
the exact path that should be near free. maybe use this at both sites

    cmd.readReply(proto.NewReaderSize(bytes.NewReader(raw), len(raw)+1))

that drops the fixed 32 KiB to the reply size and bufio clamps tiny sizes to
its 16 byte minimum, so it is safe for the smallest replies too. and if we
later see that we need zero allocations here, both proto.Reader and
bytes.Reader have Reset, so we can move to something like

    type replayReader struct {
        br bytes.Reader
        rd *proto.Reader
    }

    var replayReaders = sync.Pool{New: func() any {
        rr := &replayReader{}
        rr.rd = proto.NewReaderSize(&rr.br, 512)
        return rr
    }}

    func applyCachedReply(cmd Cmder, raw []byte) error {
        rr := replayReaders.Get().(*replayReader)
        defer replayReaders.Put(rr)
        rr.br.Reset(raw)
        rr.rd.Reset(&rr.br)
        return cmd.readReply(rr.rd)
    }

the small buffer is fine either way, readLine already assembles lines that
overflow it and the replay never peeks push headers, bigger replies just
refill from the bytes.Reader which is a cheap copy. i would do the one liner
in this PR and keep the pool for a follow up with a benchmark.

ndyakov (commented):
submitting a partial review

ndyakov (approved):
Great work @ofekshenawa!

Everything is addressed and polished. One thing that we can address in the future branch is to fix the CI.

Looks like CSC does not work with our image for RE (or may not work on some images at all, depends on a configuration (enabled/disabled on the server side)). Can we detect (in the tests) if the csc is allowed / can be enabled (do a probe) and if the probe fails - skip the tests ?

`ERR command 'client|tracking' is not allowed` observed on https://github.com/redis/go-redis/actions/runs/30464937225/job/90620192766?pr=3794

PR #3804, opened 2026-05-07
refactor(TODO-3803): Enhance key position retrieval with CommandInfo caching

### LINKED ISSUE #3803: Use CommandInfo to determine first key position in cmdFirstKeyPos
There's a TODO in command.go:
```go
// TODO: Use the data in CommandInfo to determine the first key position.
```
Right now **cmdFirstKeyPos** hardcodes a handful of special cases and falls back to position 1 for everything else. The problem is we already fetch **CommandInfo** (which includes FirstKeyPos) via COMMAND INFO and cache it, we're just not using it here.

The **keylessCommands** map also has to be kept in sync manually, which is easy to get wrong as new commands are added.

The fix would be to look up **CommandInfo** in **cmdFirstKeyPos** when available and use **FirstKeyPos** from there, keeping the current logic as a fallback for cold cache. The **eval/evalsha** family would still need special handling since their key position depends on the **numkeys** argument at runtime.

retr0-kernel:
Fixes #3803 
There's a TODO that's been sitting in **cmdFirstKeyPos** function. The function figures out which argument position holds the first key for a given Redis command and we need this to route commands to the right node. The way it works right now is by checking a hardcoded map of around 40 keyless commands and defaulting to 1 for everything else. 

The thing is we already fetch **COMMAND INFO** from Redis at startup and cache it, and that response has the exact first-key position for every command baked in. We just weren't using it here. So every time Redis adds a new keyless command, someone has to remember to update our map by hand. Easy to forget, and we've probably already missed some.

**What I changed:** I wired the cached **COMMAND INFO** data into the routing logic. The function now takes an optional `*CommandInfo` and uses `info.FirstKeyPos` directly when it's available. All the cluster and ring call sites pass it in via a small helper that peeks at the already-warm cache.

**The round-trip concern:** @ndyakov had a genuine concern that first-key resolution shouldn't add a Redis round-trip. It doesn't. To tackle that, I added a `Peek()` method on the cache that just returns whatever is already in memory, so no network call, no blocking. If the cache is cold (like still starting up or something), it returns nil and the code falls through to the old hardcoded table exactly as before. Once the cache is warm, it's just a map lookup. 

**One edge case:** `eval/evalsha` still get special-cased. Their `FirstKeyPos` in **CommandInfo** is 3, but that's only correct when `numkeys > 0`. When `numkeys == 0` there are no key arguments at all, and you can only know that by looking at the actual runtime arguments, so no cached metadata can help there. So that logic is unchanged.

I've added comments throughout the code to make it easier to follow while reviewing. Tested locally and everything looks good, but if something seems off or I've got it wrong somewhere, just let me know and I'll fix it.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster slot and multi-shard routing logic; wrong first-key positions could misroute commands, though cold-cache fallbacks preserve prior behavior.
> 
> **Overview**
> **First-key routing** now prefers cached `COMMAND INFO` metadata instead of relying only on a hand-maintained keyless map and defaulting keyed commands to argument index `1`.
> 
> `cmdFirstKeyPos` is replaced by **`cmdFirstKeyPosWithInfo`**, which still honors per-command `firstKeyPos`, the `keylessCommands` set, and **`eval`/`evalsha` `numkeys` runtime logic**, but uses **`CommandInfo.FirstKeyPos`** when a non-nil `*CommandInfo` is supplied. **`cmdsInfoCache`** gains **`Peek()`** (read-only, no Redis round-trip; nil when cold) and **`sync.RWMutex`** so concurrent peeks don’t block each other after warm-up.
> 
> **Cluster** call sites use **`cmdInfoPeek`**, batch **`Peek()`** once in **`slottedKeyedCommands`**, and **`cmdSlotWithPos`** so slot computation can reuse a resolved key index. **Multi-shard** paths pass **`firstKeyPos`** into **`createSlotSpecificCommand`** so sub-commands don’t re-resolve keys inconsistently. **Ring** still passes **`nil`** for `CommandInfo` (TODO to warm cache like cluster).
> 
> Unit test **`TestCmdFirstKeyPosWithInfo_UsesCommandInfoWhenWarm`** covers cold vs warm cache behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8752c6376d3ca7b29cff5dafdad5aaddc84cf4a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


retr0-kernel:
Hi @ndyakov ,
This PR addresses issue #3803 , been working on this one for a while. Let me know if anything is needed from my end or any questions.

command.go
```diff
@@ -4818,6 +4824,17 @@ func (c *cmdsInfoCache) Refresh() {
 	c.once = internal.Once{}
 }
 
+// Peek returns the cached CommandInfo map without triggering a Redis round-trip.
+// Returns nil when the cache is cold; callers should fall back to other heuristics.
+func (c *cmdsInfoCache) Peek() map[string]*CommandInfo {
+	if c == nil {
+		return nil
+	}
+	c.refreshLock.Lock()
```
ndyakov: @retr0-kernel let's make `refreshLock` `RWMutex` and lock it only for reading in `Peek` and `Get`.

command.go
```diff
@@ -267,35 +267,41 @@ func writeCmd(wr *proto.Writer, cmd Cmder) error {
...
 		return int(pos)
 	}
 
 	name := cmd.Name()
 
-	// first check if the command is keyless
-	if _, ok := keylessCommands[name]; ok {
-		return 0
-	}
-
```
ndyakov, line 296: i do think we can still check **first** if the command is marked as keyless.

ndyakov (commented):
left some suggestions. keep in mind there is another configuration with the default command policies and we may want to combine those. will get back to you with more information soon.

retr0-kernel:
> left some suggestions. keep in mind there is another configuration with the default command policies and we may want to combine those. will get back to you with more information soon.

I will reiterate and get back to you on these.

command.go
retr0-kernel: Ok swapped refreshLock from sync.Mutex to sync.RWMutex. Peek() only reads c.cmds so it uses RLock now, meaning concurrent routing goroutines calling it in parallel no longer block each other. Get() and Refresh() still take the write lock since they actually modify state. Sending the change for this just now.

command.go
retr0-kernel: Agreed. Sending a change where keylessCommands check now runs before the CommandInfo lookup, it's a straight O(1) map hit with no cache dependency, so it made more sense as the fast-path for that known set. CommandInfo.FirstKeyPos picks up everything else once the cache is warm, and return 1 stays as the final fallback just like before. Will send the change for this too in the commit.

retr0-kernel:
Hi @ndyakov ,
Did the changes you requested. Open to more as you suggested. Thanks for the review again.

> will get back to you with more information soon.

retr0-kernel:
Sorry for the mess. Accidentally committed with my official so just changed the authors. The code is still intact if any issues will cherry pick y changes into a new branch.

retr0-kernel:
Hi @ndyakov any updates on this? I was thinking of cherry picking for a clean change.

command.go
```diff
@@ -274,36 +274,42 @@ func writeCmd(wr *proto.Writer, cmd Cmder) error {
...
 		}
+		// CommandInfo (if available) gives the correct answer
+		// otherwise the hardcoded fallback applies.
+	}
+
+	// if the command is keyless O(1) in-memory, no round-trip needed.
+	if _, ok := keylessCommands[name]; ok {
+		return 0
 	}
+
```
ndyakov: let's first check the keyless commands (like it was previously, before doing the switch with the name).

retr0-kernel: Hi @ndyakov 
This is done. Did it like it was before.

retr0-kernel: I can see a failing testcase. I did tried researching it a bit --> https://github.com/redis/go-redis/actions/runs/25592951942/job/75133806596?pr=3804 and I think it is unrelated to this PR and our changes don't touch any of that code.

retr0-kernel:
Hi @ndyakov any updates on this?

osscluster_router.go
```diff
@@ -179,7 +179,7 @@ func (c *ClusterClient) executeMultiSlot(ctx context.Context, cmd Cmder, slotMap
 // createSlotSpecificCommand creates a new command for a specific slot's keys
 func (c *ClusterClient) createSlotSpecificCommand(ctx context.Context, originalCmd Cmder, keys []string) Cmder {
 	originalArgs := originalCmd.Args()
-	firstKeyPos := int(cmdFirstKeyPos(originalCmd))
+	firstKeyPos := cmdFirstKeyPosWithInfo(originalCmd, c.cmdInfoPeek(originalCmd.Name()))
```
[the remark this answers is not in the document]
ndyakov: this should never happen as keyless commands are checked before looking up the cache.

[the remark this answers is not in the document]
retr0-kernel: well since the keyless map check runs before we ever touch the cache, there's no window for that inconsistency to happen right? No code changes needed here I guess?

retr0-kernel:
Hi @ndyakov any actionable items from my side?

retr0-kernel:
Is this MR still ready to merge? Let me know if any changed needed from my side @ndyakov

ndyakov:
@retr0-kernel i will do a full review today or tomorrow and let you know, thank you for pinging me.

osscluster.go
```diff
@@ -1919,7 +1919,7 @@ func (c *ClusterClient) slottedKeyedCommands(ctx context.Context, cmds []Cmder)
 
 	prefferedRandomSlot := -1
 	for _, cmd := range cmds {
-		if cmdFirstKeyPos(cmd) == 0 {
+		if cmdFirstKeyPosWithInfo(cmd, c.cmdInfoPeek(cmd.Name())) == 0 {
```
ndyakov: Per loop iteration we now call `Peek()` (an RLock/RUnlock) twice for the same command - once here and again inside c.cmdSlot when cmdFirstKeyPosWithInfo(cmd, c.cmdInfoPeek(cmd.Name())). `cmd.Name()` also runs twice which is not needed.
Maybe we can peek the full map once outside the loop, then do a single map lookup per command and pass the resolved `*CommandInfo` into both the filter check and cmdSlot. Something along the lines of:

```
cachedInfo := c.cmdsInfoCache.Peek()
for _, cmd := range cmds {
    var info *CommandInfo
    if cachedInfo != nil {
        info = cachedInfo[cmd.Name()]
    }
    pos := cmdFirstKeyPosWithInfo(cmd, info)
    if pos == 0 {
        continue
    }
    slot := cmdSlotWithInfo(cmd, pos, prefferedRandomSlot) // or thread `pos` through cmdSlot
}
```

osscluster_router.go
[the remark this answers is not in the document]
ndyakov: For keyless it is safe, but for other commands - not so much. 
`executeMultiShard` computes `firstKeyPos` here, then `createSlotSpecificCommand` (line 182) independently recomputes it via another `Peek()`. These two calls are sequential within the same flow, and if the cache warms between them they can disagree — executeMultiShard extracts keys from one offset and createSlotSpecificCommand rebuilds args from a different offset, producing a malformed command. Should be rare (if it happens at all), but actually is easy to fix now. Compute `firstKeyPos` once in `executeMultiShard` and pass it down as an argument to `createSlotSpecificCommand`.

command.go
```diff
@@ -296,14 +296,20 @@ func cmdFirstKeyPos(cmd Cmder) int {
...
 		}
+		// CommandInfo (if available) gives the correct answer
+		// otherwise the hardcoded fallback applies.
 	}
+
+	// Use CommandInfo cache when warm (in-memory only, no extra round-trips).
+	if info != nil {
+		return int(info.FirstKeyPos)
+	}
+
```
ndyakov, line 317: Could you add a routing test that actually exercises this branch? Something like: a command name that is not in `keylessCommands`, has no `cmd.firstKeyPos()` set, and whose `COMMAND INFO` returns `FirstKeyPos = 0` (any module/extension keyless command works). With a cold cache the existing fallback returns 1; with a warm cache this branch should return 0. Right now nothing in the suite distinguishes the two, so we have no regression guard on the actual change.
A focused unit test against `cmdsInfoCache` with an injected fn is probably the cleanest — no need for database.

command.go
```diff
@@ -4936,6 +4942,17 @@ func (c *cmdsInfoCache) Refresh() {
 	c.once = internal.Once{}
 }
 
+// Peek returns the cached CommandInfo map without triggering a Redis round-trip.
+// Returns nil when the cache is cold; callers should fall back to other heuristics.
```
ndyakov, line 5138:
```suggestion
// Returns nil when the cache is cold; callers should fall back to other heuristics.
// Note: during the very first Get() (initial population) this call will block on
// the writer lock. After that, concurrent Peek() calls do not block each other.
// The returned map and its entries MUST NOT be mutated by the caller.
```

ndyakov (changes requested):
Thanks for sticking with this @retr0-kernel , and apologies for the delay. I went through the diff end-to-end and there are a few things to address before this can land:

1. cmdsInfoCache is never populated for `ring`. cmdsInfoCache.Get() is only called from ClusterClient.cmdInfo (osscluster.go:2294). Nothing in ring.go ever calls Get, so Peek() always returns nil for Ring and every info argument we pass into cmdFirstKeyPosWithInfo from ring code is nil. Net effect for Ring: identical routing as before, plus an extra RLock per command. We need to either populate the ring cache lazily on first routing decision (mirroring how cluster gets it via cmdInfo) or drop the ring changes from this PR and do it as a follow-up.

2. Per-loop redundant peeks in slottedKeyedCommands and executeMultiShard/createSlotSpecificCommand. See inline comments. The ring.generalProcessPipeline pattern (peek once, lookup per cmd) is what we want on the cluster side too.

3. No test covers the new path. We need at least one regression test where the cached FirstKeyPos differs from the hardcoded fallback — otherwise the suite only ever exercises the cold-cache branch.

4. Peek() may block - added a suggestion to edit the comment.

retr0-kernel:
Hi @ndyakov no issues. I will get these done and will update

retr0-kernel:
Hey @ndyakov, sorry for the slow reply. Was tied up with some other stuff this past week.

### Addressed everything from your review:
Dropped the ring changes. The cache is never warmed there since **Get()** is never called in the ring routing path, so **Peek()** always returned nil and the **RLock** per command was just noise. Both call sites in **cmdShard** and **generalProcessPipeline** now pass nil directly, identical behaviour to before. Left a TODO at both spots, fixing it properly means threading ctx into cmdShard and calling **Get()** lazily on first use like **ClusterClient.cmdInfo** does, but that touches method signatures across the ring routing path so felt cleaner to leave for a separate PR.

For **slottedKeyedCommands**, moved the **Peek()** outside the loop so it's one lock acquire for the whole batch. Added **cmdSlotWithPos** to take the pre-computed pos so we're not acquiring the lock twice per command. The cluster **getkeysinslot/countkeysinslot** special case is still there.

For **executeMultiShard/createSlotSpecificCommand, firstKeyPos** is now computed once and passed down through **executeMultiSlot**. The independent **Peek()** in **createSlotSpecificCommand** is gone so the inconsistency you flagged can't happen.

Updated the **Peek()** comment to mention the **RLock** and that it can briefly block if **Get** or **Refresh** holds the write lock.

Added **TestCmdFirstKeyPosWithInfo_UsesCommandInfoWhenWarm** in `unit_test.go`, **plain testing.T**, no Redis needed, injected cache fn. Covers cold returns 1, warm FirstKeyPos=0 returns 0, and the full before/after Get() round-trip.

Let me know if any changes or more clarifications needed.

command.go
```diff
@@ -4936,6 +4942,19 @@ func (c *cmdsInfoCache) Refresh() {
...
+// Note: during the very first Get() (initial population) this call will block on
+// the writer lock. After that, concurrent Peek() calls do not block each other.
```
ndyakov, line 5140:
```suggestion
// the writer lock. After that, concurrent Peek() calls do not block each other.
// The returned map and its entries MUST NOT be mutated by the caller.
```

ndyakov (dismissed):
Looks good, will apply the godoc suggestion and proceed.

ndyakov (approved):

PR #3807, opened 2026-05-07
fix: include sharded channels in PubSub.conn() so cluster reconnect routes to the slot owner

### LINKED ISSUE #3806: ClusterClient.SSubscribe silently re-subscribes to a random node after reconnect — PubSub.conn() ignores c.schannels
### Expected Behavior

A `*PubSub` created via `ClusterClient.SSubscribe` should reconnect to the
node owning the shard channel's hash slot after a connection failure, the
same way it does for the initial connection.

### Current Behavior

After any connection loss (server restart, transient network blip,
`CLIENT KILL`, idle timeout), the `PubSub` auto-reconnects to a **random
cluster node** and re-issues `SSUBSCRIBE` there. On any node other than the
slot owner, Redis replies `-MOVED`. The reply is never read (the `resubscribe`
path is write-only), so the failure is **silent**: the `PubSub` looks healthy,
`Ping` succeeds, `Receive`/`Channel()` keep returning, but no messages arrive
because the subscriber isn't on the shard the publishers reach.

On an N-node cluster the chance of landing on the wrong node is `(N-1)/N` per
reconnect. The only recovery is to close and recreate the `PubSub` (or restart
the process).

### Root Cause

[`pubsub.go:90`](https://github.com/redis/go-redis/blob/8cdff5946a35/pubsub.go#L90)
in `PubSub.conn()` builds the `channels` slice that's handed to the
`newConn` callback for node selection — but it only collects `c.channels`
(regular `SUBSCRIBE`), not `c.schannels` (sharded `SSUBSCRIBE`):

```go
channels := slices.Collect(maps.Keys(c.channels))
channels = append(channels, newChannels...)

cn, err := c.newConn(ctx, c.opt.Addr, channels)
```

For a `PubSub` that *only* has sharded subscriptions — the normal case for
`ClusterClient.SSubscribe` — `channels` is empty when reconnecting. The
`ClusterClient`'s `newConn` closure
([`osscluster.go:2152-2178`](https://github.com/redis/go-redis/blob/8cdff5946a35/osscluster.go#L2152-L2178))
then takes the `else` branch and picks a random node:

```go
if len(channels) > 0 {
    slot := hashtag.Slot(channels[0])
    ...
    node, err = c.slotMasterNode(ctx, slot)
    ...
} else {
    node, err = c.nodes.Random()       // ← reconnect lands here
    ...
}
```

`resubscribe()`
([`pubsub.go:127-128`](https://github.com/redis/go-redis/blob/8cdff5946a35/pubsub.go#L127-L128))
then writes `SSUBSCRIBE` over the new conn, but `_subscribe()` is
write-only and never reads the reply. The `-MOVED` from the wrong node is
either dropped, or eventually surfaces as an error from `ReceiveTimeout` —
but `isBadConn()`
([`error.go:204-210`](https://github.com/redis/go-redis/blob/8cdff5946a35/error.go#L204-L210))
returns `false` for a `MOVED` that points at a *different* address, so it
doesn't trigger another reconnect.

The *initial* connection isn't affected, because `PubSub.subscribe()` passes
the new channels through the `newChannels` argument to `conn()`, and they're
appended to the routing list. The bug only bites the **re**connect path,
where `newChannels` is `nil` and `c.schannels` is the only place the shard
channels live.

### Steps to Reproduce

1. Start a Redis cluster (≥ 2 master shards).
2. `pubsub := clusterClient.SSubscribe(ctx, "shard-chan")`.
3. Verify delivery works: `clusterClient.SPublish(ctx, "shard-chan", "x")` →
   `pubsub.ReceiveTimeout(...)` returns the message.
4. Forcibly close the PubSub connection on the slot owner — e.g.
   `CLIENT KILL TYPE pubsub` on that node — without sending `SUNSUBSCRIBE`.
5. Trigger reconnect (`pubsub.Ping(ctx)`, or just wait for the next
   `Receive` / health-check ping).
6. `SPublish` again → `ReceiveTimeout` times out, and `SPUBLISH`'s return
   value is `0` (no subscribers reached on the slot owner).

A regression test in the repo's existing ginkgo cluster harness will be
included in the PR (`osscluster_test.go`); it iterates the
kill→reconnect→publish cycle 8 times so a lucky random-node hit can't mask
the bug.

### Possible Solution

Include `c.schannels` when building the channel list passed to `newConn`:

```go
channels := slices.Collect(maps.Keys(c.channels))
channels = append(channels, slices.Collect(maps.Keys(c.schannels))...)
channels = append(channels, newChannels...)
```

Sharded channels are appended *after* regular channels so that for
`SSubscribe`-only `PubSub`s (the common case) `channels[0]` is a shard
channel and slot routing works, while `PubSub`s using only regular
`SUBSCRIBE`/`PSUBSCRIBE` see no behavior change.

The `channels` argument is unused for routing in `redis.Client.pubSub()`
(single node, fixed `addr`), `Ring.SSubscribe()` (shard chosen *before* the
`PubSub` is created, owned by a single `redis.Client`), and `SentinelClient`
(no `SSubscribe`), so the change only affects `ClusterClient`.

### Context (Environment)

- go-redis: v9.17.1 (also reproduced against v9.19.0 and `master` @
  `8cdff5946a35`)
- Server: Redis Cluster 7.x (sharded pub/sub requires ≥ 7.0)
- Production impact: on a 25-shard cluster, a transient connection blip
  caused message receive rate to drop ~90% (24/25 chance of wrong node) and
  never recover. The PubSub health-check ping kept succeeding against the
  wrong node, so the failure went undetected for ~6 hours until a process
  restart.

### Note on mixed regular + sharded subscriptions

A `PubSub` carrying *both* regular and sharded subscriptions on a cluster
client is already underdetermined (regular channels can be served by any
node; sharded channels must be on the slot owner; a single conn can't
satisfy both for arbitrary channel sets). This fix doesn't change behavior
for that case — `channels[0]` is still a regular channel and routing follows
it. `ClusterClient.Subscribe`/`PSubscribe`/`SSubscribe` each return a fresh
`PubSub`, so the mixed case only arises if the caller mixes
`Subscribe`/`SSubscribe` calls on the same `PubSub` deliberately.

igorkofman:
Fixes #3806.

### Problem

`PubSub.conn()` builds the channel list passed to `newConn` from
`c.channels` only. For a `PubSub` with only sharded subscriptions (the
normal `ClusterClient.SSubscribe` case), that list is empty on reconnect.
`ClusterClient.pubSub`'s `newConn` closure falls back to `nodes.Random()`
when `len(channels) == 0`, so the reconnect lands on a random node. The
`SSUBSCRIBE` written by `resubscribe()` to the wrong node fails silently
(`_subscribe` is write-only and `isBadConn` ignores `MOVED` to a different
address), leaving a `PubSub` that looks healthy — `Ping` succeeds,
`Receive` keeps running — but never receives another message until the
process restarts.

### Fix

Append `c.schannels` to the list, after `c.channels`. The `channels`
argument is only used for slot resolution in the cluster client; the
single-node, ring, and sentinel clients ignore it for routing, so the
change is scoped to cluster behavior. Ordering schannels after channels
keeps behavior identical for `PubSub`s that have any regular
subscriptions.

### Confidence

- Added a regression test in `osscluster_test.go` that `SSubscribe`s,
  kills the PubSub conn on every cluster node (`CLIENT KILL TYPE pubsub`),
  triggers reconnect, and asserts message delivery — repeated 8 times so a
  lucky random-node hit can't mask the bug. Without the fix the chance of
  a false pass on the 3+3 node test cluster is `(1/6)^8 ≈ 6e-7`.
- Reproduced the bug in production on a 25-shard cluster behind a managed
  Redis Cluster; the PubSub silently dropped to ~10% of expected message
  delivery after a transient connection blip and stayed degraded until a
  process restart. Confirmed the fix restores delivery after `CLIENT KILL`.
- `go build`, `go vet`, `gofmt` clean on the patched files.

---

> [!NOTE]
> **Medium Risk**
> Touches cluster PubSub reconnect/routing behavior; a bug here can silently drop messages after reconnects, though the change is small and covered by a targeted regression test.
> 
> **Overview**
> Fixes a Redis Cluster sharded PubSub reconnect bug where `PubSub.conn()` could reconnect to a random node when only `SSUBSCRIBE` channels were in use, causing resubscribe to land on the wrong shard and silently stop message delivery.
> 
> `PubSub.conn()` now includes sharded channels (`c.schannels`) when building the channel list passed to `newConn` so slot-owner resolution works on reconnect. Adds a regression test that repeatedly kills pubsub connections across all cluster nodes and verifies `SPUBLISH` continues to reach subscribers after each reconnect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7271dec6e8a48523d9059fea65693f5c5cc87090. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/osscluster_test.go
+++ b/osscluster_test.go
@@ -668,6 +668,67 @@ var _ = Describe("ClusterClient", func() {
 			}, 30*time.Second).ShouldNot(HaveOccurred())
 		})
 
+		It("should re-subscribe sharded PubSub to the slot owner after reconnect", func() {
+			const channel = "shard-reconnect-chan"
+
+			pubsub := client.SSubscribe(ctx, channel)
+			defer pubsub.Close()
+
+			expectDelivery := func() error {
+				n, err := client.SPublish(ctx, channel, "hello").Result()
+				if err != nil {
+					return err
+				}
+				if n == 0 {
+					// Nobody is subscribed on the slot owner — the
+					// PubSub reconnected to the wrong node.
+					return fmt.Errorf("SPUBLISH reached 0 subscribers")
+				}
+				msg, err := pubsub.ReceiveTimeout(ctx, time.Second)
+				if err != nil {
+					return err
+				}
+				if _, ok := msg.(*redis.Message); !ok {
+					return fmt.Errorf("got %T, wanted *redis.Message", msg)
+				}
+				return nil
+			}
+
+			// Sanity-check the initial subscription before disrupting it.
+			Eventually(expectDelivery, 30*time.Second).ShouldNot(HaveOccurred())
+
+			// Repeatedly tear down the server side of the PubSub connection
+			// and verify the client re-subscribes to the correct shard.
+			// Each cycle has a 5/6 chance of exposing a wrong-node
+			// reconnect on a 3+3 node test cluster, so 8 iterations make a
+			// false pass vanishingly unlikely.
+			for i := 0; i < 8; i++ {
+				// Kill the PubSub connection on every node so we don't have
+				// to hunt for the slot owner. CLIENT KILL TYPE pubsub only
+				// touches connections in subscribe mode, so the regular
+				// cluster connections used by `client` are unaffected.
+				for _, master := range cluster.masters() {
+					_, _ = master.ClientKillByFilter(ctx, "TYPE", "pubsub").Result()
+				}
+				for _, slave := range cluster.slaves() {
+					_, _ = slave.ClientKillByFilter(ctx, "TYPE", "pubsub").Result()
+				}
+
+				// Drive the PubSub to notice the dead socket and reconnect.
+				Eventually(func() error {
+					return pubsub.Ping(ctx)
+				}, 10*time.Second).ShouldNot(HaveOccurred())
+
+				// Without the fix the new conn went to nodes.Random(); the
+				// SSUBSCRIBE in resubscribe() landed on the wrong shard
+				// (and the resulting -MOVED reply is never read, so the
+				// failure is silent). SPUBLISH on the slot owner therefore
+				// reaches no subscribers.
+				Eventually(expectDelivery, 5*time.Second).ShouldNot(HaveOccurred(),
+					"sharded PubSub did not survive reconnect cycle %d", i)
+			}
+		})
+
 		It("should support PubSub.Ping without channels", func() {
 			pubsub := client.Subscribe(ctx)
 			defer pubsub.Close()
--- a/pubsub.go
+++ b/pubsub.go
@@ -88,6 +88,10 @@ func (c *PubSub) conn(ctx context.Context, newChannels []string) (*pool.Conn, er
 	}
 
 	channels := slices.Collect(maps.Keys(c.channels))
+	// Include sharded subscriptions so cluster newConn can resolve the
+	// correct slot owner; ClusterClient.pubSub falls back to a random node
+	// when channels is empty, which silently drops SSUBSCRIBE on reconnect.
+	channels = append(channels, slices.Collect(maps.Keys(c.schannels))...)
 	channels = append(channels, newChannels...)
 
 	cn, err := c.newConn(ctx, c.opt.Addr, channels)


ndyakov:
Hello @igorkofman , there are couple of initiatives around pub sub at the moment. I do think the biggest one right now is https://github.com/redis/go-redis/pull/3717, maybe check it and align the changes to it if needed?

ndyakov:
closing in favor of https://github.com/redis/go-redis/pull/3717

PR #3810, opened 2026-05-14
fix: make wrapped dial tcp error redirectable (#3799)

### LINKED ISSUE #3799: Dial timeout error still is not redirectable
After we fix #3786 

I think that we haven't finished with this bug yet. I've updated lib version, decided to re-check and noticed that dial tcp timeouts still not redirectable. So I think that problem is here
```go
# internal/pool/pool.go
attemptCtx, cancel = context.WithTimeout(ctx, p.cfg.DialTimeout)
```
We use context timeout here. So when dial timeout is over *net.OpError{Op: "dial"} is wrapped inside ContextDeadlineExceeded, so in shouldRetry this condition is true

```go
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
	return false
}
```

I suppose that we should move new condition before this check

```go
var opErr *net.OpError
if errors.As(err, &opErr) && opErr.Op == "dial" {
    return true 
}

if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
	return false
}
```

@ndyakov Hi! I commented in original issue https://github.com/redis/go-redis/issues/3799#issue-4356209088 so we can reopen it or discuss it here

vladisa88:
fixes #3799

Now check if it was an OpError before we check for context errors

---

> [!NOTE]
> **Medium Risk**
> Changes `shouldRetry` behavior for connection failures by retrying `net.OpError` dial errors even when wrapped with `context.DeadlineExceeded`, which could alter retry frequency under timeout conditions.
> 
> **Overview**
> Fixes retry classification so TCP *dial* failures are retried even when they appear wrapped as `context.DeadlineExceeded` (e.g., from dial timeouts), by checking for `net.OpError{Op:"dial"}` **before** context cancellation/deadline handling.
> 
> Adds `TestDialTimeoutShouldRetry` coverage to ensure dial errors (including those wrapped in deadline errors) retry, while non-dial `net.OpError` and plain context errors still do not.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd5266f1d1371788606adcfc76ffbd6bf3aea523. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/error.go
+++ b/error.go
@@ -91,6 +91,14 @@ func shouldRetry(err error, retryTimeout bool) bool {
 		return true
 	}
 
+	// Dial errors mean TCP connection was never established — safe to retry even
+	// when wrapped inside context.DeadlineExceeded (from DialTimeout context).
+	// Must be checked before the context error check below.
+	var opErr *net.OpError
+	if errors.As(err, &opErr) && opErr.Op == "dial" {
+		return true
+	}
+
 	// Check for context errors (works with wrapped errors)
 	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		return false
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
Nice, thank you @vladisa88!

PR #3813, opened 2026-05-15
feat: impl array data structure

cxljs:
Redis has the new Redis Array type.
The PR: https://github.com/redis/redis/pull/15162

---

> [!NOTE]
> **Medium Risk**
> Adds a new Redis data structure command surface plus new command result types (`UintCmd`, `UintSliceCmd`, `AREntrySliceCmd`) that integrate into the core command decoding/dispatch paths, which could affect generic command handling if misparsed.
> 
> **Overview**
> **Adds experimental support for Redis 8.8+ `array` data structure commands.** Introduces a new `ArrayCmdable` API with implementations for commands like `ARSET/ARGET/ARGETRANGE`, `ARINSERT/ARNEXT/ARSEEK`, `ARDELRANGE`, `ARSCAN`, `ARGREP` (with predicates and `WITHVALUES`), `ARRING`, `ARLASTITEMS`, `ARINFO`, and `AROP`.
> 
> **Extends the core command layer** with new reply types `UintCmd`, `UintSliceCmd`, and `AREntrySliceCmd` (and corresponding `CmdType*` cases) and wires `ArrayCmdable` into the main `Cmdable` interface. Adds a large Ginkgo test suite covering array behavior, edge cases, and large `uint64` indexes (skipped pre-Redis 8.8).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83a8b715c5f732e991584b43b41a42ec20faebbd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/array_commands.go
+++ b/array_commands.go
@@ -1,0 +1,333 @@
+package redis
+
+import (
+	"context"
+)
+
+// ArrayCmdable defines the interface for Redis Array data structure commands
+// available in Redis 8.8.0+.
+type ArrayCmdable interface {
+	ARSet(ctx context.Context, key string, index uint64, values ...string) *IntCmd
+	ARGet(ctx context.Context, key string, index uint64) *StringCmd
+	ARGetRange(ctx context.Context, key string, start, end uint64) *SliceCmd
+	ARMGet(ctx context.Context, key string, indexes ...uint64) *SliceCmd
+	ARMSet(ctx context.Context, key string, members ...AREntry) *IntCmd
+	ARInsert(ctx context.Context, key string, values ...string) *UintCmd
+	ARDel(ctx context.Context, key string, indexes ...uint64) *IntCmd
+	ARDelRange(ctx context.Context, key string, ranges ...ARRange) *UintCmd
+	ARLen(ctx context.Context, key string) *UintCmd
+	ARCount(ctx context.Context, key string) *UintCmd
+	ARNext(ctx context.Context, key string) *UintCmd
+	ARSeek(ctx context.Context, key string, index uint64) *IntCmd
+	ARInfo(ctx context.Context, key string) *MapStringInterfaceCmd
+	ARInfoFull(ctx context.Context, key string) *MapStringInterfaceCmd
+	ARScan(ctx context.Context, key string, start, end uint64, args *ARScanArgs) *AREntrySliceCmd
+	AROp(ctx context.Context, key string, start, end uint64, op AROp, matchValues ...string) *Cmd
+	ARGrep(ctx context.Context, key string, start, end string, args *ARGrepArgs) *UintSliceCmd
+	ARGrepWithValues(ctx context.Context, key string, start, end string, args *ARGrepArgs) *AREntrySliceCmd
+	ARRing(ctx context.Context, key string, size uint64, values ...string) *UintCmd
+	ARLastItems(ctx context.Context, key string, count uint64, rev bool) *SliceCmd
+}
+
+// AREntry represents an index-value pair for ARMSET.
+type AREntry struct {
+	Index uint64
+	Value string
+}
+
+// ARRange represents a start-end range for ARDELRANGE.
+type ARRange struct {
+	Start uint64
+	End   uint64
+}
+
+// AROp represents an aggregate operation for AROP.
+type AROp string
+
+const (
+	ArrayOpSum   AROp = "SUM"
+	ArrayOpMin   AROp = "MIN"
+	ArrayOpMax   AROp = "MAX"
+	ArrayOpAnd   AROp = "AND"
+	ArrayOpOr    AROp = "OR"
+	ArrayOpXor   AROp = "XOR"
+	ArrayOpUsed  AROp = "USED"
+	ArrayOpMatch AROp = "MATCH"
+)
+
+// ARScanArgs contains optional arguments for ARSCAN.
+type ARScanArgs struct {
+	Limit uint64
+}
+
+// ARGrepPredicateType defines the type of predicate for ARGREP.
+type ARGrepPredicateType string
+
+const (
+	ARGrepExact ARGrepPredicateType = "EXACT"
+	ARGrepMatch ARGrepPredicateType = "MATCH"
+	ARGrepGlob  ARGrepPredicateType = "GLOB"
+	ARGrepRegex ARGrepPredicateType = "RE"
+)
+
+// ARGrepPredicate represents a search predicate for ARGREP.
+type ARGrepPredicate struct {
+	Type  ARGrepPredicateType
+	Value string
+}
+
+// ARGrepArgs contains optional arguments for ARGREP.
+// Redis ARGREP defaults to OR when multiple predicates are given.
+// Set CombineAnd to true to combine predicates with AND instead.
+type ARGrepArgs struct {
+	Predicates []ARGrepPredicate
+	CombineAnd bool
+	Limit      uint64
+	NoCase     bool
+}
+
+// ARSet sets one or more contiguous values starting at an index in an array.
+// Returns the number of new slots that were set (previously empty).
+func (c cmdable) ARSet(ctx context.Context, key string, index uint64, values ...string) *IntCmd {
+	args := make([]any, 3, 3+len(values))
+	args[0] = "arset"
+	args[1] = key
+	args[2] = index
+	for _, v := range values {
+		args = append(args, v)
+	}
+	cmd := NewIntCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// ARGet gets the value at an index in an array.
+// Returns redis.Nil if the key or index does not exist.
+func (c cmdable) ARGet(ctx context.Context, key string, index uint64) *StringCmd {
+	cmd := NewStringCmd(ctx, "arget", key, index)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// ARGetRange gets values in a range of indexes.
+// Returns values in the range, with nil for unset indexes.
+func (c cmdable) ARGetRange(ctx context.Context, key string, start, end uint64) *SliceCmd {
+	cmd := NewSliceCmd(ctx, "argetrange", key, start, end)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// ARMGet gets values at multiple indexes in an array.
+// Returns values at the specified indexes, with nil for unset indexes.
+func (c cmdable) ARMGet(ctx context.Context, key string, indexes ...uint64) *SliceCmd {
+	args := make([]any, 2+len(indexes))
+	args[0] = "armget"
+	args[1] = key
+	for i, idx := range indexes {
+		args[2+i] = idx
+	}
+	cmd := NewSliceCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// ARMSet sets multiple index-value pairs in an array.
+// Returns the number of new slots that were set (previously empty).
+func (c cmdable) ARMSet(ctx context.Context, key string, members ...AREntry) *IntCmd {
+	args := make([]any, 2, 2+2*len(members))
+	args[0] = "armset"
+	args[1] = key
+	for _, m := range members {
+		args = append(args, m.Index, m.Value)
+	}
+	cmd := NewIntCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// ARInsert inserts one or more values at consecutive indexes.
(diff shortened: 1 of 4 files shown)


ndyakov:
@cxljs thank you for working on this. we did prepare something as well, we will review your pr and help you complete :)

cxljs:
> @cxljs thank you for working on this. we did prepare something as well, we will review your pr and help you complete :)

Thanks @ndyakov ! array is a fascinating new data structure. I took a deep dive into its implementation—the dense/sparse modes for Slice and the encoding optimizations for value pointers are both brilliant designs. I’m also thrilled to contribute go-redis support for it, even if it’s just a simple API wrapper.

array_commands_test.go
```diff
@@ -0,0 +1,1983 @@
...
+
+	"github.com/redis/go-redis/v9"
+)
+
+var _ = Describe("Array Commands", Label("array"), func() {
+	ctx := context.TODO()
+	var client *redis.Client
+
+	BeforeEach(func() {
+		SkipBeforeRedisVersion(8.8, "Redis 8.8.0 introduces support for Array")
```
ndyakov, line 20: 👍

array_commands.go
```diff
@@ -0,0 +1,333 @@
...
+	ARNext(ctx context.Context, key string) *UintCmd
+	ARSeek(ctx context.Context, key string, index uint64) *IntCmd
+	ARInfo(ctx context.Context, key string) *MapStringInterfaceCmd
+	ARInfoFull(ctx context.Context, key string) *MapStringInterfaceCmd
+	ARScan(ctx context.Context, key string, start, end uint64, args *ARScanArgs) *AREntrySliceCmd
+	AROp(ctx context.Context, key string, start, end uint64, op AROp, matchValues ...string) *Cmd
+	ARGrep(ctx context.Context, key string, start, end string, args *ARGrepArgs) *UintSliceCmd
+	ARGrepWithValues(ctx context.Context, key string, start, end string, args *ARGrepArgs) *AREntrySliceCmd
+	ARRing(ctx context.Context, key string, size uint64, values ...string) *UintCmd
+	ARLastItems(ctx context.Context, key string, count uint64, rev bool) *SliceCmd
```
ndyakov, line 40: i took a brief look and initially was not sure about the `Uint`. After some testing with ARSet - it makes sense as you can set up to max uint64. Can we add tests around the limits to make sure we are processing this correctly? I will continue with my review tomorrow.

ndyakov (commented):
left a partial review.

array_commands.go
cxljs: Previous tests didn't cover uint64 scenarios. Your review suggestion is great, thanks! I've just added the relevant test cases.

array_commands.go
```diff
@@ -0,0 +1,333 @@
...
+	ARDel(ctx context.Context, key string, indexes ...uint64) *IntCmd
+	ARDelRange(ctx context.Context, key string, ranges ...ARRange) *UintCmd
+	ARLen(ctx context.Context, key string) *UintCmd
+	ARCount(ctx context.Context, key string) *UintCmd
+	ARNext(ctx context.Context, key string) *UintCmd
+	ARSeek(ctx context.Context, key string, index uint64) *IntCmd
+	ARInfo(ctx context.Context, key string) *MapStringInterfaceCmd
+	ARInfoFull(ctx context.Context, key string) *MapStringInterfaceCmd
+	ARScan(ctx context.Context, key string, start, end uint64, args *ARScanArgs) *AREntrySliceCmd
+	AROp(ctx context.Context, key string, start, end uint64, op AROp, matchValues ...string) *Cmd
```
ndyakov: Hey @cxljs , what would you think about adding a higher level commands for the possible `AROps`. I do think they will be more useful for end users instead of working with the `*Cmd` return
```
// Numeric aggregates returned as strings (preserves arbitrary precision).
AROpSum(ctx, key string, start, end uint64) *StringCmd
AROpMin(ctx, key string, start, end uint64) *StringCmd
AROpMax(ctx, key string, start, end uint64) *StringCmd

// Bitwise aggregates over fixed-width integer elements.
AROpAnd(ctx, key string, start, end uint64) *UintCmd
AROpOr (ctx, key string, start, end uint64) *UintCmd
AROpXor(ctx, key string, start, end uint64) *UintCmd

// Counts (always non-negative).
AROpUsed (ctx, key string, start, end uint64) *UintCmd
AROpMatch(ctx, key string, start, end uint64, value string) *UintCmd
```

array_commands.go
```diff
@@ -0,0 +1,333 @@
+package redis
+
+import (
+	"context"
+)
+
+// ArrayCmdable defines the interface for Redis Array data structure commands
+// available in Redis 8.8.0+.
```
ndyakov, line 10: maybe it makes sense to note why some commands are `Uint` and others are `Int`.  By the looks of it - if it is expected to be an `index` it is `Uint` if it is a count - `int`. It makes sense, let's just document it. At the same time, the `ARDelRange` can overflow and thus uses `Uint`... Maybe it will be better to use `Uint` at all places, what do you think @cxljs ?

array_commands.go
```diff
@@ -0,0 +1,333 @@
...
+//	client.AROp(ctx, "key", 0, 100, ArrayOpMatch, "hello")
+func (c cmdable) AROp(ctx context.Context, key string, start, end uint64, op AROp, matchValues ...string) *Cmd {
+	args := []any{"arop", key, start, end, string(op)}
+	if op == ArrayOpMatch && len(matchValues) > 0 {
+		args = append(args, matchValues[0])
+	}
+	cmd := NewCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
+}
```
ndyakov: Once we have `ArOp*` maybe remove this or comment that it is a lower level approach to what `ArOp*` provide.

array_commands.go
```diff
@@ -0,0 +1,333 @@
...
+
+// ARScan iterates existing elements in a range, returning index-value pairs.
+func (c cmdable) ARScan(ctx context.Context, key string, start, end uint64, args *ARScanArgs) *AREntrySliceCmd {
+	a := []any{"arscan", key, start, end}
+	if args != nil && args.Limit > 0 {
+		a = append(a, "limit", args.Limit)
+	}
```
ndyakov, line 240: This is totally correct, but for consistency, let's use the approach from above with the `make([]any..` for creating the args:
```
capacity := 4
if args != nil && args.Limit > 0 {
    capacity += 2
}
a := make([]any, 4, capacity)
a[0] = "arscan"
a[1] = key
a[2] = start
a[3] = end
if args != nil && args.Limit > 0 {
    a = append(a, "limit", args.Limit)
}
```


**Or even without trying to calculate the capacity previously and being safe by allocating 6 slots.**

array_commands.go
```diff
@@ -0,0 +1,333 @@
...
+	cmd := NewUintSliceCmd(ctx, a...)
+	_ = c(ctx, cmd)
+	return cmd
+}
+
+// ARGrepWithValues searches array elements in a range using textual predicates.
+// Returns matching indexes and their values as index-value pairs.
+func (c cmdable) ARGrepWithValues(ctx context.Context, key string, start, end string, args *ARGrepArgs) *AREntrySliceCmd {
+	a := []any{"argrep", key, start, end}
+	a = appendGrepArgs(a, args)
```
ndyakov: same, if we can calculate the capacity and use `make` for creation of the slice. I do think this one (you have right now) is cleaner, but would prefer consistency here.

array_commands_test.go
```diff
@@ -0,0 +1,2166 @@
...
+				redis.AREntry{Index: 1, Value: "b"},
+				redis.AREntry{Index: 100, Value: "c"},
+			).Err()).NotTo(HaveOccurred())
+
+			info, err := client.ARInfo(ctx, "myarray").Result()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(info["count"]).To(Equal(int64(3)))
+			Expect(info["len"]).To(Equal(int64(101)))
+		})
+	})
```
ndyakov, line 1133: can we check ARInfoFull as well?

array_commands.go
ndyakov: something like

```

// grepArgsLen returns the number of args appendGrepArgs will append.
func grepArgsLen(args *ARGrepArgs) int {
    if args == nil {
        return 0
    }
    n := 2 * len(args.Predicates)
    if args.CombineAnd {
        n++
    }
    if args.Limit > 0 {
        n += 2
    }
    if args.NoCase {
        n++
    }
    return n
}
```

ndyakov (commented):
looks good, left some suggestions. @cxljs let me know what you think I am open for discussion if you have other suggestions.

array_commands.go
cxljs: Great suggestion! Returning *Cmd is indeed quite user-unfriendly, I've already updated the code.

Just a small detail: it might be better for the and/or/xor ops to return int64. Redis itself filters out values exceeding int64 during implementation and eventually returns an int64.

https://github.com/redis/redis/blob/ba1a4b2c8f37beec5a9be040c37c755207f0cc8d/src/t_array.c#L1350

Examples in redis-cli：
```
127.0.0.1:6379> arset key 1 -5
(integer) 1
127.0.0.1:6379> arop key 0 10 and
(integer) -5
```

array_commands.go
cxljs: I think uint64 could also be a good fit for count.  @ndyakov  What do you think? Should we change count to uint64, or just add some notes?

And there are two points where I think we should not switch to uint64:

For arop and/or/xor, I think keeping them as int64 is better.

arseek currently uses int but it only returns 0 or 1 —should we change it to *BoolCmd instead?

array_commands.go
cxljs: Yes, I removed it.

array_commands.go
cxljs: You're right, consistency is better. I've updated it.

array_commands_test.go
cxljs: I've added the test cases.

array_commands.go
ndyakov: For `ARSeek` it is noted that the return type is integer, so let's keep it int for now. It is not the most ergonomic, but will be future proofing if new return values are added.

array_commands.go
ndyakov: Your observation is correct. Also the match and used are signed long long, so `*IntCmd` should work there as well.

ndyakov (dismissed):
Thank you for the swift replies and work on this @cxljs. Looks good, let me know if you want me to wait a bit longer for any docs or polishing before merging or if it is good to go from your perspective.

command.go
```diff
@@ -1044,6 +1047,52 @@ func (cmd *IntCmd) Clone() Cmder {
...
 
+type UintCmd struct {
+	baseCmd
+
+	val uint64
+}
+
+var _ Cmder = (*UintCmd)(nil)
+
+func NewUintCmd(ctx context.Context, args ...interface{}) *UintCmd {
```
ndyakov: @cxljs feel free to make `args` `...any` since this is a new command.

command.go
```diff
@@ -1175,6 +1224,66 @@ func (cmd *IntSliceCmd) Clone() Cmder {
...
 
+type UintSliceCmd struct {
+	baseCmd
+
+	val []uint64
+}
+
+var _ Cmder = (*UintSliceCmd)(nil)
+
+func NewUintSliceCmd(ctx context.Context, args ...interface{}) *UintSliceCmd {
```
ndyakov: same here, `args` can be `...any`

command.go
```diff
@@ -8557,3 +8687,80 @@ func ExtractCommandValue(cmd interface{}) (interface{}, error) {
...
+
+// AREntrySliceCmd is a command that returns index-value pairs from ARSCAN or ARGREP.
+type AREntrySliceCmd struct {
+	baseCmd
+	val []AREntry
+}
+
+var _ Cmder = (*AREntrySliceCmd)(nil)
+
+func NewAREntrySliceCmd(ctx context.Context, args ...any) *AREntrySliceCmd {
```
ndyakov, line 8832: 👍

cxljs:
> Thank you for the swift replies and work on this @cxljs. Looks good, let me know if you want me to wait a bit longer for any docs or polishing before merging or if it is good to go from your perspective.

Added some docs explaining the supported array index range. I think this PR is ready to merge.

@ndyakov Thank you so much for your review and all the great suggestions, I really appreciate it!

ndyakov (approved):
Looks good, thank you!

PR #3816, opened 2026-05-18
feat(command): add support for INCREX

ndyakov:
add support for https://github.com/redis/redis/pull/15045

---

> [!NOTE]
> **Medium Risk**
> Adds new Redis 8.8 `INCREX` command implementations with custom reply parsing and expands the command-type dispatch logic; mistakes here could cause runtime decode errors. CI/docker test images are also changed to a custom 8.8 container tag, which may affect test environment stability.
> 
> **Overview**
> Adds client support for Redis 8.8 `INCREX` via new `IncrEXInt`/`IncrEXFloat` APIs (argument structs, expiration/`ENX` handling, and result types exposing *value* and *applied increment*), plus corresponding command implementations and reply decoding in `command.go`.
> 
> Expands test coverage with new integration tests for bounds/saturation/TTL semantics and unit tests asserting command argument construction, and updates CI/compose defaults to use a custom Redis 8.8 `client-libs-test` image tag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d401b83c7147a3332be0b43a36310a32fe3a967. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/string_commands.go
+++ b/string_commands.go
@@ -227,9 +227,14 @@ type IncrEXIntArgs struct {
 }
 
 // IncrEXFloatArgs are the arguments to IncrEXFloat (the BYFLOAT variant of
-// INCREX). BYFLOAT is implied — the By field is always sent, even when zero.
+// INCREX).
+//
+// If By is zero and HasBy is false, BYFLOAT is omitted and the server
+// increments by 1. HasLBound/HasUBound gate the optional LBOUND/UBOUND
+// clauses so that 0 is a valid bound.
 type IncrEXFloatArgs struct {
-	By float64
+	By    float64
+	HasBy bool
 
 	LBound, UBound       float64
 	HasLBound, HasUBound bool
@@ -245,8 +250,8 @@ type IncrEXFloatArgs struct {
 	ENX bool
 }
 
-// IncrEXInt Redis `INCREX key [BYINT amount] [SATURATE] [LBOUND value]
-// [UBOUND value] [EX seconds | PX ms | EXAT ts | PXAT ts | PERSIST] [ENX]`
+// IncrEXInt Redis `INCREX key [BYINT amount] [LBOUND value] [UBOUND value]
+// [SATURATE] [EX seconds | PX ms | EXAT ts | PXAT ts | PERSIST] [ENX]`
 // command.
 //
 // Atomically increments the integer value stored at key, optionally
@@ -263,14 +268,14 @@ func (c cmdable) IncrEXInt(ctx context.Context, key string, a IncrEXIntArgs) *In
 	if a.HasBy {
 		args = append(args, "byint", a.By)
 	}
-	if a.Saturate {
-		args = append(args, "saturate")
-	}
 	if a.HasLBound {
 		args = append(args, "lbound", a.LBound)
 	}
 	if a.HasUBound {
 		args = append(args, "ubound", a.UBound)
+	}
+	if a.Saturate {
+		args = append(args, "saturate")
 	}
 	args = appendIncrEXTail(args, a.Expiration, a.ENX)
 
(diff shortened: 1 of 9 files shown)


string_commands.go
```diff
@@ -197,6 +200,125 @@ func (c cmdable) IncrByFloat(ctx context.Context, key string, value float64) *Fl
...
+//
+// Available since Redis 8.8.
+// For more information, see https://redis.io/commands/increx
+func (c cmdable) IncrEXInt(ctx context.Context, key string, a IncrEXIntArgs) *IncrEXIntCmd {
+	args := make([]interface{}, 0, 14)
+	args = append(args, "increx", key)
+	if a.HasBy {
+		args = append(args, "byint", a.By)
+	}
+	if a.Saturate {
```
ofekshenawa: The design doc specifies the following argument order for the command::
```
INCREX key [BYFLOAT | BYINT increment]
           [LBOUND lower-bound] [UBOUND upper-bound]
           [SATURATE]
           [EX seconds | PX milliseconds | EXAT unix-time-seconds | PXAT unix-time-milliseconds | PERSIST]
           [ENX]
```
However, the current implementation parses SATURATE before LBOUND and UBOUND. Could you please verify that this deviation will not affect the query output?

string_commands.go
```diff
@@ -197,6 +200,125 @@ func (c cmdable) IncrByFloat(ctx context.Context, key string, value float64) *Fl
...
+
+// IncrEXFloat Redis `INCREX key BYFLOAT amount [SATURATE] [LBOUND value]
+// [UBOUND value] [EX seconds | PX ms | EXAT ts | PXAT ts | PERSIST] [ENX]`
+// command.
+//
+// Available since Redis 8.8.
+// For more information, see https://redis.io/commands/increx
+func (c cmdable) IncrEXFloat(ctx context.Context, key string, a IncrEXFloatArgs) *IncrEXFloatCmd {
+	args := make([]interface{}, 0, 14)
+	args = append(args, "increx", key, "byfloat", a.By)
```
ofekshenawa: We always serializes byfloat <By>, meaning IncrEXFloatArgs{} sends byfloat 0.
Could we either add HasBy to IncrEXFloatArgs like in IncrEXInt, or document that By is required for IncrEXFloat?

ofekshenawa (commented):
Overall looks good.
I left some questions.
Also Maybe we can add the following scenarios to the test suite. These aren't mandatory and are definitely not blockers:
- Currently, we only cover the explicit UBOUND/LBOUND paths. It would be good to also test the implicit clamping behavior (using LLONG_MAX/LLONG_MIN for BYINT and ±LDBL_MAX for BYFLOAT).
- We should add a case that verifies the TTL is applied correctly, even when the operation is rejected by the default policy.
- It would be useful to include an integration test that triggers a real server error such as running INCREX on a non-numeric key to ensure the error message is forwarded correctly.

string_commands.go
```diff
@@ -197,6 +200,133 @@ func (c cmdable) IncrByFloat(ctx context.Context, key string, value float64) *Fl
...
+		args = append(args, "ubound", a.UBound)
+	}
+	if a.Saturate {
+		args = append(args, "saturate")
+	}
+	args = appendIncrEXTail(args, a.Expiration, a.ENX)
+
+	cmd := NewIncrEXFloatCmd(ctx, args...)
+	_ = c(ctx, cmd)
+	return cmd
```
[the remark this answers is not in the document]
ndyakov: @ofekshenawa this is actually correct. i think we will have to always send byfloat even if it is 0

ofekshenawa (approved):
LGTM!

PR #3843, opened 2026-06-03
fix: ignore unknown fields in CLUSTER SHARDS response

madolson:
## Summary

The `CLUSTER SHARDS` response parser errors on any unrecognized field, breaking clients whenever the server introduces new node or shard attributes.

The [`nodes` field](https://redis.io/docs/latest/commands/cluster-shards/) contains a list of all nodes within the shard. Each individual node is a map of attributes that describe the node. Some attributes are optional and more attributes may be added in the future. The current list of attributes:

> id, endpoint, ip, hostname, port, tls-port, role, replication-offset, health

The parser should be forward-compatible. Unknown fields are now discarded via `rd.DiscardNext()`, the same pattern used elsewhere in this codebase (e.g., `FunctionListCmd`, `ModuleLoadexConfig` parsing).

Changes:
- Replace `fmt.Errorf` on unknown node-level fields with `rd.DiscardNext()`
- Replace `fmt.Errorf` on unknown shard-level fields with `rd.DiscardNext()`
- Add unit tests verifying unknown fields (string, integer, array values) are silently skipped

## Test plan

- [x] Unit tests cover unknown fields at both shard and node levels
- [x] Tests exercise string, integer, and array-typed unknown values
- [x] CI passes (no Go installed locally to verify)

---

> [!NOTE]
> **Medium Risk**
> Changes cluster topology parsing used for routing; incorrect discard logic could break cluster clients, but behavior aligns with existing DiscardNext patterns and is covered by new unit tests.
> 
> **Overview**
> The **`CLUSTER SHARDS`** reply parser no longer fails when Redis returns **unknown shard- or node-level attributes**. Unrecognized keys are **skipped** with `rd.DiscardNext()` instead of returning `fmt.Errorf`, matching forward-compatibility used elsewhere (e.g. `XInfo*` parsers).
> 
> **Tests** assert that known `slots` / `nodes` data still parses when extra fields are present (string, integer, and array-valued unknown node fields).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f4fdaf8d5a3d2986bb6222b024d8b85b177732e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -7125,7 +7125,9 @@ func (cmd *ClusterShardsCmd) readReply(rd *proto.Reader) error {
 						case "health":
 							cmd.val[i].Nodes[k].Health, err = rd.ReadString()
 						default:
-							return fmt.Errorf("redis: unexpected key %q in CLUSTER SHARDS node reply", nodeKey)
+							if err = rd.DiscardNext(); err != nil {
+								return err
+							}
 						}
 
 						if err != nil {
@@ -7134,7 +7136,9 @@ func (cmd *ClusterShardsCmd) readReply(rd *proto.Reader) error {
 					}
 				}
 			default:
-				return fmt.Errorf("redis: unexpected key %q in CLUSTER SHARDS reply", key)
+				if err = rd.DiscardNext(); err != nil {
+					return err
+				}
 			}
 		}
 	}
--- a/command_cluster_shards_test.go
+++ b/command_cluster_shards_test.go
@@ -1,0 +1,103 @@
+package redis
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/redis/go-redis/v9/internal/proto"
+)
+
+func TestClusterShardsCmdIgnoresUnknownFields(t *testing.T) {
+	// Simulate a CLUSTER SHARDS response containing unknown fields at both the
+	// shard level and node level. The 'nodes' field contains a list of all nodes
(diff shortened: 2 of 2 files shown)


ndyakov:
Hello @madolson and thank you for this contribution. It makes sense to be forward compatible and ignore the unknown fields. Were there any specific ones that triggered this error for you? If so, does it make more sense to include them in the struct now? 

Also, since we had a similar case for `XInfoConsumersCmd`  - https://github.com/redis/go-redis/blob/65d6abde0d812234d26763930b5e84ff05963c96/command.go#L3186 I was wondering does it make sense to have all the unknown fields in a map for such commands? What do you think?

madolson:
@ndyakov We hit while upgrade to Valkey 9.1. I understand that isn't Redis, but the documentation looks like they should be compatible for this dimensions, so better to make sure it doesn't break in the future!

> I was wondering does it make sense to have all the unknown fields in a map for such commands? What do you think?

That makes sense to me! I'm just trying to scope down the change, but happy to make more.

ndyakov (approved):
This looks good as it is.
@madolson no need to extend it at this time, there are other similar commands that try to parse the whole response and fail on unknown fields, which is not ideal. I am willing to include this change and take care of the rest when I decide on a proper approach. Thank you!

PR #3846, opened 2026-06-09
fix(autopipeline): return undersized slice to pool in getQueueSlice

cxljs:
When the pooled slice capacity is too small, getQueueSlice allocates a new slice but discards the old one without returning it to the pool.

---

> [!NOTE]
> **Low Risk**
> Small pooling and control-flow cleanup in autopipeline batching with no auth or API surface changes.
> 
> **Overview**
> Fixes a **sync.Pool leak** in autopipeline queue batching: when `getQueueSlice` needs a larger backing array than the pooled slice provides, it now **`Put`s the undersized slice back** before allocating a new one instead of dropping it on the floor.
> 
> Also **removes a post-semaphore empty-batch guard** in `flushBatchSlice` that called `sem.Release()` and returned when `queuedCmds` was empty—redundant with the earlier queue-empty check and risky if that path could run without a matching acquire.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0165c11d56b65da6efcf0c07417fe2f214b7e27a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/autopipeline.go
+++ b/autopipeline.go
@@ -476,11 +476,6 @@ func (ap *AutoPipeliner) flushBatchSlice() {
 		}
 	}
 
-	if len(queuedCmds) == 0 {
-		ap.sem.Release()
-		return
-	}
-
 	// Fast path for single command
 	if len(queuedCmds) == 1 {
 		qc := queuedCmds[0]
--- a/autopipeline.go
+++ b/autopipeline.go
@@ -138,6 +138,7 @@ func getQueueSlice(capacity int) []*queuedCmd {
 	slice = slice[:0]
 	// If the capacity is too small, allocate a new one
 	if cap(slice) < capacity {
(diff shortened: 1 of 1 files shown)


ndyakov:
Quite interesting, the linter is complaining but for other things in the code.

autopipeline.go
```diff
@@ -475,11 +476,6 @@ func (ap *AutoPipeliner) flushBatchSlice() {
 		}
 	}
 
-	if len(queuedCmds) == 0 {
-		ap.sem.Release()
-		return
-	}
```
ndyakov, line 481: you are correct this is dead code, it made sense on some iteration of this method, but now it is not needed.

ndyakov (approved):
thank you!

PR #3848, opened 2026-06-11
feat(streams): support explicit `LIMIT 0` in XTRIM/XADD trimming via `XTrimLimitDisabled` sentinel

TheRealMal:
## Motivation

Redis distinguishes between *omitting* the `LIMIT` clause and passing `LIMIT 0` when trimming a stream with the approximate (`~`) strategy:

- **No `LIMIT` clause** — Redis applies an implicit effort cap of `100 * stream-node-max-entries` examined entries per call.
- **`LIMIT 0`** — the effort cap is disabled entirely; Redis trims as many macro nodes as the threshold requires in a single call.

(See the [`XADD`](https://redis.io/docs/latest/commands/xadd/) / [`XTRIM`](https://redis.io/docs/latest/commands/xtrim/) docs: *"Specifying the value 0 as count disables the limiting mechanism entirely."*)

Today the library cannot express the second form. `xTrim`, `xTrimMode`, and `XAdd` all build the clause with:

```go
if limit > 0 {
    args = append(args, "limit", limit)
}
```

so `0` is silently swallowed and there is no way to send `XTRIM key MINID ~ <id> LIMIT 0` through the typed API. This matters in practice for periodic cleaners on busy streams: with the implicit default cap, a single `XTRIM ... ~` call may not reclaim a large backlog, and the only workaround is dropping down to raw `Do()`.

## Changes

Introduces a documented negative-value sentinel, following the existing `KeepTTL = -1` precedent in `commands.go` (the same "zero value already has a meaning, a third state is needed" problem for `SET`):

```go
const XTrimLimitDisabled = -1
```

The `limit` parameter of `XTrimMaxLenApprox`, `XTrimMinIDApprox`, `XTrimMaxLenApproxMode`, `XTrimMinIDApproxMode`, and the `XAddArgs.Limit` field now have three documented states:

| Value | Emitted clause | Notes |
|---|---|---|
| `0` | *(none)* | unchanged historical behavior — Redis applies its implicit default cap |
| `> 0` | `LIMIT <n>` | unchanged |
| `< 0` (`XTrimLimitDisabled`) | `LIMIT 0` | **new** — disables the effort cap |

Implementation is a single shared helper, `appendXTrimLimit`, used by `XAdd`, `xTrim`, and `xTrimMode`, replacing the three duplicated `if limit > 0` blocks.

Usage:

```go
rdb.XTrimMinIDApprox(ctx, "stream", "1716000000000-0", redis.XTrimLimitDisabled)
// XTRIM stream MINID ~ 1716000000000-0 LIMIT 0
```

## Backward compatibility

- **No public API changes**: all exported signatures and struct fields are unchanged.
- **No behavior change for existing callers**: `limit = 0` still omits the clause; positive values are emitted as before. Only previously-meaningless negative values gain a meaning.
- Exact-trim commands (`XTrimMaxLen`, `XTrimMinID`, and their `Mode` variants) take no limit parameter and still can never emit `LIMIT` — Redis rejects `LIMIT` without `~` (`ERR syntax error, LIMIT cannot be used without the special ~ option`). For `XAddArgs` the library keeps its existing stance of not validating the `Approx`/`Limit` combination client-side (matching how the conflicting `MaxLen`/`MinID` pair is handled); this is now stated explicitly in the `Limit` field docs instead of being implicit.

## Testing

Added server-less unit tests asserting the constructed command args via the existing `captureCmdable` pattern (`stream_commands_unit_test.go`):

- `TestXTrim_LimitArgs` — 13 cases covering omitted / positive / disabled limit across `XTrimMaxLenApprox`, `XTrimMinIDApprox`, their `Mode` variants, and pinning that exact-trim commands never emit `LIMIT`.
- `TestXAdd_TrimLimitArgs` — 5 cases covering the `XAddArgs.Limit` path, including clause ordering relative to the entry ID.

`go build ./...`, `go vet ./...`, and `gofmt` are clean; the pre-existing unit test suite passes.

---

> [!NOTE]
> **Low Risk**
> Additive sentinel and command-building change only; existing `limit` 0 and positive values behave the same, with coverage for emitted Redis args.
> 
> **Overview**
> Adds **`XTrimLimitDisabled` (-1)** so approximate stream trimming (`XADD` / `XTRIM` with `~`) can send an explicit **`LIMIT 0`** to Redis, disabling the default trimming effort cap—something the client could not express before because only `limit > 0` ever emitted a `LIMIT` clause.
> 
> A shared **`appendXTrimLimit`** helper centralizes the three behaviors: omit `LIMIT` when `limit == 0` (unchanged), `LIMIT n` when positive, and **`LIMIT 0`** when negative. **`XAdd`**, **`xTrim`**, and **`xTrimMode`** use it instead of duplicated checks. **`XAddArgs.Limit`** and the `*Approx*` trim APIs are documented accordingly.
> 
> New **`stream_commands_unit_test.go`** tests assert built command args via **`captureCmdable`** for omitted, positive, and disabled limits across trim variants and **`XAdd`**, including that exact (`=`) trims never emit `LIMIT`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87acd5d8dbf076683b593d2d76e14b5c91527b97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/stream_commands.go
+++ b/stream_commands.go
@@ -9,6 +9,33 @@ import (
 	"github.com/redis/go-redis/v9/internal/otel"
 )
 
+// XTrimLimitDisabled is a sentinel value for the LIMIT argument of stream
+// trimming (XAddArgs.Limit and the XTrim*Approx* commands). Passing it emits
+// an explicit "LIMIT 0", which tells Redis to disable the trimming effort cap
+// entirely. This differs from passing 0, which keeps the historical behavior:
+// no LIMIT clause is sent and Redis applies its implicit default
+// (100 * stream-node-max-entries examined entries).
+//
+// LIMIT is only valid together with the "~" (approximate) trimming flag;
+// Redis rejects LIMIT used with exact ("=") trimming.
+const XTrimLimitDisabled = -1
+
+// appendXTrimLimit appends the LIMIT clause used by XADD and XTRIM trimming:
+//   - limit > 0 emits "LIMIT <limit>";
+//   - limit < 0 (see XTrimLimitDisabled) emits "LIMIT 0", disabling the
+//     trimming effort cap;
+//   - limit == 0 omits the clause, so Redis applies its implicit default.
+func appendXTrimLimit(args []interface{}, limit int64) []interface{} {
+	switch {
+	case limit > 0:
+		return append(args, "limit", limit)
+	case limit < 0:
+		return append(args, "limit", int64(0))
+	default:
+		return args
+	}
+}
+
 type StreamCmdable interface {
 	XAdd(ctx context.Context, a *XAddArgs) *StringCmd
 	XAckDel(ctx context.Context, stream string, group string, mode string, ids ...string) *SliceCmd
@@ -73,7 +100,14 @@ type XAddArgs struct {
 	MaxLen     int64 // MAXLEN N
 	MinID      string
 	// Approx causes MaxLen and MinID to use "~" matcher (instead of "=").
-	Approx         bool
+	Approx bool
+	// Limit caps the trimming effort:
+	//   - 0 omits the LIMIT clause (Redis applies its implicit default);
+	//   - a positive value emits "LIMIT <n>";
+	//   - a negative value (see XTrimLimitDisabled) emits "LIMIT 0",
+	//     disabling the effort cap entirely.
+	// LIMIT requires Approx to be true; Redis rejects LIMIT together with
+	// exact ("=") trimming.
 	Limit          int64
 	Mode           string
 	ID             string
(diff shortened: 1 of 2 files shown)


ndyakov (commented):
Hello @TheRealMal and thank you for this. I like the approach and it is somewhat consistent with the options where we use `-1` to disable a given option. Let me rerun the failed job (it was github runner issue) and I will review the code.

ndyakov (approved):
Looks good to me, thank you @TheRealMal

PR #3851, opened 2026-06-14
feat(csc): client-side caching SharedTracking standalone connection

ofekshenawa:
> [!NOTE]
> **High Risk**
> Large CSC correctness refactor (staleness, conn lifecycle, push draining) touches the command hot path and connection pool; regressions could serve stale data or corrupt replies on partial drain reads.
> 
> **Overview**
> Introduces **`CSCStrategySharedTracking`** (default via `Options.ClientSideCacheStrategy`) and wires client-side caching around a **shared cache + per-connection `CLIENT TRACKING`**, replacing the old **synchronous idle-pool invalidation drain on cache hits** with a **background drainer** that periodically calls **`DrainIdleConns`** on idle connections.
> 
> **Freshness and lifecycle:** Cached entries are **attributed to the fetching connection** (`ConnOwnedCache` / `FulfillOwned` / `EvictByConn`), with **pool OnRemove hooks**, **per-conn close hooks** (`SetOnCscClose`), and a **recently-removed ring** so fulfills don’t leave orphans after conn retirement (`ConnMaxLifetime`, handoff). CSC is **disabled** without `ConnOwnedCache`, on **non-drainable pools** (e.g. `Client.Conn` sticky pool), and clones share **`cscActive`** so hits stop when the owner’s drainer stops; **`runtime.AddCleanup`** stops the drainer if the client is dropped without `Close`.
> 
> **Cache and protocol:** `NewLocalCache` is **sharded** with approximate LRU, optional **`MaxStaleness`** / **`DrainInterval`**, default caps when unbounded, and **invalidation/flush now drops in-flight placeholders**. Push peeking is **more robust** (larger window, split headers, `ErrPushNotificationNameTooLong`); RESP3 **`ReadBufferSize`** is clamped; **`SORT_RO` with BY/GET** is excluded from caching.
> 
> **Observability:** Process-wide **`CommandStats`** / **`CacheAdmissionRejects`** and per-client **`CSCStats()`**, plus **`docs/csc-strategy-guide.md`**. Unknown strategy values clamp to SharedTracking (extension point only—no Broadcast/PerConnection implementation in this diff).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5a0b678512bcaabd575b8e29fc3db401675b691. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


csc_integration.go
```diff
@@ -78,62 +105,119 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
+var cscDrainHandles sync.Map
+
+// cscDrainInterval returns ClientSideCacheConfig.DrainInterval, or the default
+// (cscDrainSkipWindow, 5ms) when unset.
+func (c *baseClient) cscDrainInterval() time.Duration {
+	if cfg := c.opt.ClientSideCacheConfig; cfg != nil && cfg.DrainInterval > 0 {
+		return cfg.DrainInterval
+	}
+	return cscDrainSkipWindow
+}
```
ndyakov, line 353: i think we should have a min interval here. Anything sub-ms may not work properly as of today - https://github.com/golang/go/issues/53824 anyway

redis.go
```diff
@@ -1677,15 +1751,96 @@ func (c *baseClient) peekAndProcessPushNotifications(ctx context.Context, cn *po
...
+		if err != nil {
+			return err
+		}
+		if notification, ok := reply.([]interface{}); ok && len(notification) > 0 {
+			if name, ok := notification[0].(string); ok {
+				if handler := c.pushProcessor.GetHandler(name); handler != nil {
+					if herr := handler.HandlePushNotification(context.Background(), handlerCtx, notification); herr != nil {
+						internal.Logger.Printf(context.Background(), "csc: drain: push handler %q failed: %v", name, herr)
+					}
+				}
```
ndyakov: this will swallow pubsub messages, they are push notifications in RESP3 and they do not have a handler

redis.go
```diff
@@ -1677,15 +1751,96 @@ func (c *baseClient) peekAndProcessPushNotifications(ctx context.Context, cn *po
...
+			return err // surfaced for the caller to classify (benign timeout vs fatal)
+		}
+		if typ != proto.RespPush {
+			// Not a push frame — do not consume a command reply.
+			return nil
+		}
+		reply, err := rd.ReadReply()
+		if err != nil {
+			return err
+		}
```
ndyakov: look at https://github.com/redis/go-redis/pull/3842 and maybe peek the notification name here to validate if we can handle it.

redis.go
```diff
@@ -1677,15 +1751,96 @@ func (c *baseClient) peekAndProcessPushNotifications(ctx context.Context, cn *po
...
+	if !cn.HasBufferedData() && !cn.MaybeHasData() {
+		cn.SetLastPushDrainAtNs(pool.GetCachedTimeNs())
+		return nil
+	}
+
+	err := cn.WithReaderHardDeadline(cscDrainHardReadCap, func(rd *proto.Reader) error {
+		// Built-in processor: use our loop that SURFACES read errors (it swallows them).
+		if _, builtin := c.pushProcessor.(*push.Processor); builtin {
+			return c.drainPendingPushNotifications(cn, rd)
+		}
```
ndyakov: why we need to do the type check here? if the push processor is `*push.VoidProcessor` maybe the draining make sense as well? but if it is, then does the csc make sense to be enabled at all? I wouldn't think so?

redis.go
```diff
@@ -1677,15 +1751,96 @@ func (c *baseClient) peekAndProcessPushNotifications(ctx context.Context, cn *po
...
+	}
+	cn.SetLastPushDrainAtNs(pool.GetCachedTimeNs())
+	return nil
+}
+
+// drainPendingPushNotifications dispatches the push frames currently buffered on
+// rd and STOPS when the buffer empties (it never blocks for more). Returns the
+// first read/peek error for the caller to classify; handler errors are logged and
+// draining continues. (Built-in ProcessPendingNotifications swallows read errors.)
+func (c *baseClient) drainPendingPushNotifications(cn *pool.Conn, rd *proto.Reader) error {
```
ndyakov: maybe we can return the errors from the built-in processor? I think the main reason to not return them was that since it is in the path of normal response reading, the error will be handled by the client later on.

redis.go
```diff
@@ -1677,15 +1751,96 @@ func (c *baseClient) peekAndProcessPushNotifications(ctx context.Context, cn *po
...
+	for {
+		typ, err := rd.PeekReplyType()
+		if err != nil {
+			return err // surfaced for the caller to classify (benign timeout vs fatal)
+		}
+		if typ != proto.RespPush {
+			// Not a push frame — do not consume a command reply.
+			return nil
+		}
+		reply, err := rd.ReadReply()
```
ndyakov: if the deadline is triggered here we may end up with fragmented data in the socket (and we may discard a notification that needs to be processed).

internal/pool/pool.go
```diff
@@ -1394,6 +1394,118 @@ func (p *ConnPool) Filter(fn func(*Conn) bool) error {
...
+			// unusable, or transiently outside idleConns mid-putConn). Covered by
+			// the command-path drain and/or the next round.
+			st.visited[id] = struct{}{}
+			handled++
+			continue
+		}
+
+		if err := fn(cn); err != nil {
+			// Fatal drain error (read/protocol/connection) — remove without a turn.
+			p.removeConnInternal(ctx, cn, err, false)
```
ndyakov, line 1462: if we remove a conn here (or anywhere, if we switch the underlying network connection), we should flush that part of the cache that was owned by it (or if not mapped -> the whole cache).

internal/pool/pool.go
```diff
@@ -1394,6 +1394,118 @@ func (p *ConnPool) Filter(fn func(*Conn) bool) error {
...
+			continue
+		}
+		// Min-progress: handle at least one member — drained OR reconciled — before
+		// honoring the per-cycle deadline, so a pass does a bounded amount of work
+		// and cannot reconcile an entire large round (O(round*idleConns)) while
+		// ignoring an expired ctx. A deadline-truncated round resumes on the next
+		// pass (st.round/st.visited persist).
+		if handled > 0 && ctx.Err() != nil {
+			return
+		}
```
ndyakov, line 1448: if the ctx deadline is reached, even if there are no handled maybe we should stop the loop?

csc_integration.go
```diff
@@ -78,62 +105,119 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
+		var st pool.DrainState
+		drain := func(cn *pool.Conn) error { return c.drainPushNotifications(cn) }
+		for {
+			select {
+			case <-h.stop:
+				return
+			case <-ticker.C:
+				// ctx bounds the whole pass; the drain read has its own hard deadline.
+				cycleCtx, cancel := context.WithTimeout(context.Background(), interval/2)
+				cp.DrainIdleConns(cycleCtx, &st, drain)
```
ndyakov, line 393: @ofekshenawa Correct me if I am wrong but DrainIdleConns will wait for any `drain` executing on the last aquired connection before triggering the deadline. There is no clear way for this background drainer goroutine to know if there is a drain curring at the moment. Maybe we need some type of signal? Currently the code just assumes that the draining won't overlap, but with different timeouts this can happen.

csc_integration.go
```diff
@@ -78,62 +105,119 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
+
+// stopBackgroundDrainer halts the client's drainer goroutine and JOINS it before
+// returning, so a pass in progress cannot touch the pool after baseClient.Close
+// tears it down. Idempotent; called from baseClient.Close.
+func (c *baseClient) stopBackgroundDrainer() {
+	if v, ok := cscDrainHandles.LoadAndDelete(c); ok {
+		h := v.(*cscDrainHandle)
+		close(h.stop)
+		<-h.done
 	}
```
ndyakov: if for whatever reason we stop the drainer, we should flush the cache and disable it, right?

csc_integration.go
```diff
@@ -156,23 +240,21 @@ func (c *baseClient) processCached(ctx context.Context, cmd Cmder) error {
...
-		if data, ok := c.csc.Get(ctx, key); ok {
-			if err := applyCachedReply(cmd, data); err == nil {
-				return nil
-			}
-			c.csc.DeleteByCacheKey(key)
+	// Serve hits straight from the cache.
+	if data, ok := c.csc.Get(ctx, key); ok {
+		if err := applyCachedReply(cmd, data); err == nil {
+			commandHits.Add(1)
+			return nil
```
ndyakov, line 478: we have to be sure that the cache invalidation mechanism works if we are doing that - meaning:
- is there are working background bcast connection that is invalidating? (if in bcast mode)
- are there pending invalidation to be processed on it
For the shared:
- is the drainer running at the moment?

csc_integration.go
```diff
@@ -156,23 +240,21 @@ func (c *baseClient) processCached(ctx context.Context, cmd Cmder) error {
...
 	}
 
-	if _, ok := c.csc.Get(ctx, key); ok {
-		// Drain pending invalidations and re-check so the hit path cannot
-		// serve a value the server has already marked stale.
-		c.drainPendingInvalidations(ctx)
-		if data, ok := c.csc.Get(ctx, key); ok {
-			if err := applyCachedReply(cmd, data); err == nil {
-				return nil
-			}
```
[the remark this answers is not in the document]
ndyakov: 👍 cc @ofekshenawa

csc_integration.go
```diff
@@ -78,62 +105,119 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
-// different connection could otherwise race ahead of a pending "invalidate".
-// The walk is bounded by an IdleLen snapshot and dedup'd by connection ID so a
-// rotating pool (FIFO/LIFO) is not traversed indefinitely, and each connection
-// is peeked via peekAndProcessPushNotifications so the recent-health-check
-// shortcut does not suppress drainage on this non-reply-read path.
-func (c *baseClient) drainPendingInvalidations(ctx context.Context) {
-	if c.connPool == nil || c.opt.Protocol != 3 {
+// cscDrainHandles maps *baseClient -> *cscDrainHandle for its background drainer.
+// The entry is removed (and the goroutine joined) on Close.
+var cscDrainHandles sync.Map
```
ndyakov: why don't we use a channel on the client instead of this sync map? What happens if the GC picks the *baseClient without anyone calling Close on it? Maybe using something like https://pkg.go.dev/runtime#AddCleanup to the client makes more sense and we can add Close as a cleanup func. I will let you consider (this can be implemented either way right now, I do think we should have AddCleanup for the client anyway, could be a separate PR)

local_cache.go
ndyakov, line 72: I would advice against having both of those unlimited. How is this handled in other clients? 

p.s. we can simply validate the config and have at least one of those as a default limit if both are unlimited.

csc_per_connection.go
```diff
@@ -0,0 +1,480 @@
...
+	cache := NewLocalCache(state.cfg)
+	state.putCache(cn.GetID(), cache)
+
+	id := cn.GetID()
+	cn.SetOnClose(func() error {
+		if removed := state.removeCache(id); removed != nil {
+			removed.Flush()
+		}
+		return nil
+	})
```
ndyakov: if we check the comment on `SetOnClose` it states that it will overwrite any previous fun that was set on close. This will overwrite the mechanism for streaming credentials provider. Initially it was chaining, but there was a memory leak when the streaming credential provider reauthenticated.

Maybe we can have another csc related callback on the connection to not overcomplicate the onclose?

csc_broadcast_sidecar.go
```diff
@@ -0,0 +1,401 @@
...
+// handshake (authenticating if credentials are configured) and issues
+// CLIENT TRACKING ON BCAST. On success it publishes the new conn/reader/
+// writer atomically under connMu and returns nil.
+//
+// The handshake commands are built with the same Cmder constructors and
+// argument builders the normal client uses (writeCmd + appendClientTrackingOptions),
+// so RESP3 framing, auth, and tracking-option encoding live in one place; only
+// the dedicated socket and the push-frame read loop are bespoke.
+func (s *broadcastSidecar) dialAndHandshake(ctx context.Context) error {
+	if s.opt.Dialer == nil {
```
ndyakov: now this won't work with SCH and I think this is acceptable, but we should discuss if we should prepare it to work with maintnotifications.

csc_broadcast_sidecar.go
```diff
@@ -0,0 +1,401 @@
...
+
+		s.connMu.Lock()
+		rd := s.reader
+		s.connMu.Unlock()
+		if rd == nil {
+			// Lost connection — reconnect.
+			s.ready.Store(false)
+			if !s.sleepWithDone(backoff) {
+				return
+			}
```
ndyakov: should we flush the cache on reconnect?

csc_broadcast_sidecar.go
```diff
@@ -0,0 +1,401 @@
...
+// spawns the read-loop goroutine. Returning nil from Start means the server
+// has acknowledged CLIENT TRACKING ON BCAST and the loop is live.
+func (s *broadcastSidecar) Start(ctx context.Context) error {
+	if err := s.dialAndHandshake(ctx); err != nil {
+		return err
+	}
+	s.ready.Store(true)
+	go s.readLoop()
+	return nil
+}
```
ndyakov: We should do a periodic ping/health check on that connection. The connection can disconnect on the server side and this will be undetected until the OS TCP keepalive fires.

csc_integration.go
```diff
@@ -4,21 +4,48 @@ import (
...
 	"github.com/redis/go-redis/v9/internal/proto"
 	"github.com/redis/go-redis/v9/push"
 )
 
+// RESPInvalidationBytesRead returns the cumulative byte count of key names
+// received in "invalidate" push frames since process start. This counter is
+// shared across all clients in the process and is never reset by the library.
+func RESPInvalidationBytesRead() int64 {
+	return proto.InvalidationBytesRead.Load()
+}
```
ndyakov: why do we keep that, do we need it? if so, should we also increment it in the shared handler?

csc_broadcast_sidecar.go
```diff
@@ -0,0 +1,401 @@
...
+// Shutdown signals the read loop to exit and closes the underlying socket.
+// Bounded to 5s to satisfy the spec's no-block guarantee.
+func (s *broadcastSidecar) Shutdown() {
+	s.doneOnce.Do(func() { close(s.done) })
+
+	s.connMu.Lock()
+	if s.conn != nil {
+		_ = s.conn.Close()
+	}
+	s.connMu.Unlock()
```
ndyakov: we should check in a defer as well, since there may be a running `dialAndHandshake` while we do the s.conn check initially and change the underlying conn before the read loop is done.

ndyakov (changes requested):
This was quite the read @ofekshenawa, amazing effort to achieve this and implementing three different client side caching strategies.  

I left some blocking comments around correctness. Took a look at the tests and they seem to mainly test the green (happy) path and the data structures (which are well executed!), let's focus on testing the lifecycle/eviction/reconnect paths that carry the real correctness risk (and several sharded paths).

redis.go
ofekshenawa: fixed. Drain loop peeks the name and leaves pub/sub-reserved frames buffered

redis.go
ofekshenawa: CSC can never be active with a VoidProcessor, The RegisterHandler fails in attachCSC so c.csc stays nil and no drainer starts. The type check specifically applies an error catching loop to the builtin processor, while leaving custom processors alone to run their own error handling logic

internal/pool/pool.go
ofekshenawa: It's Intentional, we force the loop to process at least one connection per run, even if the time limit has already expired. This prevents a starvation trap where a very short time limit gets entirely eaten up by setup tasks, causing the loop to stall forever without actually doing any work.

csc_integration.go
ofekshenawa: Overlapping drain cycles are impossible because a single worker handles them one at a time. If a cycle takes longer than expected, the next scheduled run simply waits or is skipped, they can never run simultaneously. Furthermore, any connection currently being checked is temporarily hidden from the rest of the system so no other process can interfere with it, and any delay is strictly limited to a fraction of a millisecond.

csc_broadcast_sidecar.go
```diff
@@ -0,0 +1,615 @@
...
+	// BCAST subscription (which would flush the cache). Fall back to a reconnect
+	// when there's no live conn (mid-handshake; the gen bump handles it) or the
+	// AUTH write fails.
+	if username, password := creds.BasicAuth(); password != "" {
+		if err := l.s.reauthConn(username, password); err == nil {
+			return
+		}
+	}
+	l.s.tearDownConn()
+}
```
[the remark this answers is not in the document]
ofekshenawa: Not a real race: only one goroutine ever reads the connection while writes are already serialized by a lock, and a connection safely supports reading and writing at the same time, so the re-auth reply just comes back as a normal message the reader ignores, exactly like the existing keepalive ping, with no corruption

csc_integration.go
```diff
@@ -71,94 +114,290 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
+			return true
+		}
+	}
+	return false
+}
+
+// registerConnEvictHook wires cscEvictOnRemoveHook when the cache supports
+// owning-connection attribution and the pool supports hooks. No-op otherwise
+// (custom caches / poolers fall back to the MaxStaleness backstop).
+func (c *baseClient) registerConnEvictHook(cache Cache) {
```
ndyakov, line 292: The problem: with the default strategy, the cache can still serve a stale value forever after a connection expires normally. I can still reproduce it on this head. Anchoring here on the hook registration since the pool-side lines are outside this diff.

How it happens:
1. A key gets cached. The owner index remembers which connection fetched it, so that connection's entries can be evicted when the connection dies. Good design, and it works: when I kill the connection with CLIENT KILL, the entries get evicted and the next GET is fresh.
2. But the eviction runs from the pool's OnRemove hook, and not every connection teardown fires that hook. When a connection simply expires (ConnMaxLifetime or ConnMaxIdleTime), the pool retires it through CloseConn (see the stale rotation in getConn, pool.go ~823), and CloseConn never calls ProcessOnRemove. Same for the MaxIdleConns overflow discard in putConn and for ConnPool.Close.
3. So on the most common teardown path (normal expiry), the entries stay in the cache, the server has forgotten it was tracking them, no invalidation will ever arrive, and MaxStaleness is off by default, so nothing else evicts them.

Repro: PoolSize 1, ConnMaxLifetime 300ms. Cache a key, wait for the connection to rotate, SET the key from another client, GET still returns the old value.

Fix options: make CloseConn fire ProcessOnRemove (or route expiry through RemoveWithoutTurn, which already fires hooks), or register the eviction with cn.SetOnCscClose in initConn the way PerConnection does. Conn.Close runs on every teardown path, so it cannot be bypassed. Please also add a regression test with a tiny ConnMaxLifetime; the current tests only cover teardowns where the hook does fire.

csc_integration.go
```diff
@@ -71,94 +114,290 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
+	if h == nil {
+		return
+	}
+	// Owner only. The hook (when present) is always registered alongside the
+	// drainer, so removing it here — before the pool is closed — can't strand it.
+	if c.cscPoolHook != nil {
+		if reg, ok := c.connPool.(poolHookRegistrar); ok {
+			reg.RemovePoolHook(c.cscPoolHook)
 		}
+		c.cscPoolHook = nil
```
ndyakov: The problem: calling Close() while another goroutine runs a command is a data race that go test -race flags.

How it happens: Close sets `c.cscPoolHook = nil` and `c.cscDrainHandle = nil` as ordinary writes. A command in flight on another goroutine reads cscPoolHook at the same time (in processCached and fulfillCached) with no synchronization. Close-while-commands-run is something the client tolerates everywhere else, so people will do it.

Fix: just do not nil the fields. RemovePoolHook already detaches the hook, and reading the stale pointer afterwards is harmless because signalStop is a sync.Once. Alternatively make the fields atomic.Pointer. Two lines either way.

push/processor.go
```diff
@@ -80,11 +80,13 @@ func (p *Processor) ProcessPendingNotifications(ctx context.Context, handlerCtx
 			break
```
ndyakov, line 83: The problem: one badly timed network packet can permanently stop invalidation delivery on a connection, so its cached entries go stale until that connection happens to be used for a command.

How it happens:
1. An invalidation message can arrive split across packets. If the drainer looks at the connection between the two halves, it sees only the start of the message header.
2. PeekPushNotificationName (proto/reader.go) reads at most min(36, Buffered()) bytes and never asks the socket for more. On the half header it returns a parse error. That error is handled safely now (the connection is kept, nothing consumed), which is correct.
3. But the half header stays at the front of the buffer, and the peek never refills, so every future drain pass hits the same parse error, even long after the second half arrived. The invalidations queued behind it are never processed. I have a test showing the handler runs once and then never again across three more passes.
4. When a command finally borrows that connection, typed commands error and the connection gets removed, which recovers. But a generic Cmd/Do can read the push frame as if it were its own reply.

Fix: let this one peek refill from the socket under the drain's existing 50us hard deadline. If the bytes are not there yet, the timeout is the same harmless stop as today; once they are, the header parses and everything unblocks. A MaxStaleness default would also cap how long the stall can matter.

csc_integration.go
```diff
@@ -71,94 +114,290 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
 	}
-	for _, cn := range borrowed {
-		c.connPool.Put(ctx, cn)
+	c.cscDrainHandle = nil
+	h.signalStop()
+	<-h.done
+	if c.cscOwnsCache && c.csc != nil {
```
ndyakov: The problem: if two clients share one cache through Options.ClientSideCache, closing one client leaves dead entries that the other client keeps serving, and nothing can ever invalidate them.

How it happens: invalidations for an entry are delivered to the connection that fetched it. When client A closes, A's connections die with it, so entries fetched by A can no longer receive invalidations. This flush only runs for a cache the client built itself (cscOwnsCache); an injected shared cache is skipped, and pool Close does not fire the OnRemove hook that would have evicted A's entries one by one.

Fix: on Close, evict the closing client's connection-owned entries even when the cache was injected. Or document that a Cache instance must not be shared between clients.

ndyakov (changes requested):
Re-reviewed 50833d0f, including live-server repros and the test suites under -race. The fixes from the last round hold up well. Verified fixed: GET after Close now errors instead of serving from the cache (checked live, including Close during a reconnect), the drain rework (a split invalidation frame no longer kills a healthy connection, a mid-frame read error still removes the connection, pubsub frames are left alone), the drainer handle living on the client with GC cleanup instead of the package map, the placeholder cap under load (and having Reserve reject its own overflow rather than aborting someone else's fetch is a nice touch), the per-conn turn leak, and the hook registration locking.

Three smaller things. A rejected re-auth on the sidecar gets logged as "discard non-push reply failed", which hides what actually happened; worth its own log line. The GC cleanup path does not re-clear the ready flag after a racing reconnect the way Shutdown does, so a client that gets collected without Close can in theory leave a clone serving stale until MaxStaleness. And a few comments say custom caches and poolers "fall back to the MaxStaleness backstop", but MaxStaleness is off by default, so today that backstop does not exist.

My two remaining blockers are inline: the CloseConn hook gap and the Close data race.

csc_integration.go
```diff
@@ -71,94 +121,347 @@ func registerInvalidateHandler(p push.NotificationProcessor, cache Cache, db int
...
+				"(FulfillOwned/EvictByConn); use redis.NewLocalCache or implement it.")
+		return
+	}
 	if err := registerInvalidateHandler(c.pushProcessor, cache, c.opt.DB); err != nil {
 		internal.Logger.Printf(ctx, "csc: failed to register invalidate handler: %v", err)
 		return
 	}
 	c.csc = cache
+	c.startBackgroundDrainer()
+	c.registerConnEvictHook(cache)
```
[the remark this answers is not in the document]
ofekshenawa: This case cannot happen because connPool is a pool.Pooler, and pool.Pooler always requires AddPoolHook and RemovePoolHook. Therefore, the hook check cannot fail for a valid pool, and cached entries will always be linked to their connection.

ndyakov (approved):
Re-reviewed d5a0b678. Both blockers from my last review are fixed, and I verified them against a live server: the connection-expiry staleness repro (tiny ConnMaxLifetime, external write, read) now returns fresh in both the lifetime and idle-timeout variants, since the eviction moved to the conn close hook that every teardown path runs; and the Close data race is gone by construction now that the hook fields are never cleared and liveness is an atomic flag (clean under a Close-vs-Get hammer with -race). The truncated-header stall is also fixed, the refilling peek unblocks my old repro once the tail arrives, and the new split-point tests are good. Reducing the PR to SharedTracking only was the right call, and the cleanup is thorough: no dangling references, the strategy enum clamps with a log, and the rewritten guide matches the code. The ConnOwnedCache contract also closes the shared-injected-cache hole.

One fast-follow and a few nits, none blocking. The custom push-processor branch on the drain path still re-pools a conn after a deadline error that may have consumed part of a frame; treating deadline-shaped errors as fatal before the builtin/custom split would close it. The removed-ring is fixed at 64 slots while the default pool can be larger, so a mass close can theoretically orphan one racing fulfill; sizing it to the pool would remove the case. The peek window grows a byte at a time, so a split header re-parses once per byte; jumping straight to Buffered() would fix the quadratic re-parse. And 13 of the 20 new wordlist entries look like leftovers from the deleted docs.

Approving. Nice work on this round.

PR #3853, opened 2026-06-16
fix(maintnotifications): keep ModeAuto fail-open

terrorobe:
## Summary

In `ModeAuto`, a single successful `CLIENT MAINT_NOTIFICATIONS` handshake currently mutates the client policy to `ModeEnabled`. From that point on, the client is fail-closed: if a later connection reaches an endpoint that rejects `CLIENT MAINT_NOTIFICATIONS`, `initConn` fails and the connection cannot be used.

Against managed/proxied Redis endpoints that can change which backend a connection lands on (failover, reconnect, rolling maintenance), this can turn an optional feature into a persistent connection-init failure loop: every new connection retries the handshake, gets `ERR unknown subcommand 'MAINT_NOTIFICATIONS'`, and fails.

This PR stops promoting `ModeAuto` to `ModeEnabled` on a successful probe. `ModeAuto` stays fail-open and can downgrade if a later connection reaches an endpoint that rejects the command. Explicit `ModeEnabled` is unchanged (still fail-closed), and `ModeDisabled` is unchanged.

## Background / motivation

The documented contract is:

- `ModeDisabled` — client does not send `CLIENT MAINT_NOTIFICATIONS ON`
- `ModeEnabled` — client sends it and interrupts the connection on error (fail-closed)
- `ModeAuto` (default) — client tries it and disables the feature on error (fail-open)

The implementation currently uses `MaintNotificationsConfig.Mode` for both the user's configured policy and the runtime "support was detected once" state. A successful auto probe only proves that one connection accepted the command at that moment; it should not make the optional feature mandatory for the rest of the client's lifetime.

This can occur with managed/proxied Redis deployments where reconnects or failovers may route a client to a backend with different command support.

## Root cause

`initConn` in `redis.go` (success branch of the handshake):

```go
} else {
    // handshake was executed successfully
    // to make sure that the handshake will be executed on other connections as well if it was successfully
    // executed on this connection, we will force the handshake to be executed on all connections
    c.optLock.Lock()
    c.opt.MaintNotificationsConfig.Mode = maintnotifications.ModeEnabled
    c.optLock.Unlock()
}
```

The promotion is also unnecessary for the stated goal. The handshake is gated by:

```go
maintNotifEnabled := c.opt.MaintNotificationsConfig != nil &&
    c.opt.MaintNotificationsConfig.Mode != maintnotifications.ModeDisabled
```

So `ModeAuto` already re-runs the handshake on every new connection. Promotion to `ModeEnabled` does not make future connections "also handshake" — they already do — it only changes the failure semantics from fail-open to fail-closed.

## Fix

Remove the `ModeAuto -> ModeEnabled` promotion on a successful handshake.

- Supported endpoint: `ModeAuto` connections still send `CLIENT MAINT_NOTIFICATIONS` and enable the feature on each connection (no change in commands sent).
- Endpoint that later rejects the command: `ModeAuto` falls through the existing default branch, disables maintnotifications upgrades for the client, and the connection succeeds (no loop).
- `ModeEnabled`: unchanged, still fails closed on error.
- `ModeDisabled`: unchanged.

This is not more chatty: while supported, the number of commands per connection is the same. After an unsupported response it is strictly fewer (no repeated failing handshakes).

## Behavior before/after (ModeAuto, command support changes across connections)

Before:

```text
conn1: CLIENT MAINT_NOTIFICATIONS -> OK     ; mode becomes ModeEnabled
conn2: CLIENT MAINT_NOTIFICATIONS -> ERR... ; ModeEnabled => fail connection
conn3: CLIENT MAINT_NOTIFICATIONS -> ERR... ; ModeEnabled => fail connection
... repeats on every new connection
```

After:

```text
conn1: CLIENT MAINT_NOTIFICATIONS -> OK     ; mode stays ModeAuto
conn2: CLIENT MAINT_NOTIFICATIONS -> ERR... ; ModeAuto  => disable + continue
conn3: (no CLIENT MAINT_NOTIFICATIONS)      ; connection succeeds
```

## Tests

Added to `internal_maint_notif_test.go`, reusing the existing `initConn` mock-server style. A new mock accepts the first `CLIENT MAINT_NOTIFICATIONS` and rejects later ones.

- `TestInitConn_ModeAuto_RemainsFailOpenAfterSuccessfulHandshake`
  - first probe succeeds; configured mode stays `ModeAuto`
  - second probe (now unsupported) does not fail `initConn`
  - third connection no longer sends `CLIENT MAINT_NOTIFICATIONS` (downgrade sticks)
- `TestInitConn_ModeEnabled_RemainsFailClosedAfterSuccessfulHandshake`
  - explicit `ModeEnabled` still fails closed when a later probe is unsupported

The first test fails on `master` (second `initConn` returns `failed to enable maintnotifications: ERR unknown subcommand 'MAINT_NOTIFICATIONS'`) and passes with this change.

## Test plan

```bash
go test -run 'TestInitConn_(HelloFallback|ModeAuto|ModeEnabled)' ./
go test -race -run 'TestInitConn|MaintNotifications|OptionsClone' ./
go test -run 'Test.*MaintNotifications|Test.*Cluster.*Maint' ./
```

All pass locally. CI should exercise the full Docker test matrix.

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -33,6 +33,11 @@ const (
 
 	// CloseReasonFailover indicates the connection was closed due to a failover event.
 	CloseReasonFailover = "failover"
+
+	// CloseReasonMaintNotificationsDisabled indicates the connection enabled
+	// maintenance notifications, but the client later downgraded and disabled
+	// maintenance notification handling for the pool.
+	CloseReasonMaintNotificationsDisabled = "maintnotifications_disabled"
 )
 
 // Metric state constants for connection state tracking.
@@ -313,6 +318,10 @@ type Stats struct {
 	PendingRequests uint32 // number of pending requests waiting for a connection
 
 	PubSubStats PubSubStats
+}
+
+type ConnRetirer interface {
+	RetireConns(ctx context.Context, conns []*Conn, reason string)
 }
 
 type Pooler interface {
@@ -1217,6 +1226,11 @@ func (p *ConnPool) putConn(ctx context.Context, cn *Conn, freeTurn bool) {
 	shouldRemove := false
 	var err error
 
+	if reason := cn.CloseOnPutReason(); reason != "" {
+		p.removeConnInternal(ctx, cn, errors.New(reason), freeTurn)
+		return
+	}
+
 	if cn.HasBufferedData() {
 		// Peek at the reply type to check if it's a push notification
 		if replyType, err := cn.PeekReplyTypeSafe(); err != nil || replyType != proto.RespPush {
@@ -1560,6 +1574,47 @@ func (p *ConnPool) Stats() *Stats {
 
 func (p *ConnPool) closed() bool {
 	return atomic.LoadUint32(&p._closed) == 1
+}
+
+func (p *ConnPool) RetireConns(ctx context.Context, conns []*Conn, reason string) {
+	if len(conns) == 0 {
+		return
+	}
+
+	idleConnSet := make(map[*Conn]struct{})
+	toClose := make([]*Conn, 0, len(conns))
+
+	p.connsMu.Lock()
+	for _, ic := range p.idleConns {
+		idleConnSet[ic] = struct{}{}
+	}
+	for _, cn := range conns {
+		if cn == nil {
+			continue
+		}
+		if _, ok := p.conns[cn.GetID()]; !ok {
+			continue
+		}
+		if _, isIdle := idleConnSet[cn]; isIdle {
+			if p.removeConn(cn) {
+				toClose = append(toClose, cn)
+			}
+			continue
+		}
+		cn.MarkCloseOnPut(reason)
+	}
+	p.connsMu.Unlock()
+
+	if hookManager := p.hookManager.Load(); hookManager != nil {
+		for _, cn := range toClose {
+			hookManager.ProcessOnRemove(ctx, cn, errors.New(reason))
+		}
+	}
+	for _, cn := range toClose {
+		p.recordConnectionMetrics(ctx, cn, reason, MetricStateIdle)
+		_ = p.closeConn(cn)
+	}
+	p.checkMinIdleConns()
 }
 
 func (p *ConnPool) Filter(fn func(*Conn) bool) error {
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -114,6 +114,11 @@ type Conn struct {
 	// to inform the goroutine using the connection why the connection was closed.
 	closeReason uberatomic.String
 
+	// closeOnPutReason marks an in-use connection for removal when it is returned
+	// to the pool. The socket is left open for the in-flight command and closed
+	// by ConnPool.Put.
+	closeOnPutReason uberatomic.String
+
 	// maintenanceNotifications upgrade support: relaxed timeouts during migrations/failovers
 
 	// Using atomic operations for lock-free access to avoid mutex contention
@@ -435,6 +440,17 @@ func (cn *Conn) IncrementAndGetHandoffRetries(n int) int {
 // IsPooled returns true if the connection is managed by a pool and will be pooled on Put.
 func (cn *Conn) IsPooled() bool {
 	return cn.pooled
+}
+
+// MarkCloseOnPut marks the connection for removal when it is returned to the pool.
+func (cn *Conn) MarkCloseOnPut(reason string) {
+	cn.closeOnPutReason.Store(reason)
+}
+
+// CloseOnPutReason returns a non-empty reason when the connection should be
+// removed instead of pooled on Put.
+func (cn *Conn) CloseOnPutReason() string {
+	return cn.closeOnPutReason.Load()
 }
 
 // IsPubSub returns true if the connection is used for PubSub.
--- a/internal/pool/pool_test.go
+++ b/internal/pool/pool_test.go
@@ -75,6 +75,25 @@ var _ = Describe("ConnPool", func() {
 		}))
 	})
 
+	It("should retire idle conns and mark in-use conns for close on put", func() {
+		idleConn, err := connPool.Get(ctx)
+		Expect(err).NotTo(HaveOccurred())
+		inUseConn, err := connPool.Get(ctx)
+		Expect(err).NotTo(HaveOccurred())
+
+		connPool.Put(ctx, idleConn)
+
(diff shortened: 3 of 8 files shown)


terrorobe:
A note on scope: this PR keeps the existing fallback model and only removes the success-path promotion from `ModeAuto` to `ModeEnabled`.

The broader state-model issue is that `MaintNotificationsConfig.Mode` currently carries both the user's configured policy and runtime capability state. A fuller design could keep the configured mode immutable and track support separately, for example:

```go
ConfiguredMode: Disabled | Auto | Enabled
RuntimeSupport: Unknown | Supported | Unsupported
```

with behavior:

```text
Disabled:
  never send CLIENT MAINT_NOTIFICATIONS

Enabled:
  send on every connection; fail the connection on error

Auto + Unknown:
  send; OK -> Supported, unsupported -> Unsupported; connection continues

Auto + Supported:
  send; OK -> continue; unsupported -> Unsupported; connection continues

Auto + Unsupported:
  skip; connection continues
  optionally re-probe after a cooldown
```

That would preserve the documented semantics without mutating the user's policy at all, while still allowing the client to rediscover support later if the deployment changes. I kept this PR smaller because the existing code already has a `ModeAuto -> ModeDisabled` fallback path; the production-impacting bug is the success path converting auto into fail-closed enabled mode.

ndyakov:
@terrorobe This is something we were currently looking into, thank you for opening the PR. Our idea was to introduce a third mode auto-enabled which will point out that there were connections where the handshake succeeded, because those should be recycled. Maybe we could still keep the Auto mode, but have a record of the connections where the handshake succeeded and mark them for removal (simply putting the `pooled` to `false` may work if they are in use at the moment, I have to check) when removing the listeners.

Just to clarify what I think should happen here:
 Since the push notification listeners are per pool, we should not have connection in the pool that both have and do not have maint notifications enabled. Actually, it won't hurt to have the handlers enabled and then some connection may not need them, but other will. However it will be a problem to have a connection receiving the push notifications and have no handler to process it. We should cleanup the connection for which the maintnotifications were enabled before disconnecting the handlers from the pool.

terrorobe:
Thanks for the guidance. The driver on our side was an Azure AMR maintenance event that left several go-redis clients stuck retrying `CLIENT MAINT_NOTIFICATIONS`. The `ModeAuto -> ModeEnabled` promotion fixed here looks like the likely trigger for that failure mode.

Do you want me to expand this PR to also address the pool/listener cleanup gaps you described, or would you prefer to take this in a different direction yourself? Happy either way!

ndyakov:
@terrorobe We had the same working theory about your case, but haven't had the chance to verify it. Please address this cleanup of connections in this PR, so we can include it as a working piece. If we decide to change the direction in the future (with the auto-enabled or something similar) we will make sure this flow continues to work (thanks for the tests). 

I will make sure we review, include and release as soon as possible.

terrorobe:
Do you have a preference for how we should handle connections that are currently in use when a maintnotifications downgrade happens?

Closing them immediately seems safest from a protocol perspective, but will surface application-visible errors for in-flight commands. Marking them unpooled is more graceful, but it looks like a larger lifecycle change since we would need to keep handlers installed until those connections return. I also do not have a great answer for blocking/long-running operations except "wait for n seconds, then close anyway."

Since the downgrades should not be a regular occurrence, disrupting in-flight commands might be an acceptable approach. Thoughts?

ndyakov:
@terrorobe As long as we hold the lock on the pool, the state of the connection should show if it is in use or not. For connections that are currently idle it is safe to close them. For connection that are in use it is better to mark them for closing and close them when they are put back in the pool. There is still a possibility that the response of the executed command is a out of bound notification that won't be processed (because we drop the handlers), but it is slim and for most cases the commands should execute normally. Where in the other scenario we are definitely end up with failed in-flight commands. I do agree that the chances of this happening are quite small, so if it turns out to be a complex problem to solve in this PR, feel free to just close the connections and we can work on another approach in the future.

terrorobe:
Pushed the connection cleanup.

On a successful `CLIENT MAINT_NOTIFICATIONS` the connection is now tracked, and on downgrade/disable the tracked connections are retired before the pool hook is removed. Idle ones are closed immediately; in-use ones are marked close-on-put and closed when they’re returned, so in-flight commands aren’t cut off. `OnRemove` untracks so the set doesn’t leak.

That leaves the slim window you mentioned where an in-use conn could still get a notification after the pool hook/listeners are gone. Closing in-use conns outright would avoid it but fail in-flight commands, so I went with close-on-put.

I also hit a race in `initConn`: the lookup + track could run after a concurrent downgrade had already closed the manager, leaving an enabled-but-untracked conn. Fixed by holding the manager read lock across the handshake and tracking. There’s a regression test that fails without it.

Tests cover idle vs in-use retire in the pool, retire-before-hook-removal + untrack in the manager, and two client-level cases: in-use conn retired on put after downgrade, and concurrent in-flight handshake + downgrade.

internal/pool/pool.go
```diff
@@ -1562,6 +1576,47 @@ func (p *ConnPool) closed() bool {
...
+	if hookManager := p.hookManager.Load(); hookManager != nil {
+		for _, cn := range toClose {
+			hookManager.ProcessOnRemove(ctx, cn, errors.New(reason))
+		}
+	}
+	for _, cn := range toClose {
+		p.recordConnectionMetrics(ctx, cn, reason, MetricStateIdle)
+		_ = p.closeConn(cn)
+	}
+	p.checkMinIdleConns()
```
ndyakov, line 1617: can we use `removeConnInternal` here ?

terrorobe: I don't think we can reuse `removeConnInternal` directly here because of the `connsMu` boundary. Calling it while holding `connsMu` would deadlock, but releasing the lock first would let an idle conn get checked out before we retire it.

This path also needs different behavior for idle vs in-use conns: idle conns are removed immediately with idle-state metrics, while in-use conns are only marked close-on-put so we don't cut off in-flight commands. `removeConnInternal` assumes the normal Remove/Put path and used-state metrics.

We can reduce the duplication though, I can pull the shared tail (OnRemove hook, removal metrics, closed metric, close) into a small finalize helper and call it from both, keeping the idle/used distinction intact.

ndyakov: It is fine as it is right now, I will check if we can improve further.

ndyakov (approved):
Thank you @terrorobe!

PR #3861, opened 2026-06-23
fix(cluster): data race appending to shared masters/slaves slices

saddamr3e:
Concurrent command routing on a ClusterClient trips the race detector:

    WARNING: DATA RACE
    Write at 0x00c00007d498 by goroutine 10:
      runtime.slicecopy()
      redis.(*ClusterClient).pickArbitraryNode()
          osscluster_router.go:497
    Previous write at 0x00c00007d498 by goroutine 40:
      runtime.slicecopy()
      redis.(*ClusterClient).pickArbitraryNode()
          osscluster_router.go:497

clusterState is published as a shared immutable snapshot, but its Masters slice is grown with appendIfNotExist and keeps spare capacity, so `append(state.Masters, state.Slaves...)` writes the slaves into the shared backing array instead of allocating. Two goroutines routing at the same time then write the same slots. The same construct sits in executeOnAllNodes, pickArbitraryNode and mapCmdsByNode, and is reachable in the default config (RoundRobinPicker, routing policies on).

Copy into a fresh slice at each site so the snapshot is never mutated. The added test fails under `-race` before the change and passes after.

---

> [!NOTE]
> **Medium Risk**
> Touches hot concurrent cluster routing paths used with default ShardPicker and routing policies; the change is a targeted concurrency fix with a race test.
> 
> **Overview**
> Fixes a **data race** in concurrent `ClusterClient` routing when code built a combined master+replica list with `append(state.Masters, state.Slaves...)`. That append can write into spare capacity on the shared `clusterState` snapshot (grown via `appendIfNotExist`), so parallel routers corrupt the same backing array.
> 
> **`executeOnAllNodes`** now copies masters and slaves into a **new slice** before fan-out. **`pickArbitraryNode`** and **read-only keyless pipeline routing** in **`mapCmdsByNode`** avoid a combined slice entirely by indexing `ShardPicker` across `len(Masters)+len(Slaves)` and choosing from `Masters` or `Slaves` by index.
> 
> Adds **`TestPickArbitraryNodeConcurrent`** to exercise `pickArbitraryNode` under concurrency (fails under `-race` before the fix).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96919052dfbb4cbcb74e70d88335344a03d5d4a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -1649,12 +1649,15 @@ func (c *ClusterClient) mapCmdsByNode(ctx context.Context, cmdsMap *cmdsMap, cmd
 				if len(state.Masters) == 0 {
 					return errClusterNoNodes
 				}
-				// For read-only keyless commands, pick from all nodes (masters + slaves)
-				allNodes := make([]*clusterNode, 0, len(state.Masters)+len(state.Slaves))
-				allNodes = append(allNodes, state.Masters...)
-				allNodes = append(allNodes, state.Slaves...)
-				idx := c.opt.ShardPicker.Next(len(allNodes))
-				node = allNodes[idx]
+				// For read-only keyless commands, pick from all nodes (masters + slaves).
+				// Index directly instead of building a combined slice, which would
+				// append into the shared snapshot's spare capacity and race.
+				idx := c.opt.ShardPicker.Next(len(state.Masters) + len(state.Slaves))
+				if idx < len(state.Masters) {
+					node = state.Masters[idx]
+				} else {
+					node = state.Slaves[idx-len(state.Masters)]
+				}
 			} else {
 				node, err = c.slotReadOnlyNode(state, slot)
 				if err != nil {
--- a/osscluster.go
+++ b/osscluster.go
@@ -1650,7 +1650,9 @@ func (c *ClusterClient) mapCmdsByNode(ctx context.Context, cmdsMap *cmdsMap, cmd
 					return errClusterNoNodes
 				}
 				// For read-only keyless commands, pick from all nodes (masters + slaves)
-				allNodes := append(state.Masters, state.Slaves...)
+				allNodes := make([]*clusterNode, 0, len(state.Masters)+len(state.Slaves))
+				allNodes = append(allNodes, state.Masters...)
+				allNodes = append(allNodes, state.Slaves...)
 				idx := c.opt.ShardPicker.Next(len(allNodes))
 				node = allNodes[idx]
 			} else {
--- a/osscluster_router.go
+++ b/osscluster_router.go
@@ -85,7 +85,9 @@ func (c *ClusterClient) executeOnAllNodes(ctx context.Context, cmd Cmder, policy
 		return err
 	}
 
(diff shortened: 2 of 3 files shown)


osscluster_router.go
```diff
@@ -494,7 +496,9 @@ func (c *ClusterClient) pickArbitraryNode(ctx context.Context) *clusterNode {
 		return nil
 	}
 
-	allNodes := append(state.Masters, state.Slaves...)
+	allNodes := make([]*clusterNode, 0, len(state.Masters)+len(state.Slaves))
+	allNodes = append(allNodes, state.Masters...)
+	allNodes = append(allNodes, state.Slaves...)
 
 	idx := c.opt.ShardPicker.Next(len(allNodes))
 	return allNodes[idx]
```
ndyakov: here, instead of creating a new slice, can't we do a shardPicker.Next(len(state.Masters)+len(state.Slaves)) and then with basic math get the underlying slice from which we should pick?

osscluster.go
```diff
@@ -1650,7 +1650,9 @@ func (c *ClusterClient) mapCmdsByNode(ctx context.Context, cmdsMap *cmdsMap, cmd
 					return errClusterNoNodes
 				}
 				// For read-only keyless commands, pick from all nodes (masters + slaves)
-				allNodes := append(state.Masters, state.Slaves...)
+				allNodes := make([]*clusterNode, 0, len(state.Masters)+len(state.Slaves))
+				allNodes = append(allNodes, state.Masters...)
+				allNodes = append(allNodes, state.Slaves...)
```
ndyakov: here, instead of creating a new slice, can't we do a `shardPicker.Next(len(state.Masters)+len(state.Slaves))` and then with basic math get the underlying slice from which we should pick?

ndyakov (commented):
@saddamr3e thank you for spotting this. Left suggestion for two of the cases. I do think we can skip the allocation.

osscluster_router.go
saddamr3e: Makes sense. Dropped the allocation: Next(len(Masters)+len(Slaves)), then idx < len(Masters) picks from Masters, else Slaves[idx-len(Masters)]. Nothing writes the shared backing array now, so the race is gone.

osscluster.go
saddamr3e: Same here, switched to the index math instead of building allNodes. idx < len(Masters) -> Masters[idx], otherwise Slaves[idx-len(Masters)].

saddamr3e:
Applied the index-math approach to both single-pick sites (pickArbitraryNode and mapCmdsByNode), no more combined slice. Left executeOnAllNodes on the copy since it executes on every node and needs the full list, but it now copies into a fresh slice rather than appending into the shared snapshot. Race test still passes under -race -count=5.

ndyakov (approved):
Thank you @saddamr3e!

PR #3862, opened 2026-06-24
fix(ring): data race on onNewNode during SetAddrs

saddamr3e:
go test -race, registering a callback while resharding:

    WARNING: DATA RACE
    Read at ... by goroutine 11:
      (*ringSharding).newRingShards()  ring.go:411
      (*ringSharding).SetAddrs()       ring.go:374
    Previous write at ... by goroutine 10:
      (*ringSharding).OnNewNode()      ring.go:347

OnNewNode appends to c.onNewNode under c.mu, but SetAddrs runs newRingShards after dropping the lock and ranges over c.onNewNode unguarded when a new shard gets created, so a reshard racing a callback registration touches the slice header from two goroutines. Snapshot the slice under the same RLock that already captures c.shards and pass it in; clusterNodes.GetOrCreateWithNodeAddress reads the equivalent list under its lock already.

---

> [!NOTE]
> **Low Risk**
> Small synchronization fix in ring sharding with a dedicated race test; behavior for callers is unchanged aside from eliminating undefined concurrent access.
> 
> **Overview**
> Fixes a **data race** when `SetAddrs` reshards the ring while another goroutine calls `OnNewNode`.
> 
> `SetAddrs` already snapshots `c.shards` under `c.mu` before `newRingShards`, but `newRingShards` iterated `c.onNewNode` without the lock while `OnNewNode` appends under the lock. The fix **copies the callback slice under the same read lock** and passes it into `newRingShards`, matching how the cluster client handles equivalent callbacks.
> 
> Adds `TestRingShardingOnNewNodeConcurrent` to exercise concurrent registration and resharding under `-race`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8202a0faeff4e765242834d38bd9b1d1a56c8703. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -369,9 +369,10 @@ func (c *ringSharding) SetAddrs(addrs map[string]string) {
 		return
 	}
 	existing := c.shards
+	onNewNode := c.onNewNode
 	c.mu.RUnlock()
 
-	shards, created, unused := c.newRingShards(addrs, existing)
+	shards, created, unused := c.newRingShards(addrs, existing, onNewNode)
 
 	c.mu.Lock()
 	if c.closed {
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
LGTM, thanks @saddamr3e

PR #3865, opened 2026-06-24
feat(autopipeline): #2 add pipeline pool

ndyakov:
Adds an **optional dedicated connection pool tuned for pipelining** — a small
pool of large-buffer connections. Independent of autopipelining: it also speeds
up manual `Pipeline()` / `TxPipelined()`, so it ships value on its own.

Pipelines benefit from large read/write buffers, but sizing the *whole* pool
that big wastes memory. A small dedicated pool of big-buffer connections gives
the win at much lower memory cost. Zero/unset options = current behavior (no
separate pool created; falls back to the main pool).

---

> [!NOTE]
> **High Risk**
> Touches core connection pooling, pipeline execution, maintnotifications/failover, and large autopipeline semantic changes (defaults, `Do`, ordering/coalescing) that can affect latency, throughput, and ordering for existing users.
> 
> **Overview**
> Adds an **opt-in dedicated pipeline connection pool** via `PipelineReadBufferSize`, `PipelineWriteBufferSize`, and `PipelinePoolSize` on `Options` (and cluster/ring/sentinel option structs). When enabled, `processPipeline` and related paths use `withPipelineConn` so manual `Pipeline()` / autopipeline batches can use larger buffers and a smaller pool without sizing the whole client pool that way; unset options keep today’s behavior (main pool only). `PoolStats` gains optional `PipelineStats`, OTel pool registration includes the pipeline pool, and maintnotifications attaches a separate pool hook to that pool (with failover filtering stale pipeline connections).
> 
> **Autopipeline** is heavily reworked alongside this: default blocking config goes back to **one ordered batch stream** (`MaxConcurrentBatches: 1`, no forced `Unordered`); standalone clients default to **one shard** with optional `NumShards` and stricter ordering rules on the async face. Coalescing drops the fixed default accumulate window in favor of **expected-arrival / in-flight backpressure** (`awaitExpectedArrivals`, RTT-scaled silence gap), **8-way enqueue striping** where reordering is safe, merged multi-stripe flushes, and safer shutdown/permit waits on `context.Background`. **`Do`** no longer batches—it runs on a normal connection; **`Submit`** is rejected on the blocking face; **`AutoFuture.WaitContext`**, `IsBlocking`, and `Config()` are added. Tests cover coalescing behavior and the pipeline pool; CI pins Go **1.26.5**; `golang.org/x/sys` is a direct dep for stripe cache-line padding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da2f94fbda482de691ed959b9970707df63a3250. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/autopipeline.go
+++ b/autopipeline.go
@@ -315,15 +315,46 @@ type AutoPipeliner struct {
 
 // apShard is one queue + flusher. Its fields are touched only by enqueuing
 // goroutines (under mu) and by its own single flusher goroutine.
-type apShard struct {
-	ap *AutoPipeliner
+// apEnqueueStripes is how many enqueue stripes a shard runs when striping is
+// safe (unordered configs, and every blocking-face shard — a blocking caller
+// waits for each command, so stripes cannot reorder its stream). The
+// enqueue mutex is the hottest lock in the engine (128 concurrent callers on
+// one shard spend ~half their CPU in lock slow paths); striping the queue
+// spreads that contention while the flusher still drains every stripe into ONE
+// merged pipeline, so batches stay deep. Ordered shards always use a single
+// stripe: with several stripes a caller's consecutive commands can land in
+// stripes on opposite sides of an in-progress drain and execute out of order.
+const apEnqueueStripes = 8
 
+// apStripe is one striped slice of a shard's enqueue queue. Each stripe has
+// its own batch-completion signal so a drain can take stripes one lock at a
+// time; every batch taken in one drain completes together after the merged
+// pipeline executes. Padded so neighbouring stripes' mutexes do not share a
+// cache line.
+type apStripe struct {
 	mu       sync.Mutex
 	queue    []Cmder
 	queueLen atomic.Int32
-	curBatch *apBatch                // completion signal for currently-queued cmds
-	notify   chan struct{}           // buffered (cap 1) enqueue wake-up
-	sem      *internal.FIFOSemaphore // per-shard concurrent-batch budget
+	curBatch *apBatch // completion signal for currently-queued cmds
+	_        [40]byte // pad against false sharing
+}
+
+type apShard struct {
+	ap *AutoPipeliner
+
+	next    atomic.Uint32           // round-robin stripe pick (unordered mode)
+	stripes []apStripe              // 1 stripe when ordered, apEnqueueStripes when Unordered
+	notify  chan struct{}           // buffered (cap 1) enqueue wake-up
+	sem     *internal.FIFOSemaphore // per-shard concurrent-batch budget
+}
+
+// stripe picks the enqueue stripe for the next command: the single stripe in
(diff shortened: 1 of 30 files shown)


autopipeline.go
```diff
@@ -282,15 +354,60 @@ type AutoPipeliner struct {
...
-	curBatch *apBatch                // completion signal for currently-queued cmds
-	notify   chan struct{}           // buffered (cap 1) enqueue wake-up
-	sem      *internal.FIFOSemaphore // per-shard concurrent-batch budget
+	curBatch *apBatch // completion signal for currently-queued cmds
+	// Pad each stripe onto its own cache line(s). Without it, one stripe's hot
+	// fields (queueLen/curBatch) share a cache line with the NEXT stripe's
+	// contended mutex, so a lock-free counter bump on stripe i invalidates the
+	// line a different core is trying to lock stripe i+1 on — false sharing
+	// that measured ~16x on a contended microbenchmark. cpu.CacheLinePad is
+	// sized per GOARCH (64 B on x86-64/arm64, 128 B on ppc64, 256 B on s390x),
```
cxljs, line 383: Just a subtle point: Go uses 128 as the cache line size for arm64. 
src: https://cs.opensource.google/go/x/sys/+/master:cpu/cpu_arm64.go;l=12

autopipeline.go
cxljs, line 639: This probably also needs special handling for a few specific commands like shutdown/monitor.

ofekshenawa (approved):
LGTM!

cxljs (commented):
I left these comments a while ago and only realized today that they were stuck as a pending draft and never actually got submitted. Apologies for the delayed feedback.

autopipeline.go
ndyakov: Good point — and it's exactly why the padding uses `cpu.CacheLinePad` from golang.org/x/sys rather than a hardcoded `[64]byte`: that type is arch-aware (128 bytes on arm64 per the source you linked, 64 on amd64), so the stripes stay on separate lines on both. The x/sys dependency was added for this.

autopipeline.go
ndyakov: You were right — fixed in #3867 (45f9ba37). SHUTDOWN and the typed Monitor both rode pipeline batches: a batched SHUTDOWN killed the server before replying (batchmates all failed with EOF and the batch was retried against a dead server), and MONITOR rebinds the shared connection into a monitor stream, desyncing every reply behind it. Both now execute directly on their own pooled connection, like blocking commands, with a routing regression test. The rest of the typed surface audits clean: Quit is a deprecated panic stub, and the connection-state commands (AUTH/SELECT/HELLO/SWAPDB/CLIENT SETNAME) live on statefulCmdable, which the AutoPipeliner doesn't expose.

PR #3868, opened 2026-06-25
fix(hooks): data race on hook state during AddHook

saddamr3e:
`AddHook` rebuilds the composed hook chain while other goroutines run commands. The read paths (`processHook`, `processPipelineHook`, `processTxPipelineHook`, and the slice readers `withProcessHook`/`withProcessPipelineHook`) touched `hs.current`/`hs.slice` without synchronization, so calling `AddHook` on a live client races on the func values and the hook slice. `dialHook` already snapshotted `current` under a lock for the same reason; the command paths didn't.

Fold `slice` + `current` into an immutable `hooksState` published through a single `atomic.Pointer`. Every read path becomes one lock-free `Load`, so the `RWMutex` is gone from the hot path entirely. `AddHook` is copy-on-write under a plain `Mutex` used only by writers, which also fixes the previously-racy unlocked slice append across concurrent `AddHook` callers.

Includes a `-race` regression test running concurrent `AddHook` writers against the command-path getters and the `withProcessHook`/`withProcessPipelineHook` slice readers. Clean under `-count=3`, plus `go vet` and `golangci-lint`.

---

> [!NOTE]
> **Medium Risk**
> Touches core client hook wiring on every dial and command path; behavior should be equivalent but concurrency semantics changed from in-place updates to snapshot publishing.
> 
> **Overview**
> Fixes a **data race** when `AddHook` runs on a live client while other goroutines execute commands: writers used to mutate `hs.current` and append to `hs.slice` while `processHook` / pipeline getters read those values without synchronization, which could tear func pointers.
> 
> **`hooksMixin` now publishes immutable hook configuration via `atomic.Pointer[hooksState]`**. `AddHook` takes `hooksMu`, clones the prior snapshot (new hook slice + `rebuild()` of chained dial/process/pipeline hooks), then `Store`s the new state. Command-path readers (`dialHook`, `processHook`, pipeline getters, and `withProcess*` slice walks) only `Load()` the pointer—no read lock on the hot path.
> 
> `initHooks` and `clone` were updated to seed or share snapshots consistently with this model. A new **`TestHooksMixinCurrentConcurrent`** stress test runs concurrent `AddHook` writers against hook readers under `-race`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a32a69444b1d030fd2c4c2d350189aa3c636292. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/hooks_current_race_test.go
+++ b/hooks_current_race_test.go
@@ -15,10 +15,11 @@ func (noopRaceHook) ProcessPipelineHook(next ProcessPipelineHook) ProcessPipelin
 }
 
 // AddHook publishes a fresh hooks snapshot via hs.state, while the command-path
-// getters (processHook, processPipelineHook, processTxPipelineHook) and the
-// AddHook slice append all touch the same hook state. Exercise concurrent
-// AddHook writers against the command-path readers; run with -race to catch a
-// torn read of the published snapshot or an unsynchronized slice append.
+// getters (processHook, processPipelineHook, processTxPipelineHook), the slice
+// readers (withProcessHook, withProcessPipelineHook) and the AddHook slice
+// append all touch the same hook state. Exercise concurrent AddHook writers
+// against both reader groups; run with -race to catch a torn read of the
+// published snapshot or an unsynchronized slice append.
 func TestHooksMixinCurrentConcurrent(t *testing.T) {
 	var hs hooksMixin
 	hs.initHooks(hooks{})
@@ -49,7 +50,12 @@ func TestHooksMixinCurrentConcurrent(t *testing.T) {
 		}()
 	}
 
-	// Reader: the command-path getters Load hs.state lock-free.
+	noopProcess := func(context.Context, Cmder) error { return nil }
+	noopPipeline := func(context.Context, []Cmder) error { return nil }
+
+	// Reader: the command-path getters Load hs.state lock-free; the slice
+	// readers (withProcessHook, withProcessPipelineHook) iterate the published
+	// snapshot's hook slice.
 	go func() {
 		defer wg.Done()
 		defer close(done)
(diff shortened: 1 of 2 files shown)


ndyakov (changes requested):
@saddamr3e The analysis is correct, this is racy as is AddHook for cluster client on `hs.slice`.

However, this is a hot path (processHook, processPipelineHook). Adding even the read mutex lock is not ideal here if we can skip it. What we could do is use atomic pointer.

If you want to address this, maybe fold `slice` + `current` into an immutable snapshot published via a single
`atomic.Pointer`. Every read path becomes one lock-free `Load`, the `RWMutex`
comes out entirely, both races are fixed. `AddHook` is copy-on-write and
serialized by a plain `Mutex` used only by writers (which also makes concurrent
`AddHook` safe — today the unlocked append is already racy across `AddHook`
callers).

saddamr3e:
Reworked along those lines. Folded slice + current into an immutable hooksState published through a single atomic.Pointer. The read paths (dial/process/pipeline/txPipeline) are now a lock-free Load and the RWMutex is gone from the hot path entirely. AddHook is copy-on-write under a plain Mutex used only by writers, so the unlocked-append race across concurrent AddHook callers is fixed too.

Updated the -race test to run concurrent AddHook writers against the command-path readers. Clean under -count=3, plus go vet and golangci-lint.

ndyakov (dismissed):
Thank you @saddamr3e!

Only nits: the race test doesn't cover the       
  `withProcessHook/withProcessPipelineHook` slice readers, and the title/description are stale

saddamr3e:
Both addressed. The race test now also drives withProcessHook/withProcessPipelineHook against the AddHook writers, and updated the title and description to match the atomic snapshot approach. Still clean under -count=3 with -race.

ndyakov (approved):
Looks good to me, fixed the failing test in another PR - https://github.com/redis/go-redis/pull/3872

PR #3870, opened 2026-06-28
Add TS.NRANGE and TS.NREVRANGE Commands

ofekshenawa:
Adds client support for `TS.NRANGE` and `TS.NREVRANGE`, two new RedisTimeSeries commands introduced in Redis 8.10 that query an explicit list of time-series keys and return a timestamp-pivoted table where each row is `[timestamp, [value_per_key]]`.

### Changes

- **New result type** `TSNRangePivotRow` with a `Values []float64` slice (one entry per input key, NaN for missing samples)
- **New option structs** `TSNRangeOptions` / `TSNRevRangeOptions` — aggregation takes one `Aggregator` per key, emitted as separate wire tokens (not comma-joined like existing range commands)
- **4 new methods** on `cmdable` and `TimeseriesCmdable`: `TSNRange`, `TSNRangeWithArgs`, `TSNRevRange`, `TSNRevRangeWithArgs`
- **CI updated** to include `8.10.x` in the benchmark and test matrices, mapped to `redislabs/client-libs-test:8.10-m01-int`

---

> [!NOTE]
> **Low Risk**
> Additive Timeseries API and reply parsing with client-side validation; no changes to auth, persistence, or existing command behavior.
> 
> **Overview**
> Adds **Redis 8.10+** client support for **`TS.NRANGE`** and **`TS.NREVRANGE`**, which query an explicit list of time-series keys and return a **timestamp-pivoted** table (`[timestamp, [value per key]]`) instead of per-key series maps.
> 
> New **`TSNRangePivotRow`** / **`TSNRangePivotRowSliceCmd`** parse that wire format (missing points as **NaN**), with **`CmdTypeTSNRangePivotRowSlice`** wired into **`ExtractCommandValue`**. **`TSNRange`**, **`TSNRangeWithArgs`**, **`TSNRevRange`**, and **`TSNRevRangeWithArgs`** build `TS.NRANGE` / `TS.NREVRANGE` with key count prefix and **`SetFirstKeyPos(2)`** for cluster routing.
> 
> **`TSNRangeOptions`** / **`TSNRevRangeOptions`** mirror other range modifiers (filters, count, aggregation buckets, etc.), but **`AGGREGATION` requires exactly one `Aggregator` per key**, emitted as **separate wire tokens**—validated client-side before send. Integration tests cover pivot ordering, NaN gaps, options, duplicate keys, and aggregator validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 108f7cd058146039b9ad6c6c14b027e184876527. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/timeseries_commands.go
+++ b/timeseries_commands.go
@@ -40,6 +40,10 @@ type TimeseriesCmdable interface {
 	TSMRevRangeWithArgs(ctx context.Context, fromTimestamp int, toTimestamp int, filterExpr []string, options *TSMRevRangeOptions) *MapStringSliceInterfaceCmd
 	TSMGet(ctx context.Context, filters []string) *MapStringSliceInterfaceCmd
 	TSMGetWithArgs(ctx context.Context, filters []string, options *TSMGetOptions) *MapStringSliceInterfaceCmd
+	TSNRange(ctx context.Context, keys []string, fromTimestamp interface{}, toTimestamp interface{}) *TSNRangePivotRowSliceCmd
+	TSNRangeWithArgs(ctx context.Context, keys []string, fromTimestamp interface{}, toTimestamp interface{}, options *TSNRangeOptions) *TSNRangePivotRowSliceCmd
+	TSNRevRange(ctx context.Context, keys []string, fromTimestamp interface{}, toTimestamp interface{}) *TSNRangePivotRowSliceCmd
+	TSNRevRangeWithArgs(ctx context.Context, keys []string, fromTimestamp interface{}, toTimestamp interface{}, options *TSNRevRangeOptions) *TSNRangePivotRowSliceCmd
 }
 
 type TSOptions struct {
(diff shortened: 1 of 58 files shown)


timeseries_commands.go
```diff
@@ -1106,3 +1134,259 @@ func (c cmdable) TSMGetWithArgs(ctx context.Context, filters []string, options *
...
+	if latest {
+		args = append(args, "LATEST")
+	}
+	if filterByTS != nil {
+		args = append(args, "FILTER_BY_TS")
+		for _, ts := range filterByTS {
+			args = append(args, ts)
+		}
+	}
+	if filterByValue != nil {
```
ndyakov: should we check for the len of those slices as well?

ndyakov (commented):
Left a question, other than that looks good. Thank you for fixing the test guards.

ndyakov (dismissed):
Looks good to me

timeseries_commands.go
```diff
@@ -273,6 +277,30 @@ type TSMGetOptions struct {
...
+	Aggregators     []Aggregator
+	BucketDuration  int
+	BucketTimestamp interface{}
+	Empty           bool
+}
+
+type TSNRevRangeOptions struct {
+	Latest          bool
+	FilterByTS      []int
+	FilterByValue   []int
```
ndyakov: should be float64

timeseries_commands.go
```diff
@@ -273,6 +277,30 @@ type TSMGetOptions struct {
 	SelectedLabels []interface{}
 }
 
+type TSNRangeOptions struct {
+	Latest          bool
+	FilterByTS      []int
+	FilterByValue   []int
```
ndyakov: FilterByValue should be min,max. If we would like to keep it a slice, lets make sure it is exactly two elements or we can define a type to make it {min float64, max float64}

timeseries_commands.go
ndyakov: only for byvalue, ts is int.

ndyakov (changes requested):
requested some changes on the types, overall looks good

ndyakov (approved):
Looks good, thank you @ofekshenawa!

PR #3874, opened 2026-07-01
fix(proto): don't over-read nil replies in Reader.Discard

saddamr3e:
Reader.Discard on a nil bulk string, feeding the reply after it:

    $-1\r\n$5\r\nhello\r\n  ->  DiscardNext() then ReadString()
    redis: invalid reply: "\r\n"   // want "hello"

replyLen returns (0, Nil) for $-1/=-1/!-1, but the guard is `err != nil && err != Nil`, so Nil falls through to the string branch and runs Discard(n+2)=Discard(2). A nil reply is only its header line (already consumed by readLine), so those 2 bytes come out of the next reply and desync the connection. On a pooled conn that means one command reads bytes belonging to another.

Reached on the read path when a RESP3 attribute frame is skipped (ReadLine -> RespAttr -> Discard/DiscardNext) and one of its values is a nil bulk string. readRawReplyBuf and readRawReplyWriteTo already early-return on err==Nil; Discard was the only consumer that did not. Fix matches them.

---

> [!NOTE]
> **Medium Risk**
> Touches low-level RESP stream alignment in the shared proto reader; incorrect behavior could corrupt pooled connections, but the change is narrow and covered by new regression tests.
> 
> **Overview**
> **`Reader.Discard`** no longer treats nil-length RESP headers (`$-1`, `=-1`, `!-1`, `*-1`, `%-1`) as bulk payloads. When `replyLen` returns `Nil`, it returns immediately instead of calling `Discard(2)`, which was stealing bytes from the next reply and desyncing pooled connections.
> 
> This aligns **`Discard`** with **`readRawReplyBuf`** / **`readRawReplyWriteTo`**, which already handled `Nil` correctly. The bug showed up when skipping RESP3 attributes whose values are nil bulk strings (`ReadLine` → `RespAttr` → `Discard`).
> 
> Tests cover discarding each nil header type followed by a bulk string, and reading through an attribute with a nil value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb3a513284088b954984ca7bc7f5068c1a6d4e26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/proto/reader.go
+++ b/internal/proto/reader.go
@@ -738,7 +738,15 @@ func (r *Reader) Discard(line []byte) (err error) {
 	}
 
 	n, err := replyLen(line)
-	if err != nil && err != Nil {
+	if err != nil {
+		if err == Nil {
+			// A nil reply ($-1, =-1, !-1, *-1, %-1) carries no payload; the
+			// header line was already consumed by readLine, so there is
+			// nothing to discard. Falling through would Discard(n+2)==2 bytes
+			// that belong to the next reply and desync the stream, matching
+			// how readRawReplyBuf/readRawReplyWriteTo already treat Nil.
+			return nil
+		}
 		return err
 	}
 
--- a/internal/proto/reader_test.go
+++ b/internal/proto/reader_test.go
@@ -347,6 +347,44 @@ func TestReader_ReadStringInto_BufferTooSmall(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


ndyakov (approved):
Good defensive fix, thank you @saddamr3e

PR #3877, opened 2026-07-05
fix(proto): avoid int overflow skipping map/attr bodies on read

saddamr3e:
The RESP3 map and attribute branches of Discard, readRawReplyBuf and readRawReplyWriteTo read key/value pairs by looping over n*2 elements, where n is the element count taken straight from the reply header (replyLen only rejects n < -1). A server sending a count above MaxInt/2, e.g. %4611686018427387904, makes n*2 overflow to a negative number, so the loop bound is negative, the body is never read, and the call returns nil. The header line was already consumed but the map bytes stay in the buffer, and since no error comes back the connection goes back to the pool and the next command reads those leftover bytes as its own reply. I hit it while checking the aggregate paths after the nil-discard fix and reproduced it against a crafted reply. Iterating over n pairs and reading two elements each keeps valid replies identical while making an oversized count fail with a read error instead of silently desyncing. The change stays inside these three readers so nothing on the call path shifts.

---

> [!NOTE]
> **Medium Risk**
> Touches low-level RESP parsing on pooled connections where a prior bug could silently misalign streams; the change is narrow and preserves behavior for valid replies.
> 
> **Overview**
> Fixes a **RESP3 map/attribute parsing bug** where `Discard`, `readRawReplyBuf`, and `readRawReplyWriteTo` advanced through key/value data using an `n*2` loop bound. When the advertised pair count is above `MaxInt/2`, that product overflows to a negative bound, the body is skipped with **no error**, and leftover bytes corrupt the next reply on a pooled connection.
> 
> The three code paths now iterate **`n` pairs** (two nested reads per pair), matching how `readMap` already works, so valid frames behave the same while absurd counts fail during reads instead of silently desyncing.
> 
> Adds **`TestReader_Discard_MapCountOverflow`** to confirm normal map discard stays aligned and crafted `%4611686018427387904` / `|4611686018427387904` headers return an error from `DiscardNext`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c929164c0b52edcf86ba6c49843024b3b831814c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/proto/reader.go
+++ b/internal/proto/reader.go
@@ -763,8 +763,14 @@ func (r *Reader) Discard(line []byte) (err error) {
 		}
 		return nil
 	case RespMap, RespAttr:
-		// Read key & value.
-		for i := 0; i < n*2; i++ {
+		// Iterate over the n key/value pairs rather than n*2 elements: a count
+		// above MaxInt/2 makes n*2 overflow to a negative loop bound, which
+		// would skip the body entirely and return nil, leaving the map bytes in
+		// the stream for the next reply to consume (a silent desync).
+		for i := 0; i < n; i++ {
+			if err = r.DiscardNext(); err != nil {
+				return err
+			}
 			if err = r.DiscardNext(); err != nil {
 				return err
 			}
@@ -857,10 +863,12 @@ func (r *Reader) readRawReplyBuf(buf []byte) ([]byte, error) {
 			}
 			return buf, err
 		}
-		for i := 0; i < n*2; i++ {
-			buf, err = r.readRawReplyBuf(buf)
-			if err != nil {
-				return buf, err
+		for i := 0; i < n; i++ {
+			for pair := 0; pair < 2; pair++ {
+				buf, err = r.readRawReplyBuf(buf)
+				if err != nil {
+					return buf, err
+				}
 			}
 		}
 		return buf, nil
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
This is an interesting finding. Were you able to hit it with an actual database @saddamr3e? The fix looks good. I am just curious how were you able to hit this.

saddamr3e:
Not against a live server, no. A normal Redis won't advertise a count that large, so hitting it needs a malicious/MITM server or a corrupted stream. I reproduced it by feeding a crafted RESP3 frame straight through the reader:

```go
input := "%4611686018427387904\r\n+k\r\n+v\r\n:42\r\n"
r := NewReader(strings.NewReader(input))
r.DiscardNext()          // nil before the fix
rep, _ := r.ReadReply()  // "k" instead of 42
```

The count comes straight from replyLen (which only rejects n < -1), and n*2 overflows negative so the skip loop never runs. DiscardNext returns nil with the map body still buffered, and the next ReadReply picks up "k" from the leaked bytes instead of the following :42. Found it while auditing the aggregate skip paths after the nil-discard fix. The committed test covers the same for the % and | headers.

PR #3884, opened 2026-07-07
fix(command): drain unexpected reply frames in readReply parsers

### LINKED ISSUE #1: add license please
under which license is this work distributed? is it safe to use it in greater projects?

saddamr3e:
1. MapStringSliceInterfaceCmd.readReply peeks the reply type, but the switch has no default, so a RESP3 null, RESP2 nil bulk, set, or any other unexpected type returns nil with that frame still buffered.
2. FunctionStatsCmd.readEngines and the LCS idx path switch on the field key inside a fixed map with no default, so an unknown field leaves its value unread.

Both cases return a nil error, so the connection goes back to the pool desynced and the next command reads the leftover bytes as its own reply. Fixed #1 to error on an unexpected type (isBadConn then drops the conn, matching sibling type switches like XStreamSliceCmd), and #2/#3 to DiscardNext the unknown value (matching the other field parsers like ClusterShards/XInfoStream). Tests drive each path through proto.NewReader.

---

> [!NOTE]
> **Medium Risk**
> Changes core RESP parsing on pooled connections; mis-handling could still corrupt subsequent commands, but the fix is narrowly scoped with regression tests.
> 
> **Overview**
> Fixes **pooled connection desync** when several `readReply` parsers stopped early without consuming the full RESP stream.
> 
> **`MapStringSliceInterfaceCmd`** now returns an error on unexpected reply types (after peeking) instead of succeeding with a partially buffered frame, so bad connections can be dropped rather than reused.
> 
> **`FunctionStatsCmd.readEngines`** and **`LCSCmd`** (IDX map path) **`DiscardNext()`** on unknown map keys and propagate **`ReadString`** errors on engine fields, keeping the reader aligned when Redis adds new fields.
> 
> Adds **`command_desync_test.go`** with synthetic RESP fixtures asserting errors vs. stream alignment for each path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8300b75d80c0653f9bc0b0af0135b8730e752ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -2384,6 +2384,11 @@ func (cmd *MapStringSliceInterfaceCmd) readReply(rd *proto.Reader) (err error) {
 				cmd.val[key] = append(cmd.val[key], data)
 			}
 		}
+	default:
+		// Any other reply type leaves the peeked frame unread. Returning nil
+		// here would put the connection back in the pool with those bytes
+		// buffered, so the next command reads them as its own reply.
+		return fmt.Errorf("redis: can't parse map-string-slice-interface reply: unexpected type %c", readType)
 	}
 
 	return nil
@@ -6620,6 +6625,10 @@ func (cmd *FunctionStatsCmd) readEngines(rd *proto.Reader) ([]Engine, error) {
 				engine.LibrariesCount, err = rd.ReadInt()
 			case "functions_count":
 				engine.FunctionsCount, err = rd.ReadInt()
+			default:
+				// Unknown field: drain its value so the reader stays aligned
+				// with the rest of the reply.
+				err = rd.DiscardNext()
 			}
 			if err != nil {
 				return nil, err
@@ -6825,6 +6834,12 @@ func (cmd *LCSCmd) readReply(rd *proto.Reader) (err error) {
 			case "len":
 				// read match length
 				if lcs.Len, err = rd.ReadInt(); err != nil {
+					return err
+				}
+			default:
+				// Unknown field: drain its value so the reader stays aligned
+				// with the rest of the reply.
+				if err = rd.DiscardNext(); err != nil {
 					return err
 				}
 			}
--- a/command.go
+++ b/command.go
@@ -6620,6 +6620,9 @@ func (cmd *FunctionStatsCmd) readEngines(rd *proto.Reader) ([]Engine, error) {
 
 		for i := 0; i < 2; i++ {
 			key, err := rd.ReadString()
+			if err != nil {
+				return nil, err
+			}
 			switch key {
 			case "libraries_count":
 				engine.LibrariesCount, err = rd.ReadInt()
--- a/command_desync_test.go
+++ b/command_desync_test.go
@@ -1,0 +1,74 @@
+package redis
+
+import (
+	"context"
+	"strings"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


ndyakov:
@saddamr3e can you check the linter and I can review the PR

saddamr3e:
Fixed. The ineffassign hit was in readEngines: the ReadString err wasn't checked before the switch reassigned err. Added the check right after, matching the LCS field loop. Lint is clean now.

command.go
```diff
@@ -6615,11 +6620,18 @@ func (cmd *FunctionStatsCmd) readEngines(rd *proto.Reader) ([]Engine, error) {
...
+			}
 			switch key {
 			case "libraries_count":
 				engine.LibrariesCount, err = rd.ReadInt()
 			case "functions_count":
 				engine.FunctionsCount, err = rd.ReadInt()
+			default:
+				// Unknown field: drain its value so the reader stays aligned
+				// with the rest of the reply.
+				err = rd.DiscardNext()
```
ndyakov, line 6634: does it make sense to log those?

command.go
```diff
@@ -6827,6 +6839,12 @@ func (cmd *LCSCmd) readReply(rd *proto.Reader) (err error) {
 				if lcs.Len, err = rd.ReadInt(); err != nil {
 					return err
 				}
+			default:
+				// Unknown field: drain its value so the reader stays aligned
+				// with the rest of the reply.
+				if err = rd.DiscardNext(); err != nil {
+					return err
+				}
```
ndyakov, line 6847: we have similar cases for other comment, for which we are not parsing the whole response. Would you like to discuss a proper way to return those fields to the user? You can find me on discord or we can do it here in the PR. This is not a blocker for the PR, rather another improvement that I think can be beneficial for some users.

ndyakov (approved):
Overall good, opened two questions that we can continue outside of this PR if we decide not to address them now.

command.go
saddamr3e: None of the existing default-discard sites in command.go log (there are ~14 doing the same silent drain), and readReply has no ctx to hand internal.Logger.Printf. It would also fire on every reply once a newer server adds a field, which gets noisy fast. If we want visibility here I'd rather add it across all the discard sites in one pass, but happy to do it in this PR if you prefer.

command.go
saddamr3e: Here works for me. The shape I'd try: instead of DiscardNext, read the unknown value with rd.ReadReply() into something like a map[string]interface{} exposed via an accessor on the cmd, so typed commands stop silently dropping fields newer servers add. Open questions are whether it should be opt-in to avoid the allocation on the common path, and it should land consistently across all the existing discard sites, so probably its own issue/PR. I can write that up if you want to go that way.

command.go
ndyakov: let's do it in a separate pr. i was thinking of adding something like `Other` map field to the responses that may return something that is not parsed and populate it with them. I will let you decide how to approach this if you would like to work on it.

command.go
ndyakov: Exactly what i was thinking, let's do it in a separate PR.

PR #3888, opened 2026-07-09
fix(command): WaitAOF should return *IntSliceCmd, not *IntCmd

### LINKED ISSUE #3887: WaitAOF returns *IntCmd but WAITAOF replies with a two-element array
## Expected Behavior

`WaitAOF` should return the two integers; WAITAOF replies with `[numlocal, numreplicas]`, so `Result()` succeeds and the caller can read both counts.

## Current Behavior

`WaitAOF` is declared as `*IntCmd`, but WAITAOF replies with an array of two integers. `IntCmd.readReply` calls `ReadInt`, which can't parse an array header, so `Result()` always fails with `redis: can't parse int reply: "*2"`. 

## Possible Solution

Return `*IntSliceCmd` from `WaitAOF` so both counts are parsed. This is a breaking return-type change. 

## Steps to Reproduce

1. Run Redis 7.2+ (WAITAOF exists since 7.2).
2. `rdb.Set(ctx, "k", "v", 0)`
3. `_, err := rdb.WaitAOF(ctx, 0, 0, 0).Result()`
4. `err` is `redis: can't parse int reply: "*2"` instead of returning `[]int64{1, 0}`.

## Context (Environment)

Affected versions: v9.19.0 and current master.

## Detailed Description

Per the WAITAOF docs (https://redis.io/docs/latest/commands/waitaof/), the reply is an array of two integers: the number of local instances and the number of replicas that fsynced the writes. Because `WaitAOF` returns `*IntCmd`, that array can't be parsed. `*IntSliceCmd` reads both values;

CipherN9:
### Problem

`WaitAOF` returns `*IntCmd`, but WAITAOF replies with a two-element array `[numlocal, numreplicas]`. So `Result()` always fails with `redis: can't parse int reply: "*2"`, and the command is unusable.

### Fix

Returns `*IntSliceCmd` so both counts are parsed. 

Also un-skips the WaitAOF test. Earlier it was skipped and its assertions did not match its own comment. Now it checks the reply parses into a two-element slice.

Fixes [#3887](https://github.com/redis/go-redis/issues/3887)

**BREAKING CHANGE:** `WaitAOF` now returns `*IntSliceCmd` instead of `*IntCmd`.

---

> [!NOTE]
> **Medium Risk**
> Breaking change to WaitAOF’s return type affects any caller using *IntCmd; behavior is otherwise a narrow command-wrapper fix with low runtime risk.
> 
> **Overview**
> **`WaitAOF`** now returns **`*IntSliceCmd`** instead of **`*IntCmd`**, matching Redis’s WAITAOF reply `[numlocal, numreplicas]` so `Result()` parses both counts instead of failing on a multi-bulk reply.
> 
> The integration test is re-enabled: it calls WAITAOF with **`numLocal=0`** (avoids AOF-disabled errors on a standalone instance), expects a **two-element** slice, and checks the replica count is **0** when no replicas are present, along with the timeout behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03baf7cbe6481957971ed6ed73a0f2d337a62fd2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -296,8 +296,8 @@ func (c cmdable) Wait(ctx context.Context, numSlaves int, timeout time.Duration)
 	return cmd
 }
 
-func (c cmdable) WaitAOF(ctx context.Context, numLocal, numSlaves int, timeout time.Duration) *IntCmd {
-	cmd := NewIntCmd(ctx, "waitAOF", numLocal, numSlaves, int(timeout/time.Millisecond))
+func (c cmdable) WaitAOF(ctx context.Context, numLocal, numSlaves int, timeout time.Duration) *IntSliceCmd {
+	cmd := NewIntSliceCmd(ctx, "waitAOF", numLocal, numSlaves, int(timeout/time.Millisecond))
 	cmd.setReadTimeout(timeout)
 	_ = c(ctx, cmd)
 	return cmd
--- a/commands_test.go
+++ b/commands_test.go
@@ -108,13 +108,14 @@ var _ = Describe("Commands", func() {
 
 		It("should WaitAOF", func() {
 			const waitAOF = 3 * time.Second
-			Skip("flaky test")
 
-			// assuming that the redis instance doesn't have AOF enabled
+			// No replicas here, so WAITAOF blocks until timeout and returns
+			// the two counts [numlocal, numreplicas]. numLocal=0 skips AOF.
 			start := time.Now()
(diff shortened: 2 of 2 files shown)


ndyakov (approved):
@CipherN9 although this is a breaking change - it is a bugfix. Looks like previously the command was not usable.

PR #3891, opened 2026-07-10
feat(pool): Dial limiter

### LINKED ISSUE #3890: Feature Request: Add built-in connection creation rate limiter with idle-retry capability for burst traffic
# Feature Request: Add built-in connection creation rate limiter with idle-retry capability for burst traffic

## Background and Problem
Currently, when burst traffic arrives and all idle connections are exhausted, the connection pool allows all pending requests to initiate TCP dialing concurrently, as long as `TotalConns < PoolSize`. This causes a **connection creation storm**:
- Hundreds of goroutines dial Redis at the same instant, causing TCP handshake spikes on both client and server side.
- In scenarios where connection creation cost (e.g. ~10ms per dial) is much higher than per-request execution cost (e.g. ~1ms), creating large numbers of connections is far less efficient than waiting briefly and reusing existing connections.
- The pool fills to `PoolSize` almost immediately, even for very short traffic spikes, after which most connections sit idle and eventually expire, creating periodic destroy-rebuild cycles.

## Why a custom Dialer is not sufficient
The common community workaround is wrapping `Dialer` with a `rate.Limiter`, but this approach has a critical limitation:

Once a request passes the idle-check and enters the "new connection" branch, **it will never go back to check for newly returned idle connections**, even while waiting for a dial token. This means:
1. Even if another request returns a connection to the idle pool a millisecond later, the waiting request will still create a brand new connection, wasting the opportunity for reuse.
2. The rate limiter only slows down connection creation, it cannot redirect requests back to the idle queue, which is the more efficient path for short-lived requests.

This is suboptimal for workloads where reuse is cheaper than creation.

## Proposed Solution
Add a **built-in connection creation rate limiter** to the connection pool with the following behavior:

1. Add a new option, e.g. `DialRateLimit` / `DialRateBurst`, to control the maximum number of new connections created per second.
2. When idle connections are exhausted and the pool would normally create a new connection, first acquire a token from the rate limiter.
3. **If a token is not immediately available, instead of blocking in the dial path, requeue the request to wait for an idle connection** (the same `waitForConn` path used when `TotalConns == PoolSize`).
4. When a connection is returned to the idle pool, wake up waiting requests as usual; if a waiting request can take the idle connection, it never needs to dial at all.
5. Only if the request waits longer than some threshold (or if idle connections remain unavailable) does it proceed to create a new connection.

In short: **when dialing would be throttled, prefer waiting for reuse over waiting to create**.

## Expected Benefits
- Smooths out connection creation spikes during traffic bursts, eliminating dial storms.
- Significantly improves overall efficiency for short-request workloads, where reuse is much cheaper than creation.
- Reduces unnecessary connection churn caused by short traffic spikes.
- Avoids the "can't go back to idle queue" limitation of custom Dialer-based rate limiting.
- Remains fully backward compatible: disabled by default, only active when explicitly configured.

## Additional Considerations
- The implementation should work with both single-node and cluster clients.
- The limiter should interact correctly with existing `PoolTimeout` semantics.
- Metrics (`PoolStats`) should expose rate-limited dial attempts for observability.
- This complements the existing `ConnMaxLifetimeJitter` feature: jitter prevents mass-expiration storms, while this feature prevents mass-creation storms.

Thank you for considering this feature. I'd be happy to discuss implementation details or contribute a draft PR if the maintainers think this is worthwhile.

ndyakov:
Adds opt-in DialRateLimit/DialRateBurst. When idle conns are exhausted and a
dial would be throttled, the request parks and waits for a returned idle conn
(preferring reuse) instead of dialing; only after PoolTimeout does it create a
new connection. Exposes PoolStats.RateLimitedDials. 

Closes #3890.

---

> [!NOTE]
> **Medium Risk**
> Substantial new concurrency in `ConnPool.Get()` and min-idle refill when the feature is on, though default-off limits blast radius; mis-tuning can add tail latency or soft-limit bypass after PoolTimeout.
> 
> **Overview**
> Adds **opt-in `DialRateLimit` / `DialRateBurst`** so the pool can pace new dials during bursts instead of dialing up to `PoolSize` immediately (addresses #3890). When enabled, a miss path that would dial first tries a token-bucket limiter; if throttled, **`Get()` parks on a FIFO wait queue** and prefers **reusing an idle connection** returned via `Put`, with **`PoolTimeout` (from `Get()` entry) as a soft escape** that still allows an unpaced dial. Idle hits stay unchanged (nil limiter check only).
> 
> **`MinIdleConns` refill** is integrated: workers take dial tokens before pool turns, use **`refillPending`** so token-waiting workers do not inflate `MaxActiveConns` / `poolSize`, and **`Close()` wakes parked waiters**. New **`PoolStats.RateLimitedDials`** (and cluster aggregation) exposes throttle volume. Options propagate through standalone, cluster, sentinel, universal clients and URL `dial_rate_limit` / `dial_rate_burst`.
> 
> CI pins **Go 1.26.5** in govulncheck so runner-cached older patches do not fail the job on stdlib vulns unrelated to the PR.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40a9a9f7f13ef0aa860e91bc335e9d4f6c52775c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ndyakov:
### DialRateLimit — measured latency / throughput / connection tradeoff

Smoke benchmark against a real Redis (local, Apple M4 Max, RESP3): **200 goroutines × 100 PINGs** arriving as one burst, `PoolSize=50`, `PoolTimeout=3s`. Every request in flight simultaneously, i.e. sustained saturation — the *worst case* for the limiter, since there are never enough idle connections for everyone.

| config | conns created | throttled Gets | timeouts | errors | ops/s | avg | p50 | p99 | max |
|---|---|---|---|---|---|---|---|---|---|
| disabled | 50 | 0 | 0 | 0 | 51,267 | 3.9ms | 3.8ms | 5.6ms | 13ms |
| rate=50/s | 50 | 0 | 0 | 0 | 45,570 | 4.4ms | 4.2ms | 8.6ms | 19ms |
| rate=20/s | **33** | 8,147 | 0 | 0 | 30,616 | 6.4ms | 6.2ms | 12.8ms | 24ms |
| rate=5/s | **13** | 5,923 | 0 | 0 | 12,386 | 15.8ms | 13.9ms | 53.6ms | 260ms |

Takeaways:

- **Disabled = unchanged baseline**: the burst instantly dials the pool to 50 connections (the "dial storm" from #3890). Hot-path benchmarks vs master show no significant delta when the feature is off (interleaved benchstat, p≥0.16, allocs byte-identical).
- **rate ≥ burst demand (50/s)**: behaves like disabled — the token bucket (default burst = rate) absorbs the storm, park machinery never engages.
- **rate=20/s**: connection count drops 34% (33 vs 50); cost is ~2× p99 under this saturated workload.
- **rate=5/s**: 74% fewer connections (13 vs 50). Under *sustained* saturation that necessarily costs throughput (fewer pipes) and tail latency (max 260ms = park waits); zero pool timeouts and zero errors — every request completed via reuse or the PoolTimeout escape hatch.

Note on interpretation: sustained saturation makes the tradeoff look maximal. The feature's target scenario is a *short* burst on an otherwise steady workload — there the pool stops inflating to `PoolSize` for a spike that reuse could absorb, and steady-state latency is untouched because the limiter is only consulted on the dial slow path (idle hits never see it).

Rule of thumb: set `DialRateLimit` at or above your expected steady-state connection-creation demand; it should throttle storms, not normal growth. `DialRateBurst` (default = rate) tunes how large a spike passes before pacing engages.

internal/pool/pool.go
```diff
@@ -525,7 +560,14 @@ func (p *ConnPool) checkMinIdleConns() {
 					}
 				}()
 
-				err := p.addIdleConn()
+				// Min-idle refill dials count against the dial rate limit too:
```
ofekshenawa: Here the refill worker takes a pool slot and then waits for its rate-limit token, holding the slot the whole time it waits. So if MinIdleConns and DialRateLimit are both on, the workers can end up sitting on all the slots while they wait, and real Get() calls have nothing left and time out. Could we grab the token first and then the slot?

ofekshenawa (commented):
Left a small comment there. Other than that, all good!

internal/pool/pool.go
```diff
@@ -509,27 +557,75 @@ func (p *ConnPool) checkMinIdleConns() {
...
+						p.dialLimiter.refund()
+						p.refillPending.Add(-1)
+						return
+					}
+					turnHeld = true
+					// Token and turn in hand: the dial starts now, so convert
+					// the reservation into live counters (bump live first so a
+					// concurrent check never sees both sides low and
+					// overspawns).
+					p.poolSize.Add(1)
```
ofekshenawa, line 623: While min idle jobs wait for a rate limit token, normal requests can fill the pool, so maybe recheck capacity before those jobs create another connection?

internal/pool/pool.go
```diff
@@ -869,84 +969,168 @@ func (p *ConnPool) getConn(ctx context.Context) (cn *Conn, err error) {
...
+		}
+		if escape {
+			// Waited past PoolTimeout without a reusable connection or a dial
+			// token. Re-check the idle pool one final time: a connection may
+			// have been returned in the same instant the park timer fired (the
+			// select can win the timer race over the wakeup signal). If so,
+			// loop back to grab it — with the deadline passed, the next
+			// waitForDialSlot escapes immediately, so this cannot loop forever.
+			if p.hasAcquirableIdleConn() {
+				continue
```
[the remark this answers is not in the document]
ofekshenawa: A hook rejected idle connection can be retried in a tight loop after the timeout, so maybe limit the retry and check for context cancellation.

ndyakov:
cc @vlady-kotsev this is an interesting task you can pick, ignoring my work so far.

PR #3893, opened 2026-07-10
fix: Override the ring client Publish to write to the correct shard's topic

### LINKED ISSUE #3889: Ring client's Publish does not publish to the correct shard

## Expected Behavior

When I instantiate a `Ring` client and call `Publish`, the message would get routed to the correct shard based on the `channel` just as `Subscribe` properly calculates its shard on the channel to subscribe properly.

## Current Behavior

`publish` and `subscribe` are listed under `keylessCommands` and `cmdFirstKeyPosWithInfo` returns 0 when publish is called. That means it will ultimately get routed via `Ring.sharding.Random` under `cmdShard` instead of falling down into `c.sharding.GetByKey(firstKey)`. 

## Possible Solution

I'm not positive it's ideal but we could override `Publish` in the ring client similarly to how `Subscribe` is overridden to get the shard to subscribe to directly. 

```
func (c *Ring) Publish(ctx context.Context, channel string, message interface{}) *IntCmd {
	shard, err := c.sharding.GetByKey(channel)
	if err != nil {
		cmd := NewIntCmd(ctx, "publish", channel, message)
		cmd.SetErr(err)
		return cmd
	}
	return shard.Client.Publish(ctx, channel, message)
}
```

This would by pass the check on the keyless entry specifically for the ring shard while leaving the other clients alone as they don't need the same type of sharding on publish. 

## Steps to Reproduce

This can be covered under a test case. 

```
var _ = Describe("Ring Publish/Subscribe sharding", func() {
	var ring *redis.Ring

	BeforeEach(func() {
		ring = redis.NewRing(redisRingOptions())
	})

	AfterEach(func() {
		Expect(ring.Close()).NotTo(HaveOccurred())
	})

	It("routes Publish to the same shard as Subscribe for a given channel", func() {
		// Use many channels to test to make sure there isn't flakiness in the case this would fail.
		// Because the default sharding (Ring.sharding.Random) is determined by a random function,
		// we want to make sure there isn't an inconsistency in the failure case
		const numChannels = 20

		for i := 0; i < numChannels; i++ {
			channel := fmt.Sprintf("ring-pubsub-test-channel-%d", i)

			pubsub := ring.Subscribe(ctx, channel)

			_, err := pubsub.Receive(ctx)

			n, err := ring.Publish(ctx, channel, "hello").Result()
			Expect(n).To(Equal(int64(1)), "channel %s: message should reach the subscriber's shard", channel)

			Eventually(func() interface{} {
				msg, _ := pubsub.ReceiveTimeout(ctx, time.Second)
				m, _ := msg.(*redis.Message)
				if m == nil {
					return nil
				}
				return m.Payload
			}, 5*time.Second).Should(Equal("hello"))

			Expect(pubsub.Close()).NotTo(HaveOccurred())
		}
	})
})
```

With the proposed fix, this will pass. Without it, this fails with the following message, effectively indicating there were no subscribers at the topic the message was published to.
```
  [FAILED] channel ring-pubsub-test-channel-1: message should reach the subscriber's shard
  Expected
      <int64>: 0
  to equal
      <int64>: 1
  In [It] at: /Users/dave.kindel/dev/go-redis/ring_test.go:835
```

## Context (Environment)

We are using the ring client currently and are expanding it work with pubsub. Without this change, we are unable to use the ring client as it currently stands and would instead need to fall back to a single redis shard to manage pubsub. 

This is reproduced via test case in the latest version of go-redis, v9.

dkindel:
fixes #3889 

> [!NOTE]
> **Medium Risk**
> Changes pub/sub routing behavior for `Ring` users who relied on the old generic publish path; fix is small but affects multi-shard messaging correctness.
> 
> **Overview**
> **Fixes ring pub/sub** where `Publish` could target a different shard than `Subscribe` for the same channel, so subscribers never saw messages.
> 
> Adds an explicit `Ring.Publish` that resolves the shard with `sharding.GetByKey(channel)` (same routing as `Subscribe` / `PSubscribe`) and delegates to that shard’s client. Shard lookup failures return an `IntCmd` with the error set instead of panicking.
> 
> Adds an integration test that exercises **20 channels** end-to-end: subscribe, publish, assert subscriber count `1`, and receive the payload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0767349616367c600c4bdd800be6ebe1028f73c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ring.go
+++ b/ring.go
@@ -698,6 +698,17 @@ func (c *Ring) SSubscribe(ctx context.Context, channels ...string) *PubSub {
 	return shard.Client.SSubscribe(ctx, channels...)
 }
 
+// Publish posts the message to the channel
+func (c *Ring) Publish(ctx context.Context, channel string, message interface{}) *IntCmd {
+	shard, err := c.sharding.GetByKey(channel)
+	if err != nil {
+		cmd := NewIntCmd(ctx, "publish", channel, message)
+		cmd.SetErr(err)
+		return cmd
+	}
+	return shard.Client.Publish(ctx, channel, message)
+}
+
 func (c *Ring) OnNewNode(fn func(rdb *Client)) {
 	c.sharding.OnNewNode(fn)
 }
--- a/ring_test.go
+++ b/ring_test.go
@@ -804,6 +804,49 @@ var _ = Describe("Ring Tx timeout", func() {
 	})
 })
 
+var _ = Describe("Ring Publish/Subscribe sharding", func() {
+	var ring *redis.Ring
+
+	BeforeEach(func() {
+		ring = redis.NewRing(redisRingOptions())
+	})
+
+	AfterEach(func() {
+		Expect(ring.Close()).NotTo(HaveOccurred())
+	})
+
+	It("routes Publish to the same shard as Subscribe for a given channel", func() {
+		// Use many channels to test to make sure there isn't flakiness in the case this would fail.
+		// Because the default sharding (Ring.sharding.Random) is determined by a random function,
+		// we want to make sure there isn't an inconsistency in the failure case
+		const numChannels = 20
+
+		for i := 0; i < numChannels; i++ {
+			channel := fmt.Sprintf("ring-pubsub-test-channel-%d", i)
+
+			pubsub := ring.Subscribe(ctx, channel)
+
+			_, err := pubsub.Receive(ctx)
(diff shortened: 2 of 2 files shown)


ndyakov (approved):
thank you @dkindel

PR #3894, opened 2026-07-12
fix(maintnotifications): data race on clusterStateReloadCallback

saddamr3e:
The maintenance-notifications Manager holds a cluster-state reload callback that a SMIGRATED push notification invokes to reload cluster state. On a ClusterClient with maintenance notifications enabled it is installed from the OnNewNode hook while a node client is being created, and it is read and called from the SMIGRATED handler on that node's connections. Every other field on Manager is a sync.Map or an atomic, but this one was a plain func value guarded on neither the write nor the read. When MinIdleConns is above zero the new node client starts background dialers during creation, and those connections process push frames while they initialize, so a SMIGRATED arriving in that window reads the callback at the same moment the OnNewNode hook writes it. go test -race reports the read in TriggerClusterStateReload against the write in SetClusterStateReloadCallback. Moving the field to an atomic pointer so the setter stores and the trigger loads closes it, matching the atomic-pointer handling already used for the hook chain in redis.go.

---

> [!NOTE]
> **Low Risk**
> Small concurrency fix using atomic load/store on one callback field; behavior unchanged aside from eliminating the race.
> 
> **Overview**
> Fixes a **data race** on the maintenance `Manager`’s cluster-state reload callback used when **SMIGRATED** push notifications arrive.
> 
> The callback was a plain field written from **OnNewNode** during node client creation and read from **TriggerClusterStateReload** on connection push handling; with **MinIdleConns > 0**, those paths can overlap during connection init. The field is now an **`atomic.Pointer`**, with **Store** on set and **Load** on trigger, consistent with other lock-free fields in this package.
> 
> Adds **`TestManagerClusterStateReloadCallbackConcurrent`** to exercise concurrent set/trigger (intended for **`-race`**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a02514d25349dbb29c2fdc8ad8bc7e30a0de89b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/maintnotifications/cluster_state_callback_race_test.go
+++ b/maintnotifications/cluster_state_callback_race_test.go
@@ -1,0 +1,56 @@
+package maintnotifications
+
+import (
+	"context"
+	"sync"
+	"testing"
+)
+
+// clusterStateReloadCallback is set from the OnNewNode hook while a node client
+// is being created (GetOrCreateWithNodeAddress) and read from
+// TriggerClusterStateReload, which runs when a SMIGRATED push notification is
+// handled on one of that node's connections (handleSMigrated). With
+// MinIdleConns>0 the node's background dial goroutines process push frames
+// during initConn, so a SMIGRATED can reach TriggerClusterStateReload while the
+// OnNewNode hook is still setting the callback. Run with -race against the
+// pre-fix tree to see the report at manager.go (read in TriggerClusterStateReload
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
thank you @saddamr3e

PR #3900, opened 2026-07-13
fix(command): reject odd-length flat replies in Z/KeyValue parsers

saddamr3e:
Odd-length RESP2 flat reply (member/score or key/value) on a pooled connection:

    ZSliceCmd.readReply err=<nil>          // reports success
    proto.Reader.ReadString => "b"         // next command reads the leftover element

The flat (non-array) branch of ZSliceCmd, KeyValueSliceCmd and ZSliceWithKeyCmd
allocates make([]T, n/2) and reads n/2 pairs without checking n is even, so an
odd count leaves the trailing element buffered and desyncs the connection for the
next command. Adds the even-length guard the sibling flat parsers (VLINKS, VSIM,
VectorAttrib) already have.

---

> [!NOTE]
> **Medium Risk**
> Touches core RESP reply parsing on pooled connections; behavior change is stricter validation (errors where malformed replies previously slipped through), which is safer but could surface latent server/proxy bugs.
> 
> **Overview**
> Fixes a **RESP2 flat-reply parsing bug** where odd-length member/score or key/value arrays could be treated as success while leaving one element unread on the connection, corrupting the next command on pooled clients.
> 
> The **flat (non-nested-array) branches** of `KeyValueSliceCmd`, `ZSliceCmd`, and `ZSliceWithKeyCmd` now **require an even element count** and return a clear error instead of partially consuming the reply—aligned with guards already used by flat parsers such as VLINKS/VSIM.
> 
> Adds **`command_flat_reply_test.go`** to assert odd-length flat replies are rejected and that valid flat and nested replies still parse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a21bf675d5e46010108b0e41100fddcd8c0e3a28. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -2030,6 +2030,9 @@ func (cmd *KeyValueSliceCmd) readReply(rd *proto.Reader) error { // nolint:dupl
 	if array {
 		cmd.val = make([]KeyValue, n)
 	} else {
+		if n%2 != 0 {
+			return fmt.Errorf("redis: got %d elements in the key-value array, wanted a multiple of 2", n)
+		}
 		cmd.val = make([]KeyValue, n/2)
 	}
 
@@ -4132,6 +4135,9 @@ func (cmd *ZSliceCmd) readReply(rd *proto.Reader) error { // nolint:dupl
 	if array {
 		cmd.val = make([]Z, n)
 	} else {
+		if n%2 != 0 {
+			return fmt.Errorf("redis: got %d elements in the sorted set array, wanted a multiple of 2", n)
+		}
 		cmd.val = make([]Z, n/2)
 	}
 
@@ -6230,6 +6236,9 @@ func (cmd *ZSliceWithKeyCmd) readReply(rd *proto.Reader) (err error) {
 	if array {
 		cmd.val = make([]Z, n)
 	} else {
+		if n%2 != 0 {
+			return fmt.Errorf("redis: got %d elements in the sorted set array, wanted a multiple of 2", n)
+		}
 		cmd.val = make([]Z, n/2)
 	}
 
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
@saddamr3e thank you for another defensive check. I do not believe this can happen right now. Also, if there is buffered data in the connection before writing a new command, this connection will be recycled. That said, for pipelines this will be a potential issue if it can happen.

Would you mind pinning the govulcheck to a version after the fix, so we can include this in master as well.

saddamr3e:
Agreed, a well-behaved server won't emit this, and for single commands the buffered-data check recycles the conn before the next write; pipelines reading replies back-to-back on the same conn are the case this guards.

On the pin: tried to push it, but my PAT is missing the workflow scope so GitHub refuses any change under .github/workflows from my side. It's a one-liner in govulncheck.yml, go-version: "1.26.x" -> "1.26.5" (GO-2026-5856 is fixed in go1.26.5, the job currently resolves 1.26.x to 1.26.4). If it's easier, apply it on master and a re-run here picks it up from the merge ref; otherwise I'll push it once I've sorted the token.

saddamr3e:
go1.26.5 landed in the setup-go manifest, so I retriggered CI hoping 1.26.x would pick it up, but no luck: the runner grabs 1.26.4 from its hostedtoolcache first (setup-go prefers the local tool cache for a range spec, the log shows "Found in cache @ /opt/hostedtoolcache/go/1.26.4/x64"). So plain re-runs stay red until the runner images update. The explicit "1.26.5" pin (or check-latest: true) in govulncheck.yml is still the fix, and it'll need to come from your side since my token still can't push workflow files.

ndyakov (approved):
Thank you @saddamr3e

PR #3901, opened 2026-07-13
fix: run MONITOR on a dedicated connection and propagate read errors

### LINKED ISSUE #3079: monitor command breaks due to ReadTimeout
## Expected Behavior

Monitor command should work.

## Current Behavior

If you do not set a large `ReadTimeout` on the client when using the monitor command then the socket timeout is hit but that error is never propagated back in any way so the channel listener is left in a blocked state.

## Possible Solution

Either:
1. Update docs to inform users that they are responsible for detecting timeouts when listening to the channel
2. A `ReadTimeout` (or any other error) should be propagated back to user via closing the monitor channel

## Steps to Reproduce

1. Start client
```
		Addr: "127.0.0.1:6379",
		// ReadTimeout: 1 * time.Minute,
	})

	stream := make(chan string, 1000)
	monitorCmd := client.Monitor(ctx, stream)
	monitorCmd.Start()

	for {
		var msg string

		select {
		case <-ctx.Done():
			fmt.Println("ctx done")
		case msg = <-stream:
			fmt.Println("message")
		}

		fmt.Println(msg)
	}
```

2. Wait a second and then via cli perform any redis operation
3. Check output from client

ankit-songara:
Fixes #3079

## Problem

Two layered bugs, the second discovered while CI's race detector ran the regression test for the first:

**1. Read errors were silently swallowed (the reported issue).** `readMonitor` discards the error from `rd.Peek(1)`, so once the connection died the goroutine spun in a hot loop forever re-peeking it, `cmd.Err()` stayed nil, and a listener blocked on the monitor channel hung indefinitely with no way to detect the failure.

**2. MONITOR ran on a pooled connection.** `MonitorCmd.readReply` spawns the reader goroutine and returns immediately, so `withConn` put the connection **back into the pool** while the goroutine was still reading from it. The pool's `putConn` health-check peek then read from the same `bufio.Reader` concurrently — a data race (CI flagged it on the first version of this PR). It also meant the client's `ReadTimeout` stayed armed on the connection, which is why a monitor on a quiet server broke after a few seconds in the first place.

## Changes

- `_process` routes `MonitorCmd` to a new `processMonitor`, which runs the command on a **dedicated, non-pooled connection** (`ConnPool.NewConn`) with **no read deadline** — matching both the doc comment's existing promise ("It needs a dedicated connection") and how MONITOR behaves in redis-cli. An idle server no longer breaks the monitor.
- On a read error the channel is closed and the error is available via `cmd.Err()`, so a blocked listener wakes up (`msg, ok := <-ch` with `ok == false`) — option 2 from the issue.
- `Stop()` closes the dedicated connection to unblock a reader waiting in `Peek` for traffic. A clean stop reports no error and leaves the channel open (unchanged from before).
- `readMonitor` no longer holds the mutex across the blocking `Peek` (that would deadlock `Stop()`); the mutex now only guards the status field. The reader goroutine is the sole reader of the connection, so no lock is needed around reads.

## Testing

Two regression tests using a fake in-process TCP server speaking just enough RESP for MONITOR — they need no real Redis and are not gated behind `RUN_MONITOR_TEST`:

- `TestMonitorConnErrorClosesChannel`: server drops the connection mid-monitor → channel closes, `Err()` is set. Hangs forever on master.
- `TestMonitorStopWithoutTraffic`: `Stop()` takes effect with zero server traffic and reports no error. Deadlocks or times out on master since the reader only observed the stop flag when data arrived.

---

> [!NOTE]
> **Medium Risk**
> Changes connection ownership and error/lifecycle behavior for MONITOR (pool bypass, close-on-stop, channel close on failure). Incorrect close/error handling could leak connections or still hang callers, but the change is isolated to this command path.
> 
> **Overview**
> Fixes MONITOR hanging forever when the backing connection dies (#3079), and stops it from racing the connection pool.
> 
> `MonitorCmd` is now routed out of `_process`/`withConn` onto a **dedicated, non-pooled** connection with **no read deadline**, so an idle server no longer times out the stream and the pool cannot peek the same `bufio.Reader` the monitor goroutine is using.
> 
> On a real read error the monitor channel is **closed** and `cmd.Err()` is set, so a blocked listener unblocks. `Stop()` closes that connection to wake a reader stuck in `Peek` (no error, channel left open). `readMonitor` no longer holds the mutex across blocking reads.
> 
> Adds two fake-TCP regression tests: connection drop mid-monitor, and `Stop()` with no server traffic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ded076baa89b207af3fb92da1975523cf6e38e47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


command.go
```diff
@@ -8054,25 +8066,39 @@ func (cmd *MonitorCmd) readReply(rd *proto.Reader) error {
...
+				// Stop closed the connection to unblock Peek: shut down
+				// cleanly instead of reporting the read error.
+				cancel()
+				return nil
+			}
+			// The connection is no longer usable; without this the loop
+			// would spin forever re-peeking a dead connection.
+			return err
+		}
+		if len(pk) != 0 && cmd.getStatus() == monitorStatusStart {
```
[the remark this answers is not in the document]
ankit-songara: Good catch — fixed in a54004dc. The ReadString error path now does the same stop-status check as the Peek path, so a Stop that lands while a line is being read shuts down cleanly instead of reporting the read error and closing the channel.

redis.go
```diff
@@ -1041,6 +1048,40 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder, attempt int) (bool
...
+		return err
+	}
+	if err := cn.WithWriter(c.context(ctx), c.opt.WriteTimeout, func(wr *proto.Writer) error {
+		return writeCmd(wr, cmd)
+	}); err != nil {
+		_ = cn.Close()
+		cmd.SetErr(err)
+		return err
+	}
+	cmd.closeConn = func() { _ = cn.Close() }
```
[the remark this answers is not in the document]
ankit-songara: Fixed in a54004dc — teardown now goes through connPool.CloseConn (with a new pool.CloseReasonMonitor reason) so the connection is removed from the pool's conns map and metrics are recorded. Conn.Close is idempotent, so the potential double-close between Stop and the reader goroutine's deferred cleanup is safe.

ndyakov (commented):
Thank you for this contribution @ankit-songara. I did close the previous fix which was not ideal anyway and I do prefer this one. Will review this one as soon as possible.

command.go
```diff
@@ -8487,6 +8495,10 @@ func (cmd *MonitorCmd) readReply(rd *proto.Reader) error {
 				err := cmd.readMonitor(rd, cancel)
 				if err != nil {
 					cmd.err = err
+					cancel()
+					// Close the channel so a listener blocked on it is unblocked
+					// and can pick up the error with cmd.Err().
+					close(cmd.ch)
```
ndyakov, line 8501: We're closing `cmd.ch` here, but the caller owns this channel (passed into `Monitor`). If it's shared between writers, or the caller closes it during teardown, this panics — "send on closed channel" / "close of closed channel". Suggest not closing a channel we didn't create, and signalling the error via `Err()` + `ctx` cancellation only (or documenting the ownership contract very loudly in `commands.go`).

command.go
```diff
@@ -8497,25 +8509,46 @@ func (cmd *MonitorCmd) readReply(rd *proto.Reader) error {
...
+				if cmd.getStatus() == monitorStatusStop {
+					// Stop closed the connection while a line was being
+					// read: shut down cleanly instead of reporting the
+					// read error.
+					cancel()
+					return nil
+				}
 				return err
 			}
 			cmd.ch <- line
```
ndyakov, line 8543: This send is uninterruptible. If the consumer stops draining and the buffer fills, the reader blocks here — not in `Peek` — and `Stop()` closing the connection only unblocks `Peek`, so the goroutine stays parked and leaks. If the conn also dies while we're stuck here, the close/`Err()` path never runs and `cmd.Err()` stays `nil`. That's #3079 again on the send side. Suggest `select { case cmd.ch <- line: case <-ctx.Done(): return nil }`.

command.go
```diff
@@ -8497,25 +8509,46 @@ func (cmd *MonitorCmd) readReply(rd *proto.Reader) error {
...
+				// Stop closed the connection to unblock Peek: shut down
+				// cleanly instead of reporting the read error.
+				cancel()
+				return nil
+			}
+			// The connection is no longer usable; without this the loop
+			// would spin forever re-peeking a dead connection.
+			return err
+		}
+		if len(pk) != 0 && cmd.getStatus() == monitorStatusStart {
```
ndyakov, line 8531: If `Start()` is deferred after `Monitor()` returns, the buffered `+OK` makes `Peek(1)` return immediately, but this `status == monitorStatusStart` guard skips the read, so we loop and busy-spin at 100% CPU until `Start()` is called. A legit "set up monitor, do work, then Start()" pattern burns a core. Better to block on the start signal than to spin.

command.go
```diff
@@ -8487,6 +8495,10 @@ func (cmd *MonitorCmd) readReply(rd *proto.Reader) error {
 				err := cmd.readMonitor(rd, cancel)
 				if err != nil {
 					cmd.err = err
```
ndyakov, line 8497: Use `cmd.SetErr(err)` here rather than assigning `cmd.err` directly — it's the repo convention (AGENTS.md) and `processMonitor` already uses it. The direct write also races any caller polling `cmd.Err()`; `-race` will flag it since only the channel-close establishes a happens-before edge.

redis.go
```diff
@@ -1609,6 +1616,47 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder, attempt int, captu
...
+	// reader goroutine.
+	closeConn := func() {
+		_ = c.connPool.CloseConn(context.Background(), cn, pool.CloseReasonMonitor, pool.MetricStateUsed)
+	}
+	if err := c.initConn(ctx, cn); err != nil {
+		closeConn()
+		cmd.SetErr(err)
+		return err
+	}
+	if err := cn.WithWriter(c.context(ctx), c.opt.WriteTimeout, func(wr *proto.Writer) error {
```
ndyakov, line 1641: After writing `MONITOR` we never read/validate the `+OK` ack. Two consequences: (1) a rejected `MONITOR` (ACL `NOPERM`, or renamed/disabled) still returns from `Monitor()` with `cmd.Err() == nil` — even the new test's `if cmd.Err() != nil` check would pass on a rejected command; (2) on success the buffered `+OK` becomes the first line delivered to the channel, so every consumer gets a spurious leading `"OK"`. Reading + validating the ack during setup fixes both.

redis.go
```diff
@@ -1609,6 +1616,47 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder, attempt int, captu
...
+	}); err != nil {
+		closeConn()
+		cmd.SetErr(err)
+		return err
+	}
+	cmd.closeConn = closeConn
+	// Timeout 0 clears any read deadline armed during the handshake:
+	// MONITOR receives no traffic while the server is idle, so the
+	// connection must not have a read deadline at all.
+	if err := cn.WithReader(c.context(ctx), 0, cmd.readReply); err != nil {
```
ndyakov, line 1652: `0` isn't actually "no deadline" when `Options.ContextTimeoutEnabled` is set: `c.context(ctx)` returns the caller ctx, and `cn.deadline(ctx, 0)` then returns `ctx.Deadline()` (conn.go), arming a real read deadline. On an idle server the next `Peek` times out and the monitor is torn down — the exact failure this PR is fixing, for that config. Same if a relaxed maintenance timeout is armed. Need to detach the deadline explicitly for this read path.

redis.go
```diff
@@ -1490,6 +1490,13 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder, attempt int, captu
 		}
 	}
 
+	// MONITOR takes over its connection indefinitely, so it must not run on
+	// a pooled connection: the pool (or another command) would read from the
+	// same connection the monitor goroutine is reading from.
+	if monitorCmd, ok := cmd.(*MonitorCmd); ok {
+		return false, nil, c.processMonitor(ctx, monitorCmd)
```
ndyakov, line 1497: Returning `false, nil, err` short-circuits the retry loop — `MaxRetries`/`retryBackoff` no longer apply to MONITOR setup. Previously it ran through `withConn` inside the loop, so a transient dial/init failure was retried. Is dropping retries here intentional? A flaky-network setup now fails on the first attempt.

redis.go
```diff
@@ -1609,6 +1616,47 @@ func (c *baseClient) _process(ctx context.Context, cmd Cmder, attempt int, captu
...
+		return err
+	}
+	// CloseConn (rather than cn.Close) removes the connection from the
+	// pool's bookkeeping and records metrics. Background context: the
+	// connection may outlive ctx and be closed much later by Stop or the
+	// reader goroutine.
+	closeConn := func() {
+		_ = c.connPool.CloseConn(context.Background(), cn, pool.CloseReasonMonitor, pool.MetricStateUsed)
+	}
+	if err := c.initConn(ctx, cn); err != nil {
```
ndyakov, line 1636: `initConn` runs the full handshake on this dedicated conn, including `TrackMaintNotificationsConn` and the CSC close/reinit hooks. A later maintnotifications downgrade/retire could then close or hand off this conn out from under `readMonitor`. Worth deciding deliberately whether a monitor conn should be tracked at all. (Minor, related: `Stop()` holds `cmd.mu` across `CloseConn`'s hooks + socket I/O — consider narrowing the lock.)

monitor_conn_test.go
```diff
@@ -0,0 +1,153 @@
...
+					}
+					if line[0] == '*' || line[0] == '$' {
+						continue
+					}
+					switch {
+					case strings.HasPrefix(strings.ToLower(line), "hello"):
+						c.Write([]byte("-ERR unknown command 'hello'\r\n"))
+					case strings.HasPrefix(strings.ToLower(line), "monitor"):
+						c.Write([]byte("+OK\r\n"))
+						select {} // keep the connection open, send nothing
```
ndyakov, line 121: `select {}` blocks this handler goroutine forever, and `ln.Close()` at teardown won't unblock it — it leaks for the lifetime of the test binary. Harmless to the assertion, but a goroutine-leak detector would flag it. A `done` channel to release the handler would be cleaner.

ndyakov (changes requested):
Took a look at this — good direction, but I don't think it fully closes #3079, and it adds a couple of new issues on the way.

The main one: the reader can still get stuck. `cmd.ch <- line` is a blocking send, and `Stop()` closing the connection only unblocks `Peek`, not a send — so a stalled consumer leaks the goroutine and `cmd.Err()` stays `nil`, which is the same hang class this PR is meant to fix. Separately, `close(cmd.ch)` closes a channel the caller owns (can panic user code), a rejected `MONITOR` (NOPERM/renamed) still returns from `Monitor()` with no error, and the "no read deadline" guarantee doesn't hold when `ContextTimeoutEnabled` is set.

Left inline comments — the top few need addressing before we continue. Marking as request changes.

PR #3899, opened 2026-07-13
test: run the Redis Enterprise suite against an external cluster

kiryazovi-redis:
## Motivation
The Redis Enterprise test set (the `!NonRedisEnterprise` specs run with `RE_CLUSTER=true`) currently assumes a passwordless Redis Enterprise instance reachable on `localhost:6379`. This change lets the same suite also run against an **external / managed** Redis Enterprise deployment — remote host, authentication, and optional TLS — without changing the existing local behaviour.

## What changed
- **Opt-in endpoint discovery.** When `REDIS_ENDPOINTS_CONFIG_PATH` is set (the same endpoints-config format already consumed by the `maintnotifications/e2e` suite), the suite resolves the target database's host, port, username, password and TLS from it and applies them to the clients it builds: `redisOptions()`, the per-suite `setupRedisClient` closures (probabilistic / json / timeseries / vectorset), the RediSearch client constructions, and the universal-client spec. `RE_DB_NAME` selects a named database (defaults to the first). **When the variable is unset, behaviour is unchanged** (`localhost:6379`, no auth), so the existing local flow is unaffected.
- **`NonRedisEnterprise` labels (with comments).** Marked the specs that cannot meaningfully run against a remote, authenticated, managed Enterprise deployment:
  - *Client-internal / local-only*: credential-provider precedence, connection-pool (`withConn`) internals, and the client `Stringer` address assertion — these need a co-located server and inspect local behaviour.
  - *OSS-vs-Enterprise behaviour*: `AUTH` error text, `CLIENT INFO` flags, `FT.CONFIG`, module `CONFIG`/`INFO` shape, `XINFO STREAM` / `XINFO STREAM FULL` fields, and the `FT.AGGREGATE` search-timeout spec (depends on the `search-timeout` module config, unsupported on Enterprise).
  - *Impractical against a remote endpoint*: the `PExpire` / `PExpireAt` TTL assertions (±100 ms tolerance vs. network RTT), the contended `races` `Watch/Unwatch` loop (10 × 1000 optimistic-lock retries on one key), and the two-key `TS.CREATERULE` / `TS.DELETERULE` (`CROSSSLOT` on a sharded database).

## Note for maintainers
`Commands server should ClientSetInfo` fails with a **nil-pointer panic** against Redis Enterprise (rather than a clean error). I labelled it `NonRedisEnterprise` to keep the suite green, but this looks like a client-robustness issue worth a closer look — happy to break it out into a separate issue.

---
[...]

PR #3905, opened 2026-07-16
test: run the Redis Enterprise suite against an external cluster

ndyakov:
Supersedes #3899 — includes @kiryazovi-redis's commit unchanged, plus three maintainer commits addressing review findings.

## From #3899

When `REDIS_ENDPOINTS_CONFIG_PATH` is set, the Redis Enterprise suite (`RE_CLUSTER=true`, `!NonRedisEnterprise`) resolves the target database's host, port, credentials and TLS from the env0 endpoints config (same format as `maintnotifications/e2e`) instead of assuming a passwordless `localhost:6379`. `RE_DB_NAME` selects a named database. Behaviour is unchanged when the variable is unset. Specs that cannot run against a remote managed cluster are labelled `NonRedisEnterprise`, each with a comment explaining why.

## Maintainer additions

- **Skip the local-TLS tests on RE runs.** `TestTLSStandalone`/`TestTLSRedissURL` are plain Go tests, so the Ginkgo label filter cannot exclude them, and they `t.Fatalf` when the docker-compose TLS server (`localhost:6666`) is absent — this alone kept the nightly red. They now skip when `RE_CLUSTER=true`.
- **Harden endpoint resolution.** An empty `dns_name` in `raw_endpoints` now falls through to the endpoint URLs instead of producing a `":port"` address that silently dials localhost; the `Options`/`UniversalOptions` overlays share timeout constants.
- **Re-enable the `should Auth` spec everywhere.** It now accepts both failure shapes — `ERR AUTH ...` (passwordless OSS) and `WRONGPASS ...` (password-protected servers such as Redis Enterprise) — and issues the two failing `AUTH` commands in separate pipelines. On a password-protected Redis 8.x, a second failed `AUTH` in the same pipeline is throttled by the server's brute-force protection and never answered, stalling the read until the client times out; the spec only passed in CI because a passwordless server's `ERR AUTH` reply doesn't count as an auth failure.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to test harness, labels, and connection setup; production client code is untouched.
> 
> **Overview**
> Enables the Ginkgo suite to target a **remote Redis Enterprise** database when `RE_CLUSTER=true` and `REDIS_ENDPOINTS_CONFIG_PATH` is set: a new `re_endpoints_test.go` loads host/credentials/TLS from the env0 config (`RE_DB_NAME` optional), and `applyREConnection` / `reNewClient` overlay those settings (with longer timeouts) across `redisOptions()` and module/search/json helpers.
> 
> Specs that assume local OSS behavior are gated with **`NonRedisEnterprise`** (module CONFIG/INFO, tight PTTL checks, CROSSSLOT timeseries rules, local TLS docker, credential-provider recording, etc.). Plain **`TestTLS*`** tests skip on RE runs because Ginkgo labels cannot filter them.
> 
> **`should Auth`** now accepts both `ERR AUTH` and `WRONGPASS`, runs the two failing AUTHs in **separate pipelines** to avoid RE brute-force throttling, and expects two pool hits. Endpoint resolution uses **`net.JoinHostPort`** for IPv6 and falls through when `raw_endpoints` has an empty DNS name.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90ae165f4720d2fbd151799a9d54fde36dacf1e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/re_endpoints_test.go
+++ b/re_endpoints_test.go
@@ -102,8 +102,11 @@ func loadREConnection() (*reConnection, error) {
 func resolveREEndpoint(db reDatabaseConfig) (string, int, bool, error) {
 	tlsEnabled := db.TLS
 	if len(db.RawEndpoints) > 0 {
-		ep := db.RawEndpoints[0]
-		return ep.DNSName, ep.Port, tlsEnabled, nil
+		// Fall through to the endpoint URLs when the raw endpoint carries no
+		// DNS name, so a sparse config fails loudly instead of dialing ":port".
+		if ep := db.RawEndpoints[0]; ep.DNSName != "" {
+			return ep.DNSName, ep.Port, tlsEnabled, nil
+		}
 	}
 	if len(db.Endpoints) > 0 {
 		u, err := url.Parse(db.Endpoints[0])
@@ -131,6 +134,16 @@ func resolveREEndpoint(db reDatabaseConfig) (string, int, bool, error) {
 	return "", 0, false, fmt.Errorf("no endpoints found in database configuration")
 }
 
+// A remote Redis Enterprise endpoint has higher latency than a local server, so
+// the overlays below use the same relaxed timeouts as redisOptions() instead of
+// the ~3s defaults; otherwise module/search/timeseries helpers can flake on
+// slower commands.
+const (
+	reDialTimeout  = 10 * time.Second
+	reReadTimeout  = 30 * time.Second
+	reWriteTimeout = 30 * time.Second
+)
+
 // applyREConnection overlays the resolved Redis Enterprise credentials and TLS
 // settings onto opt. It is a no-op when reConn has not been populated.
 func applyREConnection(opt *redis.Options) {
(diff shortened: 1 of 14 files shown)


re_endpoints_test.go
```diff
@@ -0,0 +1,195 @@
...
+		db = databases[names[0]]
+	}
+
+	host, port, tlsEnabled, err := resolveREEndpoint(db)
+	if err != nil {
+		return nil, err
+	}
+
+	return &reConnection{
+		Addr:     fmt.Sprintf("%s:%d", host, port),
```
ofekshenawa: Address construction with fmt.Sprintf("%s:%d", host, port) breaks IPv6 endpoints. Maybe use net.JoinHostPort.

ofekshenawa (dismissed):
I left a comment, not urgent

ofekshenawa (approved):

PR #3907, opened 2026-07-17
fix(command): reconcile declared entry-array length in reply parsers

saddamr3e:
Malicious-server reply driven through proto.NewReader, unpatched tree:

    command_entry_array_desync_test.go:27: stream desynced, got "7" want PONG

LatencyCmd/SlowLogCmd/HotKeysCmd/GeoSearchLocationCmd each read a fixed set of fields from an entry array but never reconcile that with the element count the server declares. When a reply declares more elements than the parser reads (LATENCY LATEST guarded `nn < 3` while reading 4 fields; SLOWLOG never drains past field 6; HOTKEYS parses only the first element of the outer aggregation array; GEOSEARCH discarded the per-entry length), the trailing frames stay buffered, readReply returns nil, and the desynced connection goes back to the pool so the next command reads the leftovers as its own reply.

Same trust boundary as the sibling parsers XAutoClaim/ClusterSlots/GeoLocationCmd, which already reconcile the count. Fix drains the extra declared frames (LATENCY/SLOWLOG/HOTKEYS) or validates the entry length via ReadFixedArrayLen like GeoLocationCmd (GEOSEARCH); LATENCY's off-by-one guard becomes `nn < 4`. Regression test drives each case through proto.NewReader and checks the sentinel reply after readReply is still aligned.

---

> [!NOTE]
> **Medium Risk**
> Touches core wire-protocol parsing for admin/geo commands; behavior changes include rejecting malformed short entries and accepting longer server replies, but the fix reduces connection-pool corruption risk from hostile or newer Redis responses.
> 
> **Overview**
> Fixes **RESP reply parsers** that could leave unread frames on the connection when the server declares more (or fewer) array elements than the client consumes. Unread bytes make `readReply` succeed while the stream is misaligned; pooled connections then corrupt the next command’s reply.
> 
> **`GeoSearchLocationCmd`** now derives a minimum entry length from `WITH` flags, **errors on short entries**, and **`DiscardNext`s** any extra fields per location. **`LatencyCmd`** tightens validation to require at least four fields per entry (was three while reading four) and drains surplus elements. **`SlowLogCmd`** drops the hard cap at seven fields and drains anything beyond the seven known slots so newer Redis (e.g. 8.10+ `CommandArgc`) and future fields stay forward-compatible. **`HotKeysCmd`** drains additional top-level aggregation array elements after the first map.
> 
> Regression coverage in **`command_entry_array_desync_test.go`** feeds crafted RESP through `proto.NewReader` and asserts a trailing **PONG** sentinel still parses, plus negative cases for undersized latency/geo entries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22430eeaff2894c6cc3b7b2c9afbcee719e3a67e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -4823,9 +4823,22 @@ func (cmd *GeoSearchLocationCmd) readReply(rd *proto.Reader) error {
 	}
 
 	cmd.val = make([]GeoLocation, n)
+	// Each element is an array of [name, ...] whose length is fixed by the
+	// requested WITH flags. Validate it (like GeoLocationCmd) instead of
+	// discarding it, so a server declaring extra elements can't leave frames
+	// on the wire and desync the connection.
+	withLen := 1
+	if cmd.opt.WithDist {
+		withLen++
+	}
+	if cmd.opt.WithHash {
+		withLen++
+	}
+	if cmd.opt.WithCoord {
+		withLen++
+	}
 	for i := 0; i < n; i++ {
-		_, err = rd.ReadArrayLen()
-		if err != nil {
+		if err = rd.ReadFixedArrayLen(withLen); err != nil {
 			return err
 		}
 
@@ -5407,6 +5420,14 @@ func (cmd *SlowLogCmd) readReply(rd *proto.Reader) error {
 				return err
 			}
 		}
+
+		// Drain any elements past the 6 this parser knows about so a server
+		// that declares a longer entry array doesn't leave frames on the wire.
+		for j := 6; j < nn; j++ {
+			if err = rd.DiscardNext(); err != nil {
+				return err
+			}
+		}
 	}
 
 	return nil
@@ -5488,8 +5509,8 @@ func (cmd *LatencyCmd) readReply(rd *proto.Reader) error {
 		if err != nil {
 			return err
 		}
-		if nn < 3 {
-			return fmt.Errorf("redis: got %d elements in latency get, expected at least 3", nn)
+		if nn < 4 {
+			return fmt.Errorf("redis: got %d elements in latency get, expected at least 4", nn)
 		}
 		if cmd.val[i].Name, err = rd.ReadString(); err != nil {
 			return err
@@ -5509,6 +5530,13 @@ func (cmd *LatencyCmd) readReply(rd *proto.Reader) error {
 			return err
 		}
 		cmd.val[i].Max = time.Duration(maximum) * time.Millisecond
+		// Drain any elements beyond the 4 this parser reads so a server that
+		// declares a longer entry array can't leave frames on the wire.
+		for j := 4; j < nn; j++ {
+			if err = rd.DiscardNext(); err != nil {
+				return err
+			}
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -4823,9 +4823,22 @@ func (cmd *GeoSearchLocationCmd) readReply(rd *proto.Reader) error {
...
+		withLen++
+	}
+	if cmd.opt.WithCoord {
+		withLen++
+	}
 	for i := 0; i < n; i++ {
-		_, err = rd.ReadArrayLen()
-		if err != nil {
+		if err = rd.ReadFixedArrayLen(withLen); err != nil {
 			return err
```
ndyakov, line 4859: I do agree that we should expect exactly withLen elements, but if the server is returning anything else, why don't just discard it instead of erroring here? My point is, this is not forwards compatible if the server decides to add another element in the array.

ndyakov (commented):
@saddamr3e thank you for looking into this. I left a question.

command.go
saddamr3e: Makes sense. Switched it to read the actual entry length and drain anything past withLen with DiscardNext, same as the other three parsers in this PR, so a newer server appending fields keeps working. Kept an error only for the short case (nn < withLen) since parsing those fields would read into the next reply. Test updated: oversized entry now parses and the sentinel stays aligned.

ndyakov:
The switch from exact-length to minimum-length + drain for GeoSearchLocation is the right call — a server that appends fields to the entry shouldn't break us, and the short-entry case still errors. Drain loops in SlowLog/Latency/HotKeys line up with what the parsers actually consume, and the PONG-sentinel tests prove the stream stays aligned. LGTM.

ndyakov (dismissed):
Fixed a merge conflict, waiting for the build and overall ready to be included on master. Thank you @saddamr3e!

command_entry_array_desync_test.go
```diff
@@ -0,0 +1,117 @@
...
+	t.Run("SlowLogCmd", func(t *testing.T) {
+		// entry declares 7 elements: the 6 known fields plus one extra
+		rd := proto.NewReader(strings.NewReader(
+			"*1\r\n*7\r\n:1\r\n:1000\r\n:50\r\n*1\r\n$3\r\nGET\r\n$5\r\n1.2.3\r\n$4\r\nname\r\n:99\r\n+PONG\r\n"))
+		cmd := NewSlowLogCmd(context.Background())
+		if err := cmd.readReply(rd); err != nil {
+			t.Fatalf("readReply: %v", err)
+		}
+		assertSentinel(t, rd)
+	})
```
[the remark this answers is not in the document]
ndyakov: Good catch — with CommandArgc as the 7th known field the 7-element entry no longer reached the drain loop. The entry now declares 8 elements (7 known + 1 extra) and asserts CommandArgc is parsed; disabling the drain loop makes the test fail on sentinel desync, so the path is covered.

ndyakov (approved):

saddamr3e:
Thanks for merging, and for sorting out the conflict along the way.

PR #3909, opened 2026-07-19
re-route whole tx pipeline on redirect/abort, not per-command

cxljs:
The cluster tx pipeline treated a MULTI..EXEC as if each command could be retried independently and EXEC always returned a *N result array. That conflicts with Redis's transaction semantics: any queue-stage redirect (MOVED/ASK/TRYAGAIN) or command error calls flagTransaction, dirties the tx, and EXEC returns -EXECABORT (or an EXEC-level MOVED/ASK/TRYAGAIN/ CLUSTERDOWN) -- the whole tx is aborted with no partial execution and no result array. The old code consequently broke six ways:

  - Scattering: on getConn/writer failure it ran the already-wrapped [MULTI..EXEC] through mapCmdsByNode, routing the keyless MULTI/EXEC to a different (round-robined) master and sending a malformed tx.
  - Double-wrap latent: mapCmdsByNode ran after wrap, and retry wrapped again; masked only by the scattering above.
  - Per-command MOVED/ASK retry via checkMovedErr/cmdsMoved, impossible once the tx is already aborted.
  - EXECABORT and EXEC-level MOVED masked as a generic "expected '*'" error, dropping the redirect addr.
  - TRYAGAIN at queue stage only SetErr'd the command, never retried.
  - A for+goroutine+WaitGroup retry loop that was pure overhead since the cmds map was always single-node.

Redesign processTxPipeline around a per-attempt txOutcome classified as success / retryMoved / retryAsk / retryTryAgain / retryConn / fatal, and re-route the whole trimmed transaction accordingly. Read replies without assuming EXEC is an array: collect the first queue-stage redirect and the first fatal command error, then classify the EXEC reply (array = success; MOVED/ASK/TRYAGAIN/CLUSTERDOWN/EXECABORT each map to a retry or fatal outcome).

ASK retries now send a single top-level ASKING before MULTI -- the flag persists across a MULTI per server semantics (it is only cleared outside MULTI), so one ASKING covers the whole tx and lets the importing slot serve at EXEC. Placing ASKING inside the MULTI queued it and left the flag unset during queueing, so keyed commands still got MOVED. The ASKING flag is preserved across a subsequent TRYAGAIN/conn retry while still following an ASK, since the slot may still be importing.

Connections that early-exit before consuming all N+2 replies (a MULTI read failure, or a non-array aggregate EXEC reply) are marked unreadReplies so releaseConn discards them rather than returning a connection with bytes left on the wire.

The pipeline-operation-duration callback now runs in a defer and reports cmp.Or(retErr, cmdsFirstErr(cmds), lastErr), so a final success after retries is no longer reported as a failure via a stale lastErr.

---

> [!NOTE]
> **High Risk**
> Changes core cluster transaction routing, retry, and connection lifecycle; incorrect retry after a committed tx could double-apply writes, though new tests explicitly guard the post-write read path.
> 
> **Overview**
> **Cluster `TxPipeline` (`processTxPipeline`) is redesigned** so each attempt returns a `txOutcome` (success, MOVED/ASK/TRYAGAIN/conn retry, or fatal) and the **entire** transaction is re-sent on redirects—not split or re-wrapped through `mapCmdsByNode`.
> 
> Reply handling no longer assumes `EXEC` is always a `*N` array: it records the first queue-stage redirect or command error, classifies `EXEC` (`EXECABORT`, `MOVED`, `ASK`, `TRYAGAIN`, `CLUSTERDOWN`), surfaces the triggering error on abort, and **skips retry** when the batch was already written but the read failed (avoids double-execution).
> 
> **ASK retries** send one **top-level `ASKING` before `MULTI`** (not inside the tx); that flag is kept across conn/`TRYAGAIN` retries on the importing node. Dirty connections with unread replies are discarded via `errTxDirtyConn` / `unreadReplies`. Pipeline duration metrics use a `defer` with `cmp.Or(retErr, cmdsFirstErr, lastErr)`.
> 
> Integration tests cover MOVED/EXECABORT errors, transient write retry, ASK+ASKING preservation, and no retry after post-commit read failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0bba1c7cc0301130f71fdc723eaaeeffc3ab38a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/osscluster.go
+++ b/osscluster.go
@@ -1837,18 +1837,108 @@ func (c *ClusterClient) TxPipelined(ctx context.Context, fn func(Pipeliner) erro
 	return c.TxPipeline().Pipelined(ctx, fn)
 }
 
-func (c *ClusterClient) processTxPipeline(ctx context.Context, cmds []Cmder) error {
-	// Only call time.Now() if pipeline operation duration callback is set to avoid overhead
+// A cluster tx pipeline sends MULTI, c1..cN, EXEC — N+2 commands, or N+3 with a
+// leading ASKING — and always receives exactly that many replies, so every
+// redirect/abort path leaves the connection clean.
+//
+// Possible reply sequences:
+//	1. Slot owned here, no migration:
+//	     +OK,  +QUEUED x N,  *N (array of N results)   -> success
+//	2. Slot already migrated away:
+//	     +OK,  -MOVED x N,  -EXECABORT                 -> re-route whole tx
+//	3. Slot in migrating state (still owned here, keys draining out). Per
+//	   cmd, the queue reply is +QUEUED / -ASK / -TRYAGAIN (keys present /
+//	   all gone / some gone); any -ASK or -TRYAGAIN dirties the tx, so
+//	   EXEC is -EXECABORT. Still N+2 replies, like the cases above:
+//	     +OK,  (+QUEUED|-ASK|-TRYAGAIN) x N,  -EXECABORT  -> follow first redirect
+//	4. Narrow race (all +QUEUED, slot moves before EXEC):
+//	     +OK,  +QUEUED x N,  -MOVED <slot> <addr>         -> re-route whole tx
+//	5. Non-cluster command error (arity / ACL / unknown):
+//	     +OK,  +QUEUED..., -ERR...,  -EXECABORT           -> surface, not retryable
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
This makes sense, thank you @cxljs!

PR #3914, opened 2026-07-21
fix(redisotel): use ObservableCounter for cumulative pool stats

### LINKED ISSUE #3612: Wrong otel metric types

`db.client.connections.timeouts` metric defined here
https://github.com/redis/go-redis/blob/5b0b228a37c83b7c7ace2fcd7cf230f61e2e3092/extra/redisotel/metrics.go#L176
is defined as `Int64ObservableUpDownCounter` but according to [otel conventions](https://opentelemetry.io/docs/specs/semconv/database/database-metrics/#metric-dbclientconnectiontimeouts)
it should be a `counter`

Also, `db.client.connections.waits`, `db.client.connections.hits `and `db.client.connections.misses` should also be `counter`

## Expected Behavior

`db.client.connections.timeouts`, `db.client.connections.waits`, `db.client.connections.hits` and `db.client.connections.misses` should be defined as `Counter`

## Current Behavior

`db.client.connections.timeouts`, `db.client.connections.waits`, `db.client.connections.hits` and `db.client.connections.misses` are defined as `UpDownCounter`

## Steps to Reproduce

1.
2.
3.
4.

## Context (Environment)

When using prometheus exporter the `UpDownCounter` is exported as a `Gauge` and it is not possible to use for example `rate` function.

## Detailed Description

Either change current metric types (which is a breaking change) or create new correct metrics.

Solaris-star:
## Summary
Fixes #3612.

`db.client.connections.timeouts`, `waits`, `hits`, and `misses` are monotonic cumulative counts, but redisotel registered them as `Int64ObservableUpDownCounter`.

Per OpenTelemetry database semantic conventions these should be counters so rate() aggregations work correctly.

## Changes
- Switch those four instruments to `Int64ObservableCounter`
- Leave gauges/up-down gauges (`usage`, idle min/max, max connections) unchanged
- Leave `waits_duration` unchanged (not listed in the issue)

## Test plan
- [x] `go test ./extra/redisotel/`

---

> [!NOTE]
> **Medium Risk**
> Opt-in changes exported metric instrument types for four pool stats, which can alter backend/Prometheus series types; default path is unchanged so existing deployments are unaffected unless they enable the flag.
> 
> **Overview**
> Adds **`WithSemConvCompliantMetrics`** so cumulative pool metrics (`db.client.connections.waits`, `.timeouts`, `.hits`, `.misses`) can be registered as **monotonic `Int64ObservableCounter`** instruments per OpenTelemetry database semconv (#3612). **Default behavior is unchanged**: those four still use **`Int64ObservableUpDownCounter`** so existing exporters (e.g. Prometheus metric type) are not broken.
> 
> Pool registration is refactored through a **`cumulativePoolStat`** helper that picks counter vs up-down counter from config. Other pool instruments (`usage`, idle min/max, max, `waits_duration`) are untouched.
> 
> Tests collect metrics via the SDK manual reader and assert non-monotonic sums by default and monotonic sums when the option is enabled. **`go.opentelemetry.io/otel/sdk/metric`** is added to `extra/redisotel` for those tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c31b3f670920538e165d0ca09c837cb60fa659f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/extra/redisotel/config.go
+++ b/extra/redisotel/config.go
@@ -33,6 +33,15 @@ type config struct {
 	mp    metric.MeterProvider
 	meter metric.Meter
 
+	// semconvCompliantMetrics, when true, registers the cumulative pool-stat
+	// metrics (waits, timeouts, hits, misses) as monotonic Counters as
+	// required by the OpenTelemetry database semantic conventions. It defaults
+	// to false to preserve backwards compatibility: existing users have these
+	// exported as UpDownCounter (which some backends surface as a Gauge), and
+	// switching the instrument type is a breaking change for them. New users
+	// that want spec-compliant metrics can opt in. See #3612.
+	semconvCompliantMetrics bool
+
 	poolName string
 
 	closeChan chan struct{}
@@ -226,5 +235,20 @@ func WithMeterProvider(mp metric.MeterProvider) MetricsOption {
 func WithCloseChan(closeChan chan struct{}) MetricsOption {
 	return metricsOption(func(conf *config) {
 		conf.closeChan = closeChan
+	})
+}
+
+// WithSemConvCompliantMetrics registers the cumulative connection-pool metrics
+// (db.client.connections.waits, .timeouts, .hits, .misses) as monotonic
+// Counters, matching the OpenTelemetry database semantic conventions.
+//
+// It is opt-in and defaults to off. By default these metrics remain
+// UpDownCounters for backwards compatibility, because changing an existing
+// instrument's type is a breaking change for current OTel users (e.g. the
+// Prometheus exporter would change the exported metric type). Enable this when
+// starting fresh or when you can tolerate the metric-type change. See #3612.
+func WithSemConvCompliantMetrics(on bool) MetricsOption {
+	return metricsOption(func(conf *config) {
+		conf.semconvCompliantMetrics = on
 	})
 }
--- a/extra/redisotel/go.mod
+++ b/extra/redisotel/go.mod
@@ -12,6 +12,7 @@ require (
 	go.opentelemetry.io/otel v1.43.0
 	go.opentelemetry.io/otel/metric v1.43.0
 	go.opentelemetry.io/otel/sdk v1.43.0
+	go.opentelemetry.io/otel/sdk/metric v1.43.0
 	go.opentelemetry.io/otel/trace v1.43.0
 )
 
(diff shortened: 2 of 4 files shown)


ndyakov (commented):
@Solaris-star this will be a breaking change for people using the otel at the moment, right? If I am correct, we cannot merge it as it is in v9.

Solaris-star:
@ndyakov you're right — swapping the instrument type outright is a breaking change for anyone currently scraping these as UpDownCounter (e.g. the Prometheus exporter would change the exported metric type), so it can't go into v9 as-is.

I've reworked it to be **non-breaking and opt-in**:

- Default behaviour is unchanged: `db.client.connections.{waits,timeouts,hits,misses}` remain `Int64ObservableUpDownCounter`.
- A new `WithSemConvCompliantMetrics(true)` option registers them as `Int64ObservableCounter` per the OTel database semantic conventions, for users starting fresh or who can tolerate the change.
- Both types satisfy `metric.Int64Observable`, so the observe callback is unchanged; a small `cumulativePoolStat` helper picks the instrument based on the flag.

Tests use the SDK manual reader to assert the actual exported aggregation type:
- `TestCumulativePoolStats_DefaultIsUpDownCounter` — default, `IsMonotonic == false`.
- `TestCumulativePoolStats_OptInIsCounter` — opt-in, `IsMonotonic == true`.

This lets v9 users get spec-compliant metrics without breaking existing dashboards. If you'd rather make compliant-by-default the behaviour in v10 and keep this as the opt-in bridge for v9, that works too — happy to align with whatever you prefer.

ndyakov (approved):
@Solaris-star this sounds better, yes. Let's keep the default compatible with the current type is something we can merge. Keep in mind `redisotel` is maintained mostly by the community nowadays. We do recommend looking at `redisotel-native` and using it for the future. If you have any questions - the redisotel-native is maintained mostly by @ofekshenawa and it should be aligned with the OpenTelemetry database semantic conventions.

Solaris-star:
Thanks @ndyakov! Glad the opt-in approach works. Noted on redisotel-native for future work — happy to keep this one community-maintainable.

ndyakov:
Thanks for reworking this as opt-in — `WithSemConvCompliantMetrics(true)` gets the semconv-correct monotonic counters without breaking existing dashboards, and the test asserting `IsMonotonic` on both paths pins the behavior nicely. LGTM.

PR #3922, opened 2026-07-23
fix(command): handle nil element in int/uint/bool slice parsers

saddamr3e:
`BITFIELD ... OVERFLOW FAIL` returns nil for an operation that overflows, so a reply can carry a nil in a non-final slot, e.g. `[255, nil, 100]`:

    *3\r\n:255\r\n_\r\n:100\r\n:999\r\n
    IntSliceCmd.readReply -> redis.Nil (only :255 read), next read = 100 not 999

`ReadInt`/`ReadUint`/`ReadBool` return `redis.Nil` on the nil element and the loop returns there, leaving the rest of the array on the wire. `redis.Nil` is not a bad conn, so the connection goes back to the pool desynced and the next command reads the leftovers. `FloatSliceCmd` and `StringSliceCmd` already store the zero value and keep going; the int/uint/bool parsers now do the same.

---

> [!NOTE]
> **Medium Risk**
> Touches core RESP parsing for pooled connections; behavior change maps nil slots to zero values rather than failing the command.
> 
> **Overview**
> Fixes **connection desync** when Redis returns a **nil in the middle of a scalar array** (e.g. `BITFIELD ... OVERFLOW FAIL` → `[255, nil, 100]`).
> 
> `IntSliceCmd`, `UintSliceCmd`, and `BoolSliceCmd` used to **abort** on `redis.Nil` from `ReadInt`/`ReadUint`/`ReadBool`, so trailing array elements stayed on the wire and the next command on a pooled connection read garbage. They now **consume the full array**, mapping nil slots to **0** / **false** (same pattern as `FloatSliceCmd` and `StringSliceCmd`).
> 
> Adds **`TestSliceCmdNilElementNoDesync`** to assert the reader is aligned for the following reply after a `*3` reply with a middle `_` nil.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f659a31e941bc753c0792a3963c60a13058c4a7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -1322,8 +1322,13 @@ func (cmd *IntSliceCmd) readReply(rd *proto.Reader) error {
 	}
 	cmd.val = make([]int64, n)
 	for i := 0; i < len(cmd.val); i++ {
-		if cmd.val[i], err = rd.ReadInt(); err != nil {
+		switch num, err := rd.ReadInt(); {
+		case err == Nil:
+			cmd.val[i] = 0
+		case err != nil:
 			return err
+		default:
+			cmd.val[i] = num
 		}
 	}
 	return nil
@@ -1382,8 +1387,13 @@ func (cmd *UintSliceCmd) readReply(rd *proto.Reader) error {
 	}
 	cmd.val = make([]uint64, n)
 	for i := range cmd.val {
-		if cmd.val[i], err = rd.ReadUint(); err != nil {
+		switch num, err := rd.ReadUint(); {
+		case err == Nil:
+			cmd.val[i] = 0
+		case err != nil:
 			return err
+		default:
+			cmd.val[i] = num
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
@saddamr3e actually pool's Put guard checks Buffered() > 0 and discards dirty conns (pool.go:1243-1248). I reproduced master behavior: command errors, but next PING/ECHO is fine, conn silently dropped. Real-world impact on master is a lost results + conn churn, plus narrow corruption window when trailing bytes haven't reached the buffer at Put time (unlikly, but true desync possible). Fix valid either way, Thank you!

saddamr3e:
Right, I'd missed the HasBufferedData() guard in Put. So on master it's mostly a lost reply plus a dropped conn, with true desync only in that window where the trailing bytes haven't reached the buffer yet. Thanks for checking it against master.

saddamr3e:
Right, I missed the HasBufferedData guard in Put, so on master the common case is a failed command plus a dropped conn rather than cross-request confusion, with true desync only when the trailing bytes are still in flight at Put time. Good to know the blast radius is smaller than I described, thanks for reproducing it.

ndyakov:
Nice catch — BITFIELD with OVERFLOW FAIL returns arrays with nil holes, and the old parsers bailed mid-array leaving the trailing elements on the wire, which poisons a pooled connection. Converting nil to the zero value matches what FloatSliceCmd/StringSliceCmd already do, and the desync test with the follow-up read is exactly the right shape. LGTM.

saddamr3e:
thanks for merging, and for pinning down the real blast radius on master

PR #3925, opened 2026-07-23
feat(search): add FT.ALIASLIST support

ndyakov:
Adds client support for `FT.ALIASLIST` (Redis 8.10): listing the aliases associated with a search index.

## API

```go
func (c cmdable) FTAliasList(ctx context.Context, index string) *StringSliceCmd
```

- The reply is an unordered, server-deduplicated collection of alias names; an index with no aliases yields an empty slice, not an error.
- RESP2 (array) and RESP3 (set) replies decode to the same result.
- The argument must be an index created with `FT.CREATE`; the server does not resolve aliases and the client does not either — `SEARCH_INDEX_NOT_FOUND`, `NOPERM`, and arity errors propagate unchanged.
- Keyless command, default routing; the reply must not be cached.

## Testing

- `TestFTAliasListArgs` pins the wire format (`FT.ALIASLIST index`, no options).
- `TestFTAliasList_Integration` runs both RESP2 and RESP3 against a live server: empty listing, two aliases, missing-index error, and alias-as-argument error. It skips when the server predates the command (`unknown command`), since the `REDIS_VERSION` float gating cannot distinguish 8.10 from 8.1. Verified against `redislabs/client-libs-test:8.10-rc2`.
- Runnable example in `example/ft-aliaslist`.

---

> [!NOTE]
> **Medium Risk**
> Touches cluster slot/key resolution for FT module commands on a cold cache; incorrect classification could misroute early requests, though behavior is aligned with the static policy table and covered by new unit tests.
> 
> **Overview**
> Adds **`FTAliasList`** (Redis 8.10+) on `SearchCmdable`, issuing `FT.ALIASLIST` with a single index name and returning alias names as a `StringSliceCmd`.
> 
> Registers **`aliaslist`** in the static FT routing policy as **keyless** (default request routing, keyless response, read-only tip). Introduces **`defaultPolicyKeyless`** and uses it in **`cmdFirstKeyPosWithInfo`** so module commands that are keyless in that table (including `ft.aliaslist`) report **no key slot while the command-info cache is still cold**, avoiding hashing the index argument to the wrong shard before policies warm up.
> 
> Covers wire format, cold-cache key-position behavior, RESP2/RESP3 integration tests (version-gated), and a runnable **`example/ft-aliaslist`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 223f80058e599e84d8b3a0cc0df896f795bffb83. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ofekshenawa (approved):

ofekshenawa (changes requested):
Maybe convert the tests into ginkgo env and apply the "SkipBeforeRedisVersion"?

ofekshenawa (dismissed):

ofekshenawa (approved):

PR #3926, opened 2026-07-23
feat(timeseries): add TS.QUERYLABELS support

ndyakov:
Adds client support for `TS.QUERYLABELS` (Redis 8.10): discovering label names and label values across time series, completing the dashboard drill-down flow next to `TS.QUERYINDEX`.

## API

```go
func (c cmdable) TSQueryLabels(ctx context.Context, filterExpr []string) *StringSliceCmd
func (c cmdable) TSQueryLabelValues(ctx context.Context, label string, filterExpr []string) *StringSliceCmd
```

- Two methods mirror the two server forms (`LABELS` / `VALUES label`), so the mandatory label argument of the VALUES form is enforced by the signature.
- `filterExpr` uses the same filter language as `TSQueryIndex` and is passed to the server verbatim, in order; passing no expressions omits the `FILTER` token (the documented query-all-series form — the server rejects a bare `FILTER`).
- The label argument is not trimmed or case-folded; matching is byte-exact.
- Replies are unordered, server-deduplicated string collections, identical under RESP2 (array) and RESP3 (set); empty replies are success.
- Server errors (`TSDB:` prefixed) propagate unchanged. Keyless command, default routing, no client-side fan-out; the reply must not be cached (`dont_cache` tip).

## Testing

- `TestTSQueryLabelsArgs` pins the wire format of both forms, including FILTER-token omission and verbatim filter pass-through.
- `TestTSQueryLabels_Integration` runs both RESP2 and RESP3 against a live server: label-name and label-value aggregation, absent-label and case-variant empty replies, unfiltered superset, and server-side filter error propagation. It skips when the server predates the command (`unknown command`), since the `REDIS_VERSION` float gating cannot distinguish 8.10 from 8.1. Verified against `redislabs/client-libs-test:8.10-rc2`.
- Runnable example in `example/ts-querylabels`.

---

> [!NOTE]
> **Low Risk**
> Additive TimeSeries API with tests and an example; no changes to existing command behavior beyond new methods.
> 
> **Overview**
> Adds client bindings for Redis 8.10+ **`TS.QUERYLABELS`**, exposing **`TSQueryLabels`** (distinct label names for filtered series) and **`TSQueryLabelValues`** (distinct values for one label). Filters reuse the same expressions as **`TSQueryIndex`**; an empty filter list omits the **`FILTER`** token so all indexed series are queried.
> 
> Shared **`appendTSFilter`** builds the optional filter segment. Coverage includes wire-format unit tests, RESP2/RESP3 integration tests (version-gated at 8.10), and a runnable **`example/ts-querylabels`** drill-down demo with **`TSQueryIndex`**. OTel example modules only bump indirect **`golang.org/x/*`** and **gRPC** deps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e419aff360c2a634742a21cc4d3add9c5c6cebc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ofekshenawa (changes requested):
Maybe convert the tests into ginkgo env and apply the "SkipBeforeRedisVersion"?

ofekshenawa (dismissed):

ofekshenawa (approved):

PR #3928, opened 2026-07-23
chore(cmd): remove dead assertUnstableCommand RESP3 path

cxljs:
Drop the now-unused Cmder.readRawReply interface method and its baseCmd implementation too. readRawReply is unexported and Cmder is a sealed interface, so this is not a public API break. baseCmd.rawVal is retained (search commands still use it directly). The deprecated UnstableResp3 option is retained as well (removing it is a API break).

---

> [!NOTE]
> **Low Risk**
> Internal cleanup of a no-op code path in the core command loop; behavior should be unchanged because the removed branch never selected raw reply parsing.
> 
> **Overview**
> Removes the legacy **unstable RESP3** branch from single-command execution in `baseClient._process`. Responses are always parsed via `cmd.readReply` instead of optionally routing through `readRawReply` when protocol is not RESP2.
> 
> Deletes `assertUnstableCommand`, which had become a stub that always returned “use normal reply parsing.” That matches RediSearch commands now having stable RESP3 parsing without the old `UnstableResp3` workaround at runtime.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e23a589c79979cec97e60de4923003ec663cacd1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -225,7 +225,6 @@ type Cmder interface {
 
 	readTimeout() *time.Duration
 	readReply(rd *proto.Reader) error
-	readRawReply(rd *proto.Reader) error
 	SetErr(error)
 	Err() error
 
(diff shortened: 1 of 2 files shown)


ndyakov (dismissed):
@cxljs I agree that `assertUnstableCommand` is no longer needed and don't see a reason to keep it, but maybe we would like to keep `readRawReply` for backwards compatibility if we decide to expose a rawVal for a specific command? What do you think?

cxljs:
Good call, thanks. I restored readRawReply in command.go.

ndyakov (approved):
Looks good, thank you @cxljs!

PR #3936, opened 2026-07-29
fix(proto): peek push notification name without demanding 36 bytes

### LINKED ISSUE #3935: PubSub.Receive hangs when the subscribe confirmation frame is shorter than 36 bytes (regression in v9.21.0 via #3842)
Note: regression was detected by our CI when a dependency bot auto-updated. Issue report and minimal reproduction was produced by Fable 5.

---

## Expected Behavior

`SUBSCRIBE` to a channel with a short name is confirmed and `PubSub.Receive` returns the `*Subscription`. Works on v9.20.0.

## Current Behavior

On v9.21.0 (and master), `PubSub.Receive` blocks forever when the channel name is 6 characters or fewer (with RESP3, the default). The connection and goroutine stay stuck until other traffic arrives on that connection.

Cause: `PeekPushNotificationName` (changed in #3842) now starts with `r.rd.Peek(36)`. `bufio.Reader.Peek(36)` blocks until 36 bytes are buffered, but the whole confirmation frame

```
>3\r\n$9\r\nsubscribe\r\n$6\r\naaaaaa\r\n:1\r\n
```

is only 35 bytes, and the server has nothing more to send — so the peek waits forever for a byte that never comes. The frame reaches 36 bytes at a 7-character channel name, which is why longer names still work. The name (`subscribe`) is fully parseable from the first 20 bytes; the parser just never gets to run because `Peek` doesn't return.

## Steps to Reproduce

1. `docker run --rm -p 6379:6379 redis:8-alpine`
2. With `github.com/redis/go-redis/v9 v9.21.0`:

```go
client := redis.NewClient(&redis.Options{Addr: "localhost:6379"})
sub := client.Subscribe(context.Background(), "aaaaaa") // ≤6 chars
_, err := sub.Receive(context.Background())             // never returns
```

3. `Receive` hangs. Change the channel to `"aaaaaaa"` (7 chars) and it returns immediately.
4. `go get github.com/redis/go-redis/v9@v9.20.0` — both names work.

<details>
<summary>Full standalone repro (tests both sides of the 36-byte boundary)</summary>

```go
// PubSub.Receive blocks forever on go-redis v9.21.0 when the RESP3
// subscribe-confirmation frame is shorter than 36 bytes, the window
// PeekPushNotificationName passes to bufio.Reader.Peek. Peek waits for
// bytes the server has no reason to send.
//
// The frame Redis answers a single-channel SUBSCRIBE with is
//
//	>3\r\n $9\r\nsubscribe\r\n $<n>\r\n<channel>\r\n :1\r\n
//
// so it reaches 36 bytes at a 7-character channel name. The two names
// below sit either side of that boundary.
//
//	docker run --rm -p 6379:6379 redis:8-alpine
//	go run .
//
// v9.20.0 confirms both. v9.21.0 confirms only the 7-character one.
package main

import (
	"context"
	"fmt"
	"os"
	"time"

)

func main() {
	client := redis.NewClient(&redis.Options{Addr: "localhost:6379"})
	defer client.Close()

	hung := false
	for _, channel := range []string{"aaaaaa", "aaaaaaa"} {
		if confirmed(client, channel) {
			fmt.Printf("ok    SUBSCRIBE %-9q confirmed\n", channel)
			continue
		}
		fmt.Printf("HANG  SUBSCRIBE %-9q never confirmed\n", channel)
		hung = true
	}
	if hung {
		os.Exit(1)
	}
}

// confirmed reports whether Redis acknowledges the subscription within a
// second. A stuck Receive is abandoned rather than waited on.
func confirmed(client *redis.Client, channel string) bool {
	sub := client.Subscribe(context.Background(), channel)
	done := make(chan struct{})
	go func() {
		_, _ = sub.Receive(context.Background())
		close(done)
	}()

	select {
	case <-done:
		return true
	case <-time.After(time.Second):
		return false
	}
}
```

## Context (Environment)

- go-redis v9.21.0 (regression introduced by #3842, the fix for #3839; still present on master), works on v9.20.0
- Redis 8.x, RESP3 (default protocol)
- Go 1.26, macOS (not OS-specific)
- Affects any `Subscribe`/`PSubscribe` where the first confirmation frame is under 36 bytes and the connection is otherwise idle

ndyakov:
## Summary

Fixes #3935 — `PubSub.Receive` hangs forever when the RESP3 subscribe confirmation frame is shorter than 36 bytes (channel names of six or fewer characters). Regression introduced in #3842 (v9.21.0).

## Root cause

`PeekPushNotificationName` started with `bufio.Reader.Peek(36)`, which blocks until 36 bytes are buffered. The confirmation frame for a short channel name (e.g. `>3\r\n$9\r\nsubscribe\r\n$6\r\naaaaaa\r\n:1\r\n`, 35 bytes) is complete, and an otherwise idle server has nothing more to send — so the peek waited for a byte that never arrives. The same applies to any short push frame on an idle connection, e.g. a `message` delivery with short channel and payload.

Existing tests missed this because they use EOF-backed readers: `Peek` returns early at EOF, while a real socket just blocks.

## Fix

Parse the notification name from the bytes already buffered (never blocks). If the parser reports a valid-but-incomplete frame prefix, block for exactly one more byte (`Peek(Buffered()+1)`) and re-parse. An incomplete prefix means the server is still committed to sending the rest of the frame, so that byte is guaranteed to arrive — the loop never waits for bytes the server will not send. The `maxPushHeaderPeek` (4096) safety cap is unchanged.

## Performance

No regression; the buffered hot path is unchanged (one zero-copy peek plus one header parse, zero allocations). `BenchmarkPeekPushNotificationName`:

| case | old ns/op | new ns/op |
|---|---|---|
| Short | ~18 | ~17 |
| Medium | ~20 | ~19 |
| Long | ~32–62 | ~19–24 |

Long names got faster because the old code grew its peek window in a second iteration (36 → 72) while the new code parses once from what is buffered.

## Testing

- New `idleConnReader`-based tests simulate an idle connection that never returns EOF: short subscribe/message/minimal/simple-string frames plus byte-by-byte delivery. All hang without the fix and pass with it.
- Full `./internal/...` suite and `-race` pass.
- Reproduction from the issue verified against a live Redis 8: v9.21.0 hangs on 1- and 6-char channels, this branch confirms all.

---

> [!NOTE]
> **Medium Risk**
> Touches low-level RESP3 read/peek logic used by Pub/Sub; behavior change is intentional but affects all push-name peeks on live connections.
> 
> **Overview**
> Fixes **#3935**, where `PubSub.Receive` could hang forever on RESP3 subscribe confirmations (and other short push frames) shorter than the old **36-byte** `Peek` window on an idle socket.
> 
> **`PeekPushNotificationName`** now parses the push name from whatever is already buffered, then only blocks for **one additional byte** at a time when the prefix is valid but incomplete—never waiting for a fixed chunk the server may never send. Docs are updated to describe this vs. the prior fixed-window growth. The **4096-byte** header cap is unchanged.
> 
> New tests use an **`idleConnReader`** (no EOF until released) plus a timeout wrapper to reproduce the hang: short subscribe/message/minimal frames and byte-by-byte TCP delivery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 278f4ec9931939ed0b053dd8fba52f2d8687dd76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ofekshenawa (approved):
LGTM!

PR #3773, opened 2026-04-11
feat(csc): add command key extraction and caching logic

ofekshenawa:
> [!NOTE]
> **High Risk**
> Large change to connection pooling, command routing, and cache correctness with many concurrency and identity-namespace edge cases; incorrect behavior would serve stale data or break tracking on pooled connections.
> 
> **Overview**
> This PR wires **server-assisted client-side caching (CSC)** for RESP3 clients: cacheable read commands are intercepted, keys are namespaced by DB and fixed ACL identity, and a **background drainer** sweeps idle pool connections for `invalidate` push frames. Entries are attributed to the serving connection (`ConnOwnedCache` / `FulfillOwned`) so closes, handoffs, and pool removal evict coverage instead of serving uninvalidated hits.
> 
> **Command surface:** `csc_commands.go` adds the cacheable-command allow-list, RESP-encoded cache keys, Redis key extraction for invalidations, and guards that reject **SELECT**, **AUTH**, state-changing **HELLO**, **RESET**, raw **SUBSCRIBE**, and **CLIENT TRACKING** when built-in CSC is active (including pipelines). Typed **`ClientTracking` / `ClientTrackingOn` / `ClientTrackingOff`** are added for dedicated connections but are rejected on CSC-enabled clients; `stringArg` now handles `*string` args.
> 
> **Lifecycle & safety:** CSC disables on non–DB-0 config, dynamic credential providers, non-owner caches, pools without idle draining or pool hooks, RESP2 fallback, or tracking rejection. Invalidate handlers stay registered but rebind on close; `runtime.AddCleanup` stops drainers if `Close` is skipped. **`Client.CSCStats()`** exposes hit/miss when the cache supports it.
> 
> **Pool/transport:** Connections gain CSC close/reinit hooks, read-pending markers, hard read deadlines for draining, and safer **MaybeHasData** / health-check behavior for TLS and opaque wrappers. The old **`internal/cache`** package is removed from this diff (implementation lives with the public cache types). Large test suites cover draining, races, multi-DB policy, and e2e invalidation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fdeaf554beb8b7cc8ca18c963c6155314017e00b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3941, opened 2026-07-29
feat(csc): add standalone client-side caching

ofekshenawa:
Combines the client-side caching work from #3748, #3773, and #3794.

---

> [!NOTE]
> **High Risk**
> Large new correctness-sensitive caching path (staleness, ACL isolation, connection lifecycle) on the hot command path; mitigated by broad unit/e2e tests and fail-closed guards.
> 
> **Overview**
> Adds **experimental client-side caching (CSC)** for standalone `*redis.Client` on **RESP3** and **DB 0**, using Redis `CLIENT TRACKING` and RESP3 `invalidate` pushes to cache allow-listed read commands in a bounded local store (`ClientSideCache` / `ClientSideCacheConfig`, default **SharedTracking** strategy).
> 
> The client wires **per-connection tracking**, a **background idle-connection drainer** for push frames, **namespace keys by DB + fixed username**, and **eviction on invalidation, flush, conn close/reinit, and pool removal** (with generation checks to avoid stale fulfills). CSC is **disabled** for non-zero DB, credential providers, RESP2 fallback, sticky pools without draining, and conflicting push handlers; **SELECT/AUTH/HELLO/RESET/raw SUBSCRIBE** and manual **CLIENT TRACKING** are rejected when built-in CSC is active. Public surface also includes typed **`ClientTracking*`** APIs (for dedicated conns), **`CSCStats()`**, README + **`example/client-side-caching`**, and pool/TLS readiness fixes for safe draining.
> 
> Smaller fixes: `stringArg` handles `*string`, OTel records **non-negative** retry counts on cache hits, and a doctest waits for RediSearch indexing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b394a939683af5d5be4562b5088591ac5bad86c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


csc_multidb_test.go
```diff
@@ -0,0 +1,192 @@
...
+// cscMultiDBAddr is the address used by the standalone CSC regression tests
+// in this file. We bypass the Ginkgo harness (which assumes a CI cluster
+// fixture) and connect directly so these tests run with just a single Redis
+// instance available. REDIS_PORT (a bare port) is the env var the suite reads
+// in main_test.go's BeforeSuite, so honor it as a best-effort way to follow
+// the suite's server. (BeforeSuite currently overrides it back to the stack
+// port, so the fallback below matches the suite's effective ":6379"; we read
+// the env var rather than the suite's redisAddr package var because standard
+// testing.T tests run in undefined order relative to BeforeSuite.)
+// REDIS_ADDR remains as a full host:port override for local runs.
```
ndyakov: the term `multidb` is going to be used for `active-active`. it is better to rename this file closer to what it is doing, maybe `csc_native_test` ?

csc_negative_cache_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"strings"
+	"sync/atomic"
+	"testing"
+
+	"github.com/redis/go-redis/v9/maintnotifications"
+)
+
+// TestProcessCached_CachesServerNilReply exercises the complete miss/fill/hit
+// path. The second GET must be answered locally even though the cached command
+// still returns redis.Nil to its caller.
```
ndyakov: can we try to combine some of the csc_*_test files together?

csc_strategy_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"testing"
+
+	"github.com/redis/go-redis/v9"
+)
+
+// TestCSCStrategyDefaultIsSharedTracking pins that the zero value of CSCStrategy is
+// SharedTracking (the default when CSC is enabled without choosing a strategy).
+func TestCSCStrategyDefaultIsSharedTracking(t *testing.T) {
```
ndyakov: let's combine in a csc_test file

options_test.go
```diff
@@ -539,3 +543,70 @@ func TestOptionsCloneMaintNotificationsRace(t *testing.T) {
...
+func (l *capturingLogger) contains(substr string) bool {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	for _, msg := range l.logs {
+		if strings.Contains(msg, substr) {
+			return true
+		}
+	}
+	return false
+}
```
ndyakov: this can be extracted in main_test or some other test util file (check if there isn't something similar already implemented).

internal/proto/reader.go
```diff
@@ -100,6 +105,14 @@ func (r *Reader) PeekReplyType() (byte, error) {
 	return b[0], nil
 }
 
+// MinRESP3ReadBufferSize is the minimum buffer size used when RESP3 push
+// notifications must be inspected without consuming them.
+const MinRESP3ReadBufferSize = 512
```
ndyakov: why is that `512`? I think even `128` should be enough and the way we are reading push notifications will handle even if the notification is bigger than the buffer, as we will read the rest from the socket.

doctests/query_em_test.go
```diff
@@ -325,12 +326,19 @@ func ExampleClient_query_em() {
...
+			"@email:{test\\@redis\\.com}",
+		).Result()
+		if err != nil {
+			panic(err)
+		}
+		if res6.Total == 1 || time.Now().After(deadline) {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
 	}
```
ndyakov: I agree this will make the test more resilient, but can't we just put a sleep of 1 second and a comment "wait for indexing" and then use the previous example as it is shown in the documentation and make more sense as a code snippet to show the use of `FTSearch`

internal/pool/conn.go
```diff
@@ -650,6 +668,18 @@ func (cn *Conn) SetOnClose(fn func() error) {
...
+// previous one. It runs on Close in addition to the SetOnClose callback.
+func (cn *Conn) SetOnCscClose(fn func() error) {
+	cn.onCscClose = fn
+}
+
+// SetOnCscReinit sets the client-side-caching pre-reinitialization hook,
+// overwriting any previous one.
+func (cn *Conn) SetOnCscReinit(fn func()) {
+	cn.onCscReinit = fn
+}
```
ndyakov, line 681: @ofekshenawa do you think there may be a case where we need some feedback from cscOnReInit? If so, let's make `fn` `func() err` and return `nil` for now.

internal/pool/conn_check.go
```diff
@@ -53,7 +62,50 @@ func connCheck(conn net.Conn) error {
...
+		if !ok {
+			return nil, false
+		}
+		conn = unwrapper.NetConn()
+		if conn == nil {
+			return nil, false
+		}
+	}
+	return nil, false
+}
```
ndyakov, line 83: 🤔, so this should help the tls conns to return the syscall.Conn since it has been wrapped ?

internal/pool/conn_check_dummy.go
```diff
@@ -14,7 +14,17 @@ func connCheck(_ net.Conn) error {
 	return nil
 }
 
-// since we can't check for data on the socket, we just assume there is some
+// There is no portable non-consuming readiness check on this platform.
+// Returning true would force every idle CSC connection through a timed read on
+// every drainer tick. The CSC drainer uses needsCscPeriodicProbe instead.
 func maybeHasData(_ net.Conn) bool {
+	return false
+}
```
ndyakov, line 22: but for maintnotifications we should keep this `true`, otherwise we won't check for them?

internal/pool/pool.go
```diff
@@ -843,6 +868,15 @@ func (p *ConnPool) getConn(ctx context.Context) (cn *Conn, err error) {
...
 		}
+		if err == ErrPoolTimeout {
+			atomic.AddUint32(&p.stats.Timeouts, 1)
+			if cb := getMetricConnectionTimeoutCallback(); cb != nil {
+				cb(ctx, nil, "pool")
+			}
+			if cb := GetMetricErrorCallback(); cb != nil {
+				cb(ctx, "POOL_TIMEOUT", nil, "POOL_TIMEOUT", true, 0)
+			}
+		}
```
ndyakov, line 879: this is correct and it will work here because we use named return variables, but i do think it looks cleaner where it was.

internal/pool/pool.go
```diff
@@ -852,29 +886,17 @@ func (p *ConnPool) getConn(ctx context.Context) (cn *Conn, err error) {
...
-			if cb := GetMetricErrorCallback(); cb != nil {
-				cb(ctx, "POOL_TIMEOUT", nil, "POOL_TIMEOUT", true, 0)
-			}
-		}
 		return nil, err
 	}
-	var waitDuration time.Duration
-	if waitTimeCallback != nil {
-		waitDuration = time.Since(waitStart)
-	}
```
ndyakov, line 870: this is supposed to be the waittime for the semaphore, not for a connection.

Also, on the drainer, since we may have used up the timeout (let's say we have waited 1/2 of the pool timeout here), we should wait only the remaining 1/2 of the pooltimeout for the drainer.

internal/pool/pool.go
```diff
@@ -945,6 +967,21 @@ func (p *ConnPool) getConn(ctx context.Context) (cn *Conn, err error) {
...
+	// wait for that bounded maintenance claim and retry. The generation closes
+	// the race where the drainer returns the connection between popIdle and this
+	// check. Normal MaxActiveConns exhaustion still proceeds to newConn and
+	// returns ErrPoolExhausted immediately, preserving the existing contract.
+	if done, retry := p.drainerWaitState(drainGeneration); done != nil {
+		if err = p.waitForDrainer(ctx, done); err != nil {
+			p.freeTurn()
+			return nil, err
+		}
+		goto retryIdle
```
ndyakov, line 987: please see https://github.com/redis/go-redis/pull/3941/changes#r3677770381

Even then, this is not exactly correct as we may wait multiple times here, so we should calculate the remaining timeout on each `waitForDrainer`

internal/pool/pool.go
```diff
@@ -1134,8 +1171,6 @@ func (p *ConnPool) waitTurn(ctx context.Context) error {
 		// Successfully acquired after waiting
 		p.waitDurationNs.Add(time.Now().UnixNano() - start.UnixNano())
 		atomic.AddUint32(&p.stats.WaitCount, 1)
-	case ErrPoolTimeout:
-		atomic.AddUint32(&p.stats.Timeouts, 1)
```
ndyakov, line 1138: since this case is now handled in the caller of waitTurn, we can drop the switch for `if err != nil`

internal/pool/pool_sticky.go
```diff
@@ -40,6 +40,11 @@ type StickyConnPool struct {
 	state atomic.Uint32
 	ch    chan *Conn
 
+	// onFirstConn runs once when this sticky pool claims a connection from its
+	// parent. CSC uses it to revoke cache ownership before the connection leaves
+	// the parent's background drainer.
+	onFirstConn func(*Conn)
```
ndyakov, line 46: should it also drop the cache entries owned by this connection?

push/processor.go
```diff
@@ -73,18 +74,22 @@ func (p *Processor) ProcessPendingNotifications(ctx context.Context, handlerCtx
...
+		// Surface a ReadReply error (unlike the boundary peek errors above,
+		// which consumed nothing): it happens mid-frame after bytes are
+		// consumed, so the conn is desynced and the CSC drainer must remove it.
+		// Normal reply-read callers log-and-ignore this and let their own read fail.
 		reply, err := rd.ReadReply()
 		if err != nil {
-			internal.Logger.Printf(ctx, "push: error reading push notification: %v", err)
-			break
+			return err
 		}
```
ndyakov, line 151: let's log it before returning, so we know the error occurs here (on a normal reply-read).

push/processor.go
```diff
@@ -165,10 +170,12 @@ func (v *VoidProcessor) ProcessPendingNotifications(_ context.Context, handlerCt
...
 		if err != nil {
-			break
-		}
-
-		if willHandleNotificationInClient(notificationName) {
+			// Name too long to peek: still consume the frame below so it isn't
+			// misread as a reply.
+			if !errors.Is(err, proto.ErrPushNotificationNameTooLong) {
+				break
+			}
```
ndyakov, line 236: if we break, we won't read the rest of the frame here, correct?

redis.go
```diff
@@ -1909,25 +2183,90 @@ func (c *baseClient) processPushNotifications(ctx context.Context, cn *pool.Conn
...
+		return false, nil
+	}
+	// Skip only when nothing is buffered (reader) AND nothing on the socket:
+	// MaybeHasData peeks only the socket, but an invalidate can sit in cn.rd.
+	readPending := cn.TakeCscReadPending()
+	periodicReadPending := cn.TakeCscPeriodicReadPending(cscFallbackProbeInterval)
+	hasData := cn.HasBufferedData() || cn.MaybeHasData()
+	if !readPending && !periodicReadPending && !hasData {
+		return false, nil
+	}
```
ndyakov, line 2251: The drainer already pays the syscall that reveals a dead socket, but `maybeHasData` collapses `io.EOF`/socket errors into "no data", so a silently-dead idle connection is re-pooled forever and its owned entries keep being served with zero invalidation coverage (the server dropped its tracking table at disconnect). Nothing else catches this: `isHealthyConn` only runs on `Get`, there is no background reaper, and cache hits are served without touching the pool — a hot fully-cached working set can serve stale values indefinitely with the default `MaxStaleness=0`.

Since the drainer visits every idle conn each round anyway, please surface the death there at no extra cost: give the drain path a probe that distinguishes `(hasData, dead)` (e.g. a `connCheckWithData`-style result from `checkSyscallConn`) and return a connection-fatal error when `dead`, so `DrainIdleConns` removes the conn and `OnRemove → EvictByConn` revokes its entries. Alternatively run `isHealthyConn` inside `drainerPop`. Either way, please add an e2e test: kill the tracking conn server-side (`CLIENT KILL`), keep the working set fully cached, write from another client, and assert the stale value stops being served within a drain round rather than never.

local_cache.go
```diff
@@ -0,0 +1,816 @@
...
+	StaleTimeout time.Duration
+
+	// DrainInterval is the background-drainer period (default 5ms; zero uses the
+	// default): how often idle pool conns are swept for buffered "invalidate"
+	// frames, roughly bounding cache-hit staleness. Values below 1ms are clamped
+	// to 1ms.
+	DrainInterval time.Duration
+
+	// MaxStaleness caps how long a cached entry is served after it became valid,
+	// regardless of invalidation. Zero (default) disables it. It is a correctness
```
ndyakov: can we have a separate
// Default: 0

also, do we want to disable it by default?

redis.go
```diff
@@ -670,20 +747,40 @@ func (c *baseClient) initConn(ctx context.Context, cn *pool.Conn) error {
...
 
-	_, initErr = conn.Pipelined(ctx, func(pipe Pipeliner) error {
+	// trackingEnabled reports whether THIS pool connection must issue
+	// CLIENT TRACKING ON during init. True when CSC (SharedTracking) is enabled:
+	// the shared cache is fed by per-connection tracking + the background
+	// drainer. Once CSC serving stops (owner Close, GC cleanup, or drainer
+	// damping), new and re-inited conns skip tracking — nothing consumes the
+	// pushes into the cache anymore.
+	trackingEnabled := !helloFallbackToRESP2 && c.cscTrackingRequested()
```
ndyakov: `initConn` also runs for dedicated pub/sub connections (`Client.pubSub()`), so every pub/sub conn sends `CLIENT TRACKING ON`, receives FLUSH invalidation frames, and gets CSC close/reinit hooks plus an `initGen` entry — for connections that can never own cache entries. Not a correctness bug (the pubsub pre-read drain consumes the frames, the close hook cleans up), but pure overhead and server-side tracking state. `cn.pubsub` is set before initConn, so:

```suggestion
	trackingEnabled := !helloFallbackToRESP2 && !cn.IsPubSub() && c.cscTrackingRequested()
```

ndyakov (changes requested):
Nice read @ofekshenawa! This looks great when it is all put together. I am leaving a partial review with few blockers, suggestions and few nits. Will continue tomorrow.

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 
 require (
 	github.com/klauspost/cpuid/v2 v2.2.10 // indirect
+	github.com/stretchr/testify v1.7.0 // indirect
 	golang.org/x/sys v0.30.0 // indirect
```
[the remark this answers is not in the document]
ndyakov: was thinking the same but not sure why this ended up here, cc @ofekshenawa

ndyakov (commented):
## Follow-up review of dab65bce

Thanks for the turnaround — I verified the fixes directly against the head commit rather than the descriptions, and the hard correctness items all check out:

- Dead idle connections are now detected by the drainer (`CheckForData` surfacing socket errors) and the `CLIENT KILL` e2e pins exactly the right behavior.
- The 50ms drain budget + non-consuming 50µs probe + `ProcessPendingNotificationsBuffered` resolves the fragmented-frame concern cleanly.
- Hard-deadline clearing, PubSub tracking exclusion, `sameCache`, the binding refcount, O(1) `drainerPop`, and all six metrics issues (including nil-conn safety on the cache-hit operation-duration path via `toConnInfo`) are all confirmed fixed.
- Test consolidation and the buffer-size question from my earlier comments are addressed as well.

**Remaining before merge** — four asks, all small (inline comments on the exact spots):

1. `releaseConn` still re-pools a connection after a mid-frame push-processing error (the TODO in `drainPushNotifications`); ~4-line fix.
2. `internal/pool/conn_check_wrapped_test.go` needs the unix build tag — it currently breaks `go test ./internal/pool` on Windows.
3. An "Experimental" note on the new API surface (README section + entry-point godocs). This is the cheap way to keep the API-shape questions (Cache interface width/contract, `CSCStats` signature, `CacheEntry` export, `ClientSideCacheConfig`/`CSC*`/`Cache*` naming, `push.WillHandleNotificationInClient`) from becoming frozen one-way doors — they can then be follow-ups instead of blockers.
4. A short README paragraph listing the commands CSC rejects (SELECT/AUTH/HELLO <ver>/RESET/raw SUBSCRIBE/CLIENT TRACKING, including inside pipelines) — users will hit these guards in their first hour.

**Follow-ups I'd file rather than block on:** same-client deterministic read-your-own-writes test (no Eventually, PoolSize=1); hook-mutex serialization on the cache-fill path (benchmark ask); SWAPDB guard; server identity in the shared-cache namespace; changelog notes for the Windows `maybeHasData` flip and the TLS unwrap on the non-CSC release path; `VoidProcessor` error-contract alignment; a handful of minor notes on this delta itself; the smaller doc/style items from the earlier round.

redis.go
```diff
@@ -848,6 +988,13 @@ func (c *baseClient) releaseConn(ctx context.Context, cn *pool.Conn, err error)
 		if err := c.processPushNotifications(ctx, cn); err != nil {
```
ndyakov, line 988: Now that the drainer treats a mid-frame push-processing error as connection-fatal — and this delta extended that rule to the probe and socket-check paths — `releaseConn` remains the one place that sees the same evidence and re-pools anyway. A mid-frame timeout typically fires on a bufio fill with the reader buffer empty, so the `HasBufferedData` guard in `putConn` won't catch it, and the next command on that conn reads a shifted reply stream. The `maybeHasData` TLS unwrap also makes this window reachable for RESP3+TLS users with CSC off, so I'd rather not merge with the `drainPushNotifications` TODO still open. Suggested fix:

```go
if err := c.processPushNotifications(ctx, cn); err != nil {
    internal.Logger.Printf(ctx, "push: error processing pending notifications before releasing connection: %v", err)
    if isBadConn(err, false, c.opt.Addr) {
        // Mid-frame read failure: the reader may be desynced; re-pooling
        // would corrupt the next command's reply (same rule as the CSC drainer).
        c.connPool.Remove(ctx, cn, err)
        return
    }
}
```

internal/pool/conn_check_wrapped_test.go
```diff
@@ -0,0 +1,102 @@
+package pool_test
```
ndyakov, line 3: This file still needs the unix build-tag guard (the same one `conn_check_test.go` has): `TakeCscPeriodicReadPending(1ms)` returning false and `MaybeHasData()` flipping true contradict the `!unix` dummy implementations (`needsCscPeriodicProbe` always true, `maybeHasData` always false), so `go test ./internal/pool` fails on Windows — and CI is Linux-only, so it won't be caught there. A small `!unix` counterpart pinning the dummy contract would be welcome too, but the tag itself is the merge-blocker.

options.go
```diff
@@ -313,12 +316,68 @@ type Options struct {
...
+	// with ClientSideCache it is the on/off switch for the feature: leave both
+	// nil to disable CSC, set either one to enable it. If ClientSideCache is also set, it
+	// takes precedence over this config.
+	//
+	// Client-side caching is disabled when CredentialsProvider,
+	// CredentialsProviderContext, or StreamingCredentialsProvider is set:
+	// provider-backed credentials can change the ACL identity after the cache
+	// namespace is selected. Fixed Username/Password values are supported and
+	// included in the cache namespace.
+	ClientSideCacheConfig *ClientSideCacheConfig
```
ndyakov, line 332: Before this ships: can we add an explicit "Experimental: this API may change in a minor release" note here, on `ClientSideCache`, and on `Cache`/`NewLocalCache`/`CSCStats`/`CSCStrategy` (plus one line in the README section)? Repo precedent is the Array API and `StreamingCredentialsProvider`, both much smaller surfaces than this one.

That one paragraph lets the open API-shape questions land as fast-follows instead of merge-blockers: trimming/documenting the `Cache` interface (4 of its 11 methods are never called by the client, the Reserve/Fulfill/Cancel invariants custom implementations must uphold are undocumented, and SharedTracking de-facto requires `ConnOwnedCache`), `CSCStats() (hits, misses uint64)` vs a growable struct like `PoolStats`, unexporting `CacheEntry`/`CacheEntryState`, the `ClientSideCacheConfig = CacheConfig` alias + `CSC*` naming split, and the uncalled `push.WillHandleNotificationInClient`. Without the experimental label, at least the irreversible ones (unexport, stats struct, dead export) need to happen before the freeze.

README.md
```diff
@@ -285,6 +286,41 @@ rdb := redis.NewClient(&redis.Options{
...
+    Protocol: 3,
+    DB:       0,
+    ClientSideCacheConfig: &redis.ClientSideCacheConfig{
+        MaxEntries: 10_000,
+    },
+})
+defer rdb.Close()
+```
+
+Client-side caching currently requires RESP3, a standalone client, and database
```
ndyakov, line 314: The limitation list still doesn't mention the command guards, and users will hit them in the first hour of real use: with CSC enabled, `SELECT`, `AUTH`, `HELLO <ver>`, `RESET`, raw `Do(ctx, "subscribe", ...)`, and `CLIENT TRACKING` are rejected — including inside pipelines, where one guarded command fails the whole pipeline — while the typed `Subscribe`/`PSubscribe` APIs remain fine (dedicated connections). Please add a short "the following commands are rejected while client-side caching is enabled" paragraph here.

Related (fine as a fast-follow under the experimental label): the sentinel errors in csc_integration.go are all unexported, so callers can't `errors.Is` them — consider exporting them or a single wrappable `ErrCommandBlockedByCSC`.

internal/pool/conn.go
ofekshenawa: No, the hook only evicts local cache state and cannot fail, so returning an error would add unused plumbing.

internal/pool/conn_check.go
ofekshenawa: Yes, it unwraps TLS and compatible wrappers until it reaches the underlying syscall-capable socket used for readiness checks.

internal/pool/conn_check_dummy.go
ofekshenawa: No, returning true would force a timed read on every check. normal resp3 reads still process maintenance notifications, while CSC uses bounded periodic probes.

push/processor.go
ofekshenawa: Yes, the peek has not consumed the frame, so leaving it buffered lets the normal read retry after the rest arrives.

local_cache.go
ofekshenawa: Yes, zero is intentional because Redis invalidations are the primary mechanism and MaxStaleness is an optional refetch cost backstop.

redis.go
```diff
@@ -1909,25 +2205,90 @@ func (c *baseClient) processPushNotifications(ctx context.Context, cn *pool.Conn
...
+	if err != nil {
+		// The built-in processor surfaces mid-frame ReadReply errors (a benign
+		// boundary peek timeout returns nil). allowTimeout=false: such an error
+		// means bytes were consumed mid-frame, leaving the conn desynced —
+		// re-pooling would corrupt the next command's reply, so remove it.
+		if _, builtin := c.pushProcessor.(*push.Processor); builtin {
+			if isBadConn(err, false, c.opt.Addr) {
+				return true, err // fatal read/protocol/connection error — remove the conn
+			}
+			return true, nil
```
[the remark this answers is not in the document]
ofekshenawa: This is no longer applicable. The built in processor logs handler errors itself and only returns read or protocol errors that are fatal to the connection.

csc_commands.go
```diff
@@ -0,0 +1,257 @@
...
+	// Numkeys pattern: numkeys at args[1], keys from args[2].
+	case "sintercard", "zdiff", "zinter", "zunion":
+		if argsLen < 3 {
+			return nil
+		}
+		numKeys, err := strconv.Atoi(cmd.stringArg(1))
+		if err != nil || numKeys <= 0 {
+			return nil
+		}
+		keys := make([]string, 0, numKeys)
```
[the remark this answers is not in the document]
ndyakov: This is correct, however @ofekshenawa maybe we can just skip raw commands (e.g. `Do`/ `DoRaw`) from CSC? As a follow up.

ndyakov (dismissed):
I think we reached a great state that can be included in master. Thank you @ofekshenawa!

internal/pool/conn_check_dummy.go
ndyakov: Fair point about the forced timed read — agreed, reverting to `true` would be worse. The pre-reply drain covers the common case on Windows, so we're mostly fine.

There is still one narrow gap though. Example:

```go
// conn executes GET; while waiting for the reply, the server pushes MOVING
// (it lands in the socket after the reply bytes)
val, _ := rdb.Get(ctx, "k").Result()
// reply is read, push stays unread; releaseConn's drain is now a no-op on
// windows (MaybeHasData == false), and the conn goes idle
// → MOVING sits unprocessed until this conn is used again, which can be minutes
```

Before this PR the release-time drain picked that up. Instead of going back to `true`, can we widen the gate to also fire on already-buffered data?

```go
if !cn.MaybeHasData() && !cn.HasBufferedData() {
    return nil
}
```

`HasBufferedData` is a pure in-memory check — no syscall, no timed read — and it recovers the case where the push bytes were already pulled into the reader buffer together with the reply. Bonus: same fix helps TLS on unix (decrypted bytes held by the wrapper that the raw-socket peek can't see).

The remaining case (push still in the socket kernel buffer, conn idles, CSC off) I'm fine accepting — but let's add a line to the release notes that maintenance notifications on non-syscall platforms may be delayed until the connection is next used.

ndyakov (commented):
Two suggestions for future improvements — intentionally posted as a single global note (not line comments) so there is nothing to resolve before merge. Neither blocks this PR.

**1. Bound `numKeys` before allocating the key slice** (`csc_commands.go`, `extractRedisKeys`, numkeys-pattern branch)

`numKeys` comes straight from the command argument, so with CSC enabled a malformed raw call like `Do(ctx, "zinter", "1000000000", "key")` allocates a slice with a ~1e9 capacity before Redis ever rejects the command — without CSC the same call just returns a server error. The read loop is already bounded by `argsLen`; the allocation should be too:

```go
if numKeys > argsLen-2 {
    return nil
}
keys := make([]string, 0, numKeys)
```

**2. Widen the release-time drain gate to already-buffered data** (`redis.go`, `peekAndProcessPushNotifications`)

```go
if !cn.MaybeHasData() && !cn.HasBufferedData() {
    return nil
}
```

`HasBufferedData` is a pure in-memory check (no syscall, no timed read). It recovers release-time push processing in two cases the raw-socket peek can't see: on Windows/wasm (where `maybeHasData` is always false, so this path is currently a no-op) whenever push bytes were already pulled into the reader buffer together with the reply, and on unix for TLS-like wrappers holding decrypted bytes after the raw socket is drained. Both otherwise wait until the connection's next use — for maintenance notifications on an idling connection that can be a while.

csc_integration.go
```diff
@@ -0,0 +1,922 @@
...
+// cscDrainHandle owns the drainer lifecycle and serializes client teardown.
+// stop signals shutdown; done is closed on exit so Close can join.
+type cscDrainHandle struct {
+	stop              chan struct{}
+	done              chan struct{}
+	stopOnce          sync.Once
+	teardownOnce      sync.Once
+	handlerCloseOnce  sync.Once
+	closeOnce         sync.Once
+	closeErr          error
```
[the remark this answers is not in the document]
ndyakov: False positive: both fields are used in `baseClient.Close` (redis.go:1397-1400) — `closeOnce` dedupes `closeResources` across repeat `Close` calls and `closeErr` preserves its result. Resolving.

ndyakov (approved):

PR #3943, opened 2026-07-31
fix(script): keep client-computed SHA-1 in Script.Load

saddamr3e:
Reproducer with a Scripter whose SCRIPT LOAD answers with a digest of its own choosing:

    --- FAIL: TestScriptLoad_KeepsClientComputedDigest_RO
        script_load_digest_test.go:46: EVALSHA_RO sent "0123456789abcdef0123456789abcdef01234567", want "7f923f79fe76194c868d7e1d0820de36700eb649"

NewScript computes hex(sha1(src)) so EVALSHA stays pinned to that script body, but Load overwrote it with the server's string. One reply repoints every later Run/RunRO on that *Script at whatever Lua the peer has cached under the returned digest, running with the caller's keys and ARGV, and EVALSHA still reports success.

Overwriting is only correct for NewScriptServerSHA, where the assignment came from in #3700; kept it on that path and made the client-side-SHA path report the mismatch instead.

---

> [!NOTE]
> **Medium Risk**
> Touches script execution identity for all `NewScript` users who call `Load`, but behavior is a correctness fix with explicit error on protocol mismatch rather than silent wrong-script execution.
> 
> **Overview**
> **`Script.Load` no longer replaces the client SHA-1 from `NewScript` with whatever `SCRIPT LOAD` returns.** For normal scripts, the digest stays `SHA-1(src)` so later `Run`/`RunRO` still call `EVALSHA` with the intended Lua body. If Redis returns a non-empty digest that does not match, `Load` surfaces **`ErrScriptDigestMismatch`** instead of silently repointing execution at another cached script (same keys/ARGV).
> 
> **`NewScriptServerSHA` is unchanged:** `Load` still stores the server digest. Empty `SCRIPT LOAD` replies (e.g. pipelined commands before `Exec`) are ignored for mismatch checks.
> 
> Unit tests extend `fakeScripter` to assert which SHA is sent on `EVALSHA`/`EVALSHA_RO` and cover mismatch, match, empty, and server-SHA paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2962ac5fa925b21d91876ef244afb996785a47bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ndyakov:
Thank you for the contribution @saddamr3e, To set expectations - I review it once we release the GA version so it won't make it in 9.22.

saddamr3e:
Sounds good, no rush. I'll keep the branch up to date in the meantime.

saddamr3e:
any update?

ndyakov:
@saddamr3e thank you for pinging me on this one, it slipped from my list. Will check it till EOW, cc @vlady-kotsev may check it as well.

script_load_digest_test.go
```diff
@@ -0,0 +1,76 @@
+package redis
```
ndyakov, line 1: can we fold this into script_server_sha_unit_test.go

script.go
```diff
@@ -60,11 +61,25 @@ func (s *Script) Hash() string {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if !s.serverSHA {
+		// s.hash is SHA-1(s.src) computed in NewScript, and SCRIPT LOAD returns
+		// that same digest. A different value is a protocol violation; keeping
+		// the local one means EVALSHA stays pinned to this script instead of
+		// running whatever the server holds under the returned digest.
+		if cmd.Val() != s.hash {
+			cmd.SetErr(fmt.Errorf("redis: SCRIPT LOAD returned digest %q, want %q", cmd.Val(), s.hash))
```
ndyakov, line 77: let's export this message as ErrScriptDigestMismatch and/or wrap so error.Is works for detecting it.

ndyakov (dismissed):
Other than the two nits, the rest looks good to me, thank you @saddamr3e! Feel free to address and ping me to merge.

script.go
ndyakov, line 77: also, on pipeline this Load can trigger before the exec, so c.Val may be empty here. So maybe the check should be 
```
if v := cmd.Val(); v != "" && v != s.hash {
```

script_load_digest_test.go
saddamr3e: Done, moved them into script_server_sha_unit_test.go and dropped the extra file.

script.go
saddamr3e: Added ErrScriptDigestMismatch and wrapped it with %w, so errors.Is(err, redis.ErrScriptDigestMismatch) works and the message still carries both digests. The test checks errors.Is now instead of just err != nil.

saddamr3e: Good catch, took that form. An empty Val now falls through with no error and s.hash untouched, with a test for it. The pipelined reply never gets compared after Exec, but that matches how Load behaved before and the local digest is still what EVALSHA sends.

saddamr3e:
@ndyakov pushed the three nits: ErrScriptDigestMismatch exported and wrapped with %w, empty Val from a pipelined Load skipped, tests folded into script_server_sha_unit_test.go. Script tests are green under -race, should be good to merge.

ndyakov:
@saddamr3e thank you, will look into this today.

PR #3956, opened 2026-08-09
fix(search): read the full RESP3 map reply in FTHybridCmd

saddamr3e:
FTHybridCmd.readReply parsed the FT.HYBRID reply with rd.ReadSlice(). Feeding it the RESP3 map reply FT.HYBRID returns on a RESP3 connection:

    connection desynced: next reply = "format", want int64(99)

ReadSlice reads a map header's declared length as an element count, so a map of N key/value pairs (2N frames) is only half consumed and the trailing frames stay buffered. parseFTHybrid still succeeds on the half list and readReply returns nil, so the connection goes back to the pool desynced and the next command reads a stray map frame as its own reply.

FTHybridCmd was the only FT.* reply parser without the PeekReplyType/RespMap branch that AggregateCmd, FTInfoCmd, FTSpellCheckCmd, FTSearchCmd and FTSynDumpCmd already use. Read a map with ReadReply and flatten it into the key/value list parseFTHybrid expects; the RESP2 array path is unchanged.

---

> [!NOTE]
> **Low Risk**
> Targeted reply-parser fix aligned with existing FT.* RESP3 handling; low blast radius beyond FT.HYBRID on RESP3 connections.
> 
> **Overview**
> Fixes **connection desync** on RESP3 when `FT.HYBRID` returns a map reply. `FTHybridCmd.readReply` no longer uses `ReadSlice` for that case; it now matches other FT parsers by peeking the reply type, reading the full map with `ReadReply`, and flattening it for `parseFTHybrid`. The RESP2 flat-array path is unchanged.
> 
> Adds unit tests that wire a synthetic RESP3 map (and RESP2 array) plus a following reply, and assert the reader stays aligned for the next command.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 717d61e7b15f0a4cbf9ba166e8926f93b0385508. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/search_commands.go
+++ b/search_commands.go
@@ -3125,9 +3125,35 @@ func parseFTHybrid(data []interface{}, withCursor bool) (FTHybridResult, *FTHybr
 }
 
 func (cmd *FTHybridCmd) readReply(rd *proto.Reader) (err error) {
-	data, err := rd.ReadSlice()
+	readType, err := rd.PeekReplyType()
 	if err != nil {
 		return err
+	}
+
+	// RESP3 returns a map, RESP2 returns an array. ReadSlice reads a map
+	// header's declared length as an element count, so it consumes only half
+	// the key/value frames and leaves the rest on the connection; ReadReply
+	// consumes the whole map. Flatten it into the key/value list parseFTHybrid
+	// expects, matching AggregateCmd/FTSearchCmd/FTSpellCheckCmd/FTSynDumpCmd.
+	var data []interface{}
+	if readType == proto.RespMap {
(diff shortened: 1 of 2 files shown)


ndyakov:
Thank you @saddamr3e, will look at this till EOW.

vlady-kotsev:
Hey @saddamr3e , everything looks correct. Thanks for the fix!

vlady-kotsev (approved):

PR #3960, opened 2026-08-10
feat(csc): derive command eligibility

ofekshenawa:
## Summary
This PR introduces metadata-driven command eligibility for client-side caching and adds an optional dynamic metadata resolver. CSC now uses Redis COMMAND metadata to determine whether a command is safe to cache and to extract every key on which its reply depends. Unknown, malformed, incomplete, blocking, nondeterministic, script-based, or explicitly non-cacheable commands fail closed: they still execute normally but are not cached. The default resolver uses a generated static snapshot with no additional network calls or background work; the optional live resolver refreshes metadata from the connected server. Following the updated HLD, CSC eligibility and Cluster routing consume the same resolved command record so flags, key positions, and policies cannot drift between separate metadata sources.
Refs CAE-3353

---

> [!NOTE]
> **High Risk**
> Changes core cluster slot/key extraction, fan-out admission, and CSC routing from shared metadata; parser or generation-mixing bugs could misroute commands or serve stale cache entries.
> 
> **Overview**
> Introduces **shared command metadata** as the single source for **client-side caching eligibility** and **Redis Cluster routing**, with a fail-closed `COMMAND` parser and updated architecture docs.
> 
> The **`COMMAND` reply parser** in `command.go` is rewritten to populate **`Tips`**, **`KeySpecs`**, and legacy wire-shape tracking; malformed or conflicting records become **nil tombstones** without breaking siblings, while framing/nesting failures fail the whole refresh. **`CLUSTER SHARDS`** now treats a **null endpoint** as “reuse the connection host.”
> 
> **Autopipeline** replaces a global “CSC active” probe with **`autopipelineCSCEligible(cmd)`** so solo cacheable commands use the `Process` path only when CSC is on **and** metadata says the command is cacheable (including overrides). A **`commandDone`** hook runs on every exit path so owners can release per-command state tied to metadata generations.
> 
> **README** adds a **cluster fan-out safety** note (multi-shard ops are not atomic). **`.claude/specs/cluster-routing.md`** is expanded to describe immutable routing decisions, metadata resolution order, pipeline/AutoPipeline admission, and topology/metadata refresh behavior. Tests add broad **`command_info_parser_test.go`** coverage and cluster test skips via **`skipIfClusterUnhealthy`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c80af6e066774c46e57007a6349b781500e4b36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


command.go
```diff
@@ -5369,134 +5393,313 @@ func (cmd *CommandsInfoCmd) String() string {
...
-		cmdInfo.FirstKeyPos = int8(firstKeyPos)
 
-		lastKeyPos, err := rd.ReadInt()
+		cmdInfo.Tips = make([]string, 0, tipsLen)
+		rawTips := make(map[string]string, tipsLen)
+		if cmdInfo.ReadOnly {
+			rawTips[routing.ReadOnlyCMD] = ""
+		}
+		for f := 0; f < tipsLen; f++ {
+			tip, err := rd.ReadString()
```
ndyakov: This used to be a `DiscardNext()`, so a `Nil` tip element here didn't matter. Now it fails the whole COMMAND reply. Could we handle `err == Nil` and skip it, the way the `Flags` loop does? Same idea for the rest of this parsing: skip flag/tip/spec-field names we don't recognize instead of erroring, and if a known field comes back as the wrong type, mark the spec incomplete (so it won't cache) rather than stopping the parse. The goal is that an unfamiliar or newer shape leaves `Command()` returning what it understood, never failing the reply.

command.go
```diff
@@ -5369,134 +5393,313 @@ func (cmd *CommandsInfoCmd) String() string {
...
-				switch s, err := rd.ReadString(); {
-				case err == Nil:
-					cmdInfo.ACLFlags[f] = ""
-				case err != nil:
-					return err
-				default:
-					cmdInfo.ACLFlags[f] = s
+			ks.Flags = make([]string, 0, fn)
+			for f := 0; f < fn; f++ {
+				flag, err := rd.ReadString()
```
ndyakov: Same here for the flags loop: a missing or `Nil` flag shouldn't error out, just skip it. Otherwise one odd element from a module or a newer server takes down all of `Command()`.

command.go
```diff
@@ -5369,134 +5393,313 @@ func (cmd *CommandsInfoCmd) String() string {
...
-					rawTips[tip] = ""
-				} else {
-					// Handle normal key:value tips
-					rawTips[k] = v
+				switch name {
+				case "keyword":
+					if ks.Keyword, err = rd.ReadString(); err != nil {
+						return err
+					}
+				case "index", "startfrom", "lastkey", "keystep", "limit", "keynumidx", "firstkey":
```
ndyakov: We list these field names here for the bounds check and then again in the `switch name` just below. Could we fold it into one? Right now adding a numeric field means touching both, and if you forget the inner switch the value gets dropped with no error.

command_metadata.go
```diff
@@ -0,0 +1,710 @@
...
+	view := buildCommandMetadataView(records, s.overrides)
+	view.live = true
+	s.publishLiveView(fpStart, view)
+	return nil
+}
+
+// commandMetadataTrusted checks two signals introduced in Redis 8.10 —
+// script_runner on EVAL_RO and the nondeterministic tip on TTL — whose
+// absence means the output cannot carry the negative eligibility signals.
+func commandMetadataTrusted(records map[string]*CommandInfo) bool {
```
ndyakov: What we actually want to know here is whether the server is 8.10 or newer. Inferring that from the `eval_ro` and `ttl` flags feels fragile: if either flag ever moves, a real 8.10 server gets treated as untrusted and stays stuck on the static snapshot. Could we read the version straight from the `HELLO` reply (or a version fetch) instead?

command_metadata.go
```diff
@@ -0,0 +1,710 @@
...
+			s.current.Store(s.static)
+		}
+		s.mu.Unlock()
+		if firstUntrusted {
+			internal.Logger.Printf(context.Background(),
+				"redis: ignoring live command metadata: the server does not expose the Redis >= 8.10 eligibility signals")
+		}
+		return nil
+	}
+	view := buildCommandMetadataView(records, s.overrides)
```
ndyakov: We rebuild the whole table and re-hash ~600 entries on every tick, even when the reply hasn't changed. Could we hash the raw records first and skip the rebuild when they match what we already have?

command_metadata.go
```diff
@@ -0,0 +1,710 @@
...
+		refresh:         make(chan struct{}, 1),
+		stop:            make(chan struct{}),
+		done:            make(chan struct{}),
+	}
+	s.fetchCtx, s.fetchCancel = context.WithCancel(context.Background())
+	s.static = buildCommandMetadataView(nil, overrides)
+	s.current.Store(s.static)
+	// An override keyed by a bare container name ("memory") is pruned in
+	// favor of its subcommand entries and would otherwise vanish silently.
+	for name, info := range overrides {
```
ndyakov, line 694: This re-derives the bare-container pruning that `buildCommandMetadataView` already did. If we return the pruned set from there, the two can't drift apart later.

command_metadata.go
```diff
@@ -0,0 +1,710 @@
...
+// metadataView returns the client's current command-metadata view; clients
+// without a store share the static default.
+func (c *baseClient) metadataView() *commandMetadataView {
+	if s := c.cmdMeta; s != nil {
+		return s.view()
+	}
+	return defaultCommandMetadataView
+}
+
+// fetchCommandMetadata retrieves live COMMAND output through the retry path,
```
ndyakov: The comment says this bypasses application hooks, but that's only the process/pipeline chain. Any connection dialed for this fetch still runs the Dial, OnConnect and reauth hooks, since we clone the app hooks just below. Worth rewording so it's clear only the HELLO/COMMAND process hooks are skipped.

command_metadata.go
```diff
@@ -0,0 +1,710 @@
...
+// bypassing application hooks and the cache.
+func (c *baseClient) fetchCommandMetadata(ctx context.Context) (map[string]*CommandInfo, error) {
+	// Metadata refresh owns a bounded internal context independently of the
+	// application's ContextTimeoutEnabled and socket-timeout choices. Run the
+	// command through a lightweight clone whose zero socket timeouts defer to
+	// that context deadline; otherwise ReadTimeout=-1 would leave a stalled
+	// COMMAND holding a pooled connection after refreshOnce stopped waiting.
+	fetchClient := c.withTimeout(0)
+	fetchClient.hooksMixin = c.hooksMixin.clone()
```
ndyakov, line 1103: Looks redundant but it's actually load-bearing: `withTimeout` doesn't run `initHooks`, so `fetchClient.hooksMixin` has a nil state, and when the fetch inits a fresh pooled conn, `initConn` calls `newConn(..., &fetchClient.hooksMixin, ...)`, which clones it and reads that state (nil would panic). Let's keep it for this PR. The cleaner fix is moving the clone into `withTimeout` so the primitive never hands back nil hooks, but that touches a shared path unrelated to this work, so I'd do it as a separate bugfix. Could you add a `TODO` here pointing at that follow-up so nobody deletes this line as dead code?

csc_commands.go
```diff
@@ -2,128 +2,304 @@ package redis
...
+	"memory|usage": {"dont_cache"},
+}
+
+// cscNegativeBits are the bits that each rule out caching.
+const cscNegativeBits = cscFlagScriptRunner | cscFlagBlocking |
+	cscTipNondeterministicOutput | cscTipDontCache
+
+// cscAckExtraKeySpecs: commands whose extra unknown key specs are handled
+// elsewhere (sort_ro's is the BY/GET patterns, rejected per invocation).
+var cscAckExtraKeySpecs = map[string]bool{
```
ndyakov: `sort_ro` is special-cased here and again at line 347, so the two have to stay in sync by hand. Could we carry this as an override record that the general extractor reads, instead of hardcoding the name?

csc_commands.go
```diff
@@ -171,87 +339,67 @@ func keyArg(cmd Cmder, pos int) (string, bool) {
...
-// argument cannot be rendered in its wire form (see keyArg).
-func extractRedisKeys(cmd Cmder) []string {
-	firstKey := cmdFirstKeyPosWithInfo(cmd, nil)
-	if firstKey == 0 {
+// cscExtractRedisKeys lists the keys the cache must watch for a call to a
+// command already resolved to meta. nil = serve this call uncached; a
+// PARTIAL list is never returned.
+func cscExtractRedisKeys(meta cscCommandMeta, cmd Cmder) []string {
+	// SORT_RO with BY/GET reads pattern keys this call cannot list.
+	if cmd.Name() == "sort_ro" && sortROHasByGet(cmd) {
```
ndyakov: Hardcoded `sort_ro` check again, same issue as the `cscAckExtraKeySpecs` entry above. Anything else with `BY`/`GET`-style pattern keys just falls through to uncached. I'd rather see this in an override record than a name check here.

redis.go
```diff
@@ -2160,6 +2199,14 @@ func NewClient(opt *Options) *Client {
 		// triggers these cleanups, which stop them. See cscRegisterCleanups.
 		cscRegisterCleanups(&c)
 	}
+	// The metadata store exists only when client-side caching attached; a
+	// non-default config on a client without CSC would otherwise be dropped
+	// with no signal.
+	if cfg := opt.CommandMetadata; cfg != nil && c.baseClient.cmdMeta == nil &&
```
ndyakov: The metadata store only gets built when client-side caching is on, so a `CommandMetadata` config on a client without CSC is dropped here with just a log. This metadata is really a shared layer: caching uses it, and cluster routing will want it too. Could we build the store whenever `CommandMetadata` is set (or on any RESP3 client) and let caching just be one consumer? Then this warning goes away and routing can read the same store later.

ndyakov (changes requested):
Thanks, great base. We still need to align this to the design.

The command metadata should be the fundamental piece, with resolvers built around it: one for CSC and one for routing. Right now it is used only for CSC, and populated only when CSC is enabled. Let's make it a shared layer that both consumers read.

The command parser should also be forwards compatible: skip shapes it doesn't recognize and degrade the command to uncached, instead of failing the whole `Command()` reply.

Details in the inline comments.

PR #3968, opened 2026-08-17
fix(pubsub): avoid panics on Channel conflict and Ring empty subscribe

### LINKED ISSUE #3761: PubSub.Channel and Ring.Subscribe/PSubscribe/SSubscribe panic instead of returning error
## Summary

Several exported methods on `PubSub` and `Ring` call `panic()` on recoverable error conditions instead of returning errors. Callers cannot anticipate or guard against these panics because the function signatures do not include error returns and the doc comments do not mention panic behavior.

**Version:** go-redis/v9 (commit 68197a9)
**Go:** 1.24

### 1. `PubSub.Channel()` / `ChannelWithSubscriptions()` — mutual exclusion panic

```go
// pubsub.go:562-597
func (c *PubSub) Channel(opts ...ChannelOption) <-chan *Message {
    c.chOnce.Do(func() {
        c.msgCh = newChannel(c, opts...)
        c.msgCh.initMsgChan()
    })
    if c.msgCh == nil {
        err := fmt.Errorf("redis: Channel can't be called after ChannelWithSubscriptions")
        panic(err)  // panics instead of returning error
    }
    return c.msgCh.msgCh
}
```

The doc comment for `ChannelWithSubscriptions` says "can not be used together with Channel or ChannelSize" but does not mention that violating this constraint causes a panic. A caller who accidentally uses both gets an unrecoverable crash.

### 2. `Ring.Subscribe()` / `PSubscribe()` / `SSubscribe()` — empty channels + shard error panic

```go
// ring.go:684-722
func (c *Ring) Subscribe(ctx context.Context, channels ...string) *PubSub {
    if len(channels) == 0 {
        panic("at least one channel is required")
    }
    shard, err := c.sharding.GetByKey(channels[0])
    if err != nil {
        // TODO: return PubSub with sticky error
        panic(err)
    }
    return shard.Client.Subscribe(ctx, channels...)
}
```

All three methods (`Subscribe`, `PSubscribe`, `SSubscribe`) have the same two panic paths: (a) empty channels slice, (b) shard lookup failure. The source even has `// TODO: return PubSub with sticky error` acknowledging the panic should be replaced.

## Reproduction

```go
package main

import (
    "context"
)

func main() {
    // Bug 1: PubSub mutual exclusion panic
    client := redis.NewClient(&redis.Options{Addr: "localhost:6379"})
    pubsub := client.Subscribe(context.Background())
    _ = pubsub.ChannelWithSubscriptions()
    _ = pubsub.Channel()
    // panic: redis: Channel can't be called after ChannelWithSubscriptions

    // Bug 2: Ring empty channels panic
    ring := redis.NewRing(&redis.RingOptions{})
    ring.Subscribe(context.Background())
    // panic: at least one channel is required
}
```

## Why This Matters

- `PubSub.Channel()`, `Ring.Subscribe()`, `Ring.PSubscribe()`, and `Ring.SSubscribe()` are exported public API methods on core types
- None of the doc comments mention panic behavior
- The panics are triggered by recoverable misuse (calling methods in wrong order, empty arguments, shard errors) — exactly the cases where returning an error would be appropriate
- The existing `// TODO` comments in ring.go confirm the maintainers already identified this as a problem

## Suggested Fix

For `PubSub.Channel` / `ChannelWithSubscriptions`: return a closed channel and log a warning, or add an error return value.

For `Ring.Subscribe` / `PSubscribe` / `SSubscribe`: return a `*PubSub` with a sticky error (as the TODO suggests), or change the signature to return `(*PubSub, error)`.

```go
// Example for Ring.Subscribe:
func (c *Ring) Subscribe(ctx context.Context, channels ...string) (*PubSub, error) {
    if len(channels) == 0 {
        return nil, fmt.Errorf("redis: at least one channel is required")
    }
    shard, err := c.sharding.GetByKey(channels[0])
    if err != nil {
        return nil, fmt.Errorf("redis: %w", err)
    }
    return shard.Client.Subscribe(ctx, channels...), nil
}
```

VedantMadane:
## Summary
Fixes #3761.

- `PubSub.Channel` / `ChannelWithSubscriptions` mutual-exclusion misuse no longer panics; returns a closed channel and logs a warning.
- `Ring.Subscribe` / `PSubscribe` / `SSubscribe` no longer panic on empty channels or shard lookup failure; return a PubSub with a sticky error (as the existing TODO suggested).

## Test plan
- [x] `TestRingSubscribeEmptyChannelsNoPanic`
- [x] `TestPubSubChannelMutualExclusionNoPanic`

---

> [!NOTE]
> **Low Risk**
> Defensive error-handling in Pub/Sub and Ring; behavior shifts from panic to errors/closed channels without changing successful subscription paths.
> 
> **Overview**
> Fixes **#3761** by replacing several Pub/Sub panics with recoverable behavior.
> 
> **Ring** `Subscribe`, `PSubscribe`, and `SSubscribe` no longer panic when channels are missing or shard lookup fails (e.g. empty ring). They return a `PubSub` built via `failedPubSub` that carries a **sticky error** surfaced on `Receive`/`conn` and logged at creation.
> 
> **PubSub** adds `stickyErr` so failed handles fail operations instead of connecting. `Channel` and `ChannelWithSubscriptions` log a warning and return an **already-closed** channel when there is a sticky error or when the two channel APIs are mixed (previously panicked). Logging uses `context.Background()` in the conflict path to avoid a race on `c.cmd`.
> 
> Tests cover empty Ring subscribe, mutual exclusion of channel helpers, and Ginkgo cases for sticky-error and conflicting channel calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4e37263356f9059fab2aaed6ec4366689236bab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pubsub.go
```diff
@@ -72,6 +77,9 @@ func (c *PubSub) connWithLock(ctx context.Context) (*pool.Conn, error) {
 }
 
 func (c *PubSub) conn(ctx context.Context, newChannels []string) (*pool.Conn, error) {
+	if c.stickyErr != nil {
+		return nil, c.stickyErr
+	}
```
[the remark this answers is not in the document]
VedantMadane: Addressed in commit 0d618f37c. Prioritized c.closed check before c.stickyErr in conn() so that Close() returns pool.ErrClosed to background receiver channels, allowing them to close cleanly.

pubsub.go
```diff
@@ -1,856 +1,878 @@
...
-						ctx, "redis: %v channel is full for %s (message is dropped)",
-						c, c.chanSendTimeout)
-				}
-			default:
-				internal.Logger.Printf(ctx, "redis: unknown message type: %T", msg)
-			}
-		}
-	}()
-}
+package redis
```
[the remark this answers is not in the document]
VedantMadane: Fixed in commit f4bdddd07. Restored LF line endings across pubsub.go and pubsub_test.go.

pubsub.go
```diff
@@ -72,6 +77,9 @@ func (c *PubSub) connWithLock(ctx context.Context) (*pool.Conn, error) {
 }
 
 func (c *PubSub) conn(ctx context.Context, newChannels []string) (*pool.Conn, error) {
+	if c.stickyErr != nil {
+		return nil, c.stickyErr
+	}
 	if c.closed {
 		return nil, pool.ErrClosed
```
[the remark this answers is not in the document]
VedantMadane: Fixed in e2a2d47 (e2a2d47004266494eb6a730e7bceb7ad00ecb4a4). Close now sets stickyErr to pool.ErrClosed (same pattern as SingleConnPool.Close) and conn() still checks closed first so Channel receivers exit on pool.ErrClosed.

pubsub.go
```diff
@@ -575,8 +583,13 @@ func (c *PubSub) Channel(opts ...ChannelOption) <-chan *Message {
 		c.msgCh.initMsgChan()
 	})
 	if c.msgCh == nil {
-		err := fmt.Errorf("redis: Channel can't be called after ChannelWithSubscriptions")
-		panic(err)
+		// Already using ChannelWithSubscriptions — return a closed channel
+		// instead of panicking so callers can recover (issue #3761).
+		internal.Logger.Printf(c.getContext(),
+			"redis: Channel can't be called after ChannelWithSubscriptions")
```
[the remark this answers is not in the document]
VedantMadane: Fixed in d4cbfb1. Conflict warnings log with context.Background() instead of getContext() so they do not read c.cmd while the other channel receiver may write it.

pubsub.go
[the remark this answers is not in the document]
VedantMadane: Follow-up in e2a2d47: added .gitattributes with *.go text eol=lf so Windows checkouts keep LF in the index. PR blobs were already LF after f4bdddd.

pubsub.go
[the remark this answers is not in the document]
VedantMadane: Reinforced in e2a2d47 (e2a2d47004266494eb6a730e7bceb7ad00ecb4a4). Close overrides stickyErr with pool.ErrClosed and the unit test asserts Receive returns redis.ErrClosed after Close on a failed Ring PubSub.

ndyakov:
@VedantMadane thank you for this, Ring is not officially supported or recommended anymore, but I will take a look when possible.

pubsub.go
```diff
@@ -217,6 +225,10 @@ func (c *PubSub) Close() error {
 		return pool.ErrClosed
 	}
 	c.closed = true
+	// Override construction-time stickyErr (e.g. Ring failedPubSub) so
+	// Channel receivers observe pool.ErrClosed and exit cleanly.
+	// Matches SingleConnPool.Close which sets stickyErr = ErrClosed.
+	c.stickyErr = pool.ErrClosed
```
ndyakov, line 231: This is dead code — `conn()` checks `c.closed` before `c.stickyErr`, and we set `closed = true` just above, so this write is never read. The `c.closed` check is also what makes the pump exit on `ErrClosed`, not this line. Can we drop it (and the comment)?

pubsub.go
```diff
@@ -575,8 +587,15 @@ func (c *PubSub) Channel(opts ...ChannelOption) <-chan *Message {
...
 	if c.msgCh == nil {
-		err := fmt.Errorf("redis: Channel can't be called after ChannelWithSubscriptions")
-		panic(err)
+		// Already using ChannelWithSubscriptions — return a closed channel
+		// instead of panicking so callers can recover (issue #3761).
+		// Use Background: getContext reads c.cmd which the other channel's
+		// Receive loop may write concurrently (data race under -race).
+		internal.Logger.Printf(context.Background(),
+			"redis: Channel can't be called after ChannelWithSubscriptions")
+		ch := make(chan *Message)
```
ndyakov, line 596: The Ring sticky-error path never logs or closes the channel — the pump just spins on the error, so `range Subscribe(ctx).Channel()` hangs forever with no output until `Close()`. The PR says it "logs a warning" but that's only this mutual-exclusion case. Can we short-circuit when `stickyErr != nil && !closed` and return a closed channel like this branch does? `Eventually(ch).Should(BeClosed())` still passes. Would also be good to mention the closed-channel-on-conflict behavior in the `Channel`/`ChannelWithSubscriptions` docs.

pubsub.go
```diff
@@ -575,8 +587,15 @@ func (c *PubSub) Channel(opts ...ChannelOption) <-chan *Message {
...
 	})
 	if c.msgCh == nil {
-		err := fmt.Errorf("redis: Channel can't be called after ChannelWithSubscriptions")
-		panic(err)
+		// Already using ChannelWithSubscriptions — return a closed channel
+		// instead of panicking so callers can recover (issue #3761).
+		// Use Background: getContext reads c.cmd which the other channel's
+		// Receive loop may write concurrently (data race under -race).
+		internal.Logger.Printf(context.Background(),
+			"redis: Channel can't be called after ChannelWithSubscriptions")
```
ndyakov, line 595: Nice — `Background()` instead of `getContext()` avoids the `c.cmd` race here 👍

ring.go
```diff
@@ -705,43 +705,54 @@ func (c *Ring) Len() int {
 	return c.sharding.Len()
 }
 
+// failedPubSub returns a PubSub that surfaces err on every operation instead of
+// panicking at construction (issue #3761).
+func (c *Ring) failedPubSub(err error) *PubSub {
```
ndyakov, line 710: Worth a log in here — if the caller only uses `Channel()` and never `Receive()`, the empty/shard-failure cases are completely silent.

ring.go
```diff
@@ -705,43 +705,54 @@ func (c *Ring) Len() int {
...
+	}
+	pubsub.init()
+	return pubsub
+}
+
 // Subscribe subscribes the client to the specified channels.
 func (c *Ring) Subscribe(ctx context.Context, channels ...string) *PubSub {
 	if len(channels) == 0 {
-		panic("at least one channel is required")
+		return c.failedPubSub(fmt.Errorf("redis: at least one channel is required"))
```
ndyakov, line 725: Can we note in the doc for `Subscribe`/`PSubscribe`/`SSubscribe` that these now return a PubSub with a sticky error instead of panicking? That was the main ask in #3761.

.gitattributes
```diff
@@ -0,0 +1,6 @@
+*.go text eol=lf
```
ndyakov, line 1: Can we drop this from the PR? Repo has no `.gitattributes` today, so it adds a repo-wide EOL policy unrelated to the pubsub fix — better as a separate maintainer decision than bundled into a bugfix.

ndyakov (changes requested):
Thanks for tackling this — approach looks right and matches the TODO/issue, and keeping the signatures non-breaking is the right call. A couple of things before merge, main ones being a dead `stickyErr` write in `Close()` and the Ring sticky-error path hanging silently when consumed via `Channel()`. Details inline.

VedantMadane:
Addressed in dc6f17ec:
- Dropped the redundant stickyErr write in Close()
- Added short-circuiting in Channel() and ChannelWithSubscriptions() when stickyErr != nil && !closed to log a warning and return an already-closed channel (so for range pubsub.Channel() completes without hanging).
- Added warning logging in Ring.failedPubSub for visibility when constructing a failed PubSub.
- Updated docstrings for Subscribe/PSubscribe/SSubscribe and Channel/ChannelWithSubscriptions.
- Removed .gitattributes.
- Rebased on latest master.

pubsub.go
VedantMadane: Dropped in c6ccfc4c — Close no longer writes stickyErr; closed is checked first in conn().

pubsub.go
VedantMadane: Addressed in c6ccfc4c: Channel/ChannelWithSubscriptions short-circuit when stickyErr is set, log a warning, and return an already-closed channel. Docs updated on both methods.

ring.go
VedantMadane: Done in c6ccfc4c — failedPubSub now logs via internal.Logger before returning the sticky-error PubSub.

ring.go
VedantMadane: Documented on Subscribe/PSubscribe/SSubscribe in c6ccfc4c: empty channels or shard lookup failure return a PubSub with a sticky error rather than panicking.

.gitattributes
VedantMadane: Removed in c6ccfc4c — .gitattributes is no longer part of this PR.

VedantMadane:
Thanks @ndyakov — all of the inline feedback should now be addressed on HEAD:

- Close no longer writes dead stickyErr
- Channel/ChannelWithSubscriptions short-circuit stickyErr with a closed channel + warning log
- failedPubSub logs on construction
- Ring Subscribe/PSubscribe/SSubscribe docs note sticky-error returns
- .gitattributes dropped from the PR

Also bumped `google.golang.org/grpc` to v1.83.1 in `example/otel` and `example/otel-metrics` to clear govulncheck GO-2026-6348 (CI failure was symbol-level on those modules only; library code itself was unaffected).

PR #3969, opened 2026-08-17
fix(maintnotifications): skip endpoint DNS detect when mode is disabled

Phalanyx:
## Summary

This is especially costly for cluster clients, which run Options.init once per node, so detect latency scales with cluster size.

`Options.init` always runs `DetectEndpointType` when `EndpointType` is empty or `auto`. That does a synchronous DNS lookup (`LookupIPAddr`, bounded at 2s) to classify internal vs external endpoints for `CLIENT MAINT_NOTIFICATIONS`.

`EndpointType` is only used when maintenance notifications are enabled (see `initConn`). When `Mode` is `disabled`, the value is never read, but every `NewClient` still paid for detection. Cluster clients construct options once per node, so a slow or broken resolver can stall startup by many seconds (N nodes × detect timeout).

Manager construction and the maint handshake are already skipped when mode is disabled. This change only skips the unused DNS-backed auto-detection in `Options.init`.

## Change

- If `MaintNotificationsConfig.Mode == ModeDisabled`, do not call `DetectEndpointType`; leave `EndpointType` unchanged.
- If mode is `enabled` / `auto`, keep existing behavior (detect when type is empty/`auto`).

## Test plan

- [x] `go test ./ -run 'TestOptionsInitSkipsEndpointDetectWhenMaintDisabled|TestNewClientSkipsEndpointDetectWhenMaintDisabled'`
- [ ] CI

## Notes / follow-up

Default config is still `ModeAuto` + `EndpointTypeAuto`, so callers that never set `MaintNotificationsConfig` still run detection. A possible follow-up is lazy detection at handshake time so `Options.init` never does DNS. However, the first operation could cause a timeout in this case.

---

> [!NOTE]
> **Low Risk**
> Narrow guard on disabled maint mode only; enabled/auto paths unchanged and covered by new tests.
> 
> **Overview**
> **`Options.init` no longer runs DNS-backed `DetectEndpointType` when maintenance notifications are disabled**, so client construction avoids synchronous resolver work that was unused because `EndpointType` is only consumed when maint mode is enabled.
> 
> When `MaintNotificationsConfig.Mode` is `enabled` or `auto`, behavior is unchanged: empty or `auto` `EndpointType` still triggers detection; an explicit type is left as-is. Tests assert that disabled mode keeps `EndpointTypeAuto` on an unresolvable address (proving detection was skipped) and that `NewClient` behaves the same.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc446b8633ccfa9784573f60eb85d445fab54bdc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/options.go
+++ b/options.go
@@ -574,13 +574,17 @@ func (opt *Options) init() {
 
 	opt.MaintNotificationsConfig = opt.MaintNotificationsConfig.ApplyDefaultsWithPoolConfig(opt.PoolSize, opt.MaxActiveConns)
 
-	// auto-detect endpoint type if not specified
-	endpointType := opt.MaintNotificationsConfig.EndpointType
-	if endpointType == "" || endpointType == maintnotifications.EndpointTypeAuto {
-		// Auto-detect endpoint type if not specified
-		endpointType = maintnotifications.DetectEndpointType(opt.Addr, opt.TLSConfig != nil)
+	// EndpointType is only consumed when sending CLIENT MAINT_NOTIFICATIONS.
+	// Skip DNS-backed auto-detection when maintenance notifications are disabled
+	// so client construction cannot stall on resolver timeouts. This is especially
+	// costly for cluster clients, which construct Options once per node.
+	if opt.MaintNotificationsConfig.Mode != maintnotifications.ModeDisabled {
+		endpointType := opt.MaintNotificationsConfig.EndpointType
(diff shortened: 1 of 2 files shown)


ndyakov (approved):
Thank you @Phalanyx, this makes complete sense.

PR #3972, opened 2026-08-18
perf(proto): add zero-copy semantics to scan, remove redundant data c…

vlady-kotsev:
## Summary

`ScanSlice` (and `StringCmd.Scan`) copied every input string into a fresh
`[]byte` before handing it to `proto.Scan`, and `makeSliceNextElemFunc`
allocated a new slice header via `v.Set(v.Slice(...))` for every element. That
cost 2 heap allocations per element on every path, even for destinations that
only parse the bytes (`*int64`, `*float64`, `*string`, `*time.Time`, …).

This PR moves the copy to where it is actually needed:

- Callers now pass the input zero-copy via `util.StringToBytes`.
- The `Scan` branches where `b` escapes — `*[]byte`, `*net.IP`, and
  `encoding.BinaryUnmarshaler` — copy at the point of escape, so no caller ever
  receives (or hands user code) bytes aliasing a string's backing array.
- `makeSliceNextElemFunc` extends the destination in place with
  `reflect.Value.Grow`/`SetLen` instead of allocating a slice header per
  element, and the pointer/non-pointer variants are collapsed into one helper.

Net effect: parsing destinations are now allocation-free per element, escaping
destinations pay exactly the one copy they always needed, and existing
semantics are preserved and now pinned by tests.

## Safety notes

- `*string` aliases the input string's bytes (string → string, both
  immutable) — safe and copy-free.
- `*[]byte` / `*net.IP` copy, since the caller receives a mutable slice.
- `encoding.BinaryUnmarshaler` receives a private copy, so implementations
  that mutate or retain their input keep working exactly as before.
- New specs assert the no-aliasing guarantees (mutating results does not
  affect source strings), empty-input semantics, in-place capacity reuse, and
  pointer-element recycling.

Also includes mechanical modernization in `internal/proto`:
`interface{}` → `any`, `reflect.Ptr` → `reflect.Pointer` and
direct `strconv.ParseInt`/`ParseUint` calls in `ReadInt`/`ReadUint` (dropping
a pointless string → `[]byte` round-trip).

## Benchmarks

`go test ./internal/proto/ -bench BenchmarkScanSlice` — linux/amd64,
Intel Core Ultra 7 155U. 1024 elements per op; destinations are reset to zero
length between iterations (except the `Grow` benchmarks, which start from a
nil slice).

| Benchmark | old ns/op | new ns/op | Δ speed | old allocs/op | new allocs/op |
|---|---:|---:|---:|---:|---:|
| ScanSliceString/size=8 | 58,676 | 14,189 | 4.1× | 2,049 | 1 |
| ScanSliceString/size=64 | 76,157 | 14,139 | 5.4× | 2,049 | 1 |
| ScanSliceString/size=512 | 174,061 | 14,303 | 12.2× | 2,049 | 1 |
| ScanSliceBytes/size=8 | 59,903 | 24,928 | 2.4× | 2,049 | 1,025 |
| ScanSliceBytes/size=64 | 75,273 | 38,506 | 2.0× | 2,049 | 1,025 |
| ScanSliceBytes/size=512 | 177,868 | 144,223 | 1.2× | 2,049 | 1,025 |
| ScanSliceInt64 | 78,344 | 29,451 | 2.7× | 2,049 | 1 |
| ScanSliceFloat64 | 97,469 | 42,548 | 2.3× | 2,049 | 1 |
| ScanSliceIP | 59,449 | 22,681 | 2.6× | 2,049 | 1,025 |
| ScanSliceBinaryUnmarshaler | 599,576 | 556,226 | 1.1× | 7,169 | 6,145 |
| ScanSlicePtr | 602,739 | 554,816 | 1.1× | 7,169 | 6,146 |
| ScanSlicePtrGrow | 668,708 | 602,002 | 1.1× | 8,206 | 7,183 |
| ScanSliceGrow | 86,760 | 31,044 | 2.8× | 2,062 | 14 |

Highlights:

- Parsing paths drop from 2 allocs/element to **1 alloc per entire scan**; the
  string path is now flat across element sizes (zero-copy) and up to 12×
  faster.
- Escaping paths (`*[]byte`, `*net.IP`) halve their allocations — the
  per-element reflect slice-header allocation is gone; only the mandatory
  copy remains.

---

> [!NOTE]
> **Medium Risk**
> Touches core reply-scanning paths used by many commands; behavior is intended to stay the same but relies on careful copy-vs-alias rules and new test coverage rather than a trivial change.
> 
> **Overview**
> **Performance-focused changes** to how Redis string replies are scanned into Go values. `ScanSlice` and `StringCmd.Scan` now pass input into `proto.Scan` via `util.StringToBytes` instead of allocating a `[]byte` per element, and `makeSliceNextElemFunc` grows destination slices in place with `reflect.Value.Grow`/`SetLen` rather than a new slice header per element.
> 
> **Safety is preserved** by copying only where the scanned bytes escape the caller: `*[]byte`, `*net.IP`, and `encoding.BinaryUnmarshaler` each get a private `make`+`copy`. `ReadInt`/`ReadUint` in `internal/proto/reader.go` parse string bulk replies with `strconv` directly (no string→`[]byte` round-trip). Types are modernized (`any`, `reflect.Pointer`).
> 
> **Tests and benchmarks** in `internal/proto/scan_test.go` lock in non-aliasing, empty-slice semantics, capacity reuse, and pointer recycling.
> 
> **CI**: `govulncheck` runs on PRs to `vkotsev/**` and uses Go `stable` instead of `1.26.x`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b3c2d5f64aea42d04cadb02000fc32052ce52c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -29,12 +29,11 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v7
         with:
-          go-version: "1.26.x"
-          check-latest: true
+          go-version: "stable"
           cache: true
 
       - name: Install govulncheck
-        run: go install golang.org/x/vuln/cmd/govulncheck@latest
+        run: go install golang.org/x/vuln/cmd/govulncheck@v1.2.0
 
       - name: Run govulncheck (all modules)
         shell: bash
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -3,7 +3,7 @@ name: govulncheck
 
 on:
   pull_request:
-    branches: [master, v9, v9.7, v9.8, "ndyakov/**", "ofekshenawa/**", "ce/**", "feature/**"]
+    branches: [master, v9, v9.7, v9.8, "ndyakov/**", "ofekshenawa/**","vkotsev/**", "ce/**", "feature/**"]
   push:
     branches: [master, v9, "v9.*"]
   schedule:
@@ -33,7 +33,7 @@ jobs:
           cache: true
 
       - name: Install govulncheck
-        run: go install golang.org/x/vuln/cmd/govulncheck@v1.2.0
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
 
       - name: Run govulncheck (all modules)
         shell: bash
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -30,6 +30,7 @@ jobs:
         uses: actions/setup-go@v7
         with:
           go-version: "1.26.x"
+          check-latest: true
           cache: true
 
       - name: Install govulncheck
--- a/command.go
+++ b/command.go
@@ -1875,7 +1875,7 @@ func (cmd *StringCmd) Scan(val interface{}) error {
 	if cmd.err != nil {
 		return cmd.err
 	}
-	return proto.Scan([]byte(cmd.val), val)
+	return proto.Scan(util.StringToBytes(cmd.val), val)
 }
 
 func (cmd *StringCmd) String() string {
--- a/internal/proto/reader.go
+++ b/internal/proto/reader.go
@@ -502,7 +502,7 @@ func (r *Reader) ReadInt() (int64, error) {
 		if err != nil {
 			return 0, err
 		}
-		return util.ParseInt([]byte(s), 10, 64)
+		return strconv.ParseInt(s, 10, 64)
 	case RespBigInt:
 		b, err := r.readBigInt(line)
 		if err != nil {
@@ -529,7 +529,7 @@ func (r *Reader) ReadUint() (uint64, error) {
 		if err != nil {
 			return 0, err
 		}
-		return util.ParseUint([]byte(s), 10, 64)
+		return strconv.ParseUint(s, 10, 64)
 	case RespBigInt:
 		b, err := r.readBigInt(line)
 		if err != nil {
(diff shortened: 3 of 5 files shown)


.github/workflows/govulncheck.yml
```diff
@@ -3,7 +3,7 @@ name: govulncheck
 
 on:
   pull_request:
-    branches: [master, v9, v9.7, v9.8, "ndyakov/**", "ofekshenawa/**", "ce/**", "feature/**"]
+    branches: [master, v9, v9.7, v9.8, "ndyakov/**", "ofekshenawa/**","vkotsev/**", "ce/**", "feature/**"]
   push:
```
ndyakov, line 7: 👍

.github/workflows/govulncheck.yml
```diff
@@ -30,10 +30,11 @@ jobs:
         uses: actions/setup-go@v7
         with:
           go-version: "1.26.x"
```
ndyakov:
```suggestion
          go-version: "stable"
```

.github/workflows/govulncheck.yml
```diff
@@ -30,10 +30,11 @@ jobs:
...
+          check-latest: true
           cache: true
 
       - name: Install govulncheck
-        run: go install golang.org/x/vuln/cmd/govulncheck@v1.2.0
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
```
ndyakov: Here it is better to keep the pinned version in the case of a major release with breaking changes.

ndyakov (commented):
Partial review for the workflow fix.

ndyakov (approved):
Nice findings, thank you @vlady-kotsev and congratulation on your first contribution 🎉!

PR #3984, opened 2026-08-24
fix(msetex): first key index handling

### LINKED ISSUE #3986: bug: msetex not routing keys with same hashtag string to same slot in clustered redis

msetex is not routing keys with identical hashtags enclosed string to same slot

## Expected Behavior

keys with same substring in the hashtag should route to same slots when using msetex

## Current Behavior

keys are being routed to random nodes due to slot getting picked by entire key's hash.

## Possible Solution

issue seems to be due to missing handling in cmdFirstKeyPosWithInfo for msetex, its defaulting to index as 1 whereas it should be 2
## Steps to Reproduce

test msetex on a clustered redis setup
1.
2.
3.
4.

## Context (Environment)

go-redis version v9.21
AWS elastic cache clustered mode enabled valkey_version:9.1.1

## Detailed Description

cmdFirstKeyPosWithInfo should have handling for msetex command. Unlike mset or msetnx which have first index = 1 which is the fallback value, index 2 needs to be returned for msetex

## Possible Implementation

[
](https://github.com/redis/go-redis/pull/3984)

shivamrustagi:
fixes #3986

> [!NOTE]
> **Medium Risk**
> Touches cluster key-position routing for MSETEX; a wrong first-key index would send writes to the wrong shard. The change is small and covered by unit tests.
> 
> **Overview**
> Fixes cluster slot hashing for `MSETEX` so the first key is taken at arg position 2 (after the command name and `numkeys`), not position 1 like `MSET`/`MSETNX`.
> 
> `MSetEX` now calls `SetFirstKeyPos(2)` on the constructed command. `cmdFirstKeyPosWithInfo` also hardcodes that fallback for raw `Do("msetex", ...)` when the command-info cache is cold. Tests cover the cold-cache path versus `mset`/`msetnx`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e5968f6f9b5bf8ed269318195eeae8f4cdc909e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -330,12 +330,11 @@ func cmdFirstKeyPosWithInfo(cmd Cmder, info *CommandInfo) int {
 		if cmd.stringArg(1) == "usage" {
 			return 2
 		}
-	// CommandInfo (if available) gives the correct answer
-	// otherwise the hardcoded fallback applies.
-
 	case "msetex":
+		// MSetEX's constructor already sets this via SetFirstKeyPos; this
+		// fallback only covers raw Do("msetex", ...) calls, which aren't
+		// guaranteed to route correctly and aren't the recommended usage.
 		return 2
-
 	}
 
 	// Use CommandInfo cache when warm (in-memory only, no extra round-trips).
--- a/command.go
+++ b/command.go
@@ -330,8 +330,12 @@ func cmdFirstKeyPosWithInfo(cmd Cmder, info *CommandInfo) int {
 		if cmd.stringArg(1) == "usage" {
 			return 2
 		}
-		// CommandInfo (if available) gives the correct answer
-		// otherwise the hardcoded fallback applies.
+	// CommandInfo (if available) gives the correct answer
+	// otherwise the hardcoded fallback applies.
+
+	case "msetex":
+		return 2
+
 	}
 
 	// Use CommandInfo cache when warm (in-memory only, no extra round-trips).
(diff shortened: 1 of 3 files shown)


rahulkushwaha12 (approved):

command.go
```diff
@@ -330,8 +330,12 @@ func cmdFirstKeyPosWithInfo(cmd Cmder, info *CommandInfo) int {
...
 			return 2
 		}
-		// CommandInfo (if available) gives the correct answer
-		// otherwise the hardcoded fallback applies.
+	// CommandInfo (if available) gives the correct answer
+	// otherwise the hardcoded fallback applies.
+
+	case "msetex":
+		return 2
+
```
ndyakov: another (an currently preferred) approach would be to set `SetFirstKeyPos(2)` in the `msetex` constructor. This works and can be kept as well to cover raw `do` executions. However, the raw `Do` usage is not guaranteed to be routed correctly, this is why we would advice using the constructors and the typed commands.

ndyakov (commented):
@shivamrustagi thank you for your contribution. Please find my suggestion and if you could - address it.

command.go
shivamrustagi: aligned @ndyakov, addressed the comment.

ndyakov (approved):

PR #3988, opened 2026-08-24
fix(log): guard global Logger and LogLevel with atomics

saddamr3e:
`go test -race`:

    WARNING: DATA RACE
    Read at 0x... by goroutine 23:
      github.com/redis/go-redis/v9.formatSec()   commands.go:43
    Previous write at 0x... by goroutine 22:
      github.com/redis/go-redis/v9.SetLogger()   redis.go:45

`internal.Logger` and `internal.LogLevel` are plain package vars. `SetLogger`/`SetLogLevel` and `logging.Enable`/`Disable` write them while command and pool goroutines read them lock-free: `Logger.Printf` at every call site (the sub-millisecond warning in `formatSec`, the background dial goroutine at `pool.go:762`) and `LogLevel.*OrAbove` at the level guards (`isHealthyConn` on the Get path). Any `SetLogLevel`/`SetLogger` during live client traffic races.

Wrap both behind small atomic holders (`atomic.Pointer[Logging]` for the logger, `atomic.Int32` for the level) that keep the existing read syntax (`Logger.Printf`, `LogLevel.DebugOrAbove`), so only the five setter sites change and the read sites stay untouched. The regression test drives the public setters against the reads and is clean under `-race`.

---

> [!NOTE]
> **Low Risk**
> Localized concurrency fix for process-wide logging globals; behavior is unchanged aside from thread-safe swaps and slightly adjusted default log file/line attribution.
> 
> **Overview**
> Fixes data races when `SetLogger`, `SetLogLevel`, and `logging.Enable`/`Disable` run while pool and command goroutines call `internal.Logger.Printf` or read `internal.LogLevel` via the `*OrAbove` helpers.
> 
> **`internal.Logger` and `internal.LogLevel`** are no longer plain package variables. They sit behind small atomic holders (`atomic.Pointer[Logging]` and `atomic.Int32`) with `Store`/`Load` and unchanged hot-path usage (`Logger.Printf`, `LogLevel.DebugOrAbove()`). Setters in `redis.go` and `logging/logging.go` now call `Store`.
> 
> **`NewDefaultLogger`** takes an explicit **calldepth** so stderr line attribution stays correct through the extra `atomicLogger.Printf` frame and through `filterLogger` wrappers (`DefaultLoggerCalldepth` vs `FilterLoggerCalldepth`).
> 
> Adds **`TestSetLoggerLogLevelConcurrentWithReads`** (intended for `go test -race`) and updates tests that swap the global logger to use `Load`/`Store`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9106f51fee3eeed6ae74848a49dbfe81de958f29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/log.go
+++ b/internal/log.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"sync/atomic"
 )
 
 // TODO (ned): Revisit logging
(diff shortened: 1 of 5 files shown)


vlady-kotsev:
Hey @saddamr3e , thanks for the fix, I'll take a look and let you know if I have any comments.

vlady-kotsev:
Hey @saddamr3e , looks good but we will need to handle the calldepth in order to report the correct callsite.
Currently with this PR fix in `DefaultLogger` we report `log.go` instead of the actual callsite.
There is also a bug in `filterLogger` not reporting the correct callsite (not introduced with this PR, but we can fix it as well). It reports `logging.go`.
What I would suggest is to pass the calldepth as a param to `NewDefaultLogger`, using 3 for Default logger (an additional stack frame introduced in this PR) and 4 for filterLogger (3 for DefaultLogger + the filtered logger stack frame).

saddamr3e:
Good catch, done. NewDefaultLogger now takes the calldepth: 3 for the plain default logger and 4 for the filter loggers, as you suggested. Verified both paths by capturing stderr and checking the reported file:line points at the caller instead of log.go/logging.go, and the race test is still clean.

internal/log.go
```diff
@@ -62,7 +68,7 @@ func newAtomicLogger(l Logging) *atomicLogger {
 // Logger calls Output to print to the stderr.
 // Arguments are handled in the manner of fmt.Print.
 // Swap it with redis.SetLogger; read it through Logger.Printf.
-var Logger = newAtomicLogger(NewDefaultLogger())
+var Logger = newAtomicLogger(NewDefaultLogger(3))
```
vlady-kotsev: Great, that's what I had in mind. Can we just convert the magic numbers (3 and 4)  to constants for readability and we are good to merge.

vlady-kotsev (changes requested):

internal/log.go
saddamr3e: Done, pulled them into DefaultLoggerCalldepth/FilterLoggerCalldepth in internal/log.go and switched all four call sites. Race test still clean.

vlady-kotsev (approved):

PR #3995, opened 2026-08-27
fix(command): guard zero-length entry arrays in reply parsers

saddamr3e:
1. TSTimestampValueSliceCmd.readReply (TS.RANGE/TS.REVRANGE) and the RESP2 branch of MapStringSliceInterfaceCmd.readReply (TS.MRANGE/TS.MGET) read a per-entry array length with ReadArrayLen and use it as make([]float64, itemLen-1) / make([]interface{}, 0, itemLen-1) with no lower bound on itemLen.
2. a server that declares a zero-length entry makes itemLen-1 == -1, so the make panics (makeslice len/cap out of range); before that the mandatory timestamp/key is read out of the empty entry, taking a frame from the next reply and desyncing the pooled connection.

Reject itemLen < 1 in both parsers so a malformed entry returns a parse error and the bad connection is dropped. Test feeds a zero-length entry (panics before, errors after) and keeps a valid reply passing.

---

> [!NOTE]
> **Low Risk**
> Narrow defensive validation on malformed server replies; normal responses unchanged and covered by new tests.
> 
> **Overview**
> Malformed TimeSeries RESP replies that declare a **zero-length** inner array could crash the client or corrupt pooled connections. Parsers for **TS.RANGE/TS.REVRANGE** samples and the RESP2 path of **TS.MRANGE/TS.MGET** (`MapStringSliceInterfaceCmd`) now reject `itemLen < 1` with a parse error instead of using `itemLen-1` in `make`, which previously panicked or read past the entry and desynced the reader.
> 
> New unit tests feed a `*0` entry (must error) and valid `*2` replies (must still parse) for both command types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f500b5e7917ef01c02c3f517ad74e21092efc7dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/command.go
+++ b/command.go
@@ -2615,6 +2615,9 @@ func (cmd *MapStringSliceInterfaceCmd) readReply(rd *proto.Reader) (err error) {
 			if err != nil {
 				return err
 			}
+			if itemLen < 1 {
+				return fmt.Errorf("redis: got %d elements in map-string-slice-interface entry, expected at least 1", itemLen)
+			}
 
 			key, err := rd.ReadString()
 			if err != nil {
--- a/command_entry_len_test.go
+++ b/command_entry_len_test.go
@@ -1,0 +1,55 @@
+package redis
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/redis/go-redis/v9/internal/proto"
+)
+
+// A per-entry array whose declared length is zero must be rejected before the
+// parser uses itemLen-1 as a slice size. Without the guard, itemLen-1 == -1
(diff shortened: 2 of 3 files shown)


vlady-kotsev (approved):

vlady-kotsev:
Hey @saddamr3e thanks for the contribution. Approved!

ndyakov:
Thank you @saddamr3e !

PR #3996, opened 2026-08-27
feat(multidb): identify members by stable id

ndyakov:
Members are addressed by an immutable, never-reused **stable id** instead of a
slice position, stored in a `map[int]*db` keyed by that id.

- `RemoveDatabase` is a map delete: survivors are never renumbered and the
  active member's id never shifts, so an operator handle and a queued callback
  stay valid across it. No reindex loop, no active-position adjustment.
- The active database is tracked by id (`-1` = none); a non-negative active is
  always a live key (removing the active is refused).
- Failover callbacks (`OnFailover`, `OnActiveDatabaseChanged`,
  `OnCircuitStateChanged`) carry stable ids — a concurrent `RemoveDatabase` can
  never make a delivered handle name a different member.
- Strategy candidates (`MultiDBDatabaseState`) are sorted by id, so selection is
  deterministic despite map iteration order.

## API (unreleased — renamed for clarity, no positional "index")
- `ActiveIndex` → `ActiveDatabaseID`
- `SetActiveIndex` → `SetActiveDatabase`
- `ForceActiveIndex` → `ForceActiveDatabase`
- `MultiDBDatabaseState.Index` → `.ID`
- `AddDatabase` returns the id; `RemoveDatabase` / `SetWeight` /
  `AddDatabaseHook` take it; an unknown id yields the new `ErrDatabaseNotFound`.

## Notes
- **Supersedes** the earlier slice + id↔position-translation approach and its
  int32-narrowing fix (commit `f3791731`): narrowing is structurally impossible
  with an int-keyed map, so that resolved thread's SHA is no longer on the
  branch.
- Stacked on #3950 (`feature/multidb/client`).

Tests: stable id survives removal; failover callback carries a stable id;
strategy sees ids (exact candidate set); a large/out-of-range id resolves to no
member; full multidb suite green with `-race`.

---

> [!NOTE]
> **High Risk**
> This is a breaking API rename plus a core change to how active membership and removal interact with failover; incorrect id handling could mis-route traffic or deliver stale handles in callbacks.
> 
> **Overview**
> **MultiDB members are now addressed by immutable stable ids** instead of slice positions. Membership lives in a `map[int]*multidbDatabase` with monotonic, never-reused ids from `AddDatabase`; removing a non-active member is a map delete with no reindexing and no adjustment of the active handle.
> 
> The **control and callback surface is renamed and re-typed** for ids: `ActiveDatabaseID`, `SetActiveDatabase`, `ForceActiveDatabase`, `MultiDBDatabaseState.ID`, and failover callbacks (`OnFailover`, `OnActiveDatabaseChanged`, `OnCircuitStateChanged`) all use stable member ids. Unknown ids return **`ErrDatabaseNotFound`** instead of out-of-range index errors.
> 
> **Failover and fallback selection** pass sorted candidate lists (by id) so map iteration order and equal-weight ties do not change behavior. Tests cover id stability across removal, callback handles, strategy input, and large/out-of-range ids.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b87a7b448c2d630b9c266ba975f359d150807072. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/multidb_core.go
+++ b/multidb_core.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -31,11 +32,11 @@ type multidbDatabase struct {
 	cb     *imultidb.CircuitBreaker
 	weight float64 // guarded by core.dbMu
 	fqdn   string
-	// idx is the database's current position in core.dbs, maintained under
-	// core.dbMu on every membership change. Callbacks read it lock-free to
-	// avoid recursive dbMu read-locking (which can deadlock with a queued
-	// writer).
-	idx atomic.Int32
+	// id is the database's stable identifier and its key in core.dbs. Assigned
+	// once at membership time (before the initial probe), immutable, and never
+	// reused, so callbacks and the public API can name a member with an int
+	// that a RemoveDatabase never invalidates or renumbers. Read lock-free.
+	id int
 
 	checks []MultiDBHealthCheck
 	policy MultiDBHealthCheckPolicy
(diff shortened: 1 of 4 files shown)


multidb_core.go
```diff
@@ -543,7 +560,15 @@ func (c *multidbCore) resetDetectorSafely() {
 	c.detector.Reset()
 }
 
-func (c *multidbCore) activeIndex() int { return int(c.active.Load()) }
+// activeIndex returns the stable id of the active database, or -1 when none is
+// selected.
+func (c *multidbCore) activeIndex() int {
```
vlady-kotsev: Here we can have naming reflecting the new behaviour but for consistency and not breaking the public API we can leave it as it is.

vlady-kotsev (approved):

multidb_core.go
ndyakov: resolved, the whole api will use `ID` not `Index`

vlady-kotsev (approved):

PR #4010, opened 2026-09-08
fix(probabilistic): tolerate new fields in *.INFO replies

elena-kolevska:
Redis 8.12 will [add](https://github.com/RedisBloom/RedisBloom/pull/1063) "cell size" to CMS.INFO (the new CELL_SIZE option for CMS.INITBYDIM / CMS.INITBYPROB). The response parser rejected any unknown key, so CMSInfo failed against the unstable image with `redis: CMS.INFO unexpected key cell size`.

This PR exposes the new value as `CMSInfo.CellSize` and, as XINFO already does, future proofs the parser by draining and skipping unknown map fields instead of erroring. It also applies the same forward-compatible default to `BF.INFO`, `CF.INFO`, `TOPK.INFO` and `TDIGEST.INFO`. 

`BF.INFO` with an explicit unknown attribute (ex. `BF.INFO key unknown-key`) still errors, since that is a user error and needs to be surfaced.

---

> [!NOTE]
> **Low Risk**
> Client-side reply parsing only; behavior change is more permissive on full INFO maps while preserving errors for invalid BF.INFO attribute requests.
> 
> **Overview**
> Makes RedisBloom **\*.INFO** response parsing forward-compatible with newer servers, and adds support for **CMS.INFO** `cell_size` (Redis 8.12).
> 
> **`CMSInfo`** now includes **`CellSize`**, populated from the `cell_size` map key (zero on older servers). For full-map replies on **`BF.INFO`**, **`CF.INFO`**, **`CMS.INFO`**, **`TOPK.INFO`**, and **`TDIGEST.INFO`**, unknown keys are **drained with `DiscardNext()`** instead of failing the command—matching the intent described for XINFO-style tolerance. **`BF.INFO key <attribute>`** still returns an error for an unknown attribute name.
> 
> Tests add integration coverage for **`CellSize`** on Redis ≥8.12 and unit tests that verify unknown fields (including nested values) are skipped without desyncing the protocol reader.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c0fe21ef1578a0025e61619654acd4120a71314. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/probabilistic.go
+++ b/probabilistic.go
@@ -877,7 +877,7 @@ func (cmd *CMSInfoCmd) readReply(rd *proto.Reader) (err error) {
 			result.Depth, err = rd.ReadInt()
 		case "count":
 			result.Count, err = rd.ReadInt()
-		case "cell size":
+		case "cell_size":
 			result.CellSize, err = rd.ReadInt()
 		default:
 			// skip unknown fields so newer servers don't break the parser
--- a/probabilistic.go
+++ b/probabilistic.go
@@ -350,10 +350,15 @@ func (cmd *BFInfoCmd) readReply(rd *proto.Reader) (err error) {
 		"EXPANSION":                &result.ExpansionRate,
 	}
 
-	// Helper function to read and assign a value based on the key
-	readAndAssignValue := func(key string) error {
+	// Helper function to read and assign a value based on the key.
+	// Unknown keys are drained and skipped when skipUnknown is set so that
+	// fields added by newer servers don't break the parser.
+	readAndAssignValue := func(key string, skipUnknown bool) error {
 		fieldPtr, exists := respMapping[key]
 		if !exists {
+			if skipUnknown {
+				return rd.DiscardNext()
+			}
 			return fmt.Errorf("redis: BLOOM.INFO unexpected key %s", key)
 		}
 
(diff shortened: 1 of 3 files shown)


probabilistic.go
```diff
@@ -867,8 +877,11 @@ func (cmd *CMSInfoCmd) readReply(rd *proto.Reader) (err error) {
 			result.Depth, err = rd.ReadInt()
 		case "count":
 			result.Count, err = rd.ReadInt()
+		case "cell size":
```
ndyakov: Is cell size really with a space in it?

ndyakov (commented):
Left one question, other than that lgtm

probabilistic.go
ndyakov: it is.

ndyakov: this will be changed to `cell_size`, we need to update the pr.

ndyakov:
```suggestion
		case "cell_size":
```

ndyakov (commented):
Let's update the field and we can include this.

probabilistic.go
elena-kolevska: Done

ndyakov (approved):

PR #4022, opened 2026-09-16
feat(autopipeline): run N full-duplex engines on one client

ndyakov:
### feat(autopipeline): run N full-duplex engines on one client

**Draft.** Depends on #4021 (which depends on #4018), so it is cut from that
branch rather than from master. Three known gaps before it could merge are
listed at the bottom.

`NumShards > 1` under `FullDuplex` now means N engines on ONE client: one pool,
one hook chain, one invalidation stream, N held wires. Commands route by key
hash, so same-key commands keep one wire and per-key order holds without an
`Unordered` opt-in. Keyless commands round-robin. A whole `FDPipelined` batch
pins to one engine, since splitting it would lose the internal order a pipeline
exists to guarantee. `TxPipeline` stays out for the reason it stayed out of
#4021: MULTI/EXEC needs connection affinity for the whole transaction.

#### The change is two-sited, which is the finding

Relaxing the `Validate` rejection alone silently turns full duplex **off**:
`fdOn` is separately gated on `nShards == 1`, so `NumShards > 1` fell through to
half-duplex round-robin shards, which are unordered, and a same-key
write-then-read returned `redis.Nil`. That is exactly what the deleted error
warned about ("reject it rather than silently falling back to half-duplex").
Both the validation and the construction gate have to move together, with
`nShards` forced back to 1 so the shard array stays flusherless. The added test
catches it: 1 engine passes, 8 fail.

#### Why this exists

The benchmark harness faked multi-connection FD by building N separate
`*redis.Client` instances and round-robining across them. That holds six sockets
per client to do I/O on one — 48 for eight wires — and keeps the routing outside
the library, so key routing could not be tested. Measured head to head at 1024
callers, 10 commands per `Exec`, one binary, 3-pass medians:

| engines | proxy (N clients) | in-client | proxy held conns | in-client held conns |
|---|---|---|---|---|
| 2 | 1,126,850 | 1,116,070 | 12 | 7 |
| 4 | 1,528,987 | 1,519,397 | 24 | 9 |
| 8 | 1,724,728 | **1,798,932** | 48 | **12** |

So in-client matches or beats N clients on throughput at a quarter of the
sockets, and preserves per-key ordering the proxy cannot.

#### What it is worth, and where it costs

Full ladder, one binary, 3-pass medians, one fixed configuration per arm:

| callers | 1 engine | 8 engines | 8 vs 1 |
|---|---|---|---|
| 8 | 206,439 | 144,017 | **-30.2%** |
| 64 | 920,157 | 752,934 | -18.2% |
| 128 | 920,342 | 1,067,123 | +15.9% |
| 1024 | 819,703 | 1,807,357 | **+120.5%** |
| 4096 | 814,346 | 1,585,368 | +94.7% |

The crossover is 128 callers. Below it, extra engines fragment the batch and
cost up to 30%. The penalty is **not** specific to go-redis: rueidis' own
default of 4 connections is 23.9% behind its own 1-conn configuration at 8
callers, and crosses over at 128 as well. The cost is intrinsic to spreading a
small batch across queues.

Latency moves both ways. At 4096 callers eight engines halve p50 (49.9 -> 24.3
ms) but raise p99.9 (59.7 -> 77.6 ms), and the p99.9/p50 ratio goes 1.2x ->
3.2x. A single FIFO cannot starve anyone, so any split gives that up.

#### Known gaps before this can merge

1. **Pipeline pool sizing.** Engines lease from the pipeline pool and
   `DefaultPipelinePoolSize` is 10, so more than ~10 engines spill to the main
   pool. The pool must be sized from the engine count.
2. **`cmdFirstKey` is crude** — it takes `args[1]` rather than using the
   command's declared key positions, which the cluster path already computes.
3. **No auto-sizing.** A fixed 8 would hand a ~30% regression to every
   low-concurrency user, so the engine count should be derived from load
   (roughly one wire per 256-512 callers fits both the ladder and the engine
   sweep). `NumShards: 0` meaning "auto" is the natural place for it.

A fourth, smaller one: `accumMin` (`max(64, window>>9)`, the depth gate on the
writer's flush wait) is counted **per engine**, so splitting into N engines
divides each engine's in-flight depth and makes the gate harder to reach. With
the default window the flush delay measures as a no-op on this path at every
width. If the delay should help fragmented multi-engine batches, `accumMin`
needs to scale with the engine count.


rueian:
Great job @ndyakov! Nice to see go-redis finally support CSC and auto pipelining. I tried to implement them once in go-redis, but gave up and then made rueidis on my own.

PR #4025, opened 2026-09-18
fix(otel): don't count Nil replies as client errors

### LINKED ISSUE #4024: Native OpenTelemetry metrics count cache misses(`redis.Nil`) as internal client errors
## Expected Behavior

A `GET` for a key that does not exist returns `redis.Nil`. This is a successful command that returned an empty reply, so it should not be counted in `redis.client.errors` (`redis_client_errors_total`).

If nil replies are recorded for visibility, they should at least be distinguishable from real failures rather than being reported as unknown internal client errors.

## Current Behavior

With native OpenTelemetry(`extra/redisotel-native`) metrics enabled, every `redis.Nil` reply increments `redis.client.errors` metrics with:

```text
error.type = UNKNOWN
db.response.status_code = UNKNOWN
redis.client.errors.internal = true
redis.client.errors.category = server
```

## Possible Solution

The error path appears to invoke the metrics error callback for every non-nil error. Since `redis.Nil` is an error, it reaches error classification, matches no known case, and falls through to the `UNKNOWN` / `internal=true` catch-all.

Two options:

**1. Preferred — do not record `redis.Nil` as an error:**

Skip the metrics error callback when `errors.Is(err, redis.Nil)`, and apply the same guard to the error attributes on `db.client.operation.duration`.

If maintaining a datapoint for nil replies is preferred, classify them explicitly (for example `error.type="NIL"`, `internal=false`, `category="cache_miss"`) instead of relying on the `UNKNOWN` fallback.

**2. Alternative — give nil replies their own non-failure category:**

```text
error.type = NIL
db.response.status_code = NIL
redis.client.errors.internal = false
redis.client.errors.category = cache_miss
```

This keeps the datapoint but lets dashboards and alerts exclude cache misses from error rates and success-rate calculations.

## Steps to Reproduce

Minimal reproduction repository: https://github.com/leonardovillela/redis-otel-native-metrics-cache-miss-error-metrics-bug

1. Clone the repository and run `make`. This starts Redis, a Go reproducer, the OpenTelemetry Collector, Prometheus and Grafana via Docker Compose.
2. The reproducer enables native metrics before creating the client:

```go
instrumentation := redisotel.GetObservabilityInstance()
err := instrumentation.Init(
    redisotel.NewConfig().
        WithEnabled(true).
        WithMetricGroups(redisotel.MetricGroupAll),
)
```

3. It then issues 10 `GET` requests per second for a key that does not exist:

```go
client := redis.NewClient(&redis.Options{Addr: "redis:6379"})

for range time.NewTicker(100 * time.Millisecond).C {
    err := client.Get(ctx, "missing-key").Err()
    if !errors.Is(err, redis.Nil) {
        log.Fatal(err)
    }
}
```

4. Open Grafana at http://localhost:3000 (`admin` / `admin`) and view the **Redis Client Observability Dashboard**. `redis.client.errors` increases by ~10/second with `error.type="UNKNOWN"` and `redis.client.errors.internal="true"`, even though every command returned the expected `redis.Nil`.

Run `make stop` to tear the stack down.

## Context (Environment)

A read-heavy service with a normal cache-miss rate reports a large, permanent client error rate and a degraded success rate. Because misses are classified as `UNKNOWN` internal errors, there is no way to separate them from genuine client failures, which makes error-rate dashboards, alerts and error budgets unusable without knowing this quirk.

- `github.com/redis/go-redis/v9`: `v9.22.0`
- `github.com/redis/go-redis/extra/redisotel-native/v9`: `v9.22.0`
- OpenTelemetry Go SDK: `v1.43.0`
- Go: `1.25`
- Redis server: Docker `redis:latest`

## Detailed Description

`redis.Nil` signals an empty reply, not a failed operation, but it is currently funnelled through the same classification path as real errors. Because it matches no known error case, it is reported as an unknown internal client error.

Cache misses are expected behavior for most Redis workloads, so counting them as errors conflates normal operation with failure.

lazerg:
A `GET` for a missing key returns `redis.Nil`, which is a successful command with an empty reply. `classifyCommandError` has no case for it, so it fell through to the `UNKNOWN` / `internal=true` catch-all and every cache miss incremented `redis.client.errors`. A read-heavy service with a normal miss rate ends up with a permanent, unfilterable client error rate.

The fix skips the native error callback when the command error unwraps to `Nil`. The same guard goes on the three other paths that emit that callback from a command's own reply: the pipeline exit in `generalProcessPipeline`, the full-duplex autopipeline reader, and the client-side-cache miss coalescer. `failReqs`/`failQueue` are untouched — they only ever carry transport and shutdown errors, and `osscluster.go` only reports MOVED/ASK.

`redisotel-native` gets the same treatment on `db.client.operation.duration`, which was tagging nil replies `error.type=redis` and `redis.client.errors.category=server`, so success-rate queries built on that histogram counted misses as failures too.

Tests cover the command path, the pipeline path, the full-duplex path and the duration histogram; each one fails without the guard, and a real Redis error is still recorded.

Closes #4024

---

> [!NOTE]
> **Low Risk**
> Observability-only change on successful empty replies; real Redis errors and transport failures are still recorded. Low risk to command execution behavior.
> 
> **Overview**
> **Treats `redis.Nil` as success for metrics**, so cache misses (e.g. `GET` on a missing key) no longer inflate client error counters or OTel failure attributes.
> 
> Adds **`isNilReply`** (`errors.Is` against `Nil`, including wrapped errors) and applies it anywhere a command reply drives error reporting: **`recordCommandError`**, pipeline exit in **`generalProcessPipeline`**, full-duplex **`reportReplyMetrics`**, and CSC miss **`emitReplyErr`**. **`redisotel-native`** skips `error.type` / error-category attributes on **`db.client.operation.duration`** for Nil replies.
> 
> New tests lock in Nil vs real Redis errors on the command, pipeline, full-duplex, and OTel histogram paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6a1ddd4fa65cb037288158ecc1b08b2b2f1a978. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lazerg:
The govulncheck failure does not come from this change. It reports vulnerabilities in required modules for every module in the repo, including examples this PR does not touch, and other open PRs fail the same check.

vlady-kotsev:
Hey @lazerg, thanks for the contribution. Such a quick fix for the issue that was opened! I will review it soon.

redis.go
```diff
@@ -1641,8 +1641,12 @@ func (c *baseClient) processWithRetry(
 // failure: err classified, the connection that served the last attempt (nil when
 // none did), and the retries spent (attempts beyond the first). Every exit that
 // ends a command with an error shares it: the two in processWithRetry and the
-// exhausted-budget return in processCached.
+// exhausted-budget return in processCached. A Nil reply is a successful command
+// with no value, so it never reaches the callback.
 func recordCommandError(ctx context.Context, err error, cn *pool.Conn, retries int) {
+	if isNilReply(err) {
```
vlady-kotsev, line 1647: I was wondering can we move this filtering into `redisotel-native -> RecordError` . This way we won't need to filter at each callsite.

vlady-kotsev (changes requested):

vlady-kotsev:
I was thinking that we could introduce a configuration flag in `extra/redisotel-native`, which will default to false but can be controlled via `func (c *Config) WithRecordNilErrors(record bool) *Config ` that will enable/disable `redis.Nil` error reporting (under a new category "NIL" => `classifyCommandError` can do something like: 
```
if errors.Is(err, Nil) {
		return "NIL", "NIL", false
	}
```

redis.go
lazerg: Done in cca0756. I moved the filter into `RecordError` and removed the checks at the callsites. `RecordError` gets only the error type string, not the error, so `classifyCommandError` now returns `NIL` for `Nil`. I also added `WithRecordNilErrors` from your other comment. It is off by default.

extra/redisotel-native/metrics.go
```diff
@@ -188,6 +188,19 @@ func (r *metricsRecorder) RecordPipelineOperationDuration(
 	r.operationDuration.Record(ctx, durationSeconds, metric.WithAttributes(attrs...))
 }
 
+// errorTypeNil is the errorType the client reports for a redis.Nil reply.
+const errorTypeNil = "NIL"
```
vlady-kotsev: Let's reuse this constant in `redis` package as well. Maybe we can define it publicly in `<root>/otel.go`

extra/redisotel-native/metrics.go
```diff
@@ -386,6 +399,10 @@ func getErrorCategoryFromString(errStr string) string {
 		return ""
 	}
 
+	if errStr == errorTypeNil {
+		return "nil"
```
vlady-kotsev: Let's reuse the `errorTypeNil` constant

extra/redisotel-native/metrics.go
vlady-kotsev: Let categorize here: 
```
	if errors.Is(err, redis.Nil) {
		return errorTypeNil
	}
```

vlady-kotsev: Let's classify: 
```
if errors.Is(err, redis.Nil) {
		return errorTypeNil
	}
```

vlady-kotsev: Let's add: 
```
if errors.Is(err, redis.Nil) {
		return errorTypeNil
}
```

vlady-kotsev (changes requested):

vlady-kotsev:
Hey @lazerg, great progress! Let's address the final comments and resync with `master` and we will be ready to merge!

extra/redisotel-native/metrics.go
lazerg: Done in 70dd598. It is now `redis.ErrorTypeNil` in `otel.go`, and both `classifyCommandError` and redisotel-native use it.

extra/redisotel-native/metrics.go
lazerg: Done in 70dd598. A Nil reply now gets the `NIL` category as well.

extra/redisotel-native/metrics.go
lazerg: Added in 70dd598.

lazerg: Added in 70dd598.

lazerg: Added in 70dd598.

extra/redisotel-native/metrics.go
```diff
@@ -113,7 +113,7 @@ func (r *metricsRecorder) RecordOperationDuration(
 		}
 	}
 
-	if err != nil {
+	if err != nil && !r.skipNilReply(err) {
```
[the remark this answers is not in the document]
lazerg: Fixed in 70dd598. With `WithRecordNilErrors` on, the duration histogram now tags a Nil reply with `NIL` for the error type, category and status code, same as `RecordError`.

lazerg:
Thanks! I merged master into the branch and pushed the changes in 70dd598.

PR #4026, opened 2026-09-19
fix(cmd): preserve millisecond expiration dates

jakezwang:
`PExpireAt`, `HPExpireAt`, and `HPExpireAtWithArgs` derive milliseconds through `time.Time.UnixNano`, which overflows outside its supported timestamp range. For example, a 2270 expiration becomes negative: Redis immediately deletes the key for PEXPIREAT and rejects HPEXPIREAT.

Use `UnixMilli` directly. Public pipeline regressions cover both sides of the nanosecond range, normal precision and pre-epoch rounding. Integration tests verify that all three methods retain data and store the exact 2270 expiration.

Validation: original-code regressions fail; unit and live Redis 7.4.5/8.10.0 checks pass with the fix. The non-Ginkgo race suite across the root module passes on Go 1.27.1, as do focused tests on Go 1.25.13, six expiration integration specs, standard vet and custom vet. Three existing golangci-lint findings reproduce on the unchanged base; the patch has no new findings. Full cluster/e2e CI was not run locally.

---

> [!NOTE]
> **Medium Risk**
> Changes how expiration-at timestamps are serialized for three commands; behavior is wrong today for out-of-range dates but could affect any caller relying on the old (buggy) encoding for edge-case times.
> 
> **Overview**
> Fixes incorrect Redis expiration timestamps when `time.Time` values fall outside Go’s nanosecond `UnixNano` range (e.g. far-future dates like 2270), which could turn into negative millisecond values and cause keys/fields to expire immediately or commands to fail.
> 
> **`PExpireAt`**, **`HPExpireAt`**, and **`HPExpireAtWithArgs`** now send **`tm.UnixMilli()`** instead of deriving milliseconds from **`UnixNano()`**.
> 
> Adds a focused unit test that asserts the encoded timestamp in command args for **`PExpireAt`**, **`HPExpireAt`**, and **`HPExpireAtWithArgs`** (post-range, pre-range, sub-ms rounding, pre-epoch). Ginkgo integration tests cover **`PExpireAt`** and hash field expiration with a 2270 deadline and expected **`PEXPIRETIME`/`HPEXPIRETIME`** values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9537c922b552c71f62eba271e598ff2620b0884. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vlady-kotsev:
Hey @jakezwang, thanks for the contribution. I will review your fix promptly.

PR #4028, opened 2026-09-20
fix(pool): wake waiters when a connection is released to the pool

### LINKED ISSUE #4027: Streaming credentials: background re-auth can miss the idle transition and never re-authenticate the connection (lost wake-up in ConnStateMachine.AwaitAndTransition)
## Expected Behavior

After a `StreamingCredentialsProvider` delivers new credentials (`OnNext`), every pooled connection is re-authenticated in place by the background re-auth worker as soon as the connection is idle, and keeps serving afterwards.

## Current Behavior

About one rotation in five (3/20, 4/20 and 5/20 in repeated runs), the connection marked for re-authentication is never re-authenticated. The worker spawned by `ReAuthPoolHook.OnPut` blocks in `ConnStateMachine.AwaitAndTransition`, waiting for the connection to become idle although it already is, until the `PoolTimeout` context expires. Then `reAuthFn(context.DeadlineExceeded)` reaches `onAuthenticationErr`, `isBadConn` treats the deadline error as a bad connection, the connection is closed with

```
redis: re-authentication failed: context deadline exceeded
```

and the next command dials a replacement. While the connection is marked, `OnGet` rejects it on every checkout, so it serves nothing and holds its pool slot; with the pool at `PoolSize`, every command in that window dials and then drops a replacement.

Goroutine captured from a stalled run on v9.22.0 (dumped ~5 s after the push, worker still waiting):

```
goroutine 44 [select]:
github.com/redis/go-redis/v9/internal/pool.(*ConnStateMachine).AwaitAndTransition(...)
	internal/pool/conn_state.go:267
github.com/redis/go-redis/v9/internal/auth/streaming.(*ReAuthPoolHook).OnPut.func1()
	internal/auth/streaming/pool_hook.go:194
created by github.com/redis/go-redis/v9/internal/auth/streaming.(*ReAuthPoolHook).OnPut
	internal/auth/streaming/pool_hook.go:159
```

Mechanism (found by instrumenting v9.22.0; the event trace of a stalled rotation is in the first comment): the worker waits for the connection's IN_USE → IDLE transition, but that transition is made by `Conn.Release()` (`internal/pool/conn.go`, Put's hot path) with a bare CAS on the state word, and `Release()` never calls `notifyWaiters`. Whenever the worker has enqueued itself before Put releases the connection, the waiter is never served; because the pool is LIFO, the connection then sits at the bottom of the idle stack and nothing transitions it again until the worker's context deadline. `TryTransitionFast`, which the comment on `Release()` points to, is a bare CAS as well. A second, narrower ordering also loses the wake-up: when `Release()` runs between the worker's failed fast path and its enqueue, there is no waiter to notify yet and the enqueue comes too late. Both orderings occur in the reproduction, and a change for either one alone leaves the other (measured below). Because the fast path is skipped whenever `waiterCount > 0`, a later `OnNext` for the same connection queues a second worker behind the stuck one and does not unstick it (verified: a second push of the same credentials fails at the same rate).

The code is unchanged on `v9.23.0-beta.1` (reproduced, 5/20) and on `master`.

## Possible Solution

Two hunks, verified together against the reproduction: 0 of 100 rotations stall, against 3 to 8 of 20 unpatched. Each alone still stalls (re-check after the enqueue only: 3, 0, 1, 1 of 20; `Release()` notifying only: 3, 4, 4, 3, 4 of 20). `notifyWaiters` begins with `waiterCount.Load() == 0`, so the cost on the Put hot path is one atomic load when nobody is waiting, the check `TryTransition` already makes.

```diff
--- a/internal/pool/conn.go
+++ b/internal/pool/conn.go
@@ -957,7 +957,14 @@
 // on this transition, update this to use TryTransitionFast().
 func (cn *Conn) Release() bool {
 	// Inline the hot path - single CAS operation
-	return cn.stateMachine.state.CompareAndSwap(uint32(StateInUse), uint32(StateIdle))
+	if !cn.stateMachine.state.CompareAndSwap(uint32(StateInUse), uint32(StateIdle)) {
+		return false
+	}
+	// The connection just became IDLE: wake anyone waiting for that state (the
+	// re-auth worker parked in AwaitAndTransition). One atomic load when nobody
+	// is waiting.
+	cn.stateMachine.notifyWaiters()
+	return true
 }
 
 // ClearHandoffState clears the handoff state after successful handoff.
--- a/internal/pool/conn_state.go
+++ b/internal/pool/conn_state.go
@@ -263,6 +263,11 @@
 	sm.waiterCount.Add(1)
 	sm.mu.Unlock()
 
+	// Re-check now that the waiter is visible: a transition that happened
+	// between the fast path above and the enqueue (through Release, which
+	// found no waiter to notify) would otherwise never wake it.
+	sm.notifyWaiters()
+
 	// Wait for state change or timeout
 	select {
 	case <-ctx.Done():
```

A regression test for the first hunk (package `pool`; fails on v9.22.0 after its 2 s deadline, passes with the change) is in the first comment. `go test ./internal/pool/... ./internal/auth/... ./auth/...` pass with both hunks applied.

*Edited the same day: this section first proposed only the second hunk. The instrumented trace showed the un-notified `Release()` as the main path; see the first comment.*

## Steps to Reproduce

1. Start a server with ACL support: `docker run --rm -d -p 127.0.0.1:63790:6379 valkey/valkey:8.1-alpine`
2. Save the program below as `main.go` in a module that requires `github.com/redis/go-redis/v9 v9.22.0`.
3. `go run . 127.0.0.1:63790`
4. Some of the 20 rotations report the original connection still listed as `rotuser1` after 3 s, and go-redis logs `re-authentication failed: context deadline exceeded` once the pool timeout expires.

```go
// Reproduces: a pooled connection marked for re-authentication after a
// StreamingCredentialsProvider push is sometimes never re-authenticated in place.
// The background re-auth worker blocks in ConnStateMachine.AwaitAndTransition
// waiting for the connection to become idle, although it already is, until the
// pool timeout expires and the connection is closed.
//
// Run against a Redis/Valkey with ACL support:
//
//	docker run --rm -d -p 127.0.0.1:63790:6379 valkey/valkey:8.1-alpine
//	go run . 127.0.0.1:63790
package main

import (
	"context"
	"fmt"
	"os"
	"strings"
	"sync"
	"time"

	"github.com/redis/go-redis/v9/auth"
)

// provider is a minimal auth.StreamingCredentialsProvider: Subscribe hands out
// the current credentials and remembers the listener; Push delivers new ones.
type provider struct {
	mu        sync.Mutex
	current   basicCreds
	listeners []auth.CredentialsListener
}

type basicCreds struct{ user, password string }

func (c basicCreds) BasicAuth() (string, string) { return c.user, c.password }
func (c basicCreds) RawCredentials() string       { return c.user + ":" + c.password }

func (p *provider) Subscribe(l auth.CredentialsListener) (auth.Credentials, auth.UnsubscribeFunc, error) {
	p.mu.Lock()
	defer p.mu.Unlock()
	p.listeners = append(p.listeners, l)
	return p.current, func() error { return nil }, nil
}

func (p *provider) Push(user, password string) {
	p.mu.Lock()
	p.current = basicCreds{user, password}
	listeners := append([]auth.CredentialsListener(nil), p.listeners...)
	p.mu.Unlock()
	for _, l := range listeners {
		l.OnNext(p.current)
	}
}

func idsForUser(ctx context.Context, admin *redis.Client, user string) (map[string]bool, error) {
	list, err := admin.Do(ctx, "CLIENT", "LIST").Text()
	if err != nil {
		return nil, err
	}
	ids := map[string]bool{}
	for _, line := range strings.Split(list, "\n") {
		if !strings.Contains(line, " user="+user+" ") {
			continue
		}
		for _, f := range strings.Fields(line) {
			if id, ok := strings.CutPrefix(f, "id="); ok {
				ids[id] = true
			}
		}
	}
	return ids, nil
}

func main() {
	if len(os.Args) < 2 {
		fmt.Println("usage: go run . <redis-addr>")
		os.Exit(2)
	}
	addr := os.Args[1]
	ctx := context.Background()
	admin := redis.NewClient(&redis.Options{Addr: addr})
	for _, u := range []string{"rotuser1", "rotuser2"} {
		if err := admin.Do(ctx, "ACL", "SETUSER", u, "on", ">"+u+"-pw", "~*", "+@all").Err(); err != nil {
			panic(err)
		}
	}
	defer admin.Do(ctx, "ACL", "DELUSER", "rotuser1", "rotuser2")

	const runs = 20
	stalled := 0
	for i := 0; i < runs; i++ {
		p := &provider{current: basicCreds{"rotuser1", "rotuser1-pw"}}
		client := redis.NewClient(&redis.Options{Addr: addr, StreamingCredentialsProvider: p, PoolSize: 2})
		if err := client.Set(ctx, "k", "v", time.Minute).Err(); err != nil {
			panic(err)
		}
		before, _ := idsForUser(ctx, admin, "rotuser1") // the one pooled connection

		p.Push("rotuser2", "rotuser2-pw") // credentials rotate; the connection is marked for re-auth

		// Keep issuing commands, as an application would. Each checkout rejects
		// the marked connection (OnGet), so a second connection serves them.
		reauthed := false
		deadline := time.Now().Add(3 * time.Second)
		for time.Now().Before(deadline) && !reauthed {
			_ = client.Set(ctx, "k", "v", time.Minute).Err()
			after, _ := idsForUser(ctx, admin, "rotuser2")
			reauthed = true
			for id := range before {
				if !after[id] {
					reauthed = false
				}
			}
			time.Sleep(50 * time.Millisecond)
		}
		if !reauthed {
			stalled++
			still, _ := idsForUser(ctx, admin, "rotuser1")
			fmt.Printf("run %2d: original connection %v NOT re-authenticated after 3s (still listed as rotuser1: %v)\n", i, keys(before), keys(still))
		}
		client.Close()
	}
	fmt.Printf("%d of %d rotations left the original connection un-re-authenticated after 3s\n", stalled, runs)
}

func keys(m map[string]bool) []string {
	out := []string{}
	for k := range m {
		out = append(out, k)
	}
	return out
}
```

## Context (Environment)

- go-redis `v9.22.0`; also reproduced on `v9.23.0-beta.1`; the relevant code is unchanged on `master`.
- Go 1.26.6, darwin/arm64; Valkey 8.1 (`valkey/valkey:8.1-alpine`).
- Use case: a `StreamingCredentialsProvider` that pushes rotated file-backed ACL credentials to a pool, relying on in-place re-authentication so that a rotation costs no reconnect. With this race, a rotation instead closes and replaces some connections after the pool timeout, and a saturated pool dials a replacement per command during that window.

avitenzer:
## Summary

A background worker that must wait for a pooled connection to become `IDLE` (the streaming-credentials re-auth worker in `internal/auth/streaming/pool_hook.go`) parks in `ConnStateMachine.AwaitAndTransition`. `Put` performs the `IN_USE → IDLE` transition through `Conn.Release()`, a bare CAS on the state word that never called `notifyWaiters`, so a worker that had enqueued itself before the release was never woken. With the default LIFO pool nothing touches that connection again, so the worker sleeps until its context expires (`PoolTimeout`), after which the re-auth error path closes the connection and the next command dials a replacement. A second ordering loses the wake-up as well: when `Release()` runs between the worker's failed fast path and its enqueue, there is no waiter to notify yet.

Trace of a stalled rotation on v9.22.0 (full details in #4027):

```
reauth worker conn[2]: waiting for IDLE, state=IDLE
conn[2] TryAcquire: IDLE->IN_USE                      <- the application checks it out (OnGet rejects it)
AwaitAndTransition: state=IN_USE waiters=0            <- the worker's fast path fails
AwaitAndTransition: enqueued waiter, state now IN_USE
conn[2] Release: IN_USE->IDLE (waiters=1)             <- Put releases it; nobody is notified
... 6 s later: AwaitAndTransition: gave up (context deadline exceeded)
```

## Changes

- `Conn.Release()` notifies waiters after a successful CAS. `notifyWaiters` starts with an atomic load of `waiterCount`, so the Put hot path pays one load when nobody is waiting, the check `TryTransition` already makes.
- `AwaitAndTransition` re-runs the notification once after enqueueing, so a transition that landed between the fast path and the enqueue is served instead of waited for.
- Two tests in `conn_state_test.go`, one per ordering. Each fails on master with `context deadline exceeded` and passes with its hunk: `WokenByRelease` needs the `Release()` change, `ReleaseBeforeEnqueue` needs the re-check.

## Verification

Reproduction from #4027 (a `StreamingCredentialsProvider` rotating credentials, `PoolSize: 2`, Valkey 8.1), 20 rotations per run:

| | stalled rotations |
|---|---|
| master (7f3b3dff) | 8, 9 |
| master + this change | 0, 0, 0 |

`go test ./internal/... ./auth/...` pass; the new tests also with `-race -count=5`.

Fixes #4027

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches the Put hot path and connection state-machine synchronization; incorrect waiter logic could affect re-auth/handoff correctness, but changes are narrow and well-tested.
> 
> **Overview**
> Fixes a race where **background workers** (streaming-credentials re-auth, handoff) parked in `AwaitAndTransition` waiting for **IDLE** were never woken when the pool returned a connection via **`Conn.Release()`** (`Put` hot path). That path only CAS’d `IN_USE → IDLE` and skipped waiter notification, so workers could sit until `PoolTimeout` and stall credential rotation (#4027).
> 
> **`Release()`** now calls `notifyWaiters()` after a successful release (cheap when `waiterCount` is zero). **`AwaitAndTransition`** also re-invokes `notifyWaiters()` right after enqueueing, covering the window where `Release()` runs between the failed fast path and the waiter becoming visible.
> 
> Two regression tests lock in both orderings (`WokenByRelease`, `ReleaseBeforeEnqueue`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb650090613d5bc103efaba31d085104d5746139. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vlady-kotsev:
Hey @avitenzer, thanks for the contribution, I will review it soon.

